PR #4, opened 2019-04-26
refactor: clean up observerMixin

thepassle:
@daKmoR Can you make sure this is correct before I continue? :=)


daKmoR:
looks good - as long as the tests pass 🙈 
=> still working on that ci...

thepassle:
@daKmoR could you take a look at this when you have some time? Cant get these tests to pass :/

```
FormatMixin
 ✖ fires `model-value-changed` for every change on the input
 ✖ fires `model-value-changed` for every modelValue change
```

PR #9, opened 2019-04-27
refactor: remove observermixin

thepassle:


thepassle:
This is ready for review now. if someone could take a look at the `updated()` in LionField specifically, because the only way I could get it to work was like this, but it seems to completely bypass the delegatemixin now, and im not sure if this is the way to go.

Also the tests in ValidateMixin.test.js now need a nextFrame and im confused as to why awaiting updateComplete isnt enough, would love to learn more :)

packages/field/src/FormatMixin.js
```diff
@@ -81,13 +80,29 @@ export const FormatMixin = dedupeMixin(
...
-        return {
-          ...super.syncObservers,
-          _onModelValueChanged: ['modelValue'],
-          _onSerializedValueChanged: ['serializedValue'],
-          _onFormattedValueChanged: ['formattedValue'],
-        };
+      _requestUpdate(name, oldValue) {
+        super._requestUpdate(name, oldValue);
+
+        switch (name) {
```
daKmoR: how about writing it like that?

```js
const params = 
const map = {
  serializedValue: () => this._onSerializedValueChanged(oldValue),
  formattedValue: () => this._onFormattedValueChanged(oldValue),
};
map[name]();
```

e.g. will also require rewrite for some of the "observer" to accept only the old value (current value should be read by this.foo)

thepassle: Sure :)

Not sure what you mean with the second part? `_onSerializedValueChanged` and `_onFormattedValueChanged` don't do anything with `oldValue` right?

daKmoR: sorry forgot the if
```js
if (map[name]) {
  map[name]();
}
```
so it will only call the function if we have something defined for it

thepassle:
@bashmish if you're doing reviews, this one has been done for a while now :p

bashmish:
@thepassle we only came to the agreement on what to do with this a few days ago. Let's have a quick chat somewhere to discuss how to proceed with this now, because we currently want to have a different ObserverMixin implementation for the internal usage and not expose it for other users (since it will not be very generic and teaching how to use it is also a challenge on the lion side where not many people xpect it to be), and we also need to make sure the alternative is documented in LitElement (preferably) or in our docs somewhere, so that people now how to refactor their code.

thepassle:
Hey @bashmish , I totally lost track of this and was reminded just now by #150 . Did you still wanna have a chat about this? I could drop by sometime next week if you want

bashmish:
@thepassle we have this in our roadmap, but no specific date yet. For now I would say let's keep this as is, no help is needed. If we need your help later, we'll happily let you know :) thanks!

jorenbroekema:
https://github.com/ing-bank/lion/pull/371 this one removes all deprecated things from lion, including ObserverMixin 🎉

thatianamg:
Thank you for your work here! We finally could remove all the observer mixins in the latest releases

PR #11, opened 2019-04-29
Split date functions

bashmish:
Too many functions in one file, so let's split them up as a preparation step to add more utilities there.


daKmoR:
Note: As browserstack is in a different timezone we need to use "save" Date operations - I did the change initially to get tests working here
https://github.com/ing-bank/lion/commit/cbbf0d49fbb4bd043723928e4068e581af281b3b

PR #12, opened 2019-04-29
chore(field): custom fields tutorial

tlouisse:
After getting some feedback that it was hard to find documentation for creating your own custom fields, I decided to create a small guide/tutorial for it.

It will also serve as a guide when we want to build our own slider, select(dropdown listbox) and combobox/autocomplete.

### DIFF
--- a/packages/field/docs/CustomFieldsTutorial.md
+++ b/packages/field/docs/CustomFieldsTutorial.md
@@ -1,0 +1,76 @@
+# Creating a custom field
+Custom fields can be created in just a few steps. All you need is an interaction element 
+(like for instance a slider, a listbox or a combobox) and connect it to the [Field](../README.md)
+functionality.
+
+## Prerequisite: an interaction element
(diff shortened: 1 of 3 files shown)


packages/field/docs/CustomFieldsTutorial.md
```diff
@@ -0,0 +1,76 @@
+# Creating a custom field
```
daKmoR: How about putting it at the end of the field readme?

tlouisse: There is no Field readme yet;) Only the Form Fundaments, which kind of gives an overview of all form controls (fields and fieldsets).

tlouisse: I created an issue to solve the 'relative link problem' for npm(https://github.com/ing-bank/lion/issues/13), so that we can keep separate files for documentation.

PR #15, opened 2019-05-01
fix(field): recall formatter on errorState change

gerjanvangeest:
This is a quick workaround for this problem https://github.com/ing-bank/lion/issues/22

### DIFF
--- a/packages/field/src/FormatMixin.js
+++ b/packages/field/src/FormatMixin.js
@@ -251,6 +251,21 @@ export const FormatMixin = dedupeMixin(
         }
       }
 
+      // Current flow works like this: setting of modelValue triggers observers for:
+      // - computation of 'formattedValue'
+      // - validation (and thus computation of 'errorState')
+      // Observers will be handled in above order. Since computation of formattedValue depends on 'errorState'
+      // (erroneous inputs should not be formatted), the formatter is handled with a not yet updated errorState.
+      //
+      // TODO: rewrite to a flow handling observers in this order: modelValue -> errorState(validation) -> formattedValue
+      // Important: currently, a situation where a value changes from valid to invalid will still lead to formatting when not
+      // required.
+      // Since the flow above involves two mixins (FormatMixin and ValidateMixin) and ValidateMixin should have no
+      // knowledge about FormatMixin, think of a proper way to do this.
+      _reformatOnErrorStateChanged() {
+        this.formattedValue = this.__callFormatter(this.modelValue);
+      }
+
       // TODO: rename to __dispatchNormalizedInputEvent?
       // This can be called whenever the view value should be updated. Dependent on component type
       // ("input" for <input> or "change" for <select>(mainly for IE)) a different event should be used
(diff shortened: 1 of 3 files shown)


packages/input-iban/src/formatters.js
```diff
@@ -7,5 +7,8 @@ import { friendlyFormatIBAN } from '@bundled-es-modules/ibantools';
  * @return {string} formatted value
  */
 export function formatIBAN(modelValue) {
+  if (modelValue === '') {
```
jorenbroekema: If this is a bug because ibantools doesn't handle `''` properly in their formatter, maybe nice to make an issue for it on their repo :)?

tlouisse: Good suggestion. It's a bug in ibantools indeed. Maybe also add a comment about it (defensive code because of ibantools) @gerjanvangeest

packages/field/src/FormatMixin.js
```diff
@@ -259,6 +259,11 @@ export const FormatMixin = dedupeMixin(
         }
       }
 
+      // TODO: refactor formatter/parser so this listener is not needed
+      _errorStateHandler() {
```
tlouisse: Maybe give it a more specific name like `_reformatOnErrorStateChanged` to prevent potential naming collisions on the prototype (since it's quite a generic name)?

In the comment, maybe add:
```
// Current flow works like this: setting of modelValue triggers observers for:
// - computation of 'formattedValue'
// - validation (and thus computation of 'errorState')
// Observers will be handled in above order. Since computation of formattedValue depends on 'errorState' 
// (erroneous inputs should not be formatted), the formatter is handled with a not yet updated errorState.
// 
// TODO: rewrite to a flow handling observers in this order: modelValue -> errorState(validation) -> formattedValue
// Important: currently, a situation where a value changes from valid to invalid will still lead to formatting when not 
// required. 
// Since the flow above involves two mixins (FormatMixin and ValidateMixin) and ValidateMixin should have no
// knowledge about FormatMixin, think of a proper way to do this.
```

packages/field/test/FormatMixin.test.js
```diff
@@ -223,9 +223,28 @@ describe('FormatMixin', () => {
       el.errorState = false;
       el.modelValue = 'bar2';
       expect(formatterSpy.callCount).to.equal(2);
-
       expect(el.formattedValue).to.equal('foo: bar2');
     });
+
+    it('will recall the formatter if errorState has changed', async () => {
```
tlouisse: Maybe create an it that describes the failing feature, since this is not necessarily a feature (but a temp solution to solve a bug) from the app developer perspective.
Could we add some code to the above test that fails in the prev state? (that doesn't reformat when invlaid input becomes valid)

packages/input-iban/src/formatters.js
gerjanvangeest: https://github.com/Simplify/ibantools/issues/15

bashmish:
@gerjanvangeest what bug are we fixing with this? I mean the test is very low level and technical, I don't see from it how the behavior is broken for the end user.

@tlouisse I've heard you had a different approach to fixing this. Can you please outline it here?

bashmish:
@tlouisse I've heard you had a different approach to fixing this. Can you please outline it here?

gerjanvangeest:
I close this one in fever of: https://github.com/ing-bank/lion/pull/84

PR #16, opened 2019-05-03
WIP: fix(core): slotContent can be instance of Element or HTMLElement

gerjanvangeest:


packages/core/src/SlotMixin.js
```diff
@@ -57,7 +57,7 @@ export const SlotMixin = dedupeMixin(
             if (!this.$$slot(slotName)) {
               const slotFactory = this.slots[slotName];
               const slotContent = slotFactory();
-              if (slotContent instanceof Element) {
+              if (slotContent instanceof Element || slotContent instanceof HTMLElement) {
```
daKmoR: can we add a test for that - seems rather important

daKmoR:
this is apparently only needed if you are on a certain version of polymer and polyfills... 🙈 

so it feels more like a workaround better to update to a working version of polymer/polyfills.

gonna close this for now - if really needed pls reopen and provide more details why it's needed

packages/core/src/SlotMixin.js
bashmish: the issue was in the application code

in short, they had this in their Polymer app which replaced global Element object

```html
<script>
  const { Element } = window.Polymer;
</script>
```

so like we decided a while ago, there is no need to change anything on our side

PR #17, opened 2019-05-03
fix(input-amount): remove randomOkValidator as default

gerjanvangeest:


tlouisse:
Good catch. So this is what the 'success message bug' was about then (it only appeared in input-amount)?

PR #18, opened 2019-05-03
chore: refactor to not use deprecated EventMixin

thepassle:


thepassle:
This can be reviewed

packages/field/src/FormatMixin.js
```diff
@@ -288,13 +280,15 @@ export const FormatMixin = dedupeMixin(
...
         this._reflectBackFormattedValueToUser = this._reflectBackFormattedValueToUser.bind(this);
+
         this._reflectBackFormattedValueDebounced = () => {
           // Make sure this is fired after the change event of inputElement, so that formattedValue
           // is guaranteed to be calculated
           setTimeout(this._reflectBackFormattedValueToUser);
         };
         this.inputElement.addEventListener(this.formatOn, this._reflectBackFormattedValueDebounced);
-
+        this.inputElement.addEventListener('input', this._proxyInputEvent.bind(this));
```
tlouisse: If I'm right `bind(this)` returns a new function and thus the reference in `disconnectedCallback` is not the same. You can solve this by doing `this._proxyInputEvent = this._proxyInputEvent.bind(this);` once in the constructor
and then just do `add/removeEventListener('input', this._proxyInputEvent);`

thepassle: Yep, you're correct, forgot to clean this one up 🙈 Fixed it now :)

PR #21, opened 2019-05-03
Feat/pointing frame

jorenbroekema:
Still WIP:

- [x] Fix IE11 bugs
- [x] Write additional tests for found edge cases
- [x] Fix failing tests
- [ ] Cleanup


packages/popup/src/LionPopup.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable class-methods-use-this, no-underscore-dangle */
```
daKmoR: line should not be needed => underscore-dangle is globally disabled and class methods should use this... (default LitElement life cycle methods are already excluded) if you have another function where you don't need this it is better to exclude it specifically

tlouisse:
Lll

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

packages/pointing-frame/src/LionPointingFrame.js
```diff
@@ -0,0 +1,278 @@
+import { LionLitElement, html, css, SlotMixin } from '@lion/core';
```
LarsDenBakker: The pointer frame is just the arrow on a popover, right? Did you see how it was implemented in the polymer popover? It seems like a lot of code + an extra web component just for a pointer.

jorenbroekema: Most of the code has to do with the UX specs of the pointer. It has to be pointing in the middle of the invoker at all times, so for center positioned popups that's easy. For the corner positions it is more complex. The pointer has to be a certain set distance from the edge of the content frame while still pointing to the middle of the invoker, and this complicates things quite a bit in that you need to give the frame some offset spacing.

The reason for the extra webcomponent is for people to be able to change the pointer svg and the distance from the edge as they please. Otherwise it could probably be styles-only.

image
This image kind of shows you how that works. The pink border is the pointing frame which position is managed by the managePosition util. Hopefully that clarifies a bit why the complexity.

I will check the polymer popover though to see if I missed an opportunity to make things simpler.

LarsDenBakker: Ah right, the moving pointer is not something that was in the old implementation. Does makes things more complex indeed.

The pointing frame isn't implemented directly by people, right? It could be a utility or mixin implemented by the popover then.

packages/overlays/stories/local-overlay-placement.stories.js
```diff
@@ -41,34 +41,36 @@ storiesOf('Local Overlay System|Local Overlay Placement', module)
...
+      'center-of-bottom',
+      'left-of-bottom',
+      'bottom-of-left',
+      'center-of-left',
+      'top-of-left',
+      'left-of-top',
+    ];
+    let placement = 'top-of-left';
+
     const togglePlacement = () => {
```
jorenbroekema: Should refactor to knobs :)

packages/pointing-frame/src/LionPointingFrame.js
jorenbroekema: Hmm you're correct I think. It wouldn't work well by itself, it should be inside something that implement LocalOverlayController like lion-popup. 

So is this close to what you're thinking? pseudo example:
```js
import { myPointer } from './myPointer.svg.js';

class MyCustomTooltip extends PointingFrameMixin(LionTooltip) {
  constructor() {
    super();
    this.pointerSvg = myPointer;
    this.pointerDistanceFromEdge = 6px;
  }
}
```

@daKmoR  thoughts?

LarsDenBakker: Yep that's it +1

packages/overlays/stories/local-overlay-placement.stories.js
```diff
@@ -122,7 +124,8 @@ storiesOf('Local Overlay System|Local Overlay Placement', module)
         default:
           placement = 'top left';
       }
-      managePosition(element, target, { placement, position: 'absolute' });
+      action('position: ')(placement);
```
daKmoR: usage of action is no longer needed... a simple `console.log` is enough and it will shop up in the action logger

packages/pointing-frame/src/LionPointingFrame.js
daKmoR: I can't think of a use case where a mixin us really helpful... having a web-component you can extend is good enough for now I would say :hugs:

people will probably only use the pointing-frame like so
```html
<lion-popup>
  <button slot="invoker">click me</button>
  <lion-pointing-frame slot="content">
     My Content // or <my-complex-form>
  </lion-pointing-frame>
</lion-popup>
```

e.g. composition should be good enough here

only if you want to make a differently looking frame you will need to extends `ComicBubbleFrame extends LionPointingFrame`
```html
<lion-popup>
  <button slot="invoker">click me</button>
  <comic-bubble-frame slot="content">
     My Content // or <my-complex-form>
  </comic-bubble-frame>
</lion-popup>
```

LarsDenBakker: So the pointing frame is also the thing that positions the content?

daKmoR: nope the frame is only the "frame" e.g. only visual stuff like the pointer and the margins...

the position is all done by "lion-popup" which is using the LocalOverlaySystem... so you can put in whatever you want... if you just want a div with a fixed background you can do so as well... the pointing-frame, however, is a little "smarter" as it displays the pointer based on which attributes get set by the Local Overlay System

packages/pointing-frame/src/LionPointingFrame.js
```diff
@@ -0,0 +1,280 @@
...
+      :host {
+        box-sizing: border-box;
+      }
+    `;
+  }
+
+  get slots() {
+    return {
+      ...super.slots,
+      // TODO: Refactor to put it on instance instead of constructor. No need for clone ;)
```
tlouisse: Coming to think of this... there actually is no need at all here for the SlotMixin :)
The general mindset should still be that we don't mess around with light dom ourselves in our components.

Unfortunately there is no way to create accessible components in shadow dom until the [form participation api](http://hangaroundtheweb.com/2018/06/html-form-participation-api-explained/) and the [AOM](https://github.com/WICG/aom/blob/gh-pages/explainer.md) are a reality.
The SlotMixin was created for accessibility purposes. 

Since the pointer is a purely visual thing there would be no problem to place the default pointer in the shadow dom.
We can just use the platform for this in our template:
```html
<slot name="pointer">
    <my-pointer-svg></my-pointer-svg>
</slot>
```
`<my-pointer-svg>` will be visible by default, unless the consuming developer defines the slot himself.

tlouisse: Also, I think the `<lion-icon>` is not needed, since it's main purpose is accessibility and a webcomponent for just an svg feels a bit heavier than needed.

packages/pointing-frame/src/LionPointingFrame.js
```diff
@@ -0,0 +1,280 @@
...
+      },
+      invokerWidth: {
+        type: Number,
+        attribute: 'invoker-width',
+      },
+      invokerHeight: {
+        type: Number,
+        attribute: 'invoker-height',
+      },
+      pointerDistanceFromEdge: {
```
tlouisse: is this supposed to be a public api? If not, we should underscore it. If so maybe a shorter name like `offset` or `pointer-offset` would be better?

packages/pointing-frame/src/LionPointingFrame.js
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+  __setFrameStyles(pointerEl) {
+    const pointingFrameEl = this.shadowRoot.querySelector('.pointing-frame');
+
+    const frameStyles = {
+      boxSizing: 'border-box',
+      fontSize: '16px',
+      color: 'black',
+      backgroundColor: 'white',
```
tlouisse: I think these are run in an update loop? The top 4 props mainly concern styling and setting them once (outside update loop) would be better

packages/pointing-frame/src/LionPointingFrame.js
jorenbroekema: If we want to go with a Mixin we should do so now rather than later I think. 

The cool thing about the mixin is that if anyone makes a toast or context menu or any local overlay basically, they can quite easily make it have a pointer by just applying the mixin to their custom overlay. So I see use cases there definitely, and it prevents that people try to use pointing-frame on its own in contexts where it doesn't work, it prevents confusion that way.

I'm not sure how someone would make a comic bubble though in this way, I guess they would have to .. extend the mixin as a ComicBubbleMixin and apply that one instead?

packages/pointing-frame/src/LionPointingFrame.js
jorenbroekema: Yes, I think pointerOffset is better indeed

jorenbroekema:
Closing this for now, but keep the branch open in case code is useful when making a pointing frame CE based on new LocalOverlay which uses Popper.

PR #23, opened 2019-05-03
WIP: refactor: remove css class mixin

thepassle:


packages/field/src/FocusMixin.js
```diff
@@ -31,17 +31,17 @@ export const FocusMixin = dedupeMixin(
        *   return this.inputElement === document.activeElement;
        */
       get focused() {
-        return this.classList.contains('state-focused');
```
bashmish: can we maybe use a property reflected in an attribute instead of writing this boilerplate code, like we have for `touched` and `dirty`?

also in the new lion-calendar we are using this now with just `[%name%]` instead of `[state-%name%]`

in terms of compatibility it will be a similar change since you are changing the API regardless (from classes to attributes)

bashmish: @daKmoR or shall we also use `[state-%name%]` in the lion-calendar?

packages/field/test/InteractionStateMixin.test.js
```diff
@@ -51,15 +51,15 @@ describe('InteractionStateMixin', async () => {
     // expect(formElement.touched).to.equal(true);
   });
 
-  it('sets a class "state-(touched|dirty)"', async () => {
+  it('sets an attribute "state-(touched|dirty)"', async () => {
```
bashmish: are we going to make it a major version bump? shall we maybe stick to half-year major updates cycle already due to the need to keep extensions of lion stable as well?

packages/field/src/FocusMixin.js
daKmoR: I would say it should just be `focused` as property / attribute then this getter goes naturally away :hugs:

packages/field/src/InteractionStateMixin.js
```diff
@@ -25,15 +24,17 @@ export const InteractionStateMixin = dedupeMixin(
            */
           touched: {
             type: Boolean,
-            nonEmptyToClass: 'state-touched',
+            reflect: true,
+            attribute: 'state-touched',
```
daKmoR: let's remove it... the default name should be good enough :hugs:

thepassle:
Made _some_ progress on this, most of it is easy enough, but the validatemixin is throwing me off enirely. The CssClassMixin updates the class on `update` time, and UpdatingElement sets atributes on `update` time as well, yet when I change it to reflected attributes, the attributes don't match the state. Gonna take some more time today/tomorrow, and I need to do more cleanup etc as well. Just fyi.

packages/field/test/InteractionStateMixin.test.js
tlouisse: It would indeed be a breaking change. In order to move on with this pr, we could consider reflecting everything still to classes in the updated callback (so that we can throw away that code in next release).

daKmoR: let's ask first if that compatibility is needed as I guess it's only used in our styling and people should not have deviated from it (let's hope)

if needed that compatibility should be on a different layer 😉

tlouisse: > if needed that compatibility should be on a different layer 😉
Ah, that's a good idea indeed

bashmish:
@thepassle hey, any news on this?

thepassle:
Havent worked on this recently, because of the discussion about it being a breaking change (do we go for it or no?), but I can pick it up again

bashmish:
@thepassle I agree with the point of @daKmoR  that this API is not used a lot, so we can technically make a breaking release (and we will), but it won't cause many issues for our users to update.

packages/field/test/InteractionStateMixin.test.js
bashmish: agreed

packages/field/src/FocusMixin.js
bashmish: let's please go for `[focused]`

tlouisse:
Hey, promised to give you an answer on this. So I did a debug session in UpdatingElement :)

This is what happens:

1. ValidateMixin.errorState = true
2.  **async update starts**: UpdatingElement.update is called
3. asyncObserver triggered (via ObserverMixin calling UpdatingElement.update instead of UpdatingElement.updated)
4. callback ValidateMixin._onErrorShowChangedAsync called
5. ValidateMixin._onErrorShowChangedAsync sets errorShow = true
6. errorShow added to UpdatingElement._reflectedProperties -> **too late, update already in progress**
7. **async update ends**: UpdatingElement._reflectedProperties is emptied

### TLDR; 
the ObserverMixin doesn’t perfectly align with UpdatingElement. It hooks into  `.update` instead of `.updated`.
The reason that it worked for CssClassMixin is because this aligns with ObserverMixin, whereas reflecting attrs rely on UpdatingElement.

### Best way to move forward 
Do those changes on top of the other (removal of ObserverMixin) branch: as long as we synchronise with updated callback and don’t use asyncObservers, it should work.

jorenbroekema:
https://github.com/ing-bank/lion/pull/228 addressed these changes in a non-breaking way (so for example, state-focused is still there, but there's a prop for it now as well). 

Let me know in case you think we should keep this open.

PR #25, opened 2019-05-04
chore: add bundlesize checks

LarsDenBakker:
I was playing around with bundlesize, and I thought it would be useful to add it to the lion repository. 

I added presets of only the core, an average application and a the entire repository. These are built with rollup and the results are analyzed with https://github.com/siddharthkp/bundlesize

The analysis is run with `yarn bundlesize`, the command reports whether it passes the determined size parameters

If you like this approach, it can also be added to the circle CI flow to give an indicator on a PR:


daKmoR:
very nice I that was on my "I want to have that at some point list" 🤗 

so yes I like it :)

`all` and `average app` is the same? so maybe just leave it at all for now? we can easily readd it once we have a deviation 👍

LarsDenBakker:
Yeah they're the same now, but I think pretty soon there will be packages which won't be in `average-app`. My thinking was to add the bundles now so that the presets are there to add to later. Up to you though. 

I could look into the circle CI thing, not sure if I have permissions to that.

daKmoR:
let's add it once we need it 🤗 

You should have access to circleci - if not pls DM me 🤗 

However, the needed bot requires access which I can't grant (only a group admin can)... I will request it on Monday 👍

bashmish:
@daKmoR any update on this?

daKmoR:
> Access request pending

so no 😭
 
@spockz anything blocking the approval?

gerjanvangeest:
woohoo it is now approved so we can revisit this 👍

daKmoR:
took a while 🙈 

but now it's there 💪

LarsDenBakker:
😁

PR #29, opened 2019-05-07
chore(wallaby.js): configure wallaby.js

bashmish:
For everyone who uses Wallaby.js this config will be very helpful.


packages/localize/test/date/parseDate.test.js
```diff
@@ -12,6 +13,10 @@ function equalsDate(value, date) {
 }
 
 describe('parseDate()', () => {
+  beforeEach(() => {
+    localizeTearDown();
+  });
```
bashmish: this was needed to make tests pass in Wallaby.js

I guess with karma it somehow worked thanks to a different order of the tests run

PR #37, opened 2019-05-13
lion-calendar

bashmish:
Reusable calendar component 📆

### DIFF
--- a/packages/calendar/package.json
+++ b/packages/calendar/package.json
@@ -1,0 +1,43 @@
+{
+  "name": "@lion/calendar",
+  "version": "0.0.0",
+  "description": "Reusable calendar component",
+  "author": "ing-bank",
+  "homepage": "https://github.com/ing-bank/lion/",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ing-bank/lion.git",
+    "directory": "packages/calendar"
+  },
+  "scripts": {
+    "prepublishOnly": "../../scripts/npm-prepublish.js"
+  },
+  "keywords": [
+    "lion",
+    "web-components",
+    "calendar"
+  ],
+  "main": "index.js",
+  "module": "index.js",
+  "files": [
+    "src",
+    "stories",
+    "test",
+    "translations",
+    "*.js"
+  ],
+  "dependencies": {
+    "@lion/core": "^0.1.4",
+    "@lion/localize": "^0.1.7"
+  },
+  "devDependencies": {
+    "@lion/button": "^0.1.7",
+    "@open-wc/demoing-storybook": "^0.2.0",
+    "@open-wc/testing": "^0.11.1",
+    "sinon": "^7.2.2"
+  }
+}
--- a/packages/localize/src/date/getMonthNames.js
+++ b/packages/localize/src/date/getMonthNames.js
@@ -1,0 +1,25 @@
+import { normalizeDate } from './normalizeDate.js';
+
+const monthsLocaleCache = {};
+
+export function getMonthNames({ locale }) {
+  let months = monthsLocaleCache[locale];
+
+  if (months) {
+    return months;
+  }
+
+  months = [];
+
+  const formatter = new Intl.DateTimeFormat(locale, { month: 'long' });
+  for (let i = 0; i < 12; i += 1) {
+    const date = new Date(2019, i, 1);
+    const formattedDate = formatter.format(date);
+    const normalizedDate = normalizeDate(formattedDate);
+    months.push(normalizedDate);
+  }
+
+  monthsLocaleCache[locale] = months;
+
+  return months;
+}
--- a/packages/calendar/README.md
+++ b/packages/calendar/README.md
@@ -1,0 +1,44 @@
+# Calendar
+
+[//]: # (AUTO INSERT HEADER PREPUBLISH)
+
+`lion-calendar` is a reusable and accessible calendar view.
+
+## Features
+
+- fully accessible keyboard navigation (Arrow Keys, PgUp, PgDn, ALT+PgUp, ALT+PgDn)
+- **minDate**: disables all dates before a given date
+- **maxDate**: disables all dates after a given date
+- **disableDates**: disables some dates within an available range
+- **selectedDate**: currently selected date
+- **centralDate**: date that determines the currently visible month and that will be focused when keyboard moves the focus to the month grid
+- **focusedDate**: (getter only) currently focused date (if there is any with real focus)
+- **focusDate(date)**: focus on a certain date
+- **focusSelectedDate()**: focus on the current selected date
+- **focusCentralDate()**: focus on the current central date
+- **firstDayOfWeek**: typically Sunday (default) or Monday
+- **weekdayHeaderNotation**: long/short/narrow for the current locale (e.g. Thursday/Thu/T)
+- **locale**: different locale for the current component only
+
+## How to use
+
+### Installation
+
+```sh
+npm i --save @lion/calendar
+```
+
+```js
+import '@lion/calendar/lion-calendar.js';
+```
+
+### Example
+
+```html
+<lion-calendar
+  .minDate="${new Date()}"
+  .maxDate="${new Date('2019/12/09')}"
+  .disableDates=${day => day.getDay() === 6 || day.getDay() === 0}
+>
+</lion-calendar>
+```
--- a/packages/calendar/index.js
+++ b/packages/calendar/index.js
@@ -1,0 +1,1 @@
+export { LionCalendar } from './src/LionCalendar.js';
--- a/packages/calendar/lion-calendar.js
+++ b/packages/calendar/lion-calendar.js
@@ -1,0 +1,3 @@
+import { LionCalendar } from './src/LionCalendar.js';
+
+customElements.define('lion-calendar', LionCalendar);
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -1,0 +1,547 @@
+import { html, LitElement } from '@lion/core';
+import { localize, getWeekdayNames, getMonthNames, LocalizeMixin } from '@lion/localize';
+import { createMultipleMonth } from './utils/createMultipleMonth.js';
+import { dayTemplate } from './utils/dayTemplate.js';
+import { dataTemplate } from './utils/dataTemplate.js';
+import { getFirstDayNextMonth } from './utils/getFirstDayNextMonth.js';
+import { getLastDayPreviousMonth } from './utils/getLastDayPreviousMonth.js';
+import { isSameDate } from './utils/isSameDate.js';
+import { calendarStyle } from './calendarStyle.js';
+import './utils/differentKeyNamesShimIE.js';
+import { createDay } from './utils/createDay.js';
+
+/**
+ * @customElement
+ */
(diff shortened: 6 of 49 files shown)


tlouisse:
Looks really good.

Few questions/suggestions:
- rename 'centralDate' to 'focusableDate', since it's more clear from the name what it does.
- why are 'goToNextMonth' and 'goToPreviousMonth' public methods? Are they supposed to be called by Application Developers? If not, protected would be better

Naming conventions for Subclassers:
- '__renderHeader' instead of '_headerTemplate' or 'headerTemplate'. I would opt for '_headerTemplate' (aligned with existing naming convention, one underscore since available for subclassers)

packages/calendar/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ing-bank/lion.git",
+    "directory": "packages/calendar"
+  },
+  "scripts": {
+    "prepublishOnly": "../../scripts/insert-header.js"
```
daKmoR:
```suggestion
    "prepublishOnly": "../../scripts/npm-prepublish.js"
```

prepublish script changed

erikkroes:
Following is a list of accessibility issues found.

## General

- Screen reader navigation doesn't work in all AT/browser combinations. Adding `role="application"` on the component fixes this issue.
- `aria-selected="true" is not announced. Adding an sr-only text "selected" would fix this.
- Min and max date are not clear for SR users. Suggested fix: if navigating beyond min or max, put focus on min or max. Add sr-only text: "First selectable date" and "Last selectable date" or similar to the min and max date.
- When there are disabled dates, these are skipped by keyboard navigation. This should not be the case as it is navigation for the grid, not just the interactive elements. Preferably, navigation should put the focus for cells with disabled buttons on the cell.

##  VO + Safari

- After navigating to another month, the date gets announced as it should. After a short pause it is followed by "2 <month-name>". The month name is most likely from the `aria-live`, as it should be. The 2 should not be announced.

## NVDA + Firefox

- When navigating to another month, sometimes another day than the navigation target gets pronounced first. This day is in the same row and column as the one you're coming from. As if it changes the month first, with the focus still on the same row and column, before putting the focus on the right cell.
- The string that is the accessible name of the button gets pronounced twice. First as the accessible name of the gridcell, then for the button. This seems to be default browser/AT behaviour, but if it can be removed, that would improve the user experience.

packages/calendar/package.json
bashmish: thanks!

bashmish:
@erikkroes I added `role="application"`. And I agree with all other points as well, but they are probably for later. Some require translations for all currently supported languages, other involve quite some rework and rethinking even (especially the one where we want to enable days for keyboard navigation while disabling the buttons behind them). Let's create individual bug report and feature requests after this is released.

bashmish:
> rename 'centralDate' to 'focusableDate', since it's more clear from the name what it does.

This date has more behavior than just make a day focusable. For example it will change the view if you provide a new value from a different month. Also if I would reverse the meaning of `focusable` applying to other days I can assume that all other days will be disabled, while in fact it does not disable anything, but just gives more priority to one of them. All in all I thing `centralDate` is a good compromise between having a good name with unexpected behavior and a bad name without any expectation (which you can read about in the documentation).

packages/localize/src/date/getMonthNames.js
```diff
@@ -0,0 +1,25 @@
+import { normalizeDate } from './normalizeDate.js';
+
+const monthsLocaleCache = {};
+
+export function getMonthNames({ locale }) {
```
tlouisse: Did we deliberately choose not to make notation here configurable?

bashmish:
> why are 'goToNextMonth' and 'goToPreviousMonth' public methods? Are they supposed to be called by Application Developers? If not, protected would be better

They were protected and used in our extensions. But then I realised that programmatically changing months can be quite useful and their API will likely never change due to simplicity of the interface and unambiguity. And I made them public.

bashmish:
> 'dayPreprocessor' would be called 'dayProcessor' I thought? It would be preprocessing in the sense that it runs before your render, post processing in the sense that it runs after the default processing. Both are in a level of abstraction that the Application Developer should not be bothered with. He only wants to process his dates, so that he can add some extra meta info in his calendar

`pre` gives me a hint that it happens before smth. As a user I can then learn that this happens before the rendering. Does it really matter for the end user? Probably not. But does not make harm either. `processor` is not a perfect name in the first place, but we could find any better :(

tlouisse:
> > why are 'goToNextMonth' and 'goToPreviousMonth' public methods? Are they supposed to be called by Application Developers? If not, protected would be better
> 
> They were protected and used in our extensions. But then I realised that programmatically changing months can be quite useful and their API will likely never change due to simplicity of the interface and unambiguity. And I made them public.

Could make sense indeed, for instance when you create your own calendar overlay and want to add some interaction from outer layer.

tlouisse:
> > 'dayPreprocessor' would be called 'dayProcessor' I thought? It would be preprocessing in the sense that it runs before your render, post processing in the sense that it runs after the default processing. Both are in a level of abstraction that the Application Developer should not be bothered with. He only wants to process his dates, so that he can add some extra meta info in his calendar
> 
> `pre` gives me a hint that it happens before smth. As a user I can then learn that this happens before the rendering. Does it really matter for the end user? Probably not. But does not make harm either. `processor` is not a perfect name in the first place, but we could find any better :(

The less implementation details we expose, the better. Maybe preprocessor and processor are equally vague:)  Can't come up with a better name, though. Good docs and an example would be sufficient, then. Same for 'centralDate' (clear description in static 'get properties')

bashmish:
> '__renderHeader' instead of '_headerTemplate' or 'headerTemplate'. I would opt for '_headerTemplate' (aligned with existing naming convention, one underscore since available for subclassers)

We never had a situation with reusable templates in the past in `lion`. Calendar is the first component where we have both reusable templates (I suggest to call them just `template functions`) which receive everything via arguments and non-reusable templates bound to the context of a web component (I suggest to call them `render functions`).

e.g. `__renderData` is a render function which uses the template `dataTemplate`

https://github.com/ing-bank/lion/pull/37/files#diff-0d8b3c4eac421dbadc591e2ed5b80f6dR280

Therefore I had to introduce a new convention here. And I don't think it is a good idea to follow the previous approach with no separation between those different things. Or it will be like:

```js
 __dataTemplate() {
    return dataTemplate(this.__data, {
      monthsLabels: getMonthNames({ locale: this.__getLocale() }),
      weekdaysShort: getWeekdayNames({
        locale: this.__getLocale(),
        style: this.weekdayHeaderNotation,
        firstDayOfWeek: this.firstDayOfWeek,
      }),
      weekdays: getWeekdayNames({
        locale: this.__getLocale(),
        style: 'long',
        firstDayOfWeek: this.firstDayOfWeek,
      }),
      dayTemplate,
    });
}
```

while we want to have smth like

```js
  _dataTemplate() { // overridable part which has a default implementation
    return dataTemplate;
  }
 __renderData() { // often private, but can be also exposed if more flexibility is needed
    return this.__dataTemplate(this.__data, {
      monthsLabels: getMonthNames({ locale: this.__getLocale() }),
      weekdaysShort: getWeekdayNames({
        locale: this.__getLocale(),
        style: this.weekdayHeaderNotation,
        firstDayOfWeek: this.firstDayOfWeek,
      }),
      weekdays: getWeekdayNames({
        locale: this.__getLocale(),
        style: 'long',
        firstDayOfWeek: this.firstDayOfWeek,
      }),
      dayTemplate,
    });
}
```

PS: Ha-ha, now when I wrote that, feels like we need better terms. Maybe `reusable template` and `bound template` can give more clarity?

tlouisse:
> > '__renderHeader' instead of '_headerTemplate' or 'headerTemplate'. I would opt for '_headerTemplate' (aligned with existing naming convention, one underscore since available for subclassers)
> 
> We never had a situation with reusable templates in the past in `lion`. Calendar is the first component where we have both reusable templates (I suggest to call them just `template functions`) which receive everything via arguments and non-reusable templates bound to the context of a web component (I suggest to call them `render functions`).
> 
> e.g. `__renderData` is a render function which uses the template `dataTemplate`
> 
> https://github.com/ing-bank/lion/pull/37/files#diff-0d8b3c4eac421dbadc591e2ed5b80f6dR280
> 
> Therefore I had to introduce a new convention here. And I don't think it is a good idea to follow the previous approach with no separation between those different things. Or it will be like:
> 
> ```js
>  __dataTemplate() {
>     return dataTemplate(this.__data, {
>       monthsLabels: getMonthNames({ locale: this.__getLocale() }),
>       weekdaysShort: getWeekdayNames({
>         locale: this.__getLocale(),
>         style: this.weekdayHeaderNotation,
>         firstDayOfWeek: this.firstDayOfWeek,
>       }),
>       weekdays: getWeekdayNames({
>         locale: this.__getLocale(),
>         style: 'long',
>         firstDayOfWeek: this.firstDayOfWeek,
>       }),
>       dayTemplate,
>     });
> }
> ```
> 
> while we want to have smth like
> 
> ```js
>   _dataTemplate() { // overridable part which has a default implementation
>     return dataTemplate;
>   }
>  __renderData() { // often private, but can be also exposed if more flexibility is needed
>     return this.__dataTemplate(this.__data, {
>       monthsLabels: getMonthNames({ locale: this.__getLocale() }),
>       weekdaysShort: getWeekdayNames({
>         locale: this.__getLocale(),
>         style: this.weekdayHeaderNotation,
>         firstDayOfWeek: this.firstDayOfWeek,
>       }),
>       weekdays: getWeekdayNames({
>         locale: this.__getLocale(),
>         style: 'long',
>         firstDayOfWeek: this.firstDayOfWeek,
>       }),
>       dayTemplate,
>     });
> }
> ```
> 
> PS: Ha-ha, now when I wrote that, feels like we need better terms. Maybe `reusable template` and `bound template` can give more clarity?

tlouisse:
> We never had a situation with reusable templates in the past in `lion`. Calendar is the first component where we have both reusable templates (I suggest to call them just `template functions`) which receive everything via arguments and non-reusable templates bound to the context of a web component (I suggest to call them `render functions`).

We have all those templates (without underscore) in the field components. They are also part of the input-datepicker now.
Whatever we call them: have a session about them and align on names :)

I will approve the merge request, so that we can continue finishing datepicker.... 👍

bashmish:
@ing-bank/lion can we please make per-line comments to make discussion threads more isolated and manageable?

packages/localize/src/date/getMonthNames.js
bashmish: nope

bashmish: it was just not needed

if you look at the implementation of both new functions, you'll see that we can add other `style` options later without changing the API, so let's keep it open for now

bashmish:
> We have all those templates (without underscore) in the field components. They are also part of the input-datepicker now.
> Whatever we call them: have a session about them and align on names :)

Indeed you are right, we have this distinction already in forms.

https://github.com/ing-bank/lion/blob/master/packages/field/src/FormControlMixin.js#L272-L287

Those are templates in the sense I described them: none of them uses the context. So I think we are on the same page here. Let's just document this convention somewhere.

packages/localize/src/date/getMonthNames.js
bashmish: @tlouisse please create an issue to implement support for other notations of month names

Issue #36, opened 2019-05-13
[overlays] [proposal] improvements 


tlouisse:
While building the datepicker, I encountered a few missing features/shortcomings in the Overlay System:

**- auto a11y for invokers**
The overlay system takes care of all accessibility, the only thing that it doesn't take care of is the attrs for global overlays (aria-haspop="dialog" and aria-expanded="false/true" (should be toggled programatically when global overlay is non-modal)

**- hidesOnOutsideClick support for global overlays.**
A modal overlay shouldn't be closeable, but a global overlay is not neccessarily a modal overlay and should thus support this feature like localOverlay does (example: material design datepicker).

**- No horizontal centering of dialog**
Currently, the global dialog opens to the left of the screen. I think the default should be horizontally centered?

**- No vertical centering of dialog**
Currently, it  opens ‘below the end of the page’. That means there is no relation to invoker element. Think the default should be either a fixed number from top of window or always vertically centered

**- Not closing on esc when clicking on backdrop -> focus on body**
A modal dialog allows to click on backdrop and then body becomes the `document.activeElement`.

**- elementToFocusAfterHide -> doesn’t work**
Didn't check thoroughly, but I think it's not working (not when provided as arg to show method, also not when provided as prop 'elementToFocusAfterHide' to config.)

**- Smarter defaults: always need to add these to my own styling:**
    - Position: relative so that it is positioned above backdrop
    - Background: white (Just like HTML body. You can override to transparent or other color if needed.)

**- Shadow dom encapsulation by default**
The only way to make content put in global overlay not leak styles, is to put it in the web component you created yourself.

Imagine this: 
```js
    const myCtrl = overlays.add(
      new ModalDialogController({
        contentTemplate: () => html`<div class="overlay"><div class="overlay-heading">bla</div></div>`,
      }),
    );
```
Now this overlay should be aware of its context. If it should be used in an app that could be placed in whatever legacy context, it can easily conflict when generic names are used. (a legacy app with `.overlay { background: yellow }` would cause inward leaking. 
Something like this already happened to one of our consumers. 

The current solution, always providing our own web component, requires a lot more boilerplate (it's a bit cumbersome if you want to provide an api for Subclassers in the input-datepicker component). Also, if you provide content projection inside your web component, you would be exposed to global scope again. I think providing one layer of encapsulation above this, would solve this problem.

[...]

PR #38, opened 2019-05-13
feat(overlays): allow adding invoker nodes to GlobalOverlayController

jorenbroekema:
From https://github.com/ing-bank/lion/issues/36

Adds the following features for our users:

- Allow to pass invoker nodes (one or multiple) in the GlobalOverlayController constructor
- Allow to add invoker nodes (one or multiple) at a later stage directly on the controller instance
- In both cases click event listeners and a11y will be automatically set up for you :)!

### DIFF
--- a/packages/overlays/src/GlobalOverlayController.js
+++ b/packages/overlays/src/GlobalOverlayController.js
@@ -40,6 +40,10 @@ export class GlobalOverlayController {
     this._isShown = false;
     this._data = {};
     this._container = null;
+
+    if (finalParams.invokerNodes) {
+      this.addInvokerNodes(finalParams.invokerNodes);
+    }
   }
 
   get isShown() {
@@ -48,7 +52,7 @@ export class GlobalOverlayController {
 
   /**
    * Syncs shown state and data.
-   * @param {object} options optioons to sync
+   * @param {object} options options to sync
    * @param {boolean} [options.isShown] whether the overlay should be shown
    * @param {object} [options.data] data to pass to the content template function
    * @param {HTMLElement} [options.elementToFocusAfterHide] element to return focus when hiding
@@ -78,7 +82,22 @@ export class GlobalOverlayController {
    * Hides the overlay.
    */
   hide() {
+    if (this._expandedInvokerNode) {
+      this._expandedInvokerNode.setAttribute('aria-expanded', 'false');
+    }
     this._createOrUpdateOverlay(false, this._data);
+  }
+
+  addInvokerNodes(nodes) {
+    nodes.forEach(node => {
+      node.addEventListener('click', () => {
+        node.setAttribute('aria-expanded', 'true');
+        this._expandedInvokerNode = node;
+        this.show(node);
+      });
+      node.setAttribute('aria-haspopup', 'dialog');
+      node.setAttribute('aria-expanded', 'false');
+    });
   }
 
   /**
--- a/packages/overlays/stories/global-overlay.stories.js
+++ b/packages/overlays/stories/global-overlay.stories.js
@@ -26,6 +26,35 @@ const globalOverlayDemoStyle = css`
 
 storiesOf('Global Overlay System|Global Overlay', module)
   .add('Default', () => {
+    const buttonEl = document.createElement('button');
+    buttonEl.innerText = 'Open overlay';
+
+    const overlayCtrl = overlays.add(
+      new GlobalOverlayController({
+        contentTemplate: () => html`
+          <div class="demo-overlay">
+            <p>Simple overlay</p>
+            <button @click="${() => overlayCtrl.hide()}">Close</button>
+          </div>
+        `,
+        invokerNodes: [buttonEl],
+      }),
+    );
+    return html`
+      <style>
+        ${globalOverlayDemoStyle}
+      </style>
+      <a href="#">Anchor 1</a>
+      ${buttonEl}
+      <a href="#">Anchor 2</a>
+      ${Array(50).fill(
+        html`
+          <p>Lorem ipsum</p>
+        `,
+      )}
+    `;
+  })
+  .add('Adding invoker node later', () => {
     const overlayCtrl = overlays.add(
       new GlobalOverlayController({
         contentTemplate: () => html`
@@ -41,15 +70,16 @@ storiesOf('Global Overlay System|Global Overlay', module)
       <style>
         ${globalOverlayDemoStyle}
       </style>
-      <a href="#">Anchor 1</a>
       <button
-        @click="${event => overlayCtrl.show(event.target)}"
-        aria-haspopup="dialog"
-        aria-expanded="false"
+        @click=${event => {
+          const buttonEl = document.createElement('button');
+          buttonEl.innerText = 'Open overlay';
+          overlayCtrl.addInvokerNodes([buttonEl]);
+          event.target.insertAdjacentElement('afterend', buttonEl);
+        }}
       >
-        Open overlay
+        Click here to generate an invoker to your overlay
       </button>
-      <a href="#">Anchor 2</a>
       ${Array(50).fill(
         html`
           <p>Lorem ipsum</p>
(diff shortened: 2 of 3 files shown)


packages/overlays/src/GlobalOverlayController.js
```diff
@@ -78,9 +82,24 @@ export class GlobalOverlayController {
...
   }
 
+  addInvokerNodes(nodes) {
+    nodes.forEach(node => {
+      node.addEventListener('click', () => {
+        node.setAttribute('aria-expanded', 'true');
+        this._expandedInvokerNode = node;
+        this.show(node);
+      });
+      node.setAttribute('aria-haspopup', 'dialog');
```
tlouisse: Not sure if this applies to all global overlays. 
It could in theory also be 'listbox', 'grid', 'menu' or 'tree' (https://www.w3.org/TR/wai-aria/#aria-haspopup).
Maybe 'true' would be the safest option. @erikkroes, what do you think?

packages/overlays/src/GlobalOverlayController.js
```diff
@@ -78,9 +82,24 @@ export class GlobalOverlayController {
...
   hide() {
+    if (this._expandedInvokerNode) {
+      this._expandedInvokerNode.setAttribute('aria-expanded', 'false');
+    }
     this._createOrUpdateOverlay(false, this._data);
   }
 
+  addInvokerNodes(nodes) {
+    nodes.forEach(node => {
+      node.addEventListener('click', () => {
```
tlouisse: Here it's really important to also remove the eventListener if ctrl is destroyed/invokerNode is disattached from controller (if it would be a child of overlay it would most probably handled by browser), or else a memory leak is created.

jorenbroekema: I'll add a disconnect method and provide an example + test

packages/overlays/src/GlobalOverlayController.js
```diff
@@ -78,9 +82,24 @@ export class GlobalOverlayController {
...
+        node.setAttribute('aria-expanded', 'true');
+        this._expandedInvokerNode = node;
+        this.show(node);
```
tlouisse: not really sure if it should return focus to this element always. Maybe we should add a guard that checks if elementToFocusAfterHide is also provided and if so, it wins?
This default behavior could stay then, and the check should be added on hide.

packages/overlays/src/GlobalOverlayController.js
daKmoR: the logic should be the other way around... 
on open you should find your invoker (that's why you need to provide the invoker as an argument to open) and add the attribute right?

you never know how the invoker is triggered? could be hover, click, space, problematically...

tlouisse: The argument is the element to focus after hide, right? This usually is, but doesn't have to be, the invoker. Click, space, enter are supported by role=button and all should result in those attrs. Hover usually triggers a Local overlay with role=tool tip. I think it's safe to add this to the global overlay ctrl. (also local, unless its explicitly told to behave as a tooltip)

daKmoR: > This usually is, but doesn't have to be, the invoker.

imho it has to be the invoker or I don't understand this at all 🙈 
even for click, space, enter on a button you will need to provide the button as an argument... 

for which attributes are actually set we can have options (but pls in a later iteration when we add options for backdrop colors and stuff)

jorenbroekema: It seems problematic to have the GlobalOverlayController instance be agnostic of its invokers. Before open, you already want the invokers to have `aria-haspopup='dialog'` or `aria-haspopup='true'`, and also `aria-expanded='false'`. How would you automate this if the GlobalOverlayController has no knowledge of its invoker before being opened?

daKmoR: I think we should encourage providing invokerNodes when creating the overlay (however it should not be a requirement
```js
      const controller = new GlobalOverlayController({
        contentTemplate: () =>
          html`
            <p>Hello, World!</p>
          `,
        invokerNodes: [buttonEl, buttonEl2],
        // will set aria-haspopup="true"
      });
```

then if you open an overlay it should behave like that
```js
it('uses the first invoker by default') {
  controller.open()
  expect(buttonEl.getAttribute('aria-expended')).to.equal('true');
  expect(buttonEl2.getAttribute('aria-expended')).to.equal('false');
  controller.close();
  expect(buttonEl).to.be.focused; // psoido code
}

it('will accept a custom invoker via open') {
  const myExtraButton = document.createElement('button'); 
  controller.open(myExtraButton);
  expect(myExtraButton.getAttribute('aria-expended')).to.equal('true');
  controller.close();
  expect(myExtraButton ).to.be.focused; // psoido code
}

// maybe also add
it('still works without an empty invokers list') {
it('will accept a custom invoker via close') {
```

jorenbroekema:
```js
it('uses the first invoker by default') {
  controller.open()
```
Are you saying to automatically set up aria, but not event handler? leave that part to the user?

jorenbroekema: Maybe you can share how you expect the public API to be when someone sets the invokerNodes in the constructor and then puts the invoker in his lit template?

```js
const buttonEl = document.createElement('button');
buttonEl.innerText = 'Open overlay';

const overlayCtrl = overlays.add(
  new GlobalOverlayController({
    contentTemplate: () => html`
      <div class="demo-overlay">
        <p>Simple overlay</p>
        <button @click="${() => overlayCtrl.hide()}">Close</button>
      </div>
    `,
    invokerNodes: [buttonEl],
  }),
);
buttonEl.addEventListener('click', () => {  overlayCtrl.open() });

return html`
  ${buttonEl}
  <p>Lorem ipsum</p>
`;
```
Like this?

bashmish: @jorenbroekema not super important, but when you have a node reference already, it is more natural to use it instead of `event.target`

```
buttonEl2.addEventListener('hover', event => overlayCtrl.open(buttonEl2));
```

but I don't think this is how people want to create their buttons, people will use LitElement and lit-html and create template with it, for them it will be just an extra step to query select a button node, instead of this we created an API when you pass the invoker to the `open` function and use `event.target` for this right in your lit-html template with  `@click` and `@hover`

do I understand correctly that you want `invokerNodes` to update `aria-expanded` for all invokers (not just the one being used to open) in the case of non-modal overlays?

jorenbroekema: No, only the node that caused the overlay to open should toggle its aria-expanded attr., at least that's how I understand it

tlouisse:
Created an overview of all possible Overlay occurrences to make.
It also addresses the question whether a11y should be handled by the local/globalOverlay and how it affects the automatic addition of click behaviour to the invoker. 
Goal is to make it more clear what is the total end scope that should be handled by our generic layers: https://github.com/ing-bank/lion/blob/feat/richSelect/packages/overlays/docs/Occurrences.md

Feedback welcome ;)

jorenbroekema:
Okay so I read through this again and it seems like there is no agreement as of yet for the exact approach. Currently there seems to be no urgent use case, so I would personally consider this low prio until we need it.

As @tlouisse overlay occurrences document points out, it's not simple. He suggests to set smart defaults depending on the type of overlay, which can be overridden in the options object, here's a snippet of it that I think is particularly relevant to this PR:
```
These options are suggested to be added to the current ones:

- {Boolean} isModal - sets aria-modal and/or aria-hidden="true" on siblings
- {Boolean} isGlobal - determines the connection point in DOM (body vs handled by user) TODO: rename to renderToBody?
- {Boolean} isTooltip - has a totally different interaction - and accessibility pattern from all
other overlays, so needed for internals.
- {Boolean} handlesUserInteraction - sets toggle on click, or hover when `isTooltip`
- {Boolean} handlesAccessibility -
  - For non `isTooltip`:
    - sets aria-expanded="true/false" and aria-haspopup="true" on invokerNode
    - sets aria-controls on invokerNode
    - returns focus to invokerNode on hide
    - sets focus to overlay content(?)
  - For `isTooltip`:
    - sets role="tooltip" and aria-labelledby/aria-describedby on the content
```

gerjanvangeest:
This is currently done in a different way. Closing this one.

PR #43, opened 2019-05-15
exposed property and PropertyDeclarations from lit-element

sebastiangora:
It would be nice to have access to property decorator and PropertyDeclarations object for typing in typescript files

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -30,6 +30,8 @@ export {
   CSSResult,
   defaultConverter,
   notEqual,
+  property,
+  PropertyDeclarations,
 } from 'lit-element';
 // ours
 export { dedupeMixin } from './src/dedupeMixin.js';


daKmoR:
thank you for your contribution ❤️ 

we cherry picked and released it via https://github.com/ing-bank/lion/pull/47

daKmoR:
@sebastiangora just a heads up but we will need to remove the `PropertyDeclarations` export again :see_no_evil:  it is "just" a type and therefore will not be available in "just" javascript

sebastiangora:
ok, so we will use property decorator and remove PropertyDeclarations typings  from our project :)

daKmoR:
for types you can still get them from lit-element directly... however you will always need to make sure you have exactly the same version we have or ignore eslint import for types :grimacing:

Issue #42, opened 2019-05-15
[lion-form] throws error on submitted


daKmoR:
https://stackblitz.com/edit/lion-form-9r5wtq

PR #44, opened 2019-05-15
fix(form): native submit event should not trigger an error

daKmoR:
closes https://github.com/ing-bank/lion/issues/42

### DIFF
--- a/packages/form/src/LionForm.js
+++ b/packages/form/src/LionForm.js
@@ -18,16 +18,22 @@ export class LionForm extends DelegateMixin(LionFieldset) {
     };
   }
 
+  constructor() {
+    super();
+    this.__boundSubmit = this._submit.bind(this);
+    this.__boundReset = this._reset.bind(this);
+  }
+
   connectedCallback() {
     super.connectedCallback();
-    this.addEventListener('submit', this._submit);
-    this.addEventListener('reset', this._reset);
+    this.addEventListener('submit', this.__boundSubmit);
+    this.addEventListener('reset', this.__boundReset);
   }
 
   disconnectedCallback() {
     super.disconnectedCallback();
-    this.removeEventListener('submit', this._submit);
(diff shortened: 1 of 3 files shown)


packages/form/src/LionForm.js
```diff
@@ -18,16 +18,23 @@ export class LionForm extends DelegateMixin(LionFieldset) {
...
+    this.addEventListener('submit', this.__boundSubmit);
+    this.addEventListener('reset', this.__boundReset);
   }
 
   disconnectedCallback() {
     super.disconnectedCallback();
-    this.removeEventListener('submit', this._submit);
-    this.removeEventListener('reset', this._reset);
+    this.removeEventListener('submit', this.__boundSubmit);
+    this.removeEventListener('reset', this.__boundReset);
```
bashmish: let's be more consistent :)

image

packages/form/test/lion-form.test.js
```diff
@@ -37,4 +37,22 @@ describe('<lion-form>', () => {
       firstName: 'Foo',
     });
   });
+
+  it('works with the native submit event (triggered via a button)', async () => {
+    let isSubmitted = false;
+    const mySubmit = () => {
+      isSubmitted = true;
+    };
```
bashmish: maybe use a spy here and `spy.calledOnce` in the end to make test even more readable and prevent double firing of the event on the lion-form (which I expect to happen if not properly handled)

bashmish: I would also double check other tests and if they test this with `calledOnce`, I'm really afraid this MR can cause double submit errors

packages/form/src/LionForm.js
daKmoR: empty line removed

packages/form/test/lion-form.test.js
daKmoR: used sinon now

Issue #39, opened 2019-05-14
[button] tabindex should not be set if already set


LarsDenBakker:
`LionButton` always sets a tabindex. It should not do this if the user already defined a tabindex himself. 

https://github.com/ing-bank/lion/blob/master/packages/button/src/LionButton.js#L149

PR #45, opened 2019-05-16
fix(button): do not override user provided tabindex

daKmoR:
closes https://github.com/ing-bank/lion/issues/39

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -8,6 +8,14 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
         type: Boolean,
         reflect: true,
       },
+      role: {
+        type: String,
+        reflect: true,
+      },
+      tabindex: {
+        type: Number,
+        reflect: true,
+      },
     };
   }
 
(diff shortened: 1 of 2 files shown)


packages/button/src/LionButton.js
```diff
@@ -8,6 +8,14 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
         type: Boolean,
         reflect: true,
       },
+      role: {
```
jorenbroekema: Why do we want to allow our users to override the role of lion-button to be anything other than button?

daKmoR: imho we should provide good defaults... but we should not enforce these defaults... if anyone wants to do something stupid or needs to apply a workaround for something... we should not prevent it...  

in probably 99% of the cases no one will set it themselves... so its a good default and users generally do not need to know/care about it... but if they do they still can

packages/button/src/LionButton.js
```diff
@@ -125,12 +133,13 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
   constructor() {
     super();
     this.disabled = false;
+    this.role = 'button';
+    this.tabindex = 0;
```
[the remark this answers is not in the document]
tlouisse: @dakmor Sorry, forgot to submit my review and now it's already merged... Maybe worth considering to align with platform here?

PR #46, opened 2019-05-16
Use native ES modules in karma

daKmoR:


packages/calendar/test/utils/getFirstDayNextMonth.test.js
```diff
@@ -1,11 +1,17 @@
...
-    expect(formatDate(getFirstDayNextMonth(new Date('2000/03/10')))).to.be.equal('01/04/2000');
+    expect(
+      isSameDate(getFirstDayNextMonth(new Date('2001/01/01')), new Date('2001/02/01')),
+    ).to.equal(true);
+    expect(
+      isSameDate(getFirstDayNextMonth(new Date('2001/10/10')), new Date('2001/11/01')),
+    ).to.equal(true);
+    expect(
+      isSameDate(getFirstDayNextMonth(new Date('2000/03/10')), new Date('2000/04/01')),
+    ).to.equal(true);
```
bashmish: these tests failed because of a different order of execution of tests and because they depended on the page locale

I refactored them to be locale independent, so now they work

packages/calendar/test/utils/getLastDayPreviousMonth.test.js
```diff
@@ -1,11 +1,17 @@
...
-    expect(formatDate(getLastDayPreviousMonth(new Date('2000/03/10')))).to.be.equal('29/02/2000');
+    expect(
+      isSameDate(getLastDayPreviousMonth(new Date('2001/01/01')), new Date('2000/12/31')),
+    ).to.equal(true);
+    expect(
+      isSameDate(getLastDayPreviousMonth(new Date('2001/10/10')), new Date('2001/09/30')),
+    ).to.equal(true);
+    expect(
+      isSameDate(getLastDayPreviousMonth(new Date('2000/03/10')), new Date('2000/02/29')),
+    ).to.equal(true);
```
bashmish: same as above with getFirstDayNextMonth

packages/input-datepicker/test/test-utils.js
```diff
@@ -14,7 +14,14 @@ export class DatepickerInputObject {
...
     this.invokerEl.click();
-    return this.calendarEl ? this.calendarEl.updateComplete : false;
+    const completePromises = [];
+    if (this.overlayEl) {
+      completePromises.push(this.overlayEl.updateComplete);
+    }
+    if (this.calendarEl) {
+      completePromises.push(this.calendarEl.updateComplete);
+    }
+    return Promise.all(completePromises);
```
bashmish: expects were run before the overlay finished first render, this fixes that

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -306,7 +306,10 @@ export class LionInputDatepicker extends LionInputDate {
 
   async __openCalendarOverlay() {
     this._overlayCtrl.show();
-    await this._calendarElement.updateComplete;
+    await Promise.all([
+      this._calendarOverlayElement.updateComplete,
+      this._calendarElement.updateComplete,
+    ]);
```
bashmish: this is needed to fix `this._calendarElement.focusCentralDate();` in `_onCalendarOverlayOpened`, otherwise it is run before the overlay finished first render and the button within the calendar can not be focused yet since it is not yet in the viewport DOM

not sure how it works in the old karma setup, must be smth with transpilation which could have fixed it somehow, but it is good that we now have it working natively

package.json
```diff
@@ -41,7 +41,7 @@
     "test:legacy:watch": "karma start --legacy --auto-watch=true --single-run=false",
     "test:update-snapshots": "karma start --update-snapshots",
     "test:prune-snapshots": "karma start --prune-snapshots",
-    "test:bs": "karma start karma.bs.config.js --legacy --coverage",
+    "test:bs": "karma start karma.bs.config.js --coverage",
```
bashmish: I also had to remove `--legacy` here since it is not compatible with new native ES modules

PR #48, opened 2019-05-17
fix(validate): fix isDateDisabled validator

gerjanvangeest:


daKmoR:
I will merge it later together with a core fix

packages/validate/src/validators.js
```diff
@@ -80,7 +80,7 @@ export const minMaxDateValidator = (...factoryParams) => [
   ...factoryParams,
 ];
 
-export const isDateDisabled = (value, isDisabledFn) => isDisabledFn(value);
+export const isDateDisabled = (value, isDisabledFn) => isDate(value) && !isDisabledFn(value);
```
daKmoR: that value is a function right? so adding `isDate(value)` here will always fail?

gerjanvangeest: No, now it gives an error if the value of the input field is empty. The function you are referring to is the `isDisabledFn`.

PR #51, opened 2019-05-20
fix(field): move `_registerFormElement` to `firstUpdated`

### LINKED ISSUE #50: [field] form elements are not being registered in Edge
For some weird reason, in Edge (IE is fine) parent form elements don't seem to receive the `form-element-register` event and thus doesn't register its child elements. When using derived fieldsets as   `radio-group` some of the methods and properties don't work as expected.

tmsns:
Fixes #50.

Moved the form element registration to the `firstUpdated` event. This allows its parents to receive the event  in Edge.

Not 100% happy with it, as it seems to be a workaround for Edge. I would prefer knowing why Edge is not getting the events.

### DIFF
--- a/packages/field/src/FormControlMixin.js
+++ b/packages/field/src/FormControlMixin.js
@@ -92,12 +92,10 @@ export const FormControlMixin = dedupeMixin(
         super.connectedCallback();
         this._enhanceLightDomClasses();
         this._enhanceLightDomA11y();
-        this._requestParentFormGroupUpdateOfResetModelValue();
-      }
-
-      firstUpdated(changedProperties) {
-        super.firstUpdated(changedProperties);
-        this._registerFormElement();
+        this.updateComplete.then(() => {
+          this._registerFormElement();
+          this._requestParentFormGroupUpdateOfResetModelValue();
+        });
       }
 
       /**
--- a/packages/field/src/FormControlMixin.js
+++ b/packages/field/src/FormControlMixin.js
@@ -92,8 +92,12 @@ export const FormControlMixin = dedupeMixin(
         super.connectedCallback();
         this._enhanceLightDomClasses();
         this._enhanceLightDomA11y();
-        this._registerFormElement();
         this._requestParentFormGroupUpdateOfResetModelValue();
+      }
+
+      firstUpdated(changedProperties) {
+        super.firstUpdated(changedProperties);
+        this._registerFormElement();
       }
 
       /**
--- a/packages/fieldset/test/lion-fieldset.test.js
+++ b/packages/fieldset/test/lion-fieldset.test.js
@@ -150,6 +150,7 @@ describe('<lion-fieldset>', () => {
     expect(Object.keys(fieldset.formElements).length).to.equal(3);
 
     fieldset.appendChild(newField);
+    await fieldset.updateComplete;
     await nextFrame();
     expect(Object.keys(fieldset.formElements).length).to.equal(4);
 
@@ -721,6 +722,9 @@ describe('<lion-fieldset>', () => {
               <!-- group referred by: #msg_l1_g (local) -->
             </lion-fieldset>
           `);
+          const domElement = await dom;
+          await domElement.updateComplete;
+          await domElement.querySelector('[name=l2_g]').updateComplete;
           await nextFrame();
           return dom;
         };


tmsns:
Just rechecked the tests and it seem some fail now. :/

tmsns:
Of course, by having moved the registration away from the connected callback, there is an issue now. When removing and (re)adding an element to its parent, the connected callback is called, but the element is not registered anymore.

I think there are multiple ways now to go on:
- adding elements on both connected and firstUpdated, but we then need to handle multiple registrations (which might not be what we want)
- having some condition on when to execute on firstUpdated registration
- finding out why edge doesn't like to receive the `form-element-register` event

Any pointers, @daKmoR @jorenbroekema?

tlouisse:
> Of course, by having moved the registration away from the connected callback, there is an issue now. When removing and (re)adding an element to its parent, the connected callback is called, but the element is not registered anymore.
> 
> I think there are multiple ways now to go on:
> 
> * adding elements on both connected and firstUpdated, but we then need to handle multiple registrations (which might not be what we want)
> * having some condition on when to execute on firstUpdated registration
> * finding out why edge doesn't like to receive the `form-element-register` event
> 
> Any pointers, @daKmoR @jorenbroekema?

This is probably the reason it doesn't work in Edge without native shadow dom support: https://github.com/Polymer/lit-element/issues/658
The shadydom shim prevents slotted content from sending events before the shadowRoot content is rendered.
After first render (firstUpdated callback) it's possible to send the event.
The connectedCallback issue (for moving nodes around) would probably be solved by this:
```js
connectedCallback() {
  super.connectedCallback();
    // first render happened, now send register event
  });
}

```

tmsns:
> This is probably the reason it doesn't work in Edge without native shadow dom support: [Polymer/lit-element#658](https://github.com/Polymer/lit-element/issues/658)
> The shadydom shim prevents slotted content from sending events before the shadowRoot content is rendered.
> After first render (firstUpdated callback) it's possible to send the event.
> The connectedCallback issue (for moving nodes around) would probably be solved by this:
> 
> ```js
> connectedCallback() {
>   super.connectedCallback();
>   this.updateComplete.then(() => {
>     // first render happened, now send register event
>   });
> }
> ```
This is indeed the underlying issue. It seems that the `requestAnimationFrame` is not working for Edge, while it is working for IE.

I'm trying out your workaround in `connectedCallback`, surrounding `_registerFormElement` with the `updateComplete` promise. An although it works for Edge, the (same) tests are still failing. My first guess for this was that the additional Promise is requiring some more waiting in the tests. I'm able to fix a first test by adding an additional await there.

```javascript
    ...
    fieldset.appendChild(newField);
    await fieldset.updateComplete; // added
    await nextFrame();
    expect(Object.keys(fieldset.formElements).length).to.equal(4);
    ...
```

What do you guys think of this? Adding these two `await`s seems a bit "too much", but it doesn't work otherwise. :-) It also means that a user of the fieldset should wait twice like this when doing some DOM manipulation.

tlouisse:
> An although it works for Edge, the (same) tests are still failing. My first guess for this was that the additional Promise is requiring some more waiting in the tests. I'm able to fix a first test by adding an additional await there.

Is it still needed to wrap register in requestAnimationFrame now?
https://github.com/ing-bank/lion/blob/28590dbf523bf72ed7930eba92e6ab59e70ffdac/packages/field/src/FormControlMixin.js#L150

You could try to delete the rAF and delete `await nextFrame();` from the test (after having tested everything cross browser).

bashmish:
I think some actions should be done only after elements are registered, while others can be done after `updateComplete`. So in the test and in the production code it is OK to wait more if needed for the registration. What becomes a problem is that this moment when all elements are registered is not clear now from the API, so there need to be a hook for that, on the form and the filedset, maybe smth like `await form.formElementsRegistered` and `await fieldset.formElementsRegistered`.

tmsns:
Some updates:

- The tests uncovered some broken functionality: `resetModelValue` was empty. So I've added `_requestParentFormGroupUpdateOfResetModelValue` as well within the promise.
- It seems we can remove `requestAnimationProfile`, indeed. This means that the `await nextFrame()` in some tests can also be removed. When using `fixture()`, the code waits automatically on `updateComplete`. (node_modules/@open-wc/testing-helpers/src/litFixture.js:27)
As there are quite a lot of `await nextFrame` in the tests, I'll need to see which ones are related to the `formElements` to know which ones to remove.
- You have a point @bashmish. With the code I changed the element registration now always runs after `updateComplete`. But it indeed doesn't mean that every registered `updateComplete` would have registered elements. If you register on `updateComplete` before `connectedCallback` the code will not have registered form elements. Your proposed `formElementsRegistered` would fix this (and btw always will run after `updateComplete`). I hope this still makes sense? :see_no_evil:
So, should I add `formElementsRegistered` as well?
- Currently only Chrome is configured with Karma. What do you guys recommend to run the tests cross browser?

tmsns:
Bad news, the `updateComplete` change is breaking IE. (with or without `requestAnimationFrame`)

So, I've got a way to fix it, but it's pretty ugly: replacing the use of `requestAnimationFrame` by `updateComplete` in case of Edge. :/

What do you think?

tlouisse:
> Bad news, the `updateComplete` change is breaking IE. (with or without `requestAnimationFrame`)
> 
> So, I've got a way to fix it, but it's pretty ugly: replacing the use of `requestAnimationFrame` by `updateComplete` in case of Edge. :/
> 
> What do you think?

Not so nice, but needed apparently. Best way is to abstract all the 'ugliness' away in one method that is called from connectedCallback? (like before with rAf, and then add the updateComplete there with a check for Edge)

bashmish:
For the proper abstraction we still need a way to wait till form elements are registered regardless of the browser. Otherwise we will need to at least have double `await` in tests so that they pass in Edge too.

bashmish:
Closing this due to https://github.com/ing-bank/lion/issues/50#issuecomment-495548832

PR #53, opened 2019-05-21
Feat/textarea length indicator

panczer:
New feature for textarea to show text length indicator under textarea input with max value that cannot be exceeded. Pasted text is shortened to max length.


jorenbroekema:
Hey @panczer , thank you for taking the time to contribute! We're happy to see this. We've talked about this feature quite a bit with the team and there's a couple of things:
- Accessibility, it's rough to make it truly accessible, but there is something we can do to make it as good as possible with aria-describedby and aria-live, perhaps @erikkroes can give some details about this
- UX, there are a lot of options UX-wise to display a char counter and how it should behave:
  - Display location, should it be to the right of the label, underneath the textarea, or inside the textarea
  - Display method, e.g. should it be counting up like `${currentLength} / ${maxLength}` as you propose, or should it be like the twitter method, where it starts at max, and counts down. And if you go over the limit, should it be red?
  - Behavior: mostly whether you should be blocked from typing once you hit the maximum length (kinda how native maxlength works). The consensus in our team is that this is generally a bad experience.

So there is quite a lot to be figured out on how to approach this that our team needs to align on.

### So to move forward:
I would suggest to move this behavior and way of displaying to your extension of our textarea for now, I assume you need this for your application.

I will also look into creating a PR for something like polymer's preventInvalidInput which will act like a kind of preprocessor before parsing and formatting, which is useful for filtering out invalid values. This would mean that you could put a preventInvalidInput method override on your textarea (which extends lion textarea) and it will filter out the invalid input before the user can even see it appear on their screen. This would be way easier than doing `lengthChanged()` in `updated()` yourself to filter it out. We see quite a lot of use-cases where exactly this behavior is needed (input-amount filtering out non-digits being another example).

jorenbroekema:
Hey @panczer , as stated in my last message, we will at this moment not commit on the proposed UX for showing a max length indicator and for the blocking input a proposed solution can be found here https://github.com/ing-bank/lion/pull/88. Do you mind if we close this PR?

Issue #22, opened 2019-05-03
[field] values switching from invalid to valid are not formatted (and vice versa)


tlouisse:
## Problem
Fields containing valid modelValues should be run through a formatter, whereas invalid modelValues should not. 
Values switching from invalid to valid are not formatted (after [this pull request](https://github.com/ing-bank/lion/pull/15/files#diff-8a0f76e26a132ed6bb832fc7ff0518c8R273
) is merged they are) and values switching from valid to invalid will be formatted (but should not).

The pr mentioned provides a temporary workaround, but a proper solution is needed.

### Current flow
Setting of modelValue triggers observers for:
 - computation of 'formattedValue'
 - validation (and thus computation of 'errorState')

Observers will be handled in above order. Since computation of formattedValue depends on 'errorState' (erroneous inputs should not be formatted), the formatter is handled with a not yet updated errorState.
 
### Solution
Rewrite to a flow handling observers in this order: `modelValue -> errorState(validation) -> formattedValue`.
Important: in the referenced pr, a situation where a value changes from valid to invalid will still lead to formatting when not required.
Since the flow above involves two mixins (FormatMixin and ValidateMixin) and ValidateMixin should have no knowledge about FormatMixin, think of a proper way to do this.

It might be as easy as changing the order of mixins applied within LionField: if the ValidateMixin runs before the FormatMixin, the errorState should be computed by the time the formatter runs.

bashmish:
Can we close this now after [this fix](https://github.com/ing-bank/lion/pull/84)? @tlouisse @gerjanvangeest

tlouisse:
If you're 100% sure that the issue is gone (or more importantly doesn't have side effects.) Maybe add the tests as Joren suggested, merge everything to master and then close?

bashmish:
@tlouisse it's been merged by @jorenbroekema already and I assumed that [this is the test](https://github.com/ing-bank/lion/pull/84/files#diff-8589890f1ad950710df0434b24582b97R331) Joren was talking about

tlouisse:
Ah, missed that. The test 'title' has a small 'typo', but test looks good, so this issue can be closed

PR #55, opened 2019-05-21
fix(field): store/sync a value instead of delegating

daKmoR:


tlouisse:
This is about the modelValue and value being out of sync, right? (or more precise, LionInput.value and LionInput.$$slot('input').value being out of sync).

Before we do a proper proposal for a rewrite of DelegateMixin, maybe we can do a quick fix like this (pseudo code):

```js
// LionField...

set value(v) {
   this.__value = v;
   // sync with $$slot('input') when available
}

get value() {
   return this.__value;
}
connectedCallback() {
   // for all cases to work, we need to make sure we add logic before the SlotMixin does its thing:
   // Now, listen for slotChanged event and sync the 'cached' this.value to $$slot('input');
   super.connectedCallback();
}
```

A proper fix makes sure delegated values that are in 'external' dom not managed by LionField (in this case the light dom created by SlotMixin are synced whenever  external dom is available for the first time.

daKmoR:
> This is about the modelValue and value being out of sync, right?

yes

> (or more precise, LionInput.value and LionInput.$$slot('input').value being out of sync).

no

e.g. it's about that you set a `modelValue` at one time and the `lion-input` actually has that `modelValue` but it's not visible to the user as the `value` (both for `lion-input` and `input`) is still empty

tlouisse:
> yes
> 
> > (or more precise, LionInput.value and LionInput.$$slot('input').value being out of sync).
> 
> no
> 
> e.g. it's about that you set a `modelValue` at one time and the `lion-input` actually has that `modelValue` but it's not visible to the user as the `value` (both for `lion-input` and `input`) is still empty

True, but if I remember right `modelValue -> reflectbackFormattedValueToUser (sets this.value) -> this.value will sync to $$slot('input').value`.
Only the last step will go wrong I guess

bashmish:
@daKmoR is this related to https://github.com/ing-bank/lion/issues/22 ?

daKmoR:
I guess these are both side effects of the underlying problem of syncing of `lion-input.value` and `nativeInput.value`

tlouisse:
https://github.com/ing-bank/lion/issues/22 is purely about the order of application of mixins. Because validateMixin is applied after FormatMixin, the observed `_modelValueChanged` responds later ->
as a consequence we don't have a correct errorState at the time we want to format.

Seems unrelated to the syncing of the `value` from LionInput to native `<input>`

tlouisse:
About the failing test: it accidentally instantiates `ing-input` instead of `lion-input`. If I change that, the test just passes.
@ruphin do you still have this issue?

tlouisse:
Update: spoke to @ruphin and the bug seems to be the other way around: 
setting `.value` programmatically is not synced to `.modelValue`.

This pr handles:
- fix the test
- move all value 'loop' logic to FormatMixin (it previously was on LionField)
- make value a prop so that it will maintain state on the host instead of on the child




Still TODO: make nice commit history....

packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -55,16 +70,19 @@ describe('<lion-fieldset>', () => {
...
     await nextFrame();
     fieldset.formElements['gender[]'][0].modelValue = { value: 'male' };
     fieldset.formElements['hobbies[]'][0].modelValue = { checked: false, value: 'chess' };
     fieldset.formElements['hobbies[]'][1].modelValue = { checked: false, value: 'rugby' };
 
     expect(Object.keys(fieldset.formElements).length).to.equal(3);
     expect(fieldset.formElements['hobbies[]'].length).to.equal(2);
+
     expect(fieldset.formElements['hobbies[]'][0].modelValue.value).to.equal('chess');
-    expect(fieldset.formElements['gender[]'][0].modelValue.value).to.equal('male');
```
bashmish: why is this removed?

packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -9,19 +9,34 @@ import {
...
+  class extends LionInput {
+    parser() {
+      return this.modelValue;
+    }
+
+    formatter(mv) {
+      return mv && mv.value !== undefined ? mv.value : mv;
+    }
+  },
+);
```
bashmish: please use `defineCE` helper from open-wc to prevent name collisions in our tests

remember that we run all tests (more than 700 now) on the same page without iframes?

packages/field/test/FormatMixin.test.js
```diff
@@ -124,7 +124,12 @@ describe('FormatMixin', () => {
     expect(fooFormat.modelValue).to.equal('string');
   });
 
-  it('synchronizes inputElement.value as a fallback mechanism', async () => {
+  it('converts value to modelValue (for platform api comtatibility)', async () => {
```
bashmish: `comtatibility` => `compatibility`

packages/field/src/LionField.js
```diff
@@ -136,60 +123,24 @@ export class LionField extends FormControlMixin(
...
-    Event Handlers */
-
+  // Needed for choice-inputs and select (IE11) as an alternative to the input event.
+  // TODO: add in FormatMixin and remove here and in lion-select
   _onChange() {
     if (super._onChange) super._onChange();
     this.dispatchEvent(
       new CustomEvent('user-input-changed', {
         bubbles: true,
+        composed: true,
```
bashmish: previously this bubbled, meaning that it will reach all parent elements in the same shadow root where the host (custom input) is located, so for forms and custom forms inputs which are entirely located in the same shadow root (by our convention) it was enough

what changed? do we need some elements extending from LionField to be in a deeper shadow root than the rest of the form? can you show the code you are targeting, like a DOM structure with all elements and shadow roots, roughly?

tlouisse: We fire same event in different places (with composed), there is no reason to do it without in this particular case. Or more generic, there is no reason to be inconsistent throughout our code

packages/fieldset/test/lion-fieldset.test.js
tlouisse: Chances are slim and readability will go down, but will do ;)

packages/field/src/FormatMixin.js
```diff
@@ -226,6 +237,49 @@ export const FormatMixin = dedupeMixin(
         this._calculateValues({ source: 'serialized' });
       }
 
+      _onValueChanged({ value }) {
+        if (super._onValueChanged) super._onValueChanged();
```
jorenbroekema: Use brackets { 

}

Note: should probably set a linting rule to make it consistent, in Poland it was pointed out by someone that we are inconsistent on this.

bashmish: fully agree with having a linter rule active for this

I saw too many times how this ended up with bugs eventually

bashmish: I think we'll work on this later, there is a lot to change here and this MR is not about removing all the technical debt, but rather fixing one specific issue with the logic, so most of the code is just moved from one file to another

packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -6,22 +6,38 @@ import {
...
 import '../lion-fieldset.js';
 
+const complexInputTagString = defineCE(
+  class extends LionInput {
+    parser() {
+      return this.modelValue;
+    }
+
+    formatter(mv) {
+      return mv && mv.value !== undefined ? mv.value : mv;
```
bashmish: this makes all text inputs in tests to behave like special case inputs, and according to @tlouisse this was implemented to fix some tests that contained text inputs with complex model values (❗️) and started failing after the fix

I think this is a symptom of wrong tests in the first place and those wrong tests should have been fixed instead of introducing this special cases for every single test in this file

also I think we have pretty bad practice in this test file where we reuse a template (`inputSlot` defined below) in every test while not every element defined in it is tested, I'd rather have a needed templated defined in each test separately and make tests more isolated, also preventing other issue like the one described earlier in this comment

packages/field/src/FormatMixin.js
```diff
@@ -300,6 +358,18 @@ export const FormatMixin = dedupeMixin(
...
       }
 
+      // Needed for choice-inputs and select (IE11) as an alternative to the input event.
+      // TODO: add in FormatMixin and remove here and in lion-select
+      _onChange() {
+        if (super._onChange) super._onChange();
+        this.dispatchEvent(
+          new CustomEvent('user-input-changed', {
+            bubbles: true,
+            composed: true,
```
bashmish: you didn't only copy, but also added `composed: true`

this is not needed, and adding it for some consistency is a bad idea, because you create an extra event which will bubble outside of shadow root while this is not expected by developers... so this fix to consistency creates additional problems without the real need for it

for now I would just leave it as it was before (without `composed: true`) and in the long run revisit all events we dispatch and double check if they need to bubble and bubble outside of shadow roots or not, and the decision in each case will be individual based on our API needs and users needs

packages/field/src/LionField.js
tlouisse: To simplifiy this merge request, I will put this back in original file (and state). In a later refactor, we can move this logic to one single, appropriate place with shared functionality for ChoiceInput and Select

packages/field/src/FormatMixin.js
tlouisse: Agreed. I would agree with a linting rule for this

packages/fieldset/test/lion-fieldset.test.js
tlouisse: Fair point 👍  I will look at this later again, when I have an overview of the impact of all pending form core updates

packages/field/src/FormatMixin.js
tlouisse: Fair point again.
I will do an attempt to document everything Event related:
- no CustomEvents
- bubble only when needed
- composed only when needed

bashmish: @tlouisse ha-ha, add one more :D https://twitter.com/JonathanAschan/status/1138744064835280897

daKmoR:
ok this needs a better approach... might be superseded by https://github.com/ing-bank/lion/pull/260/commits

PR #56, opened 2019-05-21
Fix date dependent test (fixes #54)

### LINKED ISSUE #54: (calendar) test fails on master branch
One of the test for lion-calendar is failing because it depends on actual day but is compared to constant day.

Test name: `is on day closest to today, if today (and surrounding dates) is/are disabled`

AssertionError: expected 21 to equal 17

bashmish:
fixes #54

### DIFF
--- a/packages/calendar/test/lion-calendar.test.js
+++ b/packages/calendar/test/lion-calendar.test.js
@@ -815,7 +815,10 @@ describe('<lion-calendar>', () => {
 
         it('is on day closest to today, if today (and surrounding dates) is/are disabled', async () => {
           const el = await fixture(html`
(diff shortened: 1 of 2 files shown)


tmsns:
Hey Mikhael, I see sometimes the `centralDate` property, sometimes `useFakeTimers` to mark the current date. Is there some guideline on when to use what? :-)

bashmish:
@tmsns at the time when tests were written, we didn't have it, and later on when it was added we didn't refactor all the tests to use it. Most of the tests that need the calendar to be opened by default on a certain month need to use `centralDate`, and the timer should only be used if you test how calendar behaves for different `today` dates, because the timer effectively changes the `today`.

tmsns:
Great, thanks for the insights!

PR #61, opened 2019-05-22
feat(overlays): add 4 new positions to managePosition util

tlouisse:
Added a merge request on behalf of @jorenbroekema and will drop a review below...


tlouisse:
## Review

Some of those don't concern your specific contribution, but more Overlays in general. Anyways, I reviewed all positioning logic. Let's see if we can pick up the non breaking things. At least have a place to look back at  later :)

### Namings
Having globally(instead of locally) consistent and uniquely identifiable names and definitions for the exact same concepts, highly improves readability when browsing through many methods in different utils.

**_Proposal for dimension/element names:_**
- `pEl`, `rEl`,  `pElBounds`, `rElBounds` (instead of el, relEl, elRect and relRect. 
Names are confusing and similar, especially since 'rect' and 'rel' sound and look quite similar).
- relativeTo -> `rEl` (for consistency)
- positionedHeight basically is just `pElHeight`. This would make it more clear for me.

**_Proposal for placement names:_**
- `const placement` (string): sometimes we call it position, let’s call it always ‘placement’. Also, always be clear about preferred and final/actual placement. 
- `const placement` (object) -> call placementObj or ‘finalPlacementObj’ to not confuse it with string
- returnPosition -> `finalPlacement` ?
- we have a method getPlacement, maybe call it `getFinalPlacement`?

May feel a bit overdone when you are already too much into the code, but for 'first time readers' it will be easier to quickly understand the code and identify relations between different methods.

———————————————————————————————————————————————


### manage-position.js

#### Attrs
- `el.setAttribute('invoker-width’)` etc. are probably not needed. Not nice to create tight couplings with elements. Also, a rEl is not necessarily an invoker.
- Also not sure why `el.setAttribute('position’)` is needed.
Seems like those elementAttributes are placed with a specific context in mind, but this tool should know nothing about its context.
If a dom marker would be really needed, a `data-managed-position` attr or smth would be more appropriate. But I don’t think it’s needed.

#### Perf
For performance it’s better to minimise chances of invalidating styles and layouts. As a best practice:
- Read layout (boundingRect etc) once and put in ref variable
- Write styles in abject first and then apply the whole object on an el.

Since this code is performed synchronously and we are in control of invalidation, it would probably not make a difference right now, but with interfering outside context (like we had in pointing frame) it introduces risks and batching is the better practice.

#### Readability / maintainability
calcAbsolutePosition and calcFixedPosition are very nice, but almost similar. Legend documentation is really good. 
Maybe we can merge them together and put one ‘normalization’ step in-between that:
- Converts offsetTop to top
- Converts offsetHeight to height
Etc etc…
And a nice image on top (in a `manage-position.md`) to make it even more clear: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight

#### other
- For verticalMargin and horizontalMargin, please add that they reflect distance from rEl to el. 

_**Snapping**_
Why ‘(positionedWidth - pc.relRect.width) -(pc.relEl.offsetWidth || 0) -‘   Is only needed for absolute? 
I might need to run it in a debugger and compare to fully understand this part.

**_Nice to have for later_**
- improvement for window resize listeners: delegate to central instance shared by all our components, like a responsiveUtil to mitigate perf probs and align debounces and renders.
-  in disconnect: relativeTo.style.removeProperty('box-sizing') should preferably preserve all initial inline style props and then restore, instead of override and remove
- Make snapping behavior more flexible and allow to go beyond edge when limited viewport space.

———————————————————————————————————————————————

### get-position.md
As a Subclasser, when reading this doc for the first time, it can be a bit confusing to talk about primary/secondary whenever the api says ‘left-of-bottom’. 
Understandable from an Application Developer perspective. though.
First intuition is that 'left' is primary and 'bottom' is secondary. We might want to consider something like Popper.js does, like 'bottom-start' and 'top-end'.

if we stick with current namings, we could maybe add:
```
 From the config, the position argument is evaluated and extracted into a vertical and horizontal argument.
For instance, placement=`top-of-left` would be translated into: 
  - primary: ‘bottom’
  - secondary: ‘top’
  - horizontalIsPrimary: true
Similarly, placement=`left-of-top` would be translated into: 
  - primary: ‘top’
  - secondary: ‘bottom’
  - horizontalIsPrimary: false.
Finally, placement=`center-of-right` would translate to:
  - primary: ‘right’
  - secondary: ‘center’
  - horizontalIsPrimary: true
 When just one argument is given, it is automatically assumed that a center alignment is wanted for that direction.
So, placement=`bottom` would result in:
  - primary: ‘bottom’
  - secondary: ‘center’
  - horizontalIsPrimary: false
```

- Image: make it a link, so relative paths are intercepted by ‘npm prepublish script’
- Small typos -> the flow is _as_ follows

- What still misses in the step is the order in which alternatives will be executed. Also, the fact that when we have ‘horizontalIsPrimary’, we never consider vertical as primary value and vice versa. This is not only relevant for SubClassers, also for Application Developers.

#### nice to haves

- We can add step by step, concrete ‘screenshotted’ examples (like you did before, but now with nice styling:)), to make the flow more clear, without even having to read the code first.

———————————————————————————————————————————————




### Overlays (LocalOverlayController etc...)
- Rename ‘position’ to `positioning-technique` -> it's easily mistaken for placement
- Should we call updatePosition with the object we got back from managePosition? Probably meant to be used like that?

#### nice to haves (for later)

-> get rid of all tech debt (polymer) align with render cycles within  lit etc. boot.js mixing.js seem to be imported (by debounce) but not used for instance. Key-codes is not needed (see calendar approach). All the tab utils could probably be elevated to a utils package, since reusable in more contexts. For now ok, since not exposed in index

tlouisse:
As a specific addition, needed for the dropdown (of the select in this particular case):
We need to be able to specify the `verticalMargin` and `horizontalMargin` on GlobalOverlayController.
Would only be a matter of passing these on to `managePosition`

jorenbroekema:
My thoughts on your review:

## Namings

Yes please, all of it. It really bugs me as well even as a frequent user. I agree with all changes apart from `getPlacement` to `getFinalPlacement`, I think it should be `getPlacementObj` since it returns an object with horizontal (string), vertical (string), horizontalIsPrimary (boolean), if I recall correctly.

## manage-position.js
### Attrs
Sadly, `invoker-width` is needed for calculating the padding in the PointingFrame CE. Once we move the logic for this up to the positioning utils layer, this tight coupling can be much less tight, but for now we are kind of forced to pass the invoker-width to the pEl in case that it is a <pointing-frame> 

Same for `position`, it needs to be passed down to pEl in case that it is a <pointing-frame> because inside the pointing-frame we need to know the position to determine paddings / positionings of the frame and the pointer. It's a property that reflects the attribute, and it causes the rerender when it changes. If position on the lion-popup changes, it trickles down to the pointing-frame through the manage-position util and triggers the pointing-frame rerender. Removing it would break that, currently.

### Perf
I try to apply those notes you have, but if I missed it and it can be improved somewhere it may be worth trying to do that yes. I do think that once we move the logic up, we can improve this more easily.

### Readability / maintainability
Nice suggestion on making absolute / fixed position calcs more concise, I agree.

## get-position.md

I definitely think there's room for improvement on the API. `secondary-of-primary` is the best me and Goffert could think of back then, but splitting it in 2 parameters with -start and -end may be better, and it's also more common I think (thinking of CSS Flex / Grid here). 

For the alternative directions in the flipping behavior, I think this needs a bit of rethink in how it is structured, the logic is not compatible with how we changed it to support 12 positions versus how it was when it only supported 4. 

Currently flipping behavior is very specific and where it will and will not flip to, as considerations. It will flip in its primary direction and secondary direction, separately, but never combined. E.g. right-of-top will never flip to left-of-bottom even if the preferred direction is impossible and left-of-bottom has the most available space. It would flip to either right-of-bottom (primary flip) or center-of-top or left-of-top (secondary flip).

## Overlays
I think `positioning-technique` is quite verbose but I don't have a great alternative. I do agree 'position' is too easily confused with placement in this. Maybe `css-position`. 

All other things that I did not comment on specifically, it probably means that I agree or am indifferent.

Let me know if you want me to make changes or if you want to do them, I won't have time until 31st of May, and after that only on 6th of June due to events :<

tlouisse:
`getPlacementObj` is better than `getPlacement` indeed, the addition of `final` (or `actual`) would make the distinction between input (`placement`) and output (possibly different `placement`) more clear. Calling input `preferred-placement` would work as well, but you want to keep apis concise.

I was already afraid they where needed for pointing frame :) For me an extra indication this shouldn't be a web component. This should never hook into the api of a specific web component, unless explicitly stating so (but we should not want this at all).

Ah, now the term 'flipping' is suddenly confusing for me :)
From `right-of-top` to `right-of-bottom` I would call a flip, from `right-of-top` to `left-of-top` I would call a shift :)
Mainly because primary and secondary get new meanings here.

Let's see who picks up and when. I don't need it right now for the Select dropdown, let's first create an overview for all overlays with:
- what is a task of GlobalOverlayController/LocalOverlayController (invoker coupling, a11, etc?)
- what [occurrences](https://github.com/ing-bank/lion/blob/feat/richSelect/packages/overlays/docs/OverlayOverview.md) of overlays do we build and what should there web component api look like?
- how do the Specific controllers (DropdownController?) fit in here?
- expose managePosition api
- enhance positioning for dropdowns
- create some kind of ResponsiveOverlayController that allows to create dropdowns that switcht to bottom sheets on mobile (and maybe also from local to global)

tlouisse:
@jorenbroekema : closing this one, since we chose Popper in favour of our own solution

PR #67, opened 2019-05-24
[steps] Allow extensions to introduce new non-default slots

maxevilmind:
Previous implementation used .children to get steps which is not going to work in cases when there are elements inside \<lion-steps\> which are not actually \<lion-step\>.

Added getter returns a list of elements inside the default slot which have to be \<lion-step\> or possibly it's extensions.

### DIFF
--- a/packages/steps/src/LionSteps.js
+++ b/packages/steps/src/LionSteps.js
@@ -63,7 +63,7 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
 
   firstUpdated() {
     super.firstUpdated();
-    this._max = this.children.length - 1;
+    this._max = this.steps.length - 1;
   }
 
   next() {
@@ -74,12 +74,17 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
     this._goTo(this.current - 1, this.current);
   }
 
+  get steps() {
+    const defaultSlot = this.shadowRoot.querySelector('slot:not([name])');
+    return defaultSlot.assignedNodes().filter(node => node.nodeType === Node.ELEMENT_NODE);
+  }
+
   _goTo(newCurrent, oldCurrent) {
     if (newCurrent < 0 || newCurrent > this._max) {
       throw new Error(`There is no step at index ${newCurrent}.`);
     }
 
-    const nextStep = this.children[newCurrent];
+    const nextStep = this.steps[newCurrent];
     const back = newCurrent < oldCurrent;
 
     if (nextStep.passesCondition(this.data)) {
@@ -99,8 +104,8 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
   }
 
   _changeStep(newCurrent, oldCurrent) {
-    const oldStepElement = this.children[oldCurrent];
-    const newStepElement = this.children[newCurrent];
+    const oldStepElement = this.steps[oldCurrent];
+    const newStepElement = this.steps[newCurrent];
     const fromStep = { number: oldCurrent, element: oldStepElement };
     const toStep = { number: newCurrent, element: newStepElement };
 
--- a/packages/steps/test/lion-steps.test.js
+++ b/packages/steps/test/lion-steps.test.js
@@ -27,6 +27,21 @@ async function checkWorkflow(steps, expected) {
 }
 
 describe('lion-steps', () => {
+  it('has "steps" getter that returns default slot elements', async () => {
+    const el = await fixture(html`
+      <lion-steps>
+        <lion-step>Step 1</lion-step>
+        <lion-step initial-step>Step 2</lion-step>
+        <other-step-element>Step 3</other-step-element>
+        <steps-extension-element slot="new-slot">e.g. steps indicator</steps-extension-element>
+      </lion-steps>
+    `);
+    expect(el.steps.length).to.equal(3);
+    expect(el.steps[0].tagName).to.equal('LION-STEP');
+    expect(el.steps[1].tagName).to.equal('LION-STEP');
+    expect(el.steps[2].tagName).to.equal('OTHER-STEP-ELEMENT');
+  });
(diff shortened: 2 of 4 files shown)


bashmish:
Hey, thanks for your contribution! I think we gonna merge it, but I would suggest some fixes in the implementation.

But let me first share some knowledge with you. We try to make Lion components as slim and unopinionated as possible up until it makes them reusable and generic enough. In this case that means that we designed `<lion-steps>` to be used only when all the children are `<lion-step>` elements, so this is why we never supported any other elements within `<lion-steps>`.

If you want to integrate `<lion-steps>` with some other presentational component, e.g. show the current step and the number of remaining steps next to it, it is better to do via a sibling of `<lion-steps>`. So the DOM will look like this and the integration can be then done on a different level.

```html
<lion-steps-progress></lion-steps-progress>
<lion-steps>
  <lion-step><lion-step>
  <lion-step><lion-step>
</lion-steps>
```

Now why do I still want to accept you change? Because I can imagine that you might want to extend LionSteps and add support for other features via slots, because then it will be easier to integrate it with the steps controller, e.g.:

```html
<lion-steps-extended>
  <lion-steps-progress slot="progress"></lion-steps-progress>
  <lion-step><lion-step>
  <lion-step><lion-step>
</lion-steps-extended>
```

And I think this also makes sense and falls under our goal to make generic and reusable components.

packages/steps/src/LionSteps.js
```diff
@@ -55,6 +55,12 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
     ];
   }
 
+  get steps() {
+    let steps = this.children;
+    steps = Array.from(steps).filter(step => step.nodeName === 'LION-STEP');
+    return steps;
```
bashmish: taking into account [my explanation here](https://github.com/ing-bank/lion/pull/67#issuecomment-495569853), I would suggest you to change this to using [assignedNodes](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedNodes) on the default slot, because this is a safe assumption that we will have only `<lion-step>` elements in it

packages/steps/src/LionSteps.js
```diff
@@ -56,8 +56,13 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
   }
 
   get steps() {
-    let steps = this.children;
-    steps = Array.from(steps).filter(step => step.nodeName === 'LION-STEP');
+    const slots = this.shadowRoot.querySelectorAll('slot');
+    const slot = Array.from(slots).filter(element => element.name === '')[0];
```
bashmish: I believe you can do it with just one line `const slot = this.shadowRoot.querySelector('slot[name=""]')`

packages/steps/src/LionSteps.js
```diff
@@ -56,8 +56,13 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
...
+    if (!slot) {
+      return [];
+    }
```
bashmish: there is no situation where you don't have a default slot, this is an internal requirement of your own component which is guaranteed, no need to do defensive programming here

packages/steps/src/LionSteps.js
```diff
@@ -56,8 +56,13 @@ export class LionSteps extends ObserverMixin(LionLitElement) {
...
+    let steps = slot.assignedNodes();
+    steps = steps.filter(step => step.nodeType !== 3);
```
bashmish: first of all, it is better to use `Node.TEXT_NODE` instead of a magical number

second, the right logic will be `steps.filter(step => step.nodeType === Node.ELEMENT_NODE);`, because you need only element nodes (you don't need comments, right? so just ignoring text nodes is not enough)

third you are filtering nodes, not steps, while trying to find steps, so better to use `node` as an argument name, because it is more precise `nodes.filter(node => node.nodeType === Node.ELEMENT_NODE);`

and in general it is better not to use `let` unless very necessary, and here it can be just a one liner

I would write the entire getter like this:

```js
  get steps() {
    const slot = this.shadowRoot.querySelector('slot[name=""]');
    return slot.assignedNodes().filter(node => node.nodeType === Node.ELEMENT_NODE);
  }
```

packages/steps/src/LionSteps.js
maxevilmind: changed the whole method implementation to a simplistic one, so it is no longer applicable

packages/steps/src/LionSteps.js
maxevilmind: this is no longer applicable due to the new implementation of the method

packages/steps/src/LionSteps.js
maxevilmind: made it simpler in the new implementation, no longer applicable

packages/steps/test/lion-steps.test.js
```diff
@@ -27,91 +27,101 @@ async function checkWorkflow(steps, expected) {
...
 describe('lion-steps', () => {
+  it('has "steps" getter that returns default slot elements', async () => {
+    const el = await fixture(html`
+      <lion-steps>
+        <lion-step>Step 1</lion-step>
+        <lion-step initial-step>Step 2</lion-step>
+        <other-step-element>Step 3</other-step-element>
+        <steps-extension-element slot="new-slot">e.g. steps indicator</steps-extension-element>
+      </lion-steps>
+    `);
```
bashmish: @maxevilmind this is a new light DOM API which we support now. If you want your new component to be in Light DOM, it must be a different slot. While default slot can be definition only have step elements (`<lion-step>` or it's extensions)

but you also have an option to not use Light DOM of `<lion-steps>` at all and put your extra feature as a sibling of lion-steps (even before this MR is merge)

unfortunately other elements inside the default slot will not be supported

Issue #41, opened 2019-05-15
[localize] not possible to load namespace for a locale


LarsDenBakker:
Localize has a facility to override the locale used for localizing. However, it is not possible to trigger loading a namespace for a specific locale because it is always retrieved from the `this`: https://github.com/ing-bank/lion/blob/master/packages/localize/src/LocalizeManager.js#L66

LarsDenBakker:
This is not entirely correct, it is possible to pass a custom locale to `msg`: https://github.com/ing-bank/lion/blob/master/packages/localize/src/LocalizeManager.js#L83

The global locale should be the default, but there will be cases when a user needs to override this. For example a page might need to display a date or time in different timezones than the locale of the page.

In our specific use case we have a special page in our application where content managers can preview a piece of content they've prepared in multiple languages side by side before publishing it to production. For this we need to localize something in a different locale than the global locale.

bashmish:
In other words we can't really use this feature of the msg without allowing to load data for it. Yep, seems valid to me now. I thought that our LocalizeMixin allows to load namespaces when a component has its own locale (we support this in multiple custom form elements), so we probably have to fix that properly as well. I think we'll start with writing some specs for it to see the entire flow.

PR #69, opened 2019-05-25
Allow custom locale when loading namespaces

LarsDenBakker:
Closes https://github.com/ing-bank/lion/issues/41

### DIFF
--- a/packages/localize/test/LocalizeManager.test.js
+++ b/packages/localize/test/LocalizeManager.test.js
@@ -116,6 +116,26 @@ describe('LocalizeManager', () => {
       });
     });
 
+    it('can load a namespace for a different locale', async () => {
+      setupFakeImport('./my-component/nl-NL.js', { default: { greeting: 'Hello!' } });
+
+      const manager = new LocalizeManager();
+      manager.locale = 'en-US';
+
+      await manager.loadNamespace(
+        {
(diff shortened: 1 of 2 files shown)


packages/localize/test/LocalizeManager.test.js
```diff
@@ -116,6 +116,25 @@ describe('LocalizeManager', () => {
...
+    it('can load a namespace for a different locale', async () => {
+      setupFakeImport('./my-component/nl-NL.js', { default: { greeting: 'Hello!' } });
+
+      const manager = new LocalizeManager();
+
+      await manager.loadNamespace(
+        {
+          'my-component': locale => fakeImport(`./my-component/${locale}.js`),
+        },
+        'nl-NL',
```
bashmish: I would prefer to keep this consistent with `msg` and use an opts object here with `{ locale: 'nl-NL' }`. Not that this will necessarily be needed, but I remember that we wanted to extend options later with extra features and hence made it an object in `msg`, and I'm afraid that people will automatically expect this APIs to be mirrored.

Besides that I like the MR. I will update it accordingly myself and check with my team if they agree with my proposed API change.

LarsDenBakker: Sounds good to me!

Issue #71, opened 2019-05-25
[icon] icon loading system


LarsDenBakker:
I proposed this system already internally, moving it to github now.

The current icon system is a potential footgun for performance. I'd like to propose a change which improves performance and maintains a good developer experience.

In the current system the easiest way to use icons is by importing them synchronously. This method is also the default promoted/documented method. Sync imports will include the icons with the initial bundle of the application which for large applications can really add up to quite an amount.

Applications often have areas which are dynamic where it is not known beforehand what kind of content and icons will be shown. For example a user might be able to choose a profile picture from a set of icons, or the application might display a list of items where each item might have an icon based on some data from an API. In these case the list of possible icons is basically 'everything', resulting in all icons having to be imported and increasing the bundle size.

It is possible to use dynamic imports to load the icons, but this requires quite some extra boilerplate and specific knowledge. Also different teams might work on different areas of the application, and end up rewriting the same dynamic loading logic for the same icons.

Because lazy loading is not the default promoted and documented method most people will end up with the synchronous icon system anyway and (hopefully) only later find out that they need to refactor this to improve performance.

I propose a system similar to the localization system, and similar to the way [iron-icon](https://github.com/PolymerElements/iron-icon) works. Developers use the icon element and specify a specific `icon-id`:
```html
<lion-icon icon-id="lion:space:meteor"></lion-icon>
```

The icon id is made up of a specific set of parts: `<namespace>:<collection>:<icon>`. For each namespace a loader is defined which then lazily loads the icons on demand. For example:
```javascript
icons.addIconResolver('lion', async (collection, iconName) => {
  switch (collection) {
    case 'bugs':
      return (await import('./icons/bugs-collection.js'))[iconName];
    case 'space':
      return (await import('./icons/space-collection.js'))[iconName];
    default:
      throw new Error(`Unknown collection: ${collection}`);
  }
});
```

In the example above a space icon is requested, so the space collection is imported and the meteor icon is returned. When another space icon is requested, the space collection was already imported so it won't be loaded again. It's better to bundle a set of icons together and import it, instead of doing separate imports for each icon.

Benefits:
- Icons are always lazy loaded and never end up in the main bundle.
- Icons are always loaded on demand
  - For example if you display icons in a menu, the menu isn't rendered until it is opened so the icons are not imported until the menu is opened as well.
- There isn't any special syntax to learn, so people get performance for free.
[...]

PR #72, opened 2019-05-25
feat(icon): add icon lazy loading system

LarsDenBakker:
Closes https://github.com/ing-bank/lion/issues/71

I would call `icons` `iconManager` but I thought I'd keep it consistent with the other singletons.

### DIFF
--- a/packages/icon/index.js
+++ b/packages/icon/index.js
@@ -1,1 +1,2 @@
 export { LionIcon } from './src/LionIcon.js';
+export { icons } from './src/icons';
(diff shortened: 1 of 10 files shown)


jorenbroekema:
This is pretty cool because it makes icons "truly" dynamic as compared to what we have now, and it arguably makes the user API better too at least in my opinion.

So first thank you for this contribution! It's a big one. But that also means that we as a team really need to look at this and agree on the approach. We'll definitely get back to you on this, might need to have a session together ;).

LarsDenBakker:
Closing this, see https://github.com/ing-bank/lion/issues/71

PR #74, opened 2019-05-27
Feat/rich select

tlouisse:
WIP...


packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+  it('can preselect an option', async () => {
+    const el = await fixture(html`
+      <lion-select-rich .modelValue="${'nr2'}">
+        <lion-listbox slot="input">
+          <lion-option value="nr1">Item 1</lion-option>
+          <lion-option value="nr2">Item 2</lion-option>
+        </lion-listbox>
+      </lion-select-rich>
+    `);
+    expect(el._listboxElement.value).to.equal('nr2');
```
jorenbroekema: Can you explain why the listboxElement has a `value` property equaling the selected value, instead of just `.modelValue`?

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+  });
+
+  it('shows content of the selected option inside the invoker', async () => {
+    const el = await fixture(html`
+      <lion-select-rich>
+        <lion-listbox slot="input">
```
jorenbroekema: I think calling this slot "input" might lead to our users conflating it with the invoker, because I imagine the invoker looks similar to an input field, and it's where the selected options ends up in.

Suggestion: call it `options`?

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+      it('opens the listbox with [space] key', async () => {
+        const el = await fixture(html`
+          <lion-select-rich></lion-select-rich>
+        `);
+        el.dispatchEvent(new KeyboardEvent('keydown', { key: 'Space' }));
+        await el.updateComplete;
+        expect(el._listboxElement.style.display).to.be.equal('inline-block');
+      });
+
+      it('opens the listbox with [arrow down] key and changes focus to listbox', async () => {
```
jorenbroekema: https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

Here, the focus goes to the second option when doing arrow-down (as the first option is preselected by default). Enter / space puts the focus on the first option.

Do we want similar behavior? Namely, having the first option pre-selected by default? And then I guess listbox should not be focusable.

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+        `);
+        el.click();
+        expect(el._listboxElement.style.display).to.be.equal('inline-block');
+        el.dispatchEvent(new KeyboardEvent('keydown', { key: 'Esc' }));
+        await el.updateComplete;
+        expect(el._listboxElement.style.display).to.be.equal('none');
+        expect(el._invokerElement.focus).to.be.true;
+      });
+
+      it('closes the listbox with [tab] key once opened, focuses on next element', async () => {
```
jorenbroekema: May also want to add a test for [tab] that the listbox will have the last focused option as the selected one as it loses focus and closes. This is what happens at least in the WAI-ARIA

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+  describe('Accessibility', () => {
+    it('has the right references to its inner elements', async () => {
+      const el = await fixture(html`
+        <lion-select-rich></lion-select-rich>
+      `);
+      expect(el.getAttribute('aria-labelledby')).to.equal('ID_REF-label ID_RED-invoker');
+      expect(el.getAttribute('aria-haspopup')).to.equal('listbox');
+    });
+
+    it('notifies when the listbox is expanded or now', async () => {
```
jorenbroekema: `expanded or not`

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -0,0 +1,193 @@
...
+        expect(richSelectEl._listboxElement.style.display).to.be.equal('inline-block');
+        richSelectEl.dispatchEvent(new KeyboardEvent('keydown', { key: 'Tab' }));
+        await richSelectEl.updateComplete;
+        expect(richSelectEl._listboxElement.style.display).to.be.equal('none');
+        expect(nextFocusableEl._invokerElement.focus).to.be.true;
+      });
+    });
+  });
+
+  describe('Accessibility', () => {
```
jorenbroekema: Maybe add a test also for aria-activedescendant if we want that.

packages/select-rich/test/lion-listbox-invoker.test.js
```diff
@@ -0,0 +1,76 @@
...
+    const myNode = document.createElement('div');
+    myNode.textContent = 'foo';
+    el.selectedElement = myNode;
+    await el.updateComplete;
+    expect(el).lightDom.to.equal('foo');
+  });
+
+  it.skip('should render multiple values based on selectedElements', async () => {});
+
+  describe('Subclassers', () => {
```
jorenbroekema: Am I correct to assume this will eventually either be removed from the test (if we decide not to implement a subclass ourselves) or transformed into a multi-select test? For early BDD I guess this makes sense for now :)

packages/listbox/test/lion-listbox.test.js
```diff
@@ -0,0 +1,284 @@
...
+          <lion-option value="ccc">Ccc</lion-option>
+        </lion-listbox>
+      `);
+      el.querySelectorAll('lion-option')[0].focus();
+      el.dispatchEvent(new KeyboardEvent('keydown', { key: 'C' }));
+      await el.updateComplete;
+      expect(el.querySelectorAll('lion-option')[2].focused).to.be.true;
+    });
+
+    it('switch focus to specific element with multiple [character] keys', async () => {
```
jorenbroekema: In my opinion, as soon as it goes further than the 1st character key toggling through options that start with this character, it becomes a search, and should be its own component, probably part of auto-complete. This is also not a feature of the default WAI-ARIA listbox it seems.

Which reminds me, have we thought about how an auto-complete would work structurally? Will it be an extension of LionRichSelect where the invoker is overridden to have some extra behavior? I guess the ListboxBehavior will also be overridden to filter out options that don't match the search?

I know that's not part of the MVP but it makes sense to think about how it would be done before we set in stone the foundations.

packages/select-rich/test/lion-select-rich.test.js
jorenbroekema: Nvm, it's tested in listbox ;)

packages/listbox/lion-optgroup.js
```diff
@@ -0,0 +1,3 @@
+import { LionOptgroup } from './src/option/LionOptgroup.js';
+
+customElements.define('lion-optgroup', LionOptgroup);
```
jorenbroekema: I would prefer `lion-option-group` to be more coherent with `lion-option`. Having 4 fewer chars is not worth the loss of coherence imo

packages/listbox/src/option/LionOption.js
```diff
@@ -0,0 +1,96 @@
...
+      `,
+    ];
+  }
+
+  get value() {
+    return this.__value || this.textContent;
+  }
+
+  set value(v) {
+    this.requestUpdate('value', this.value);
```
jorenbroekema: Shouldn't these two lines be the other way around? Otherwise it requests an update with the old value right?

packages/overlays/test/DropdownConfig.test.js
```diff
@@ -0,0 +1,34 @@
+import { expect } from '@open-wc/testing';
+// import { OverlayController } from '../index.js';
+// import { withDropdownConfig } from '../src/DropdownConfig.js';
+
+const OverlayController = null; // wrapper around Global and LocalOverlayController
+
```
jorenbroekema: Not sure if this really needs its own file? Good idea though to test the config feature and its options all to be working.

packages/select-rich/test/lion-select-rich.test.js
gerjanvangeest: That is sort of already tested here: https://github.com/ing-bank/lion/pull/74/files#diff-3a1ca2a47600fe15d53986d5c9665db3R58, but adding a test like `selects an option on setting focus` there makes sense to me.

packages/listbox/test/lion-listbox.test.js
gerjanvangeest: The WAI-ARIA listbox example has this feature. It is usefull if you have a listbox with country names for instance.

For the autocomplete it will be the same as the `lion-rich-select` only with an input instead of a button. The listbox options will indeed by filtered out dynamically.

packages/select-rich/test/lion-select-rich.test.js
jorenbroekema: Ah yea that should cover it

packages/listbox/test/lion-listbox.test.js
jorenbroekema: Ahh I see now, there's a timeout that resets it to the first char. I think we want that too then, and we should have a test for it :). Nothing more annoying than mistyping the first char and not being able to go back

packages/listbox/lion-optgroup.js
jorenbroekema: Ahhh so the native element is called optgroup, better to stay aligned with that then :see_no_evil:

packages/select-rich/test/lion-select-rich.test.js
tlouisse: Hey Joren, thanks for your questions: forces me to think through everything again and see if we got it right.

I’m writing a bigger explanation(should end up in docs) that addresses all of your questions (and similar questions), but basically it comes down to this:

- .modelValue is only needed on Field layer
- `<lion-listbox>` is equivalent to `<select>` and follows api
- slot="input" is _presentational_ from Application Developer perspective, it is the _interactive element_ from the Subclasser perspective
- combobox with autocomplete functionality has been taken into account, I can send you the design document on Wednesday
- apis are a compromise between accessibility/DX/performance and of course how they fit in the bigger picture (a.k.a. the form system)

For difference between Subclassers and Application Developers:  https://github.com/ing-bank/lion/blob/docs/shadowBem/packages/core/docs/definitions.md

The goal of every form web component we build is to end up as a FormControl. More specific, either as a Field or a Fieldset.
Please see https://github.com/ing-bank/lion/tree/master/packages/field and https://github.com/ing-bank/lion/blob/master/packages/field/docs/CustomFieldsTutorial.md.
Small recap: a FormControl provides a normalized api(including modelValue), validation, formatting, interaction states and an accessible html structure for labels, feedback, help texts and addons. 

Thorough explanation (docs) will follow ;)

tlouisse: This is the more thorough explanation:  https://github.com/ing-bank/lion/blob/feat/richSelect/packages/select-rich/docs/DesignConsiderations.md

Hopefully addresses all questions you (and myself in the beginning, and probably consuming developers) have. Suggestions for naming/api improvements are welcome

packages/select-rich/test/lion-select-rich.test.js
tlouisse: Yeah, though to find the right compromise between AD and SC. Hopefully this docs answers it: https://github.com/ing-bank/lion/blob/feat/richSelect/packages/select-rich/docs/DesignConsiderations.md

packages/select-rich/test/lion-select-rich.test.js
tlouisse: Makes sense I guess. Maybe add an extra it for this?

packages/select-rich/test/lion-listbox-invoker.test.js
tlouisse: It is mainly meant to show the intended use of our lion layer, regardless of multiselect. See https://github.com/ing-bank/lion/blob/feat/richSelect/packages/select-rich/docs/DesignConsiderations.md for more context

packages/overlays/test/DropdownConfig.test.js
tlouisse: Mainly serves as an example of how to implement Responsive Overlays with config objects instead of controllers.
Not finished yet...

packages/listbox/src/option/LionOption.js
tlouisse: It should pass the old value, but it first needs to store it in a var, set the new value and then do requestupdate, so this needs to change indeed

daKmoR:
implemented via https://github.com/ing-bank/lion/pull/156

PR #83, opened 2019-05-31
Add throwing an exception for missing namespace

pavelKozlov:
Add throwing an exception for missing namespace in localization key instead of failing with getting "Cannot call method 'split' of undefined".

### DIFF
--- a/packages/localize/test/LocalizeManager.test.js
+++ b/packages/localize/test/LocalizeManager.test.js
@@ -405,5 +405,14 @@ describe('LocalizeManager', () => {
       manager.addData('en-GB', 'overridden-ns', { greeting: 'Hello!' });
       expect(manager.msg(keys)).to.equal('Hello!');
     });
+
+    it('throws a custom error when namespace prefix is missing', () => {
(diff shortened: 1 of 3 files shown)


packages/localize/test/LocalizeManager.test.js
```diff
@@ -405,5 +405,14 @@ describe('LocalizeManager', () => {
       manager.addData('en-GB', 'overridden-ns', { greeting: 'Hello!' });
       expect(manager.msg(keys)).to.equal('Hello!');
     });
+
+    it('gently handle missing namespace prefix', () => {
```
jorenbroekema: `it('throws a custom error when namespace prefix is missing', () => ` might be more descriptive / accurate

pavelKozlov: Agree. Let me change it.

jorenbroekema:
Hey, I took the liberty of changing it to a single chore commit (since it's not a bug or unexpected behavior), and generating a new yarn.lock since we needed that anyway ;). The contribution is still in your name of course.

Will merge now, thanks!

PR #84, opened 2019-05-31
fix(field): formats value after validation update

gerjanvangeest:
As explained [here](https://github.com/ing-bank/lion/issues/22) changing the order of `ValidationMixin` and `FormatMixin` fixed the problem.

The `InteractionStateMixin` has to be done after the `FormatMixin` otherwise the field(set)s will falsy be set to `dirty`.


jorenbroekema:
Nice, I would add tests that checks that this flow indeed works properly. E.g. create an instance of an input with validation and formatting, set it to invalid, check formattedValue, set it to valid, check it again, etc.

PR #85, opened 2019-06-01
Fix linting scripts

### LINKED ISSUE #31: lint:eclint script not working in Windows
The lint:eclint script currently don't work on windows machines.

The find command and the $ only works in unix environments, so probably it can be simplified and ignore unwanted files in the editorconfig file.

jorenbroekema:
- Fixes linting scripts for windows (fixes #31)
- Fixes a problem with prettier I had where double quotes are needed on Windows, 
- Uses .gitignore now instead of us needing to maintain the glob pattern ourselves :) (though exclude for assets folder was still necessary)
- Checks markdown against .editorconfig as well now.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{html,js}]
+[*.{html,js,md}]
 block_comment_start = /**
 block_comment = *
 block_comment_end = */
--- a/package.json
+++ b/package.json
@@ -43,12 +43,12 @@
     "test:prune-snapshots": "karma start --prune-snapshots",
     "test:bs": "karma start karma.bs.config.js --coverage",
     "lint": "run-p lint:*",
-    "lint:eclint": "eclint check $(find . \\( -name '*.html' -o -name '*.js' -o -name '*.css' \\) -type f -not -path '*/\\.*' -not -path '*node_modules/*' -not -path '*assets/*' -not -path '*coverage/*')",
+    "lint:eclint": "git ls-files | xargs eclint check",
     "lint:eslint": "eslint --ext .js,.html .",
-    "lint:prettier": "prettier '**/*.js' --list-different || (echo '↑↑ these files are not prettier formatted ↑↑' && exit 1)",
+    "lint:prettier": "prettier \"**/*.js\" --list-different || (echo '↑↑ these files are not prettier formatted ↑↑' && exit 1)",
     "format": "npm run format:eslint && npm run format:prettier",
     "format:eslint": "eslint --ext .js,.html . --fix",
-    "format:prettier": "prettier '**/*.js' --write"
+    "format:prettier": "prettier \"**/*.js\" --write"
   },
   "lint-staged": {
     "*": [
--- a/assets/.editorconfig
+++ b/assets/.editorconfig
@@ -1,0 +1,1 @@
+root = true
--- a/packages/field/docs/CustomFieldsTutorial.md
+++ b/packages/field/docs/CustomFieldsTutorial.md
@@ -1,23 +1,23 @@
 # Creating a custom field
-Custom fields can be created in just a few steps. All you need is an interaction element 
+Custom fields can be created in just a few steps. All you need is an interaction element
 (like for instance a slider, a listbox or a combobox) and connect it to the [Field](../README.md)
 functionality.
 
 ## Prerequisite: an interaction element
-An interaction element provides the means for the end user to enter a certain value, just like 
+An interaction element provides the means for the end user to enter a certain value, just like
 native elements provide in this (think of `<input>`, `<textarea>` and `<select>`).
-An example of a non native element is the 
+An example of a non native element is the
 [slider design pattern](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#slider)
 described here.
 
-For this tutorial, we assume we have a component `<my-slider>` that exposes its value via property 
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -43,12 +43,12 @@
     "test:prune-snapshots": "karma start --prune-snapshots",
     "test:bs": "karma start karma.bs.config.js --coverage",
     "lint": "run-p lint:*",
-    "lint:eclint": "eclint check $(find . \\( -name '*.html' -o -name '*.js' -o -name '*.css' \\) -type f -not -path '*/\\.*' -not -path '*node_modules/*' -not -path '*assets/*' -not -path '*coverage/*')",
+    "lint:eclint": "git ls-files | xargs eclint check",
```
bashmish: it does not cover `-not -path '*assets/*' -not -path '*coverage/*'`, and those files we don't want to lint

.editorconfig
```diff
@@ -3,9 +3,11 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
 end_of_line = lf
 insert_final_newline = true
+
+[*.{!md}]
+indent_size = 2
 trim_trailing_whitespace = true
```
bashmish: why do you want these exceptions for Markdown?

also this change is not related to Windows fixes, so I would expect to see them in a different commit with explanation in a commit message somewhere

package.json
jorenbroekema: Ahhh true, I will look into that

.editorconfig
jorenbroekema: Previously we specified the file extensions to run eclint on. In the new script, you don't specify it, so it would check markdown as well and complain quite a bit about it.

In markdown, you don't want these rules:
- If you specify indent_size = 2 for markdown, you won't be able to do codeblocks that have comment blocks inside like this:
```js
/**
 * <-- Indent of 1 instead of 2, eclint will complain
 */
const myJsVar = 'foo';
```
- The reason we don't want to trim trailing whitespace in markdown, I forgot what it was exactly but there was a very specific reason at some point where we needed to have a space and it was auto trimmed wrongly. @daKmoR  I believe you may remember what this was, is this still relevant now?

jorenbroekema: So yeah there is *some* relevance but I do agree having it in two separate commits would be better so I can always change that once we agree how to proceed here

bashmish: first point is applicable to all files containing such multiline commets, this is why we have this exception

```
[*.{html,js}]
block_comment_start = /**
block_comment = *
block_comment_end = */
```

so instead of disabling this rule completely for Markdown, I would just add `.md` to this last config with `html` and `js` extensions

bashmish: regarding the second point: never encountered any need to have explicit space in the end of any line in Markdown, but has encountered hundred times how people accidentally left space or multiple spaces there for no reason, so unless @daKmoR has a good reason, I would rather not do it

at the same time, if we include prettier by default for Markdown, all these problems will be gone, since prettier has very nice formatting rules for Markdown which also fix this sort of problems like trailing space

bashmish: so for now let's focus on Windows fixes and leave Markdown changes for next time (next MR maybe where we can continue discussing how to do it properly?)

package.json
jorenbroekema: Alright, I changed `git ls-files` to exclude the assets folder in the result, so now it does not get checked by eclint.

.editorconfig
jorenbroekema: The thing is, the eclint script changed to check all file extensions instead of just .html, .js and .css

So we can't really split it in another MR because this MR will not pass the build test for eclint without these markdown changes. 

image

jorenbroekema: Myeah, I checked where we have trailing whitespace in md (only 1 file) and it really makes no difference, so I removed it 👍

See latest version of the changes, should address all your concerns, apart from the splitting it up in 2 MRs

package.json
bashmish: did you consider this option? https://github.com/editorconfig/editorconfig/issues/228#issuecomment-163535825

this is a recommened way to exclude folders for eclint by its community https://github.com/jedmao/eclint/issues/80

and we even used this approach in the past internally, for assets as well, before @daKmoR changed it to a shell command alternative

not a big deal for me if it works, just want to make sure this is understandable `git ls-files ':!:assets' | xargs eclint check` and we know how to easily add extra excluded patterns?

jorenbroekema: To exclude more patterns you would just append more ':!:<glob-pattern>' statements and that should work. `git ls-files ':!:assets' ':!:foo'`

I don't mind changing it to excluding folders by placing an editorconfig in the folder but my personal preference is having it more centralized by an exclude pattern in a single config line.
Multiple dummy .editorconfig files with a single line in those folders we need to exclude, it intuitively feels more like a hacky workaround, I'm surprised it's encouraged.

jorenbroekema: Never mind, ':!:assets' like that breaks on Windows. Dummy editorconfig it is 🎉

bashmish:
Exactly the same source code with exactly the same coverage reports.

- passed https://circleci.com/gh/ing-bank/lion/491
- failed https://circleci.com/gh/ing-bank/lion/495

will rerun now again... this MR has bad karma, in the direct and indirect meaning of that word 🤣

bashmish:
One difference I noticed:
- https://circleci.com/gh/ing-bank/lion/491 ✔ 711 tests completed
- https://circleci.com/gh/ing-bank/lion/495 ✔ 764 tests completed

After rerun
- https://circleci.com/gh/ing-bank/lion/496 ✔ 711 tests completed

any clue? (cc @daKmoR)

looks like a glitch in karma runner

anyway, now everything passed, gonna merge now

PR #87, opened 2019-06-04
fix(calendar): init centralDate from selectedDate when today is disabled

bashmish:
The test is failing here https://circleci.com/gh/ing-bank/lion/488 because we have month day 4 disabled in that test and today is June 4, this causes a side effect where `centralDate` is not initialised from `selectedDate`. This side effect is a bug on its own and I'm fixing it here.

### DIFF
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -374,7 +374,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
   }
 
   __disableDatesChanged() {
-    this.__ensureValidCentralDate();
+    if (this.__connectedCallbackDone) {
+      this.__ensureValidCentralDate();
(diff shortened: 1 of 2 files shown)


packages/calendar/src/LionCalendar.js
```diff
@@ -374,7 +374,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
   }
 
   __disableDatesChanged() {
-    this.__ensureValidCentralDate();
+    if (this.__connectedCallbackDone) {
+      this.__ensureValidCentralDate();
```
bashmish: the side effect was caused by initial value of `disableDates` causing this method call and update of `centralDate` before is it even initialised with `selectedDate` value

actually we just need a guard here to not update `centralDate` before full initialisation has finished

PR #88, opened 2019-06-05
feat(field): add preprocessor for blocking invalid input

jorenbroekema:
### Rationale
Have a preprocessor for manipulating or processing user input before it gets reflected back as a view value. The parser is not supposed to change the view-value, it should only parse, hence why the suggested "preprocessor" is its own function. I considered the polymer way (prevent-invalid-input with allowed-pattern) where you give a regex pattern, but this is more flexible and alligned with how we do our formatters, parsers, and serializers.

Useful feature for easily disallowing / blocking the user from invalid input. Saves the user from having to do a bunch of rendering logic to get the same behavior, if we do it under the hood in LionField.

### Use cases
- Max length in textarea:
```js
this.preprocessor = v => v.slice(0, this.maxChars);
```
- Only digits + separators in input-amount:
```js
this.preprocessor = v => (v.match(/[0-9,\ .]+/g) || []).join('');
```
There's probably more.

### Usage / Public API
Main way to use it is by applying FormatMixin (or extending a component which has applied it) and then setting `this.preprocessor` in the constructor.

Currently without extending you can simply do this: 
```js
const lionInput = document.createElement('lion-input');
lionInput.label = 'No vowels allowed';
lionInput.preprocessor = v => (v.match(/[^aeoiuy]+/g) || []).join('');
return html`${lionInput}`;
```
Or this:
```js
<lion-input .preprocessor=${v => (v.match(/[^aeoiuy]+/g) || []).join('')}></lion-input>
```

### TODO:
- [x] Implementation
- [x] Tests
- [x] Apply to input-amount

### Future / other PRs
- Also preprocesses modelValue when imperatively set (In a separate PR https://github.com/ing-bank/lion/pull/55)
- Documentation (In a separate PR for FormatMixin docs https://github.com/ing-bank/lion/pull/149)

### DIFF
--- a/packages/field/src/FormatMixin.js
+++ b/packages/field/src/FormatMixin.js
@@ -133,6 +133,19 @@ export const FormatMixin = dedupeMixin(
       }
 
       /**
+       * === Formatting and parsing ====
+       * To understand all concepts below, please consult the flow diagrams in the documentation.
+       */
+
+      /**
+       * @param {String} value - the raw value from the <input> after keyUp/Down event
+       * @returns {String} preprocessedValue: the result of preprocessing for invalid input
+       */
+      preprocessor(v) {
+        return v;
+      }
+
+      /**
        * Converts formattedValue to modelValue
        * For instance, a localized date to a Date Object
        * @param {String} value - formattedValue: the formatted value inside <input>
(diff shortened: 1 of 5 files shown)


jorenbroekema:
Bunch of tests failing btw due to this, I am aware :P, just making this draft to discuss it. I tested it in storybook and it works there.

jorenbroekema:
WIP due to https://github.com/ing-bank/lion/pull/55 . Once this is fixed, need to ensure (and add tests) that programmatically setting the input value also gets preprocessed.

bashmish:
We'll close this for now. Please talk to @thatianamg about when we gonna pick this up.

jorenbroekema:
Closing, because implementation will be quite different after FormatMixin refactors.

PR #95, opened 2019-06-12
chore: poc for select as radio group

daKmoR:
This is intended to show that a "rich" select is basically the same as the RadioGroup.
We already have all the ways of grouping and setting options/values on children - we should reuse that.
- the `lion-option` would just be a special version of lion-radio (e.g. visually select the full line instead of only the radio)
- doing so should give a lot of a11y out of the box 🤗 
- and with the LocalOverlay and the focustrap it seems to work pretty nicely
- still aria attributes needs to be adjusted
- the wrapping div of the options will need to become a `lion-listbox` (which is only needed for visual and a11y as far as I can see - e.g. no extra logic or grouping)

The multi select would probably be a separate element `lion-select-rich-multiple`
- the `lion-option-multiple` would just be a special version of lion-checkbox
- ...

adding additional features like `lion-option-group` (an extension of LionField to already get the grouping) or `lion-option-separator` should be possible at a later stage as well

### DIFF
--- a/packages/radio-group/stories/index.stories.js
+++ b/packages/radio-group/stories/index.stories.js
@@ -2,13 +2,18 @@ import { storiesOf, html } from '@open-wc/demoing-storybook';
 // just a POC
 // eslint-disable-next-line
 import { overlays, LocalOverlayController } from '@lion/overlays';
+import { LitElement } from '@lion/core';
 
 import '../lion-radio-group.js';
 import '@lion/radio/lion-radio.js';
 import '@lion/form/lion-form.js';
 import { LionRadioGroup } from '../src/LionRadioGroup.js';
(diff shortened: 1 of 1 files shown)


tlouisse:
So, after a discussion I think we agreed that:
- radio, checkbox, option have common parent ChoiceInputMixin
- automatically follows they should support complex values
- listbox parent should be fieldset without role=group

@daKmoR OK if I document the above and close this issue?

daKmoR:
sounds good to me 🤗

tlouisse:
Updated docs here: https://github.com/ing-bank/lion/blob/feat/richSelect/packages/select-rich/docs/DesignConsiderations.md

PR #96, opened 2019-06-12
fix(button): indicate visually the active state on enter/space

bashmish:
This is a custom implementation to make the active state visually noticeable when Enter or Space are used to trigger click on it.

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -87,6 +87,11 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
           fill: whitesmoke;
         }
 
+        :host(:active) .btn,
+        .btn[active] {
+          background: grey;
+        }
+
         :host([disabled]) {
           pointer-events: none;
         }
@@ -140,12 +145,12 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
 
   connectedCallback() {
     super.connectedCallback();
-    this.__setupKeydownDelegation();
+    this.__setupDelegation();
   }
 
   disconnectedCallback() {
     super.disconnectedCallback();
-    this.__teardownKeydownDelegation();
+    this.__teardownDelegation();
   }
 
   __clickDelegationHandler(e) {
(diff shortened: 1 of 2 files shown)


packages/button/test/lion-button.test.js
```diff
@@ -117,6 +117,8 @@ describe('lion-button', () => {
       `);
 
       pressSpace(form.querySelector('lion-button'));
+      await aTimeout();
+      await aTimeout();
```
bashmish: Although this is not nice, with the current implementation of helpers from `@polymer/iron-test-helpers/mock-interactions.js` there is no other way to determine the moment when `pressSpace/pressEnter` finish their job.

packages/button/src/LionButton.js
```diff
@@ -140,32 +145,42 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
...
-  __teardownKeydownDelegation() {
+  __teardownDelegation() {
     this.removeEventListener('keydown', this.__keydownDelegationHandler);
+    this.removeEventListener('keyup', this.__keyupDelegationHandler);
   }
 
   __keydownDelegationHandler(e) {
+    if (e.keyCode === 32 /* space */ || e.keyCode === 13 /* enter */) {
+      e.preventDefault();
+      this.shadowRoot.querySelector('.btn').classList.add('active');
```
bashmish: I could of course cache `this.shadowRoot.querySelector('.btn')`, but decided not to since this is a super minor pre optimisation for this particular case and will only make code more complex.

packages/button/src/LionButton.js
```diff
@@ -140,32 +145,42 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
...
+    }
+  }
+
+  __keyupDelegationHandler(e) {
     // Makes the real button the trigger in forms (will submit form, as opposed to paper-button)
     // and make click handlers on button work on space and enter
     if (e.keyCode === 32 /* space */ || e.keyCode === 13 /* enter */) {
       e.preventDefault();
+      this.shadowRoot.querySelector('.btn').classList.remove('active');
```
bashmish: One of the most interesting findings during my work on this is that we can have this simple logic of removing `.active` from `.btn` without adding extra logic for corner cases when `keyup` does not happen on `lion-button`. e.g. if someone keeps pressing Enter (keydown) and then clicks somewhere, so this logic of removal will not be run and the `lion-button` will be stuck in the active state. Why? Because this is exactly how native button works with `:active`, it also remembers the active state in this situation.

So in both native and this custom implementation this can be undone only after another focus on the button and another sequence of keydown/keyup. So we are aligned with the platform here. Surprising, but yeah :) Platform is sometimes pretty surprising.

(I didn't do extensive crossbrowser testing of how native button behaves though, but I also know that it often behaves differently on different browsers, so consistency is almost impossible to achieve anyway)

packages/button/src/LionButton.js
tlouisse: Shouldn't it be a modifier (attribute) instead of a regular class?

packages/button/src/LionButton.js
```diff
@@ -87,6 +87,11 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
           fill: whitesmoke;
         }
 
+        :host(:active) .btn,
```
tlouisse: If  understand right, this line will only apply to clicks on the host, the one beneath will apply to enter/click/space?

packages/button/src/LionButton.js
tlouisse: Interesting. Almost sounds like a feature. Although would be interesting to see what other browsers do, since it wouldn't be too hard for us to fix this, I would say, platform bug

packages/button/src/LionButton.js
bashmish: This line applies to all clicks, the second line applies to Enter/Space.

packages/button/src/LionButton.js
bashmish: Let's please keep it for the other MR, because fixing browser inconsistencies is an issue on its own which we are not even sure about.

packages/button/src/LionButton.js
bashmish: This was impossible to customize active style before, still impossible now. We can investigate this, but it is not part of this bugfix MR.

packages/button/src/LionButton.js
tlouisse: would be nicer to have all styles combined in one prop from Subclasser perspective (click interception). We can consider this once we would apply those modifiers on the host I think.

packages/button/src/LionButton.js
tlouisse: For next button update: from Subclasser perspective (who theoretically should be able to override css) would be better to refer to unique identifiers instead of classes

PR #98, opened 2019-06-13
Support en-PH locale for polymer-cli

carlocanlas:
Updated switch cases.

### DIFF
--- a/packages/validate/src/ValidateMixin.js
+++ b/packages/validate/src/ValidateMixin.js
@@ -50,6 +50,7 @@ export const ValidateMixin = dedupeMixin(
                   return import('../translations/en-GB.js');
                 case 'en-US':
                   return import('../translations/en-US.js');
+                case 'en-PH':
                 case 'en':
                   return import('../translations/en.js');
                 case 'es-ES':
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -31,6 +31,7 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
               return import('../translations/de.js');
             case 'en-AU':
             case 'en-GB':
+            case 'en-PH':
             case 'en-US':
             case 'en':
               return import('../translations/en.js');
--- a/packages/input-datepicker/src/LionCalendarOverlayFrame.js
+++ b/packages/input-datepicker/src/LionCalendarOverlayFrame.js
@@ -59,6 +59,7 @@ export class LionCalendarOverlayFrame extends LocalizeMixin(DomHelpersMixin(LitE
               return import('@lion/overlays/translations/en-GB.js');
             case 'en-US':
               return import('@lion/overlays/translations/en-US.js');
+            case 'en-PH':
             case 'en':
               return import('@lion/overlays/translations/en.js');
             case 'es-ES':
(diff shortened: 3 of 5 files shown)


packages/validate/translations/en-PH.js
```diff
@@ -0,0 +1,11 @@
+import en from './en.js';
+
+export default {
+  ...en,
+  error: {
+    isDate: 'Please enter a valid date (MM/DD/YYYY).',
+  },
+  warning: {
+    isDate: 'Please enter a valid date (MM/DD/YYYY).',
```
bashmish: to clarify: is this a standard date format everywhere in Philippines?

packages/validate/src/ValidateMixin.js
```diff
@@ -48,6 +48,8 @@ export const ValidateMixin = dedupeMixin(
                   return import('../translations/en-AU.js');
                 case 'en-GB':
                   return import('../translations/en-GB.js');
+                case 'en-PH':
```
bashmish: you need to add a similar case to each component that has such a switch statement for the compatibility with `polymer-cli`, to find all the places in our code you can use this: https://github.com/ing-bank/lion/search?q=%22case+%27en%27%22&unscoped_q=%22case+%27en%27%22

you don't need to create a file everythere btw unless you know it has slightly different translations, so often you can just fallback to `en.js`, e.g.

```
case 'en-PH':
case 'en':	
  return import('../translations/en.js');
```

packages/validate/translations/en-PH.js
carlocanlas: Yes, for short format dates like in the datepicker input

packages/validate/src/ValidateMixin.js
carlocanlas: Done updating the files

packages/validate/translations/en-PH.js
bashmish: `Intl.DateTimeFormat` enforces a different standard

```js
new Intl.DateTimeFormat('en-PH').format(new Date('2019/06/30'))
// "30/06/2019"
new Intl.DateTimeFormat('en-GB').format(new Date('2019/06/30'))
// "30/06/2019"
new Intl.DateTimeFormat('en-US').format(new Date('2019/06/30'))
// "6/30/2019"
```

I'm not sure if we can or should deviate from it

bashmish:
@carlocanlas can you please also sign the CLA, there is a comment above about this: https://github.com/ing-bank/lion/pull/98#issuecomment-501572002

packages/validate/translations/en-PH.js
bashmish: so the rest of this MR is OK, but this is not good, in fact you'll have same problem we discussed previously when user will enter date in the format of `MM/DD/YYYY` in the `lion-date`, but because of how Intl works the date will be invalid, because for `en-PH` Intl would expect to see `DD/MM/YYYY`

bashmish: looks like you have 2 different format in the country, more context here: https://en.wikipedia.org/wiki/Date_and_time_notation_in_the_Philippines

and they are predefined in locales, e.g.:
- http://www.localeplanet.com/icu/en-PH/index.html => Pattern: dd/MM/y
- http://www.localeplanet.com/icu/fil-PH/index.html => Pattern: M/d/yy

so with our current design you can't have format `M/d/yy` if you want to use English (en-PH)

carlocanlas: Noted on this, and apologies on our part as we've overlooked such resources. We'll be sticking to DD/MM/YYYY.

PR #99, opened 2019-06-13
fix(textarea): if rows is defined, set height accordingly

danychi:
the prop `rows` was not setting the height for the textarea, introduced a simple fix

### DIFF
--- a/packages/textarea/src/LionTextarea.js
+++ b/packages/textarea/src/LionTextarea.js
@@ -72,6 +72,10 @@ export class LionTextarea extends ObserverMixin(LionInput) {
     const offset = cs.boxSizing === 'border-box' ? paddingOffset + borderOffset : 0;
 
     this.inputElement.style.maxHeight = `${lineHeight * this.maxRows + offset}px`;
+
+    if (this.rows) {
+      this.inputElement.style.height = `${lineHeight * this.rows + offset}px`;
+    }
   }
 
   static get styles() {


packages/textarea/src/LionTextarea.js
```diff
@@ -72,6 +72,10 @@ export class LionTextarea extends ObserverMixin(LionInput) {
     const offset = cs.boxSizing === 'border-box' ? paddingOffset + borderOffset : 0;
 
     this.inputElement.style.maxHeight = `${lineHeight * this.maxRows + offset}px`;
+
+    if (this.rows) {
+      this.inputElement.style.height = `${lineHeight * this.rows + offset}px`;
+    }
```
bashmish: I might be mistaken, but I remember that was our conscious choice to not support`rows` and expose only `min/maxLength` as a minimal sufficient API, right @tlouisse @daKmoR ?

if you wan to have a proper height, please try setting `minLength` and `maxLength` to the same value and share if that solves your use case?

danychi: It seems like `minLength` doesn't do anything

bashmish: sorry, I was mistaken :( even twice: not `min/maxLength` of course, but rather `rows` and `maxRows`

so you need smth like 

```
<lion-textarea .rows="${3}" .maxRows="${3}"></lion-textarea>
```

or with attributes (mind the different notation of maxRows => max-rows)

```
<lion-textarea rows="3" max-rows="3"></lion-textarea>
```

here is a working example https://stackblitz.com/edit/js-425m9h?file=index.js

does this fix your problem with height?

danychi: This does, we can close the issue

bashmish:
Closing since there is an alternative that works as expected. Example: https://stackblitz.com/edit/js-425m9h?file=index.js

PR #101, opened 2019-06-14
Popper.js implementation in LocalOverlay

jorenbroekema:
Showcasing how Popper.js could work inside localoverlay

WIP:
- [x] Add lazy load behavior in localoverlay
- [x] Add demos to showcase some interesting features
- [x] Create issue for overflow feature when no space is available. Need to investigate how.. Maybe it's a popper feature already, maybe we need to make a custom modifier for it, etc.
- [x] Cleanup


packages/overlays/src/LocalOverlayController.js
```diff
@@ -128,10 +126,15 @@ export class LocalOverlayController {
...
   }
 
-  _createOrUpdateOverlay(shown = this._prevShown, data = this._prevData) {
+  async _createOrUpdateOverlay(shown = this._prevShown, data = this._prevData) {
     if (shown) {
       this._contentData = { ...this._contentData, ...data };
 
+      // preload popper before render, so the user does not see a different view between render and popper being applied
+      if (!this.popperModule) {
+        await this.__preloadPopper();
```
tlouisse: Isn't it better to lazyload then do a loading on demand?
The first popper on the page will have a huge delay with 6kb to be loaded and processed first

packages/overlays/src/LocalOverlayController.js
```diff
@@ -128,10 +126,15 @@ export class LocalOverlayController {
...
+      }
```
bashmish: my expectation was that the async nature of positioning does not impact the sync nature of the interface

or to be more precisely: we can tolerate the fact that the overlay still has wrong position while all the other things behave correctly, like you can pass and render data, A11Y attributes are all set, etc...

now it feels like that this is not correct always, e.g. line 370 in this test should not be needed

https://github.com/ing-bank/lion/blob/04af7fdd7506c2a6a9af3efb6876179bc0b0d3cd/packages/overlays/test/LocalOverlayController.test.js#L366-L373

in general I would expect only the tests that check the position to wait, and all other to continue to work synchronously

jorenbroekema: Ah it makes more sense now reading this. You're absolutely right I think, just because it hasn't positioned correctly yet, doesn't mean we cannot already test non-positioning related stuff. I have removed unnecessary awaits from the test, let me know if it's more like you would expect now

packages/overlays/src/LocalOverlayController.js
jorenbroekema: Now lazy-loads in constructor ;), should make it more smooth when the user clicks it with shitty 3g network because hopefully it start preloading when the LocalOverlayController was instantiated

packages/overlays/package.json
```diff
@@ -32,7 +32,8 @@
     "*.js"
   ],
   "dependencies": {
-    "@lion/core": "^0.1.9"
+    "@lion/core": "^0.1.5",
+    "popper.js": "^1.15.0"
```
tlouisse: After seeing the presentation you sent about security (https://www.youtube.com/watch?v=2cyib2MgvdM), maybe it's better to have only fixed versions for external dependencies we don't control?

packages/overlays/src/LocalOverlayController.js
```diff
@@ -1,36 +1,77 @@
...
+        ...finalModifiers,
+        ...params.placementConfig.modifiers,
+      };
+    }
+
+    this.placementConfig = {
+      placement: finalPlacement.placement,
+      positionFixed: finalPlacement.positionFixed,
+      modifiers: finalModifiers,
+    };
```
tlouisse: This is quite hard to read for me :) Can we replace all of the above with this?

```js
    this.placementConfig = {
      placement: 'top',
      positionFixed: false,
      ...(params.placementConfig  || {}),
      modifiers: {
        keepTogether: {
          enabled: false,
        },
        preventOverflow: {
          enabled: true,
          boundariesElement: 'viewport',
          padding: 16, // viewport-margin for shifting/sliding
        },
        flip: {
          boundariesElement: 'viewport',
          padding: 16, // viewport-margin for flipping
        },
        offset: {
          enabled: true,
          offset: `0, 8px`, // horizontal and vertical margin (distance between popper and referenceElement)
        },
        arrow: {
          enabled: false,
        },
        ...(params.placementConfig && params.placementConfig.modifiers  || {}),
      },
    };
```

packages/overlays/src/LocalOverlayController.js
```diff
@@ -1,36 +1,77 @@
...
+    this.hidesOnEsc = params.hidesOnEsc;
+    this.hidesOnOutsideClick = params.hidesOnOutsideClick;
+    this.trapsKeyboardFocus = params.trapsKeyboardFocus;
+
     /**
      * A wrapper to render into the invokerTemplate
      *
      * @property {HTMLElement}
      */
     this.invoker = document.createElement('div');
```
tlouisse: I know it's not part of your merge request, but this thing probably also needs a rewrite or at least some docs

packages/overlays/src/LocalOverlayController.js
```diff
@@ -1,36 +1,77 @@
...
+      };
+    }
+
+    this.placementConfig = {
+      placement: finalPlacement.placement,
+      positionFixed: finalPlacement.positionFixed,
+      modifiers: finalModifiers,
+    };
+
+    this.__preloadPopper();
```
LarsDenBakker: I don't think it's a good idea to eagerly preload like this, there is no guarantee this will be run after the application has fully loaded so you're still blocking the initial load. It's better to let applications control this behavior, perhaps with some central manager within lion which can do other preloading too.

tlouisse: It's indeed a good idea to get more fine grained control, so we can allow the app developer to decide whether it should compete with other lazily loaded resources.
I assume the first meaningful paint depends on sync resources only or not @LarsDenBakker ?

LarsDenBakker: Depends on the definition of meaningful. A banking application is useless until you fetch the account data.

Unfortunately there is no programmatic hook in the platform for determining these landmarks, though I think they will add these things over time.

We could also considering using a prefetch link instead of importing it from the constructor. Prefetch tells the browser you will use a resource at some point, and the browser will fetch it when it's not busy with other things. It will not parse/run the javascript, so it's never blocking the main thread.

jorenbroekema: Good point, but then we do indeed need this generic lion-preloader manager for resource control before I can remove it here.

I'll look into prefetch though as an improvement for the current situation :)

packages/overlays/src/LocalOverlayController.js
jorenbroekema: Whoa this is some Javascript spread syntax magic I was not aware of, works fine and is indeed much more readable

packages/overlays/src/LocalOverlayController.js
```diff
@@ -71,8 +99,8 @@ export class LocalOverlayController {
    * @param {boolean} [options.isShown] whether the overlay should be shown
    * @param {object} [options.data] overlay data to pass to the content template function
    */
-  sync({ isShown, data } = {}) {
-    this._createOrUpdateOverlay(isShown, data);
+  async sync({ isShown, data } = {}) {
```
jorenbroekema: So yeah, we have "async sync" now 🙈  but it seems necessary to allow user to do a sync where they want to make sure that positioning (using popper, which is async since loading the module is lazy loaded) has finished updating...

packages/overlays/src/LocalOverlayController.js
```diff
@@ -113,7 +141,22 @@ export class LocalOverlayController {
     this.isShown ? this.hide() : this.show();
   }
 
-  _createOrUpdateOverlay(shown = this._prevShown, data = this._prevData) {
+  // Popper does not export a nice method to update an existing instance with a new config. Therefore we recreate the instance.
+  // TODO: Send a merge request to Popper to abstract their logic in the constructor to an exposed method which takes in the user config.
+  updatePlacementConfig(config = this.placementConfig) {
+    this.placementConfig = {
+      ...this.placementConfig,
+      ...config,
```
daKmoR: this will override all modifiers right?

how about moving the logic from the constructor to here and then call this method in the constructor?

if needed there could be a flag to not do the create/destroy?

packages/overlays/package.json
tlouisse: Doesn't outweigh importance of  preventing potential versio  conflicts with other deps loaded by App developer

packages/overlays/src/LocalOverlayController.js
jorenbroekema: Done, but it always recreates popper instance upon updating config now. Whether the user sees the difference depends on whether it is shown or not. I did not see a use case where you want to update the placement config but not update the popper instance with it.

packages/overlays/src/LocalOverlayController.js
```diff
@@ -214,4 +277,15 @@ export class LocalOverlayController {
...
   }
+
+  __createPopperInstance() {
+    const Popper = this.__popperModule.default;
+    this._popper = new Popper(this.invokerNode, this.contentNode, {
+      ...this.placementConfig,
+    });
+  }
+
+  async __preloadPopper() {
```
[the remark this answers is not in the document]
jorenbroekema: Done!

packages/overlays/src/LocalOverlayController.js
jorenbroekema: For now we will merge it as is, but we'll make sure to look into smarter way to fetch/load this (preloader manager, prefetch, etc.)

LarsDenBakker: Then I think you shouldn't preload it in the constructor, and only on use for now. otherwise you're introducing an extra load to the inital load of apps with this.

tlouisse: Loading 6k on demand on a slow mobile connection? That would lead to  bad UX when you expect something to happen on a button press. 
Wouldn't it be possible to lazy load the LocalOverlay in your app for now?

LarsDenBakker: So will loading extra 6k on startup, right? I can't control how everyone uses the local overlay, so I can't lazy load it in that way.

tlouisse: By the time the users presses the invoker button, the 6k should have loaded, so the user experience is just smooth there. 

For now we could say responsibility of every individual feature to be performant. 

Or maybe you could give an example for prefetching dynamic imports?
This shows how to do it for webpack builds at least:
https://medium.com/front-end-weekly/webpack-and-dynamic-imports-doing-it-right-72549ff49234
But that's not what you had in mind?

packages/popup/test/lion-popup.test.js
```diff
@@ -44,6 +45,21 @@ describe('lion-popup', () => {
...
+        <lion-popup>
+          <div slot="content" class="popup">Hey there</div>
+          <lion-button slot="invoker">Popup button</lion-button>
+        </lion-popup>
+      `);
+      await el._controller.show();
+      expect(el._controller._popper.options.placement).to.equal('top');
+      el.placementConfig = { placement: 'left' };
+      // placementConfig setter calls updateConfig method which recreates popper instance and this process is async..
+      await aTimeout();
```
bashmish: isn't it better to hide/show and await for it instead of a hacky timeout?

packages/tooltip/test/lion-tooltip.test.js
```diff
@@ -25,6 +25,7 @@ describe('lion-tooltip', () => {
       const eventMouseEnter = new Event('mouseenter');
       invoker.dispatchEvent(eventMouseEnter);
       await el.updateComplete;
+      await aTimeout();
```
bashmish: this is very brittle implementation IMO

we have to find another way to wait till the work is done

packages/popup/src/LionPopup.js
```diff
@@ -4,29 +4,44 @@ import { overlays, LocalOverlayController } from '@lion/overlays';
 export class LionPopup extends UpdatingElement {
   static get properties() {
     return {
-      position: {
-        type: String,
+      placementConfig: {
+        type: Object,
```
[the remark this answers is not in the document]
jorenbroekema: Since lion is still v0, we will do a minor release (v0.3.0 if I'm not mistaken).

packages/popup/src/LionPopup.js
```diff
@@ -4,29 +4,44 @@ import { overlays, LocalOverlayController } from '@lion/overlays';
...
+  set placementConfig(config) {
+    this._placementConfig = {
+      ...this._placementConfig,
+      ...config,
+    };
+
+    if (this._controller && this._controller._popper) {
+      this._controller.updatePlacementConfig(this._placementConfig);
+    }
+  }
```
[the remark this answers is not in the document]
jorenbroekema: Hmm actually that's kinda true and also there's nothing to rerender because there is no render() in this class. So perhaps we can and should remove it then :\? At the same time, for clarity's sake, it's nice that it conventionally shows the exposed property as a "user endpoint" like that.

packages/popup/stories/index.stories.js
```diff
@@ -40,6 +40,8 @@ const popupDemoStyle = css`
     background-color: black;
     border-radius: 4px;
     padding: 8px;
+    /* To display on top of elements with no z-index that are appear later in the DOM */
+    z-index: 1;
```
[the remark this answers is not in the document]
jorenbroekema: Hmmm yeah I think you're right actually. Definitely would consider this a "smart default" that covers most use cases, will change

packages/popup/src/LionPopup.js
```diff
@@ -4,29 +4,44 @@ import { overlays, LocalOverlayController } from '@lion/overlays';
 export class LionPopup extends UpdatingElement {
   static get properties() {
     return {
-      position: {
-        type: String,
+      placementConfig: {
```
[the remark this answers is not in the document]
jorenbroekema: I think the only purpose of appending "Config" is to make clear that it's a (potentially large) configuration object and not just a simple primitive. Having said that, I am not sure if it is necessary to be that obvious and it makes the API less clean (though I guess this is part preference). I'm fine with having a quick vote ;)

packages/tooltip/src/LionTooltip.js
```diff
@@ -1,31 +1,23 @@
...
-export class LionTooltip extends UpdatingElement {
-  static get properties() {
-    return {
-      position: {
-        type: String,
-      },
-    };
-  }
-
+export class LionTooltip extends LionPopup {
```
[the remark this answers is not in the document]
jorenbroekema: I was considering that, but could not think of any sensible name for it 🙈  Popup is pretty damn generic already. @tlouisse  ideas? 

LionLocalOverlayBase maybe?

packages/popup/test/lion-popup.test.js
jorenbroekema: good point, actually, `await el._controller.show();` should be enough already by itself.

packages/tooltip/test/lion-tooltip.test.js
jorenbroekema: All aTimeout()s introduced by this PR can be removed fortunately :). They were needed I think when it was still loading on-demand instead of eagerly in constructor.

packages/popup/src/LionPopup.js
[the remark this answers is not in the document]
jorenbroekema: Changed to popperConfig

packages/popup/src/LionPopup.js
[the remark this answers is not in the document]
jorenbroekema: Leaving it in for now, as third party tools (like open-wc storybook helper), and maybe docs generation, and other stuff may take use of this.

packages/popup/src/LionPopup.js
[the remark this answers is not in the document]
bashmish: 0.3.0 is breaking for NPM when 0.x.x

but we agreed that it's OK to make a breaking release

packages/popup/stories/index.stories.js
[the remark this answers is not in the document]
bashmish: we'll work on proper z-index support in futere, so that users don't have to deal with it manually in any case, but for now your change is fine :+1:

packages/tooltip/src/LionTooltip.js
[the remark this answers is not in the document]
bashmish: I'm OK to keep it like this for now, though I hope as soon as we have 3 or more we will make a new base class, or if users ask about one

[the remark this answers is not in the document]
jorenbroekema: Yeah agreed

packages/popup/src/LionPopup.js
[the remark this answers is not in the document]
ruphin: You can use getters and setters with or without having the property in `static get properties`. If you have it in `properties`, the element will trigger a render when the setter is called. Alternatively, you can call `this.requestUpdate()` in the setter, but in general, I think placing the property in `properties` is the preferred method if the element needs to re-render when the property changes.

For properties that do not trigger renders, getter and setter without having it in `properties` is fine.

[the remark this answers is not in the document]
tlouisse: Our current DelegateMixin also creates getters and setters under the hood. From a technical perspective, there is no need to add them in the properties list. From a readability/'tooling alignment' perspective there is and that's why we want to readd them in properties in the next version of DelegateMixin

Issue #92, opened 2019-06-11
[input-datepicker] Readonly property not working properly


gafortuno:
I added `readonly` property in my `lion-input-datepicker`, but I am still able to select a new date value.

```
<lion-input-datepicker label="Date" readonly></lion-input-datepicker>
```

bashmish:
I confirm that the button for the calendar still allows to select a date and change the input's value.

PR #105, opened 2019-06-18
fix(input-datepicker): disable invoker when readonly

tlouisse:
Fixes https://github.com/ing-bank/lion/issues/92

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -197,8 +197,13 @@ export class LionInputDatepicker extends LionInputDate {
    */
   _requestUpdate(name, oldValue) {
     super._requestUpdate(name, oldValue);
+
     if (name === 'disabled') {
       this.__delegateDisabled();
+    }
+
+    if (name === 'disabled' || name === 'readOnly') {
+      this.__toggleInvokerDisabled();
     }
   }
 
@@ -206,11 +211,14 @@ export class LionInputDatepicker extends LionInputDate {
    * TODO: [delegation of disabled] move this to LionField (or FormControl) level
    */
   __delegateDisabled() {
-    if (this.delegations.target()) {
-      this.delegations.target().disabled = this.disabled;
+    if (this.inputElement) {
+      this.inputElement.disabled = this.disabled;
     }
+  }
+
+  __toggleInvokerDisabled() {
     if (this._invokerElement) {
-      this._invokerElement.disabled = this.disabled;
+      this._invokerElement.disabled = this.disabled || this.readOnly;
     }
   }
 
@@ -220,6 +228,7 @@ export class LionInputDatepicker extends LionInputDate {
   firstUpdated(c) {
     super.firstUpdated(c);
     this.__delegateDisabled();
+    this.__toggleInvokerDisabled();
   }
 
   /**
--- a/packages/input/src/LionInput.js
+++ b/packages/input/src/LionInput.js
@@ -7,12 +7,21 @@ import { LionField } from '@lion/field';
  * @extends LionField
  */
 export class LionInput extends LionField {
-  get delegations() {
+  static get properties() {
     return {
-      ...super.delegations,
-      target: () => this.inputElement,
-      properties: [...super.delegations.properties, 'readOnly'],
-      attributes: [...super.delegations.attributes, 'readonly'],
+      ...super.properties,
+      /**
+       * A Boolean attribute which, if present, indicates that the user should not be able to edit
+       * the value of the input. The difference between disabled and readonly is that read-only
+       * controls can still function, whereas disabled controls generally do not function as
+       * controls until they are enabled.
+       *
+       * (From: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly)
+       */
+      readOnly: {
+        type: Boolean,
+        attribute: 'readonly',
+      },
     };
   }
 
@@ -29,5 +38,23 @@ export class LionInput extends LionField {
         return native;
       },
     };
+  }
+
+  _requestUpdate(name, oldValue) {
+    super._requestUpdate(name, oldValue);
+    if (name === 'readOnly') {
+      this.__delegateReadOnly();
+    }
+  }
+
+  firstUpdated(c) {
+    super.firstUpdated(c);
+    this.__delegateReadOnly();
+  }
+
+  __delegateReadOnly() {
+    if (this.inputElement) {
+      this.inputElement.readOnly = this.readOnly;
+    }
   }
 }
--- a/packages/input-datepicker/stories/index.stories.js
+++ b/packages/input-datepicker/stories/index.stories.js
@@ -49,6 +49,17 @@ storiesOf('Forms|Input Datepicker', module)
   .add(
     'Disabled',
     () => html`
-      <lion-input-datepicker disabled></lion-input-datepicker>
+      <lion-input-datepicker label="Disabled" disabled></lion-input-datepicker>
+    `,
+  )
+  .add(
+    'Readonly',
+    () => html`
+      <lion-input-datepicker
+        help-text="Notice that it's not possible to open the calendar on readonly inputs"
+        label="Readonly"
+        readonly
+        .modelValue="${new Date()}"
+      ></lion-input-datepicker>
     `,
   );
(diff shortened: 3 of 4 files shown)


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -213,12 +216,19 @@ export class LionInputDatepicker extends LionInputDate {
     }
   }
 
+  __onReadOnlyChanged() {
+    if (this._invokerElement) {
+      this._invokerElement.disabled = this.disabled || this.readOnly;
+    }
+  }
```
bashmish: for disabled you have this

```js
  __delegateDisabled() {
    if (this.delegations.target()) {
      this.delegations.target().disabled = this.disabled;
    }
    if (this._invokerElement) {
      this._invokerElement.disabled = this.disabled;
    }
  }
```

why not do the same for readonly and also set `this.delegations.target().readOnly`?

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -213,12 +216,19 @@ export class LionInputDatepicker extends LionInputDate {
...
+
   /**
    * TODO: [delegation of disabled] move this to LionField (or FormControl) level
    */
   firstUpdated(c) {
     super.firstUpdated(c);
     this.__delegateDisabled();
+    this.__onReadOnlyChanged();
```
bashmish: `__onReadOnlyChanged` is not the best name when you need to call it from multiple place like here in `firstUpdated`, and not only from `_requestUpdate`

I would rather give it a name that suits both needs, and I think the current name for `disabled` is way better in that regard, but my comment above shows that you don't really delegate it atm... I think you should delegate as well (explained above) and then the name will be `__delegateReadOnly`, so this can be resolved too

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -196,8 +196,11 @@ export class LionInputDatepicker extends LionInputDate {
    */
   _requestUpdate(name, oldValue) {
     super._requestUpdate(name, oldValue);
+
     if (name === 'disabled') {
       this.__delegateDisabled();
```
bashmish: would be nice to have a demo in the storybook for `Readonly` in addition to `Disabled` that we currently have

I checked how they would look like and probably smth goes wrong with the current implementation

image

packages/input/src/LionInput.js
```diff
@@ -7,12 +7,21 @@ import { LionField } from '@lion/field';
  * @extends LionField
  */
 export class LionInput extends LionField {
-  get delegations() {
+  static get properties() {
     return {
-      ...super.delegations,
-      target: () => this.inputElement,
-      properties: [...super.delegations.properties, 'readOnly'],
-      attributes: [...super.delegations.attributes, 'readonly'],
```
bashmish: I'm not sure what you gonna do with this, so maybe will revisit after the WIP status changes to done

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: the invokerElement (button) is not configured as delegation target. The inputElement is

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: Readonly is not delegated, but used as a syncObserver (we need to consider both disabled and readonly instead of just putting it blindly on invokerElement). It's a common pattern throughout our code to handle all logic that needs to be executed in an observer '__onPropXChanged'

packages/input/src/LionInput.js
tlouisse: If you know how the Delegate Mixin works, you know that it dynamically creates getters and setters. They override the ones creating by Update Mixin, disallowing us to hook into their observer hooks

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: Good idea. Will do

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: I changed it to `__toggleDisabledOfInvokerOnReadonlyChanged`. Please change and push if you don't agree ;)

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: But `delegations.target()` should be called `this.inputElement` to be forwards compatible with refactoring about DelegateMixin

packages/input-datepicker/src/LionInputDatepicker.js
bashmish: does not explain why you don't delegate 'readonly' to the input element? I didn't ask to set such an attribute on the invoker, but rather delegate and set it on the input element

let's agree on the name after we agree on the logic here

packages/input-datepicker/src/LionInputDatepicker.js
bashmish: does not explain why you don't delegate 'readonly' to the input element? I didn't ask to set such an attribute on the invoker, but rather delegate and set it on the input element

by "delegate" I don't mean using DelegateMixin, I mean literally the delegation itself

let's agree on the name of this method after we agree on the logic here

packages/input/src/LionInput.js
```diff
@@ -7,12 +7,21 @@ import { LionField } from '@lion/field';
...
+       * the value of the input. The difference between disabled and readonly is that read-only
+       * controls can still function, whereas disabled controls generally do not function as
+       * controls until they are enabled.
+       *
+       * (From: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly)
+       */
+      readOnly: {
+        type: Boolean,
+        attribute: 'readonly',
+      },
```
bashmish: do we also want `reflect: true` for `readOnly `?

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -186,8 +186,8 @@ export class LionInputDatepicker extends LionInputDate {
   get delegations() {
     return {
       ...super.delegations,
-      properties: super.delegations.properties.filter(p => p !== 'disabled'),
-      attributes: super.delegations.attributes.filter(p => p !== 'disabled'),
+      properties: super.delegations.properties.filter(p => p !== 'disabled' && p !== 'readOnly'),
+      attributes: super.delegations.attributes.filter(p => p !== 'disabled' && p !== 'readonly'),
```
bashmish: you already removed `readOnly`/`readonly` from delegation in LionInput (which I still don't understand to be honest)

but if so, why do you need to make this exception `p !== 'readonly'` here?

packages/input/src/LionInput.js
bashmish: ok, makes sense then

how about other LionInput based component which you are not changed in this MR? how would they work if `readonly` is not delegated?

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: No, because delegation is `source[propX] = target[propX] `. This was `source[propX] = target[propY] `, so observer logic

packages/input/src/LionInput.js
tlouisse: In the end we might want to. But I think it would be adding functionality not there before (not related to this small bugfix). We can reconsider a proper solution once we do all delegation via a new convention/mixin

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: Good catch. Will remove that part

packages/input-datepicker/src/LionInputDatepicker.js
bashmish: I found out why there is a difference between how disabled and readonly buttons looks like

for disabled we have this style:

```css
            :host(.state-disabled) .input-group ::slotted(*) {
              color: var(--disabled-text-color, #adadad);
            }
```

not for readonly though

this is just a side-effect of curse, there was no intention to make a different colour of the button, but it is a bug, because in both cases the button should look like disabled and in both cases we need consistent styles which we just don't define and face a strange inconsistent side effect

bashmish: I will create another but report after we finish this

PR #110, opened 2019-06-19
fix(input-amount): add support for negative numbers

vkachan:
In general input-amount field should support negative numbers


bashmish:
Looks good to me :+1:

I pushed wrong stuff to your master and can't revert it now. I will make a new MR with your changes and some fixes from me.

PR #113, opened 2019-06-19
feat(input-range): add input-range component

manolakis:
Creates the `lion-input-range` component.

This component extends `LionField` instead `LionInput` because `LionInput` contains the [readonly attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly) that belongs to text input fields and the range is a numeric input that doesn't support it.


manolakis:
I'm not able to run the tests again. It's strange because I run it several twice locally and it worked fine (I know, it's the typical phrase "it works in my machine" 😞). However, I would like to know the reason why the test failed. 🤔

daKmoR:
browserstack was flacky yesterday - it works fine now so I rerun the test and it's green 🤗

manolakis:
thanks @daKmoR! 👏

manolakis:
Why are those deploys failing? do I need to do something in the PR? 🤔

erikkroes:
We have a local contribution coming up which is more extensive. Closing this version.

PR #116, opened 2019-06-20
Update LionIcon to no longer render 'undefined'

### LINKED ISSUE #115: [icon] Renders 'undefined'
When you create a new `<lion-icon>` element, the `svg` property is initially undefined, and it renders nothing (blank space).

When I explicitly set the `svg` property to the value `undefined` (after it being set as something else) it renders the text `"undefined"` instead of rendering blank space.

I expect the element to render nothing when I set the `svg` property to `undefined`.

ruphin:
Fixes #115


bashmish:
@ruphin I added tests to verify if your change does not impact the current behavior. And they don't fail which is unexpected to me. Maybe I'm doing smth wrong in tests, so will need to have another look a little bit later. If you want to speed things up, please help with testing this.

ruphin:
I updated the entire class a bit. I verified that the test cases we discussed are handled correctly. 
I don't know why the tests were producing unexpected results. The test setup you are using seems unnecessarily complex to me, I used something simple like this and it worked for me:

```js
let resolveA;
  const a = new Promise( resolve => {
    resolveA = () => resolve('thing');
  });
  let resolveB;
  const b = new Promise( resolve => {
    resolveB = () => resolve('other');
  });
```

bashmish:
I needed these promises 4 times and just didn't want to repeat myself (tests grow pretty quickly if you repeat this 4 times), and we use real SVG strings and not some fake ones, because we do blackbox testing in our code with real input data. Hence the complexity. Although I don't see it too or unnecessarily complex. Also take a look at other tests, they are built with the same idea in mind.

I like the current implementation and it looks the same in terms of behavior, although the refactoring and actual fix should be in seprarate commit. And in the spirit of TDD this should be

1. add a failing test for the bug
2. fix the test
3. do the refactoring

1+2 can be first commit, 3 can be a second

I don't see a test for 1 anywhere, the tests I added are more related to 3 to prevent refactoring from breaking other things

bashmish:
I added a test for this

```js
it('does not render "undefined" if changed from valid input to undefined', async () => {
  const el = await fixture(
    html`
      <lion-icon .svg=${heartSvg}></lion-icon>
    `,
  );
  await el.updateComplete;
  el.svg = undefined;
  await el.updateComplete;
  expect(el).lightDom.to.equal('');
});
```

It fails in master and passes after your changes here.

Now I have better understanding of what we are fixing here, should have started with that. And I disagree with the need for this fix. It seems like we are programming in a very protective style here expecting all kinds of wrong input. For example I can give you another example where you'll have problems even with the current fix: `<lion-icon .svg=${true}></lion-icon>` will pass your condition and render a string "true". Shall we also take this into account? What other wrong inout we should take into account and prevent from rendering? A lot. Shale we render fake string with non-SVG content? Shall we render SVG which have syntax errors? We can of course make a very strict check validating everything, but is the cost and effort worth the result?

We designed this component to be used with properly formatted SVG strings. Everything else should never be passed.

If you need to show/hide the icon, you should do it on the level of `lion-icon` element and not the SVG that it renders.

So I'm thinking towards closing this MR. We can move tests for race conditions and your refactoring to another MR if needed.

ruphin:
This is most certainly a bug.

`svg` is part of the state of the element. With the current implementation I can have two identical `<lion-icon>` elements with identical properties and state, and one of them shows nothing and the other shows the text `"undefined"`. It's not about correcting malformed input to the `svg` property, it's about consistency of how you handle state.

Here is an example where you run into this problem:

```js
const items = [
  {
    icon: '<svg>...</svg>',
    title: 'Something with icon',
  },
  {
    title: 'Something without icon',
  }
];

render(items.map(item => 
  html`
    <div>
      <lion-icon .svg=${item.icon}></lion-icon><span>${item.title}</span>
    </div>`
  ), container)
```

This code works just fine, and items without an icon render a blank icon. However, when we re-render and some items may no longer have an icon because the list changed, we randomly start rendering `"undefined"` for _some_ of the icons. My template is fully delcarative, so it should not have any problems, but `<lion-icon>` breaks, because the element has a hidden "I once had a non-undefined value in my svg property" state that I cannot access. I don't see how this can be considered not a bug. As it is now the element behaves in a non-declarative way, and I cannot deterministically set it's behaviour without destroying it and re-creating it to ensure it is in the right state.

----------------------

If you decide that a `<lion-icon>` element that has an undefined `svg` property should render the text `"undefined"`, then it should also render `"undefined"` in every case where `svg` is undefined, including when you create a new icon without setting the `svg` property initially.

If you decide that a `<lion-icon>` element with an undefined `svg` property should be blank, then it should be blank in all cases where `svg` is undefined.

Personally I don't see why anyone would consider the first scenario to be correct, but if you choose to do that for whatever reason, it's fine with me. I just need it to be consistent.

bashmish:
My comment was maybe too broad, so I will restate the main point from there: writing this code

`<lion-icon .svg=${undefined}></lion-icon>`


or this code

`<lion-icon></lion-icon>`

is not right, because `lion-icon` does not make sense without right input for this property. We don't interpret such scenario as "blank icon" (what is a blank icon?) That's why we just didn't care about consistency there. Consistency is important, but is it that important for the wrong usage of component? I still doubt that.

bashmish:
> If you decide that a <lion-icon> element that has an undefined svg property should render the text "undefined", then it should also render "undefined" in every case where svg is undefined, including when you create a new icon without setting the svg property initially.

I can't believe I'm saying this, but yeah, this seems even more helpful for the end user, because it will be an indication that the component API is misused.

ruphin:
Personally, I think handling `undefined` is not necessarily violating the API. `undefined` is literally the absence of a value, and not a special value type that you have to coerce into String. The API is that the `svg` property contains a String (or any Object), in which case that String is inlined as the used svg image, or the `svg` property is undefined, in which case the content is empty.

There are obvious cases why it is desirable from a user perspective to have this behaviour. The main reason is that the `<lion-icon>` element takes space in the layout. An empty `<p>` still has margins and takes space in the layout, and an empty `<lion-icon>` should do the same. A common use case is a list of icons with labels. I want the labels to align, so the icons need to always take a predictable amount of space, even if no icon is defined for a specific label. If I have to conditionally render `<lion-icon>`, I need to either wrap it in another element with a defined and predictable size, or have some kind of replacement `<lion-icon-spacefiller>` element that I can render instead if the icon is undefined. This is incredibly inconvenient.

There are plenty of cases where showing nothing is the desired behaviour, but it is _never_ correct to paint the text `"undefined"`.

I know the `` element coerces `undefined` to string in the `src` property, but I consider that behaviour broken. Lets not do things "because that's the way it always has been done", but make decisions based on reason.

It doesn't even take extra code to handle the `undefined` case like this. If we agree that the element should be consistent, i.e. it should render `"undefined"` in the `<lion-icon></lion-icon>` case, that will take additional code, which is probably just as much as the code required to handle `undefined` gracefully.

I have made my case. You can decide what behaviour you would like this element to have. My main priority is making it deterministic, so if you make a choice I will refactor this PR to implement whichever option you choose.

bashmish:
> There are obvious cases why it is desirable from a user perspective to have this behaviour. The main reason is that the `<lion-icon>` element takes space in the layout. An empty `<p>` still has margins and takes space in the layout, and an empty `<lion-icon>` should do the same. A common use case is a list of icons with labels. I want the labels to align, so the icons need to always take a predictable amount of space, even if no icon is defined for a specific label. If I have to conditionally render `<lion-icon>`, I need to either wrap it in another element with a defined and predictable size, or have some kind of replacement `<lion-icon-spacefiller>` element that I can render instead if the icon is undefined. This is incredibly inconvenient.

I agree with this use case and agree with this change then. Please rebase with the master and update the history according to https://github.com/ing-bank/lion/pull/116#issuecomment-504456044, then I can do a final check and release.

PR #119, opened 2019-06-21
Return sign to the front of the currency formatted value in Dutch

### LINKED ISSUE #117: [localize] The € sign is put after the number in Dutch
In the current situation the formatNumber displays the amount of 200.000 euro in Dutch:

- 200.000 EUR (this is correct)
- 200.000 € (this is incorrect, it should be € 200.000)

bverhoef:
fixes #117


packages/localize/test/formatNumber.test.js
```diff
@@ -29,7 +29,7 @@ describe('formatNumber', () => {
     expect(formatNumber(123456.789, { currency: 'EUR', ...currencyCode })).to.equal(
       '123.456,79 EUR',
     );
-    expect(formatNumber(123456.789, { currency: 'JPY', ...currencySymbol })).to.equal('123.457 ¥');
+    expect(formatNumber(123456.789, { currency: 'JPY', ...currencySymbol })).to.equal('¥ 123.457');
```
bashmish: this issue is targeting EUR https://github.com/ing-bank/lion/issues/117

and explanation giving here explains the logic for EUR only https://www.reference.com/business-finance/euro-sign-before-after-number-13773f16a6f25d98

so why change it for every other currency? do we have really good understanding of the rules for other currencies like JPY in this case?

bverhoef: absolutely: see https://taaladvies.net/taal/advies/vraag/1178/euro_valutateken_voor_of_achter_het_bedrag/

All currency symbols should before in dutch language.

jorenbroekema: I can confirm the source is credible and indeed specifies this behavior for all currency symbols according to the source.

bashmish: Thanks!

I improved tests a lot and see now very well what was broken. We did normalization wrong, so it is good now to fix it, so that we only normalize the currency code for Dutch and don't touch the sign.

Issue #108, opened 2019-06-18
Run tests in all supported browsers on BrowserStack


ovidiu1:
### Note: if i'm running using Chrome ( bs_win10_chrome_latest) all the tests are running.

I try to run all the tests on IE11 (bs_win10_ie_11), and i got something like this:

```
λ npm run test:bs

> @lion/root@ test:bs D:\dev\lab\lion
> karma start karma.bs.config.js --coverage


START:
18 06 2019 21:28:51.979:WARN [filelist]: Pattern "D:/dev/lab/lion/__snapshots__/**/*.md" does not match any file.
18 06 2019 21:28:53.547:INFO [karma-server]: Karma v4.1.0 server started at http://0.0.0.0:9876/
18 06 2019 21:28:53.548:INFO [launcher]: Launching browsers bs_win10_ie_11 with concurrency unlimited
18 06 2019 21:28:53.607:INFO [launcher]: Starting browser IE 11.0 (Windows 10) on BrowserStack
18 06 2019 21:29:00.208:INFO [launcher.browserstack]: IE 11.0 (Windows 10) session at https://automate.browserstack.com/builds/a48343450ca0f66a4f1ac1da96d4536c314fe859/sessions/69d851431e3a82dc9cc70a483c21ff536a188c53
18 06 2019 21:29:05.135:INFO [IE 11.0.0 (Windows 10.0.0)]: Connected on socket RDJX-pPfYMpMcPdXAAAA with id 79747197

Finished in 0.003 secs / 0 secs @ 21:29:07 GMT+0300 (Eastern European Summer Time)

SUMMARY:
√ 0 tests completed

=============================== Coverage summary ===============================
Statements   : Unknown% ( 0/0 )
Branches     : Unknown% ( 0/0 )
Functions    : Unknown% ( 0/0 )
Lines        : Unknown% ( 0/0 )
================================================================================
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @lion/root@ test:bs: `karma start karma.bs.config.js --coverage`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @lion/root@ test:bs script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```
I think is a problem with browserstack, also i found [this issue](https://github.com/DevExpress/testcafe/issues/2480).

bashmish:
There are 2 issues here.

[...]

PR #120, opened 2019-06-23
chore: allow running tests in IE11 on BrowserStack

ovidiu1:
I have create a new npm script to run all tests in ie11 browser - browserstack


bashmish:
I don't mind merging this as as since it has some value: you are allowed to run tests in IE11 on browserstack, but it does not close the issue #108, the problem is broader there, including the fact that we need to fix flaky tests as well. Also `fix(scope)` format assumes you put either a subpackage name as a scope, of if it impacts the entire monorepository, then you just leave it empty. Third, this is a `chore` type of commit because it does not impact the released code. So taking into account all of this please change the commit message to `chore: allow running tests in IE11 on BrowserStack` and I'll approve.

bashmish:
I've realised that I was wrong. `:legacy` is for testing transpiled code everywhere, rather than for testing old browsers. So in fact the final logic will be like this:

1. modern config runs in browsers that support ES modules (modern Chrome, Firefox, Safari, Edge)
2. legacy config runs in all browsers (the ones above + IE11)

The reason it needs to be like this is because often the code is transpiled and run in production on every browser, while some people might have differential builds and serve different code based on browser agent sniffing. So basically transpiled code needs to work everywhere.

So I'm closing this MR. if you want to run tests for yourself, please use this terminal command:

```sh
yarn test:bs --legacy --browsers=bs_win10_ie_11
```

or more precisely - with `BROWSER_STACK_USERNAME` and `BROWSER_STACK_ACCESS_KEY`

```sh
BROWSER_STACK_USERNAME=%name% BROWSER_STACK_ACCESS_KEY=%key% yarn test:bs --legacy --browsers=bs_win10_ie_11
```

PR #121, opened 2019-06-24
feat(validate): add custom localize namespace to a validator instance

tlouisse:
TODO in feature merge request:
- add separate stories folder for validation, including this examples

### DIFF
--- a/packages/validate/src/ValidateMixin.js
+++ b/packages/validate/src/ValidateMixin.js
@@ -425,6 +425,7 @@ export const ValidateMixin = dedupeMixin(
         return this.constructor.__getLocalizeKeys(
           `error.${data.validatorName}`,
           data.validatorName,
+          data.validatorConfig.namespace,
         );
       }
 
@@ -432,11 +433,16 @@ export const ValidateMixin = dedupeMixin(
         return this.constructor.__getLocalizeKeys(
           `warning.${data.validatorName}`,
           data.validatorName,
+          data.validatorConfig.namespace,
         );
       }
 
       getInfoTranslationsKeys(data) {
-        return this.constructor.__getLocalizeKeys(`info.${data.validatorName}`, data.validatorName);
+        return this.constructor.__getLocalizeKeys(
+          `info.${data.validatorName}`,
+          data.validatorName,
+          data.validatorConfig.namespace,
+        );
       }
 
       /**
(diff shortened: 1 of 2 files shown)


jorenbroekema:
Implementation looks good but on the overall idea I think we should consider what this does to namespace conflict risk. Currently it's possible if you have for example 2 custom elements that both use LocalizeMixin but have the same namespace. Now it will also be *possible* for each validator that's created, whereas before, it wasn't possible (because you would still use lion-validate namespace in the end).

So another possibility could be to create an interface for our users that eventually puts everything in a custom-validator namespace. Just to name one possible solution. Then you would document this namespace and the risk of conflicts through validators would be mitigated that way.

Maybe I'm completely off on this one, but this is a concern from me, other than that I'd be completely fine with this since the current way is really tedious, and this would make it way better. Then again, when we do a full refactor of the combination of validatemixin-localizemixin, we may have a way better view of the big picture and how we approach it in the future, and simultaneously address the issue this PR addresses, in possibly a better way. So my question is really, is the value worth the extra risk (let me know if im wrong about it), and is it worth it pushing this forward now when we know there's a "big refactor" coming? And that should probably be discussed among the team

tlouisse:
- There is just one namespace per application where you would put your validator overrides. Similar to what we already documented, here: https://github.com/ing-bank/lion/blob/master/packages/validate/docs/tutorials/CustomValidatorsTutorial.md#adding-an-error-message
But now without the need to add them to non Custom Validators, with a more friendly api.
So just like any localize namespace that an Application Developer creates, he should be aware that his translations should be put on a unique namespace, exactly like any other translation he would use in his app

- The same risks we introduce by recommending people to extend components. Only then we create a lot of other additional risks:
 - once you extend, you take the role of 'Subclasser' instead of 'Application Developer'. This means people will feel comfortable changing much more then what we allow based on our semVer (public api) release cycles. It's risky, uncontrollable and this 'flexibility' for app developers will make us extremely **inflexible** at the core, since we don't know what people overrode.
- we loose track of how people use our components (harder to detect when our components are not used directly).

Ideally, for me, translation should happen on a different (non ui) layer and be provided to the component by App Developers (default translations we can offer as a separate import). This should first be discussed though and the more straightforward api we have now, the more straightforward a migration path would be.

jorenbroekema:
> There is just one namespace per application where you would put your validator overrides.

Yes, that's our recommendation, but this PR seems to allow people to put whatever string they like, not limiting it to one namespace per app, but potentially one namespace per validator (I can totally see users doing this outright). correct?

I do agree with the example of risk you mention, but that doesn't automatically mean we should be fine with introducing this risk as well. It depends on the value (where extending ES6 classes has enormous value), whether it is worth it. I wonder what the rest has to say about that, I personally would lean towards waiting until the "big refactor", as it may lead to deprecate this API altogether, but since this feature is very nice short-term value and the risk is somewhat limited, I'd also be fine with just implementing this right now. So consider me neutral ;)

daKmoR:
> I personally would lean towards waiting until the "big refactor"

jup, agree on that - imho no need to introduce a "new" more complex logic on top of what we have if we require them to learn something different in a few months anyway... (also it's more stuff we would need to currently maintain)

tlouisse:
> Yes, that's our recommendation, but this PR seems to allow people to put whatever string they like, not limiting it to one namespace per app, but potentially one namespace per validator (I can totally see users doing this outright). correct?

Would be a matter of documenting this right. But I don't see a difference with any other translation in their app.

Common dudes, choose the lesser of the two 'evils'. It's just a no brainer

jorenbroekema:
Aah like that, it wasn't clear to me that the extension-risk you mentioned was directly related to this. But now I understand, it is the only "alternative" to this problem, so then my risk argument is wrong, in fact it changes one risk to another risk, and probably the risk we mitigate is worse, so then it's worth it from a risk perspective. 

And this is really not hard to maintain I would say so I don't necessarily agree with that argument. Everything that brings a lot of value to our users is probably something we need to maintain, and when it comes to value to maintainence efforts balance, on this one, value easily wins I would say.

tlouisse:
Thanks for the support:) Even better: maybe we can do the best of both worlds by agreeing what the new validation api should be, so we can adjust this code and be fully forwards compatible? Could you do some proposal for a new Validator api (preferably backwards compatible, especially on extension layer)?

Issue #112, opened 2019-06-19
[proposal] allow lit-html in SlotMixin 


tlouisse:
When the SlotMixin was designed, we aimed for a template system agnostic solution.
Since we embraced lit-html, let’s leverage its power also in SlotMixin.

It would change this:

```js
label: () => {
  const label = document.createElement('label');
  label.textContent = this.label;
  return label;
},
'help-text': () => {
  const helpText = document.createElement('div');
  helpText.textContent = this.helpText;
  return helpText;
},
```
to this:
```js
label: () => html`<label>”${this.label}”</label>`,
'help-text': () => html`<div>”${this.helpText}”</div>`,
```

  It’s more concise and readable.
It would also be more in line with our existing templates (considering the fact that the slots are part of the template and are defined outside the main `render()` for accessibility reasons only.)

@daKmoR @bashmish since you were both involved in the creation of this mixin as well, do you think it would be a nice improvement?

bashmish:
For me it is less about the syntax and more about understanding how it works and why it is needed and making a proper abstraction with the minimal needed functionality. If you look at this example from LionButton:

```js
  get slots() {
    return {
      ...super.slots,
      _button: () => {
        if (!this.constructor._button) {
          this.constructor._button = document.createElement('button');
          this.constructor._button.setAttribute('slot', '_button');
          this.constructor._button.setAttribute('tabindex', '-1');
        }
        return this.constructor._button.cloneNode();
      },
    };
  }
```

Can you answer immediately:
- Why is it cloned? It is for performance, but can we abstract this so that users don't need to repeat this code? I think we can.
- Why do you need  `.setAttribute('slot', '_button');`? In fact you don't and this is a mistake in our code in LionButton. So we kinda misunderstand ourselves how it is supposed to be used.
- Why does it have to start with `_button`? Because we call it `private` slot and we want to keep it separate from others. Maybe we should enforce it, because it is a pretty important signal and we should not allow the usage without `_`.

Also when we look at your proposal and apply it to LionButton:

```js
  get slots() {
    return {
      ...super.slots,
      _button: () => html`<button tabindex="-1"></button>`,
    };
  }
```

Can we say if this will be rerendered when `this.label` changes? If we keep the current implementation it won't, but the expectation will be that it will, like if it is a `render` function.

If lit-html syntax and some improvements to the implementation will make it more clear and prevent misuse and bugs, then I think we should support it. But I would first make a POC and look at this functionality from different angles, not only the syntax.

[...]

PR #124, opened 2019-06-25
feat(core): allow lit-html in SlotMixin

tlouisse:
Fixes https://github.com/ing-bank/lion/issues/112

### DIFF
--- a/packages/core/src/SlotMixin.js
+++ b/packages/core/src/SlotMixin.js
@@ -1,5 +1,5 @@
 /* eslint-disable class-methods-use-this */
-
+import { render, TemplateResult } from '../index.js';
 import { dedupeMixin } from './dedupeMixin.js';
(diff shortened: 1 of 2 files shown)


jorenbroekema:
Since no issue was mentioned, can you provide a use case for this where supporting TemplateResult in SlotMixin is important?

tlouisse:
> Since no issue was mentioned, can you provide a use case for this where supporting TemplateResult in SlotMixin is important?

Ah, right. See https://github.com/ing-bank/lion/issues/112
It's an improvement of readability/maintainability and Developer Experience. One of the reasons that we don't construct our shadow dom like this.

bashmish:
We are closing this for now. Can you please talk to @thatianamg ?

daKmoR:
Let’s use renderOffline (rename to renderAsNode) instead

Issue #94, opened 2019-06-11
[form][input] Reset not working after submitting invalid inputs


joanntaguinod:
## Steps to Replicate
1. Go to [form reset demo](https://lion-web-components.netlify.com/?path=/story/forms-form--form-submit-reset)
2. Enter invalid inputs for first name and last name
3. Click Submit button
4. Click Reset button

## Expected Behaviour 
Clicking reset button should empty the input elements

## Current Behaviour
Invalid input is retained in `value` propert.  However, when you inspect the `modelValue`, it is already cleared. Even if you manually delete the value, the element is putting it back.

jorenbroekema:
Thanks for creating this issue!

I found some additional weird behavior on that demo:
1. Set input to some string (fewer than 15 char)
2. Press submit, looks fine.
3. Reset (does not work)
4. Press submit again (you suddenly get error message that you would normally get if the value was empty)

This seems to suggest that reset *does* clear the modelValue, but this is not synced back to the <input> viewValue. 

This bug may or may not be influenced by https://github.com/ing-bank/lion/pull/55 , so probably good to wait for this and check if it is still broken.

bashmish:
I confirm this bug as well. Also when calling `querySelector('lion-form').reset()` or `querySelector('lion-form').resetGroup()`.

tlouisse:
There is a fix in this branch: https://github.com/ing-bank/lion/compare/fix/formIssues
(ModelValue wasn't formatted, because we lately check for `errorState` as a condition for formatting. This check needs to be there only as user feedback(when typing in the field), not when modelValues are set  imperatively, like with the reset.)

Currently, there are more pending form bugs. When they are all solved and all individual fixes work together, this fix will be released

joanntaguinod:
Is the demo using the latest release/master? I was still able to replicate the issue on the form reset story

bashmish:
@joanntaguinod I checked on the Netlify and locally, and looks like the bug was fixed, but the storybook was not updated on the Netlify.

We are going to investigate what is going wrong with the Netlify storybook deployment.

Issue #129, opened 2019-06-27
[form][field]  `.formattedValue` should be set to undefined when not being able to format


tlouisse:
Currently, this happens:

```js
        if (this.modelValue instanceof Unparseable) {
          return this.modelValue.viewValue;
        }
        if (this.errorState) {
          return this.inputElement ? this.value : undefined;
        }
        return this.formatter(this.modelValue, this.formatOptions);
      }
```
https://github.com/ing-bank/lion/blob/master/packages/field/src/FormatMixin.js#L186

It should not return the current value of the inputElement and sync it back. Whenever the formattedValue returns undefined, it should be decided in the sync down step to leave inputElement.value as is

daKmoR:
I added a failing test... (pushed a commit)
```js
  it('leaves partial (and therefore unparseable) inputs untouched', async () => {
    const el = await fixture(html`
      <lion-input-date .value=${'2017/06'}></lion-input-date>
    `);
    expect(el.value).to.equal('2017/06');
  });
```

jorenbroekema:
Could you confirm whether this issue is still open @tlouisse  ?

tlouisse:
Yep, it should just return `undefined` when invalid(errorState) , and leave `_inputNode.value` untouched

daKmoR:
Task to do:

```js
<lion-date-input>
el.modelValue = 'foo';
```

current:
```js
el.formattedValue // ''
el.value // ''
```

wanted:
```js
el.formattedValue // undefined
el.value // ''
```

tlouisse:
The idea was to make `formattedValue` "conceptually consistent".
When formatting didn't take place, this would be resembled by `undefined` (this means current modelValue can't be formatted). 
As a consequence, the formattedValue would not be synced to `_inputNode.value` (== viewValue). (a redundant step)
 
I changed my mind on this topic for a few reasons:
- If we have an Unparseable modelValue, we also set formattedValue to modelValue.viewValue (this is needed, because _inputNode.value might be empty or different). 
- formattedValue can be read via `myForm.formattedValue` to get all (scheduled) viewValues <= important one as we noticed recently 

If we think of `.formattedValue` as the 'scheduled viewValue'(after blur by default), it actually makes sense to store the current viewValue in the formattedValue.

So my notion of "conceptually consistent" changed and if it were up to me, we can close this issue and keep the current behavior :)

PR #132, opened 2019-06-27
fix(field): format conditionally on user input only

tlouisse:
fixes https://github.com/ing-bank/lion/issues/94

- See for explanation: https://github.com/ing-bank/lion/pull/132/files#diff-8a0f76e26a132ed6bb832fc7ff0518c8R202

- Note that we discriminate two flows: https://github.com/ing-bank/lion/pull/132/files#diff-8a0f76e26a132ed6bb832fc7ff0518c8R11

- `__preventDownwardsSync ` has been renamed to `__isHandlingUserInput `

- other doc changes are cosmetic (indenting)

### DIFF
--- a/packages/field/test/FormatMixin.test.js
+++ b/packages/field/test/FormatMixin.test.js
@@ -157,6 +157,25 @@ describe('FormatMixin', () => {
     expect(formatEl.inputElement.value).to.equal('foo: test');
   });
 
+  it('reflects back .formattedValue immediately when .modelValue changed imperatively', async () => {
+    const el = await fixture(html`
+      <${elem} .formatter="${value => `foo: ${value}`}">
+        <input slot="input" />
+      </${elem}>
+    `);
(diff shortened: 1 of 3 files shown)


packages/field/test/FormatMixin.test.js
```diff
@@ -216,12 +235,12 @@ describe('FormatMixin', () => {
       expect(formatterSpy.callCount).to.equal(1);
 
       el.errorState = true;
-      el.modelValue = 'bar';
+      mimicUserInput(el, 'bar');
       expect(formatterSpy.callCount).to.equal(1);
-      expect(el.formattedValue).to.equal('foo: init-string');
+      expect(el.formattedValue).to.equal('bar');
```
tlouisse: this should become `undefined`. Will be fixed here: https://github.com/ing-bank/lion/issues/129

PR #134, opened 2019-06-28
Prevent users from committing lockfiles with invalid resolved source

jorenbroekema:
This checks for lockfiles generated using a commonly used product in companies that do dependency resolution through a proxy rather than directly from yarn or npm official registry. Such a lockfile would be unusable for users without access to this domain.

### DIFF
--- a/scripts/yarn-lock-scan.js
+++ b/scripts/yarn-lock-scan.js
@@ -1,0 +1,10 @@
+const fs = require('fs');
+const path = require('path');
+
+const yarnLockPath = './yarn.lock';
+const data = fs.readFileSync(path.resolve(yarnLockPath), 'utf8');
+if (data.match(/artifactory/g)) {
+  throw new Error(
+    'Artifactory references in your yarn.lock! Please make sure you are using the official yarn or npm registry when downloading your dependencies!',
+  );
+}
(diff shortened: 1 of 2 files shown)


scripts/yarn-lock-scan.js
```diff
@@ -0,0 +1,16 @@
...
+
+const yarnLockPath = './yarn.lock';
+
+try {
+  const data = fs.readFileSync(path.resolve(yarnLockPath), 'utf8');
+  if (data.match(/artifactory/g)) {
+    console.error(
+      "Error: Artifactory references in your yarn.lock! Please make sure you're using the official yarn or npm registry when downloading your dependencies!",
+    );
+    process.exit(1);
```
bashmish: you could just do this in 1 line

```
throw new Error('Error: Artifactory references in your yarn.lock! Please make sure you're using the official yarn or npm registry when downloading your dependencies!');
```

scripts/yarn-lock-scan.js
```diff
@@ -0,0 +1,16 @@
...
+  }
+} catch (e) {
+  console.error('Error', e.stack);
```
bashmish: if smth goes wrong during `readFileSync`, we need to fail the script, but this will prevent it from failing and only output the erro

PR #135, opened 2019-06-28
Fix update popper config method

jorenbroekema:
I noticed a small bug, when for example updating a modifier while your viewport is pushing on the popper element already, it will not position correctly when you change this modifier.

This fixes it.

### DIFF
--- a/packages/overlays/test/LocalOverlayController.test.js
+++ b/packages/overlays/test/LocalOverlayController.test.js
@@ -230,9 +230,9 @@ describe('LocalOverlayController', () => {
       });
 
       await controller.show();
-      // 16px displacement due to default 16px viewport margin both horizontal and vertical
       expect(controller.content.firstElementChild.style.transform).to.equal(
         'translate3d(16px, 16px, 0px)',
+        '16px displacement is expected due to both horizontal and vertical viewport margin',
       );
     });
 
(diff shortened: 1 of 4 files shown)


packages/overlays/test/LocalOverlayController.test.js
```diff
@@ -445,6 +445,42 @@ describe('LocalOverlayController', () => {
       expect(controller._popper.options.modifiers.offset.offset).to.equal('0, 32px');
       expect(contentChild.style.transform).to.equal('translate3d(14px, -82px, 0px)');
     });
+
+    it.only('updates positioning correctly during shown state when config gets updated', async () => {
```
gerjanvangeest: Remove the `only`

packages/overlays/test/LocalOverlayController.test.js
```diff
@@ -445,6 +445,42 @@ describe('LocalOverlayController', () => {
...
+      });
+      await fixture(html`
+        <div style="position: absolute; top: 300px; left: 100px;">
+          ${controller.invoker} ${controller.content}
+        </div>
+      `);
+
+      await controller.show();
+      const contentChild = controller.content.firstElementChild;
+      expect(contentChild.style.transform).to.equal('translate3d(14px, -58px, 0px)');
```
gerjanvangeest: Can we add a `msg` as a second argument to explain what is happening?

jorenbroekema: Done ;)

PR #136, opened 2019-06-30
fix(input-datepicker): wrong styles of the invoker

### LINKED ISSUE #123: [input-datepicker] Wrong styles of the invoker button when `readonly` or `disabled`
image

I found out why there is a difference between how disabled and readonly buttons looks like

for disabled we have this style:

```css
            :host(.state-disabled) .input-group ::slotted(*) {
              color: var(--disabled-text-color, #adadad);
            }
```

not for readonly though

this is just a side-effect of course, there was no intention to make a different colour of the button, but it is a bug, because in both cases the button should look like disabled and in both cases we need consistent styles which we just don't define and face a strange inconsistent side effect

_Originally posted by @bashmish in https://github.com/ing-bank/lion/pull/105#discussion_r295801113

ovidiu1:
Fixes #123

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -219,6 +219,7 @@ export class LionInputDatepicker extends LionInputDate {
   __toggleInvokerDisabled() {
     if (this._invokerElement) {
       this._invokerElement.disabled = this.disabled || this.readOnly;
+      this.classList[this._invokerElement.disabled ? 'add' : 'remove']('state-disabled');
     }
   }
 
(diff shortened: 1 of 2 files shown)


jorenbroekema:
Hi Ovidiu, 

With the latest refactors when we removed delegation mixin, I'm not entirely sure how this PR is affected.

state-${state} classes are now deprecated, and states are properties that reflect to attributes, so styling should be done by hooking into those attributes, instead of using 'state-something' css classes. 

Maybe you can amend your PR to just add styles for the invoker button on attribute disabled to have `color: graytext`? The disabled attribute is already correctly delegated to the button so I think this is all that's necessary to fix https://github.com/ing-bank/lion/issues/123 .

jorenbroekema:
Actually I just read that color: graytext is deprecated in CSS3. 

So perhaps we should just not care too much about this, since in the end it is up to the user to style his lightDOM, we just give the tools (attributes) to allow him to do so, and this is now possible. So perhaps we should close it. What do you think?

jorenbroekema:
Since it's been 4 weeks, I am closing this, assuming it is no longer necessary. Please open another issue/PR if you still need this.

PR #137, opened 2019-07-01
feat(overlays): create DropdownConfig

gerjanvangeest:


packages/overlays/src/LocalOverlayController.js
```diff
@@ -145,6 +146,9 @@ export class LocalOverlayController {
       this.contentNode.style.zIndex = 1;
       this.invokerNode.setAttribute('aria-expanded', true);
 
+      if (this.isDropdown) {
```
gerjanvangeest: I think it should be called something like `this.inheritsInvokerWidth`, any other ideas?

tlouisse: Please look at the document you submitted yourself: https://github.com/ing-bank/lion/blob/master/packages/overlays/docs/OverlaySystemScope.md

I would consider this part of the placement config and expect namings to be in line with those already chosen by Popper. From a generic perspective, there is not always an invoker, but a relative positioning element

tlouisse: This doc suggests 'bottom-fill', in line with Popper placement. As @jorenbroekema argued, Popper doesn't adjust dimensions, solely positions. In that case we can treat it as some kind of preprocessor and then place it as a 'bottom-middle' (at the current level (direct member of config object)).

packages/overlays/src/LocalOverlayController.js
```diff
@@ -145,6 +149,9 @@ export class LocalOverlayController {
       this.contentNode.style.zIndex = 1;
       this.invokerNode.setAttribute('aria-expanded', true);
 
+      if (this.inheritsReferenceObjectWidth) {
+        this.contentNode.style.width = `${this.invokerNode.getBoundingClientRect().width}px`;
```
tlouisse: We should also think about min width and max width here

packages/overlays/stories/local-overlay-placement.stories.js
```diff
@@ -120,6 +122,41 @@ storiesOf('Local Overlay System|Local Overlay Placement', module)
         ${popupController.invoker} ${popupController.content}
       </div>
     `;
+  })
+  .add('DropdownConfig', () => {
+    const popupController = overlays.add(
+      new LocalOverlayController(
+        Object.assign(DropdownConfig(), {
```
tlouisse: Any reason not to use spread operator?

packages/overlays/src/DropdownConfig.js
```diff
@@ -0,0 +1,17 @@
+export const DropdownConfig = () => ({
+  isGlobal: false,
+  hidesOnOutsideClick: true,
+  hidesOnEsc: true,
+  popperConfig: {
+    placement: 'bottom',
+    inheritsReferenceObjectWidth: true,
```
tlouisse: New prop should go one level up. Also see general discussion. I think this object should only contain original Popper config as stated in their docs.

packages/overlays/src/LocalOverlayController.js
gerjanvangeest: I will investigate this

PR #139, opened 2019-07-01
Enforce icon security using tagged templates

bashmish:
BREAKING CHANGE: Icon definition requires a function using a tag passed via arguments:

```js
// myicon.svg.js

// before
export default '<svg>...</svg>';

// after
export default tag => tag`<svg>...</svg>`;
```

Application developers have an alternative shortcut to use in-place with lit-html:

```js
// MyComponent.js

// before
render() {
  return html`
    <lion-icon .svg="${'<svg>...</svg>'}"></lion-icon>
  `;
}

// after
render() {
  return html`
    <lion-icon .svg="${html`<svg>...</svg>`}"></lion-icon>
  `;
}
```


packages/icon/test/lion-icon.test.js
```diff
@@ -6,6 +6,47 @@ import hammerSvg from './hammer.svg.js';
 import '../lion-icon.js';
 
 describe('lion-icon', () => {
+  it('supports svg icon as a function which recieves a tag function as an argument and returns a tagged template literal', async () => {
+    const iconFunction = tag => tag`<svg id="svg"></svg>`;
```
tlouisse: Maybe the test should give a better example than an id, because of potential a11y problems that arise (because icon is put in light dom)

packages/icon/src/LionIcon.js
```diff
@@ -84,17 +84,17 @@ export class LionIcon extends LitElement {
   set svg(svg) {
```
tlouisse: can we rename the param to `svgObject`?

packages/icon/src/LionIcon.js
```diff
@@ -111,8 +111,24 @@ export class LionIcon extends LitElement {
...
-  _renderSvg(svgOrModule) {
-    const svg = svgOrModule && svgOrModule.default ? svgOrModule.default : svgOrModule;
-    this.innerHTML = svg;
+  _renderSvg(svg) {
+    this.__validateSvg(svg);
+    render(svg, this);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  __unwrapSvg(wrappedSvg) {
```
tlouisse: maybe make this and the helper below static?

packages/icon/src/LionIcon.js
bashmish: sure

packages/icon/src/LionIcon.js
bashmish: ok

packages/icon/test/lion-icon.test.js
bashmish: I don't mind, can you suggest smth concise and simple?

tlouisse: maybe another attr like `data-identifier`

bashmish: done

bashmish:
@daKmoR I had to use `feat` commit type and change `BREAKING CHANGE` to `BREAKING` in the commit message to "explain" our build to bump the MINOR version, because `BREAKING CHANGE` caused `lion-icon@1.0.0` which is not intended and `fix` was obviously not enough to bump the MINOR. Not ideal of course. I guess this is due to the lerna way of working with conventional commits, but would be nice to opt-out of such behavior when needed till we are on `0.x.x` and make it possible to manually specify the version, like `lion-icon@0.2.0` in this case.

Issue #130, opened 2019-06-27
[form] errorState is not recomputed properly


tlouisse:
## Description

Having an `lion-form` with one `lion-input` marked as required, when this `lion-input` stops being rendered the form shouldn't reflect an error state.

## Expected Behavior

`lion-form` can unregister inputs that stops being rendered in the form, so then the `errorState` is correct when the form is submitted.

## Current Behavior

`lion-form` doesn't unregister inputs that stops being rendered in the form, so then the `errorState` doesn't reflect the real situation.

## Steps to Reproduce

I have created a web component and a test for reproducing the bug, here the steps in the case that you want to test it manually:

1- Click **Submit**, `errorState` should be true because the ing-input "name" is required.

2- Click in the button **Toggle is name visible**

3- The ing-input "name" is not rendered anymore

4- Click **Submit**, the `errorState` should be false since there isn't any required input within the form

PR #140, opened 2019-07-01
Fix/reset validation state

tlouisse:
fixes https://github.com/ing-bank/lion/issues/130

### DIFF
--- a/packages/validate/src/ValidateMixin.js
+++ b/packages/validate/src/ValidateMixin.js
@@ -386,10 +386,7 @@ export const ValidateMixin = dedupeMixin(
        */
       validate() {
         if (this.modelValue === undefined) {
-          this.constructor.validationTypes.forEach(type => {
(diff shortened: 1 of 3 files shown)


packages/validate/src/ValidateMixin.js
```diff
@@ -385,7 +385,13 @@ export const ValidateMixin = dedupeMixin(
        * Other transitions (from Warning/Info) are not followed by a success message
        */
       validate() {
-        if (this.modelValue === undefined) return;
+        if (this.modelValue === undefined) {
+          this.constructor.validationTypes.forEach(type => {
```
jorenbroekema: Perhaps it would be clearer to read if line 389-392 were its own function '_resetValidationStates'

tlouisse: Yeah, bit more readable indeed. Done

Issue #131, opened 2019-06-27
[input-date] Empty date value on form data is `Unparseable`


tlouisse:
## Expected Behavior

On previous versions, I was able to use the `<lion-input-date>` without a default value (`.modelValue`), and leave it blank when submitting the form since it's optional.

## Current Behavior

Upon loading of the page that contains the form, if I try to submit the form (with empty date), the date input from the form data is identified as `Unparseable`. If I try to input a valid date, then remove the value, upon submitting the form, it is then able to allow an empty date input.

## Steps to Reproduce

1. Add an `lion-input-date` element to a form.

2. Leave the input date empty

3. Submit the form

If you check the contents of the form data from the submit method, it contains a date property with `Unparseable` as its value.

PR #141, opened 2019-07-01
Fix/handle empty model values

tlouisse:
fixes https://github.com/ing-bank/lion/issues/131

### DIFF
--- a/packages/field/src/FormatMixin.js
+++ b/packages/field/src/FormatMixin.js
@@ -171,11 +171,37 @@ export const FormatMixin = dedupeMixin(
       }
 
       __callParser(value = this.formattedValue) {
-        let result;
-        if (typeof value === 'string') {
-          result = this.parser(value, this.formatOptions);
+        // A) check if we need to parse at all
+
+        // A.1) The end user had no intention to parse
+        if (value === '') {
+          // Ideally, modelValue should be undefined for empty strings.
+          // For backwards compatibility we return an empty string:
+          // - it triggers validation for required validators (see ValidateMixin.validate())
+          // - it can be expected by 3rd parties (for instance unit tests)
+          // TODO: In a breaking refactor of the Validation System, this behaviot can be corrected.
+          return '';
         }
-        return typeof result !== 'undefined' ? result : new Unparseable(value);
+
+        // A.2) Handle edge cases We might have no view value yet, for instance because
+        // inputElement.value was not available yet
+        if (typeof value !== 'string') {
+          // This means there is nothing to find inside the view that can be of
+          // interest to the Application Developer or needed to store for future form state
+          // retrieval.
+          return undefined;
+        }
+
+        // B) parse the view value
+
+        // - if result:
+        // return the successfully parsed viewValue
+        // - if no result:
+        // Apparently, the parser was not able to produce a satisfactory output for the desired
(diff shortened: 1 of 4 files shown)


packages/field/src/FormatMixin.js
```diff
@@ -171,9 +171,26 @@ export const FormatMixin = dedupeMixin(
...
+
+        if (typeof value !== 'string') {
+          // This means there is nothing to find inside the view that can be of
+          // interest to the Application Developer or needed to store for future form state
+          // retrieval.
+          return undefined;
+        }
+
+        const result = this.parser(value, this.formatOptions);
+        if (!result) {
```
jorenbroekema: I feel like we're doing double work here. We return a new Unparseable if the result is falsy, and then we have another check that returns a new Unparseable of result is if type 'undefined', which would already be caught by the falsy check above. We should combine the conditions into 1 conditional.

packages/field/src/FormatMixin.js
```diff
@@ -171,9 +171,26 @@ export const FormatMixin = dedupeMixin(
       }
 
       __callParser(value = this.formattedValue) {
-        let result;
-        if (typeof value === 'string') {
-          result = this.parser(value, this.formatOptions);
+        if (value === '') {
+          // Ideally, modelValue should be undefined for empty strings.
+          // For backwards compatibility we return an empty string.
```
jorenbroekema: Can you explain the backwards compatibility? We did not have a value = empty string check before right?
I also would expect the parser to parse `''` --> `''` , is this not the case?

packages/field/src/FormatMixin.js
```diff
@@ -171,9 +171,26 @@ export const FormatMixin = dedupeMixin(
...
+          return '';
+        }
+
+        if (typeof value !== 'string') {
```
jorenbroekema: Scenario? When could this happen? I thought native inputs values were always of type String.

packages/field/src/FormatMixin.js
tlouisse: No the parser (let's say we have a date) either returns a valid Date or undefined.
Before we introduced Unparseable, it used to return '' indeed. This bugfix restores that behaviors
From a 'conceptual' perspective (whether we have a value in our model) undefined would fit better. Think we can do this small change in the next breaking release)

packages/field/src/FormatMixin.js
tlouisse: More exceptional cases (extending layers) that call `__callParser` externally. Although chances are slim and it's mainly there for backwards compatibility.

packages/field/src/FormatMixin.js
tlouisse: That's what happens when you have to cherry-pick commits you worked on two weeks ago.
In this branch everything was still right: https://github.com/ing-bank/lion/blob/fix/formIssues/packages/field/src/FormatMixin.js#L219

Good catch though

packages/field/src/FormatMixin.js
tlouisse: In the next release it should become 
`if (typeof value !== 'string' || value = '') {
  return undefined;
}`

And I agree that the typeof check doesn't make sense if you expect a string always.
I think it also covers for edge cases (this.value is delegated to this.inputElement.value but inputElement not available yet). Defensive code that's there for backw. compat.

PR #142, opened 2019-07-01
Add Popper documentation

jorenbroekema:


tlouisse:
Nice docs +1 You only need to get rid of `. only`

PR #147, opened 2019-07-02
fix(field): return undefined for non formattable values

tlouisse:
fixes https://github.com/ing-bank/lion/issues/129

### DIFF
--- a/packages/field/src/FormatMixin.js
+++ b/packages/field/src/FormatMixin.js
@@ -252,8 +252,11 @@ export const FormatMixin = dedupeMixin(
         // imperatively, we DO want to format a value (it is the only way to get meaningful
         // input into `.inputElement` with modelValue as input)
 
-        if (this.__isHandlingUserInput && this.errorState && this.inputElement) {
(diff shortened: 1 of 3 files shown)


daKmoR:
I added a failing test... (pushed a commit)
```js
  it('leaves partial (and therefore unparseable) inputs untouched', async () => {
      <lion-input-date .value=${'2017/06'}></lion-input-date>
    `);
    expect(el.value).to.equal('2017/06');
  });
```

tlouisse:
> I added a failing test... (pushed a commit)
> 
> ```js
>   it('leaves partial (and therefore unparseable) inputs untouched', async () => {
>     const el = await fixture(html`
>       <lion-input-date .value=${'2017/06'}></lion-input-date>
>     `);
>     expect(el.value).to.equal('2017/06');
>   });
> ```

Cool. Since this also happens on master I will extract this as a separate bug in a separate branch.

daKmoR:
this has been addressed via other merge requests 👍 

PS: if something is still missing please feel free to reopen or create another issue or merge request

PR #148, opened 2019-07-03
fix(field): restore serialized and unparseable values

tlouisse:


jorenbroekema:
Got a story on our backlog for this that:
- has a link to this PR for reference
- has the tests copied in the story
- Blocked until we are done refactoring formatmixin

daKmoR:
we should use serializeValue for restoring from database

PR #150, opened 2019-07-04
refactor: Remove ObserverMixin from LionSteps

JaySunSyn:


bashmish:
Hey @JaySunSyn , this is quite a breaking change and we already have [another MR](https://github.com/ing-bank/lion/pull/9) created a long time ago with these fixes. We planned this update for all components altogether and will implement finish it later when our consumers are ready for these breaking changes, not the best time to break things now for just one component.

JaySunSyn:
Didn't see the other PR - makes sense to close this one. (Not sure though why this is a breaking change though)

bashmish:
@JaySunSyn because we consider methods starting with 1 underscore as "protected" by convention, meaning they can be used on the extension layer. And you changed the interface there, which makes sense taking into account your refactoring: `ObserverMixin` has pretty weird interface there currently.

image

JaySunSyn:
Do you then consider two underscores as private?

bashmish:
@JaySunSyn yes, one `_` is "protected" and two `__` is "private", convention-wise

PR #151, opened 2019-07-08
fix(field): make sure RegistrationSystem works well with ShadyDom

gerjanvangeest:
This introduces an "update" to the registration system
It is now 2 way... 
1. the form (Registrar) Element which fires an even on the body when it is ready to accepts registrations
2. the field (Registering) Element which waits for the body event and then fires it's registration

The change is needed as ShadyDom can only catch events once fully rendered... e.g. the child needs to wait until it's parent is ready... 

it is implemented via 2 mixins
1. FormRegistrarMixin
2. FormRegisteringMixin

### DIFF
--- a/packages/field/test/FormRegistrationMixins.test.js
+++ b/packages/field/test/FormRegistrationMixins.test.js
@@ -1,0 +1,106 @@
+import { expect, fixture, html, defineCE, unsafeStatic } from '@open-wc/testing';
+import sinon from 'sinon';
+import { LitElement, UpdatingElement } from '@lion/core';
+
+import { FormRegisteringMixin } from '../src/FormRegisteringMixin.js';
+import { FormRegistrarMixin } from '../src/FormRegistrarMixin.js';
+
+describe('FormRegistrationMixins', () => {
+  before(async () => {
+    const FormRegistrarEl = class extends FormRegistrarMixin(UpdatingElement) {};
+    customElements.define('form-registrar', FormRegistrarEl);
+    const FormRegisteringEl = class extends FormRegisteringMixin(UpdatingElement) {};
+    customElements.define('form-registering', FormRegisteringEl);
+  });
+
+  it('can register a formElement', async () => {
+    const el = await fixture(html`
+      <form-registrar>
+        <form-registering></form-registering>
+      </form-registrar>
+    `);
+    await el.registrationReady;
+    expect(el.formElements.length).to.equal(1);
+  });
+
+  it('supports nested registrar', async () => {
+    const el = await fixture(html`
+      <form-registrar>
+        <form-registrar>
+          <form-registering></form-registering>
+        </form-registrar>
+      </form-registrar>
+    `);
+    await el.registrationReady;
+    expect(el.formElements.length).to.equal(1);
+    expect(el.querySelector('form-registrar').formElements.length).to.equal(1);
+  });
+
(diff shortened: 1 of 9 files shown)


packages/field/src/RegisterFormElementMixin.js
```diff
@@ -0,0 +1,59 @@
...
+       * Makes sure our parentFormGroup has the most up to date resetModelValue
+       * FormGroups will call the same on their parentFormGroup so the full tree gets the correct
+       * values.
+       *
+       * Why next frame?
+       * @see {@link this._registerFormElement}
+       */
+      _requestParentFormGroupUpdateOfResetModelValue() {
+        this.updateComplete.then(() => {
+          if (this.__parentFormGroup) {
```
daKmoR:
```suggestion
          if (this.__parentFormGroup && this.__parentFormGroup._updateResetModelValue) {
```

daKmoR: we probably also want to add a test for it?

gerjanvangeest: Ah, yes. let me look into it.

packages/field/src/RegisterFormElementMixin.js
```diff
@@ -0,0 +1,59 @@
...
+      connectedCallback() {
+        super.connectedCallback();
+        this._registerFormElement();
+        this._requestParentFormGroupUpdateOfResetModelValue();
+      }
+
+      /**
+       * Fires a registration event in the next frame.
+       *
+       * Why next frame?
```
tlouisse: next frame -> updateComplete

packages/field/src/RegisterFormElementMixin.js
```diff
@@ -0,0 +1,59 @@
...
+          );
+        });
+      }
+
+      /**
+       * Makes sure our parentFormGroup has the most up to date resetModelValue
+       * FormGroups will call the same on their parentFormGroup so the full tree gets the correct
+       * values.
+       *
+       * Why next frame?
```
tlouisse: next frame -> updateComplete

packages/field/test/FormRegistrationMixins.test.js
```diff
@@ -0,0 +1,102 @@
...
+    const el = fixtureSync(html`
+      <${tagWrapper} @form-element-register=${registerSpy}>
+        <form-registering></form-registering>
+      </${tagWrapper}>
+    `);
+    await el.registrationReady;
+    expect(registerSpy.callCount).to.equal(1);
+  });
+
+  it('can by caught by listening in the appropriate dom', async () => {
```
jorenbroekema:
```suggestion
  it('can be caught by listening in the appropriate dom', async () => {
```

packages/field/src/FormRegistrationMixins.js
```diff
@@ -0,0 +1,77 @@
...
+      firstUpdated(c) {
+        super.firstUpdated(c);
+        this.__setupRegistrationHook();
+      }
+
+      __setupRegistrationHook() {
+        if (document.body.__lionFormRegistrationReady) {
+          this._registerFormElement();
+        } else {
+          document.body.addEventListener('lion-form-registration-ready', () => {
```
jorenbroekema: I think this creates a problem if I have more than 1 lion-form on the page. Can we add a unique identifiers to it to support multiple forms?

PR #158, opened 2019-07-12
fix(field): no DomHelpers for FormControl and apply all needed mixins

daKmoR:
- requirement 1 for Rich Select
- let's merge it all to develop and then release

### DIFF
--- a/packages/field/src/FormControlMixin.js
+++ b/packages/field/src/FormControlMixin.js
@@ -1,4 +1,4 @@
-import { html, css, nothing, dedupeMixin } from '@lion/core';
+import { html, css, nothing, dedupeMixin, SlotMixin } from '@lion/core';
 import { ObserverMixin } from '@lion/core/src/ObserverMixin.js';
 
 /**
@@ -14,7 +14,7 @@ import { ObserverMixin } from '@lion/core/src/ObserverMixin.js';
 export const FormControlMixin = dedupeMixin(
   superclass =>
     // eslint-disable-next-line no-shadow, no-unused-vars
-    class FormControlMixin extends ObserverMixin(superclass) {
+    class FormControlMixin extends ObserverMixin(SlotMixin(superclass)) {
       static get properties() {
         return {
           ...super.properties,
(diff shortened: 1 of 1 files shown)


packages/field/src/FormControlMixin.js
```diff
@@ -14,7 +14,7 @@ import { ObserverMixin } from '@lion/core/src/ObserverMixin.js';
 export const FormControlMixin = dedupeMixin(
   superclass =>
     // eslint-disable-next-line no-shadow, no-unused-vars
-    class FormControlMixin extends ObserverMixin(superclass) {
+    class FormControlMixin extends ObserverMixin(SlotMixin(superclass)) {
```
bashmish: why?

packages/field/src/FormControlMixin.js
```diff
@@ -76,7 +76,31 @@ export const FormControlMixin = dedupeMixin(
...
+          this.__helpTextElementCache = [...this.children].find(el => el.slot === 'help-text');
+        }
+        return this.__helpTextElementCache;
+      }
+
+      get feedbackElement() {
+        if (!this.__feedbackElementCache) {
+          this.__feedbackElementCache = [...this.children].find(el => el.slot === 'feedback');
+        }
+        return this.__feedbackElementCache;
```
bashmish: some thoughts

1. why remove $$slot helper which does the caching and reimplement it everythere? same situation as with other helpers/mixins we are removing for some reason
2. even if we remove it, I would not bother and do this optimization, it hardly effects the performance, or if it will, we will make optimisations later
3. and most important: why not just `this.querySelector('[slot="my-slot-name"]')`?

packages/field/src/FormControlMixin.js
```diff
@@ -76,7 +76,19 @@ export const FormControlMixin = dedupeMixin(
       }
 
       get inputElement() {
-        return (this.$$slot && this.$$slot('input')) || this.querySelector('[slot=input]'); // eslint-disable-line
+        return this.__getDirectSlotChild('input');
+      }
+
+      get labelElement() {
```
tlouisse: I know it's already merged, but....:

- Are we sure none of our consumers uses $$slot?
- Shouldn't these be underscored/protected (meant to be used by Subclassers)?
- I think in all other components you introduced `_my**Node**` instead of `my**Element**`. Maybe align on this as well?

packages/field/src/FormControlMixin.js
```diff
@@ -574,5 +588,9 @@ export const FormControlMixin = dedupeMixin(
       addToAriaDescription(id) {
         this._ariaDescribedby += ` ${id}`;
       }
+
+      __getDirectSlotChild(slotName) {
```
tlouisse: This helper will be needed in every component querying slots. Maybe add to SlotMixin?

PR #160, opened 2019-07-14
fix(validate): remove deprecated CssClassMixin

daKmoR:
- requirement 3 for Rich Select
- let's merge it all to develop and then release

=> has 4 commits
- chore(validate): add tests for state css classes
- fix(validate): remove deprecated CssClassMixin
- fix(validate): add attributes for validation states
- fix(field): remove unused EventMixin from FormatMixin

probably easier to review them individually

### DIFF
--- a/packages/validate/test/ValidateMixin.test.js
+++ b/packages/validate/test/ValidateMixin.test.js
@@ -143,6 +143,58 @@ describe('ValidateMixin', () => {
     expect(otherValidatorSpy.calledWith('foo')).to.equal(true);
   });
 
+  // classes are added only for backward compatibility - they are deprecated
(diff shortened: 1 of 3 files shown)


packages/validate/test/ValidateMixin.test.js
```diff
@@ -143,6 +143,100 @@ describe('ValidateMixin', () => {
     expect(otherValidatorSpy.calledWith('foo')).to.equal(true);
   });
 
+  // classes are added only for backward compatibility - they are deprecated
```
tlouisse: Do we need tests for attribute reflection?
From the App Developer perspective it makes sense to keep them in specs, since we only reflect a subset of our props, for styling purposes...

Issue #152, opened 2019-07-08
[localize] Chrome translate feature breaks LocalizeManager


hanmarkslag:
## Description

When you use the Chrome translate feature, you can translate a webpage to whatever language that you can choose from the list that Chrome provides. With this the `<html lang>` attribute will change accordingly. The `LocalizeManager` will then use the new value in the `get locale()` function from the `<html lang>` attribute and hereby will provide a language that might not be supported by an app. Moreover the main reason to use Chrome translate would be that there is no support for such a language by app authors, so we practically break Chrome translate where it is of the most use. 

## Steps to reproduce

It's difficult to reproduce it with our current storybook setup, but still these steps can be repeated in many apps where our components are used:

1. Open a page
2. Translate this page using Google Chrome translate feature to smth unsupported (e.g. Portuguese), Google Chrome automatically sets  `<html lang="pt">`
3. Render a component having translations (let's assume it supports English only)
4. The component will try to load Portuguese translations instead of English, because Google Chrome set the `<html lang="pt">`
5. But we don't support Portuguese and the `LocalizeManager` gives an error

So basically instead of the error in 5 we want to load English and let Chrome translate the English into Portuguese.

## Proposal

A solution could be to provide a list of supported languages with a (supported) fallback language or (when the LocalizeManager is able to detect a unsupported language), to be able to fallback to a supported fallback language.

bashmish:
I believe we need to implement a fallback language to support multiple scenarios:

1. The one described above
2. The apps where some components were translated to English, but NOT YET to other languages the app needs to support, so that the app continues to work with some parts still in English.

I think we should make this configurable, so that app develops can set a fallback language. Not sure if we want to have English as a default, or let it empty and keep the current default behavior.

PR #164, opened 2019-07-15
Support fallback locale

bashmish:
We need this to support multiple scenarios:

1. The apps where some components were translated to English, but NOT YET to other languages the app needs to support, so that the app continues to work with some parts still in English (now it throws and breaks which is not what most people would like to see)
2. Fix [this bug](https://github.com/ing-bank/lion/issues/152) which breaks Google Chrome translate

So this is both a new feature and a bug fix :P

For people who don't like such a default I implemented it as a configurable option. They will be able to opt-out from this behavior if needed, or change the default (currently 'en-GB') by changing the singleton using `setLocalize`.


hanmarkslag:
Thanks for the solution!

PR #181, opened 2019-07-16
[wip] [field] initial sync faling test

tlouisse:
might be related to/fixed together with: https://github.com/ing-bank/lion/pull/55

todo:
- make test case generic for formatMixin/LionField
- change commit name


gerjanvangeest:
behavior is correct now - test still fails but apparently wrong... if a real issue please create one

PR #182, opened 2019-07-17
Update testing configs for better DX

bashmish:
This will improve DX when debugging in older browsers, specifically you'll be able to look at individual files in the debugger like if you are working with native ES modules.


jorenbroekema:
Hmm so the popper tests fail after this, I guess something in the compatibility-mode changes how the elements are rendered.

bashmish:
Yeah, @LarsDenBakker do you have any idea why the rendered elements are in the bottom instead of the top of the page while the have absolute position in styles?

LarsDenBakker:
Nope, no idea about that. Are you sure this is different between compatibility mode and non-compatibility mode?

You can put a debugger statement inside your test and run karma in debugger mode in the browser. Then you can just inspect what's going on.

For these cases I think it's actually better if the open-wc fixture prepends the elements to the body (inject it on top) instead of appending (injecting to bottom). that way the css is more deterministic.

What do you think @daKmoR ?

daKmoR:
I thought about putting `position: absolute; top: 0; left: 0;` on the fixture wrapper... adding it to the top of the body might be even nicer 🤔

bashmish:
@daKmoR I believe it was always like this with mocha regardless of where you add a fixture wrapper in the DOM, but now I understand that that was probably a side effect of some mocha default styles which might have been changed. I like your idea, but we still need to investigate what changed causing these failures, and I also would prefer to not see mocha output jumping in the browser because smth is added before it.

jorenbroekema:
Looks like it's failing to run the tests now due to @open-wc's eslint config requiring eslint 6 or higher, which requires a specific node version (at least 10.13 and up and it has 10.12) on circle-ci, is this something we can fix on our end?

Kinda wanted to spend some time to see if I can make a workaround for these popper tests. @daKmoR  any news on whether you guys want to change the way fixtures work in open-wc with the whole prepending / appending to body etc?

LarsDenBakker:
Made a proposal for differently positioned fixtures https://github.com/open-wc/open-wc/pull/744

daKmoR:
ok I updated it with the following things
- chore: update @open-wc/testing to latest version => finally with added type definitions for nice docu in vs code; also speed optimizations; and allowing to use (polyfilled) es modules even on IE11 when debugging
- chore(form-system): add a main file for node.resolve => having a main in package.json point to a file that does not exist is bad
- chore: update ci to node 10.16; use local chrome for speed => it's way fast and as we use chrome only for now let's go for it
- chore: update linting => we updated the config but still used the old eslint version so some rules didn't apply fixed that

all in all this reduces the test time from ~3:00 to ~1:30 so that is a very nice improvement 💪

Issue #184, opened 2019-07-18
[textarea] max-rows allows more rows than it should


mario-aleo:
Hey there :)

I was checking out the storybook and I've seen that the Stop Growing sample on the textarea max-rows was alloying more rows than it should, so I've downloaded the textarea to try it and I've found out that if you have an initial value defined to the textarea when it's created or if you update the max-rows attribute after inputting some rows the max-rows is going to allow more rows than expected, it seems like the rows written aren't considered on the count, so if the max-row is 4 and the initial value is "1\n2\n3" (3 lines) it going to grow until line 7 (3 initial + 4 post-written) before creating the scroll.

Let me know if there's anything I can help with :)

Tested on:
* Chrome 75.0.3770.142
* Opera 62.0.3331.72
* Edge Dev 77.0.218.4
* Firefox Dev: 69.0b5

bashmish:
@mario-aleo thanks for reporting this bug 👍 I confirm it 🙈 

You can definitely help by creating a MR with a fix 💪
Please follow our [contribution guildelines](https://github.com/ing-bank/lion/blob/master/CONTRIBUTING.md). I recommend to start with tests which cover this situation and fail and then implement a fix.

mario-aleo:
@bashmish nice, that sound like fun, I'll be glad to help :D

PR #189, opened 2019-07-20
Calculate max-height based on empty state

mario-aleo:
Update setTextareaMaxHeight to fix the bug reported on #184 

Let me know if there's anything that should be improved, I'll be glad to help and work on it :)


mario-aleo:
@bashmish Thank you, I was worried about the input dispatch, I'm glad that there was another way, also next time I'll make sure to commit the test before the fix, thanks for reordering it :D

Issue #165, opened 2019-07-15
[textarea] Looks resizable but isn't.


erikkroes:
The textarea element shows a small icon in the bottom right corner in several browsers. This icon indicates that it should be resizable by dragging this icon. Except it can not be dragged. 
Suggestion to make it either resizable like it suggests, or make sure the icon doesn't show. I believe the second was the intention.

PR #198, opened 2019-07-22
[textarea] Disable user resize behavior

mario-aleo:
Hey guys, I'm hoping not being a bother :)
I was taking a look at #165 and I've seen that the autosize library applies a resize horizontal style inline if the textarea element doesn't disable it first:

```
const style = window.getComputedStyle(ta, null);

if (style.resize === 'vertical') {
	ta.style.resize = 'none';
} else if (style.resize === 'both') {
	ta.style.resize = 'horizontal';
}
```

I thought that applying the resize CSS property as none inline was a bad thing, so I've added it on the slotted ` .form-control` and I've needed to move the autosize call to the firstUpdate lifecycle hook to make sure the styles were applied before calling the library.
I've could have kept it on the connectedCallback and awaited the updateComplete but it looked too noisy for me.

### DIFF
--- a/packages/textarea/src/LionTextarea.js
+++ b/packages/textarea/src/LionTextarea.js
@@ -40,7 +40,16 @@ export class LionTextarea extends ObserverMixin(LionInput) {
   get slots() {
     return {
       ...super.slots,
-      input: () => document.createElement('textarea'),
+      input: () => {
+        const input = document.createElement('textarea');
+
+        // disable user resize behavior if browser supports it
+        if (input.style.resize !== undefined) {
+          input.style.resize = 'none';
+        }
+
+        return input;
+      },
     };
   }
 
--- a/packages/textarea/test/lion-textarea.test.js
+++ b/packages/textarea/test/lion-textarea.test.js
@@ -2,6 +2,11 @@ import { expect, fixture, html } from '@open-wc/testing';
 
 import '../lion-textarea.js';
 
+function hasBrowserResizeSupport() {
+  const textarea = document.createElement('textarea');
+  return textarea.style.resize !== undefined;
+}
+
 describe('<lion-textarea>', () => {
   it(`can be used with the following declaration
   ~~~
@@ -15,6 +20,16 @@ describe('<lion-textarea>', () => {
     const el = await fixture(`<lion-textarea></lion-textarea>`);
     expect(el.rows).to.equal(2);
     expect(el.maxRows).to.equal(6);
+  });
+
+  it('disables user resize behavior', async () => {
+    if (!hasBrowserResizeSupport()) {
+      return;
+    }
+
+    const el = await fixture(`<lion-textarea></lion-textarea>`);
+    const computedStyle = window.getComputedStyle(el.inputElement);
+    expect(computedStyle.resize).to.equal('none');
   });
 
   it('supports initial modelValue', async () => {


mario-aleo:
Oh, I've forgotten to ask, should it be 'snap tested' or tested if the resize style is equaled none on a expect test?

bashmish:
@mario-aleo thanks for another helpful investigation 👍 

I think we should make a test simple: check that after rendering it has `resize = 'none' `. Can you explain what you meant by "snap tested", I'm just curious since I've never heard this term?

Can you please merge all commits into the same fix-commit? The idea behind `chore` is that it does not touch production code, so it can be used for fixes in documentation and tests, but if you touch the production code and it impacts the released package code, it should be "fix" or "feat". Also if you fix one thing and write tests for it, all should be in one commit.

mario-aleo:
@bashmish sorry, my bad, I'll merge them :D
It was my habit of squeezing words, "Snaptest" is the snapshot test, but I'll keep the idea of checking with expect since I didn't find anything related to snapshot test :)

mario-aleo:
@bashmish done, thanks for the help, I'll keep everything that you said in mind to make sure that I can smoothly help the project :D

bashmish:
@mario-aleo I checked tests in all supported browsers and looks like your fix broke IE11 :(

These 2 tests fail now:
- starts growing when content is bigger than "rows" and stops growing after property "maxRows" is reached
- stops growing after property "maxRows" is reached when there was an initial value

We run tests in IE11 and some other browsers manually since we can't make many flaky tests stable in all supported browsers (IE11, 2 latest versions of Edge, Chrome, Firefox and Safari). You can use `yarn test:legacy:watch` and open a page `http://127.0.0.1:9876/debug.html`. You can get IE11 in a VM using this https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

Will you be able to continue working on this to make it fully cross-browser including IE11?

mario-aleo:
@bashmish oh, got it, I'll start to use it to test in other browsers.
Well, I was taking a look and I've found that the resize is not supported on IE and Edge, at least not until the Chromium version (my bad, I've checked there instead of the Microsoft version).
Before I gave up, I'll try to raid the depths of the internet dungeon (aka documentation) and, hopefully, find an 'ms-' option or something like it... anyhow, I'll let you know when I'm done ;D

mario-aleo:
@bashmish for some awkward reason the use of `firstUpdated` or `updateComplete` was messing with the slot render on IE, making it be rendered latter than it should, by doing so it started that bug since it couldn't calc what was the necessary max-height to be applied.
Unfortunately, the only wait to still apply the resize css property before the autosize library is to apply it directly to the element, I was avoiding it as much as I could, but I think it's the cleaner way that's left.

mario-aleo:
My commits have been unsquashed, I'm squashing it back to keep it clean

bashmish:
> I was avoiding it as much as I could, but I think it's the cleaner way that's left.

Yeah, I don't like this either because then you can't change it easily via CSS using the fact that the textarea is in the Light DOM, but from the practical point of view there should be no problem. Who is going to change the `resize` CSS property while it is designed not be be resizable and it won't work as expected anyway? No one :) So let's keep like you did.

mario-aleo:
@bashmish Totally agree, thanks :D
Also, thank you for being so patient, I'm having a good time trying to help :D
Oh, before I forget, IE and Edge doesn't have the resize property 'cause they don't have that user resize behavior, so it's fine and it's not going to dispatch any error or something like it, but it's going to fail on the test when running on those browsers, I've searched for the last few minutes if there was any "if IE/Edge then" on the tests are written so I could use it and make it pass on these cases, but I couldn't find anything like, so... should I leave it this way or is there any pattern that I should use in those cases?

bashmish:
Thanks for your help! 👍 

While doing cross-browsers tests I found out that `lion-textarea` is broken in Edge :( If you want to continue with improvements for it, you can take a look here: https://github.com/ing-bank/lion/issues/203

mario-aleo:
@bashmish Nice, I'll take a look at #203 ;)
Thanks for the help with the test, I was unsure of committing a test with an if :)

PR #199, opened 2019-07-23
fix(overlays): fix popper for IE11 by using esm dist target

jorenbroekema:
https://github.com/FezVrasta/popper.js#dist-targets

I'm under the impression we should use ESM, when reviewing please double check my assumption ;). Works in IE11 as far as I can tell


daKmoR:
uh nice - makes it work on IE11 also for the select 👍 

however it has one failing test now `creates a popper instance on the controller when shown, keeps it when hidden`

can you fix it 🤗

PR #202, opened 2019-07-23
Update to latest open-wc linting and testing

jorenbroekema:
Testing didn't change/break anything for us

For linting, it takes care of this breaking change from open-wc https://github.com/open-wc/open-wc/pull/550


package.json
```diff
@@ -9,9 +9,9 @@
     "@commitlint/cli": "^7.0.0",
     "@commitlint/config-conventional": "^7.0.0",
     "@open-wc/demoing-storybook": "^0.2.0",
-    "@open-wc/eslint-config": "^0.4.5",
+    "@open-wc/eslint-config": "^1.0.0",
     "@open-wc/prettier-config": "^0.1.0",
-    "@open-wc/testing": "^0.12.5",
+    "@open-wc/testing": "^2.0.0",
```
bashmish: it has to be upgraded in all other packages as well, please search in all 'package.json' files

and I think it makes more sense to enforce the latest with `^2.0.6`

PR #204, opened 2019-07-24
feat(button): reduce to minimal css; use DisabledWithTabIndexMixin

daKmoR:
this is hopefully the last requirement for select 🙈 💪 

- add a DisabledMixin and DisabledWithTabIndexMixin that can reliably manage disabled and tabindex (e.g. user can change in between the tabindex (but it will not be "reflected" e.g. disabled will always have tabindex -1) and after disabled gets "revoked" it goes to the last set state and not the original one)
- use that DisabledWithTabIndexMixin in Button
- add before/after template for button (which we will then use/override in the select invoker)
- reduce the stylings of the Button to be very minimalist so we can actually extend them where needed

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@lion/button",
   "version": "0.1.48",
-  "description": "A button that is easily stylable and accessible in all contexts",
+  "description": "A button that is easily styleable and accessible in all contexts",
   "author": "ing-bank",
   "homepage": "https://github.com/ing-bank/lion/",
   "license": "MIT",
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -48,22 +48,20 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
           display: inline-block;
           padding-top: 2px;
           padding-bottom: 2px;
-          height: 40px; /* src = https://www.smashingmagazine.com/2012/02/finger-friendly-design-ideal-mobile-touchscreen-target-sizes/ */
+          min-height: 40px; /* src = https://www.smashingmagazine.com/2012/02/finger-friendly-design-ideal-mobile-touchscreen-target-sizes/ */
           outline: 0;
           background-color: transparent;
           box-sizing: border-box;
         }
 
         .btn {
-          height: 24px;
+          min-height: 24px;
           display: flex;
           align-items: center;
           position: relative;
-          border: 1px solid black;
-          border-radius: 8px;
-          background: whitesmoke;
-          color: black;
+          background: #eee; /* minimal styling to make it recognizable as btn */
           padding: 7px 15px;
+          outline: none; /* focus style handled below, else it follows boundaries of click-area */
         }
 
         :host .btn ::slotted(button) {
@@ -81,27 +79,20 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
           padding: 0;
         }
 
-        :host(:focus) {
-          outline: none;
-        }
-
         :host(:focus) .btn {
-          border-color: lightblue;
-          box-shadow: 0 0 8px lightblue, 0 0 0 1px lightblue;
+          /* if you extend, please overwrite */
+          outline: 2px solid #bde4ff;
         }
 
         :host(:hover) .btn {
-          background: black;
-          color: whitesmoke;
-        }
-
-        :host(:hover) .btn ::slotted(lion-icon) {
-          fill: whitesmoke;
+          /* if you extend, please overwrite */
+          background: #f4f6f7;
         }
 
         :host(:active) .btn,
         .btn[active] {
-          background: grey;
+          /* if you extend, please overwrite */
+          background: gray;
         }
 
         :host([disabled]) {
@@ -109,10 +100,10 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
         }
 
         :host([disabled]) .btn {
+          /* if you extend, please overwrite */
           background: lightgray;
-          color: gray;
-          fill: gray;
-          border-color: gray;
+          color: #adadad;
+          fill: #adadad;
         }
       `,
     ];
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -22,11 +22,23 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
   render() {
     return html`
       <div class="btn">
+        ${this._renderBefore()}
         <slot></slot>
+        ${this._renderAfter()}
         <slot name="_button"></slot>
         <div class="click-area" @click="${this.__clickDelegationHandler}"></div>
       </div>
     `;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  _renderBefore() {
+    return html``;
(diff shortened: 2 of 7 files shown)


packages/field/test/DisabledMixin.test.js
```diff
@@ -0,0 +1,173 @@
...
+    expect(el.disabled).to.be.true;
+
+    el.makeRequestToBeDisable();
+    el.disabled = false;
+    el.retractRequestToBeDisable();
+    expect(el.disabled).to.be.false;
+  });
+
+  describe('Can manage TabIndex', () => {
+    it('has an initial tabIndex of 0', async () => {
```
gerjanvangeest: Why does it set tabindex=0 by default?
Is everything that gets disabled also a focussable element?

packages/button/src/LionButton.js
```diff
@@ -69,50 +76,33 @@ export class LionButton extends DelegateMixin(SlotMixin(LionLitElement)) {
...
         }
 
         :host([disabled]) .btn {
+          /* up to Application Developers and SubClassers */
           background: lightgray;
-          color: gray;
-          fill: gray;
-          border-color: gray;
+          color: #adadad;
+          fill: #adadad;
```
bashmish: if we start removing unnecessary styles, maybe we can remove even more

this is the difference in Chrome so far between lion and native one:

image

of course the size difference is for A11Y, but I don't see why we need other differences? for cross browser consistency?

packages/field/src/DisabledMixin.js
```diff
@@ -0,0 +1,118 @@
...
+        this.__internalSetDisabled(true);
+      }
+
+      retractRequestToBeDisable() {
+        this.__requestedToBeDisabled = false;
+        this.__internalSetDisabled(this.__restoreDisabledTo);
+        if (this.constructor.disableMixinManagesTabIndex) {
+          this.__internalSetTabIndex(this.__restoreTabIndexTo);
+        }
+      }
```
bashmish: can you explain why `makeRequestToBeDisable` and `retractRequestToBeDisable` are needed and especially are public?

packages/field/src/DisabledMixin.js
```diff
@@ -0,0 +1,118 @@
...
+            type: Number,
+            reflect: true,
+            attribute: 'tabindex',
+          },
+        };
+      }
+
+      static get disableMixinManagesTabIndex() {
+        return true;
+      }
```
bashmish: do I understand correctly that it is difficult to split the logic of disabled and tabindex, so you decided to have them in one? can we then maybe consider a different name for this mixin?

I see `disableMixinManagesTabIndex` flag, but it is enabled by default which should be reflected in the name

or otherwise maybe make it `false` by default and keep the name `DisabledMixin`?

packages/button/src/LionButton.js
```diff
@@ -1,28 +1,37 @@
...
+      <slot name="icon-before"></slot>
+    `;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  _afterTemplate() {
+    return html`
+      <slot name="icon-after"></slot>
+    `;
+  }
```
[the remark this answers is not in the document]
daKmoR: we will override it in the `select-invoker` to introduce the arrow down

packages/button/package.json
```diff
@@ -32,7 +32,8 @@
     "*.js"
   ],
   "dependencies": {
-    "@lion/core": "^0.1.11"
+    "@lion/core": "^0.1.11",
+    "@lion/field": "^0.1.36"
```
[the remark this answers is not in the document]
bashmish: we can put it into the core

packages/field/src/DisabledMixin.js
```diff
@@ -0,0 +1,118 @@
...
+          this.__internalSetTabIndex(0);
+        }
+      }
+
+      makeRequestToBeDisable() {
+        this.__requestedToBeDisabled = true;
+        this.__restoreDisabledTo = this.disabled;
+        this.__restoreTabIndexTo = this.tabIndex;
+        this.__internalSetDisabled(true);
+      }
```
bashmish: looks like a bug to me, because if `makeRequestToBeDisable` is called 2 times in a row, you will loose the original state

maybe just add a guard `if (this.__requestedToBeDisabled) { return; }` im the beginning of the method

maybe start with a test for this case :)

packages/field/src/DisabledMixin.js
bashmish: turned out that I misunderstood the name of the method, so we also agreed to rename this to `managesTabIndexViaDisableMixin` where `ViaDisableMixin` - is a suffix to namespace the method, because `managesTabIndex` is too generic

and indeed the value will be `false` by default, and then we can keep the name `DisabledMixin`

packages/button/src/LionButton.js
bashmish: we discussed this and carefully analyzed all the CSS of the lion-button, and found only 1 thing which is not needed at all for the generic layer - grey `background-color`, but if we remove it, then the demos of `lion-button` will be difficult to understand, so we decided to keep some minimal styling even if not strictly needed for Lion in such cases where removing it is more troubles than keeping it and overriding on the extension layer

packages/button/src/LionButton.js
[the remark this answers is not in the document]
bashmish: we agreed to keep these templates empty by default (just return a blank lit-html template in both), because we just need a hook in the template, not slots themselves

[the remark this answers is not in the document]
daKmoR: done

packages/button/package.json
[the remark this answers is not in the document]
daKmoR: done

packages/field/src/DisabledMixin.js
daKmoR: changed 💪

packages/field/src/DisabledMixin.js
daKmoR: added test which showed that a guard was needed 👍

packages/core/src/DisabledMixin.js
```diff
@@ -0,0 +1,122 @@
...
+  superclass =>
+    // eslint-disable-next-line no-shadow
+    class DisabledMixin extends superclass {
+      static get properties() {
+        return {
+          disabled: {
+            type: Boolean,
+            reflect: true,
+          },
+          // we us a property here as if we use the native tabIndex we can not set a default value
```
bashmish: 'us' => 'use'

https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker 😉

daKmoR: I have it installed... but us is a correct word and it did not mark it as "wrong" 🙈

daKmoR: changed 👍

packages/field/test/DisabledMixin.test.js
daKmoR: The DisabledMixin only manages disabled by default. You can opt into also managing tabIndex which you probably only should do for elements you actually want to be focusable

packages/field/src/DisabledMixin.js
daKmoR: needed for "contracts" between select -> options or in the future also fieldset -> fields

PR #206, opened 2019-07-24
[textarea] Wait for textarea slot before init autoresize

### LINKED ISSUE #203: [textarea] Autoresize is broken in Edge 18
Looks like `autoresize` library does not work in Edge. The textarea grows without any limits and starts with 2 rows always, even if there is initial content with more lines of text and `maxRows` allows more than 2 rows.

I checked the state of `lion-textarea` component before we introduced 2 recent fixes ([fix 1](https://github.com/ing-bank/lion/pull/189) and [fix 2](https://github.com/ing-bank/lion/pull/198)) and it was broken as well, so basically it must have been broken from the very beginning or at least for a long time now.

Currently we have 2 tests that fail because of that, so the fix should come down to making those 2 tests work:

- starts growing when content is bigger than "rows" and stops growing after property "maxRows" is reached
- stops growing after property "maxRows" is reached when there was an initial value

Browsers:
- Edge 18 (at least, maybe <18 as well)

mario-aleo:
Fixes #203 

I've seen this behavior before, unfortunately, I've misled myself to the thought that it was working.

On Edge, on the `connnectedCalback`, when the lion-textarea tries to apply the `autosize` and `setTextareaMaxHeight`, the `inputElement` isn't rendered yet, and it creates this bug since both functions need the properties related to the rendered state of the element.
I've tried to wait for the `slotchange` event on the `firstUpdated` but the element is only going to be rendered on the next update so it needs to await the `updateComplete` before calling those functions.

Right now It's working and passing the tests on all the browsers that I've tested (Chrome, Opera, Firefox, Edge, and IE), but I've marked it as `WIP` so I can take a look if there is a less "cumbersome" way to accomplish it.

### DIFF
--- a/packages/textarea/src/LionTextarea.js
+++ b/packages/textarea/src/LionTextarea.js
@@ -62,8 +62,7 @@ export class LionTextarea extends ObserverMixin(LionInput) {
   connectedCallback() {
     // eslint-disable-next-line wc/guard-super-call
     super.connectedCallback();
-    this.setTextareaMaxHeight();
-    autosize(this.inputElement);
+    this.__initializeAutoresize();
   }
 
   disconnectedCallback() {
@@ -101,7 +100,39 @@ export class LionTextarea extends ObserverMixin(LionInput) {
     ];
   }
 
+  get updateComplete() {
+    if (this.__textareaUpdateComplete) {
+      return Promise.all([this.__textareaUpdateComplete, super.updateComplete]);
+    }
+    return super.updateComplete;
+  }
+
   resizeTextarea() {
     autosize.update(this.inputElement);
+  }
+
+  __initializeAutoresize() {
+    if (this.__shady_native_contains) {
+      this.__textareaUpdateComplete = this.__waitForTextareaRenderedInRealDOM().then(() => {
+        this.__startAutoresize();
+        this.__textareaUpdateComplete = null;
+      });
+    } else {
+      this.__startAutoresize();
+    }
+  }
+
+  async __waitForTextareaRenderedInRealDOM() {
+    let count = 3; // max tasks to wait for
+    while (count !== 0 && !this.__shady_native_contains(this.inputElement)) {
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise(resolve => setTimeout(resolve));
+      count -= 1;
+    }
+  }
+
+  __startAutoresize() {
+    autosize(this.inputElement);
+    this.setTextareaMaxHeight();
   }
 }


mario-aleo:
Sorry, hadn't much time to finish it yesterday, but now it's done :)
It should be working like a charm, I've it tested on:
Opera 62.0.3331.72
Firefox 69.0b7
Chrome 075.0.3770.142
Edge 18.17763
IE 11.615.17763.0

packages/textarea/src/LionTextarea.js
```diff
@@ -59,17 +59,18 @@ export class LionTextarea extends ObserverMixin(LionInput) {
...
-    this.setTextareaMaxHeight();
-    autosize(this.inputElement);
-  }
-
   disconnectedCallback() {
     autosize.destroy(this.inputElement);
   }
 
+  async firstUpdated() {
+    if ('HTMLSlotElement' in window === false) {
```
bashmish: don't IE11 and Edge have this class from the webcomponents polyfills?

packages/textarea/src/LionTextarea.js
```diff
@@ -104,4 +105,15 @@ export class LionTextarea extends ObserverMixin(LionInput) {
...
     autosize.update(this.inputElement);
   }
+
+  _slotPolyfillRendered() {
+    return new Promise(resolve => {
+      this.shadowRoot
+        .querySelector('slot[name="input"]')
+        .addEventListener('slotchange', async () => {
+          await this.updateComplete;
+          resolve();
```
bashmish: you said we need to wait for an `updateComplete` when the slot content has changed

I have difficulties to understand why, because it seems like if slot content has changed, then both the custom element + its shadow root are already rendered (otherwise there will be no slot and you count not listen) and the slot child is also rendered (because it is just a`<textarea>` added to it causing the `slotchange` event)

in other words, what is different before and after `updateComplete` which impacts the initialisation of autosize?

packages/textarea/src/LionTextarea.js
```diff
@@ -59,17 +59,18 @@ export class LionTextarea extends ObserverMixin(LionInput) {
...
-    super.connectedCallback();
-    this.setTextareaMaxHeight();
-    autosize(this.inputElement);
-  }
-
   disconnectedCallback() {
     autosize.destroy(this.inputElement);
   }
 
+  async firstUpdated() {
```
bashmish: that makes a lot of sense to finally have the initialisation in `firstUpdated`, I really don't understand how it worked before with just `connectedCallback` anywhere, not to say about Edge/IE11

what I find interesting here is that we now have `async firstUpdated()`, or in other words if someone needs to extend `LionTextarea`, they might think they can `await super.firstUpdated()` to be sure it was initialised

I think this is not a typical usage of this LitElement life-cycle hook, so it is better to not make this method async and just use a promise inside (but do not return it), so that public API does not provide a possibility to do `await super.firstUpdated()`

if someone will ever need such a hook, we can always add it later, but it should not be in `firstUpdated`, maybe it will be a modified `updateComplete`, I mean smth that is expected to be async by the user of LitElement

packages/textarea/src/LionTextarea.js
mario-aleo: Seems awkward right? I've thought it too, but It doesn't appear to be registering the slot element as HTMLSlotElement, at least I've doubled checked and it was not part of the window object after the polyfill is loaded

packages/textarea/src/LionTextarea.js
mario-aleo: I don't clearly understand either, I didn't dig deep enough on the webcomponents polyfill, but it's probably related to what is done on the LitElement's `connectedCallback`, I think it has a slightly different rendering behavior when the polyfill is needed because of the ShadyCSS:
```
connectedCallback() {
    super.connectedCallback();
    // Note, first update/render handles styleElement so we only call this if
    // connected after first update.
    if (this.hasUpdated && window.ShadyCSS !== undefined) {
      window.ShadyCSS.styleElement(this);
    }
}
```
Maybe I'll try to ask Justin Fagnani or the polymer team on twitter about the subject to see if they can clear it out for us, or/and do some tests and maybe open an issue to them, but don't worry, I'll not talk about the project since it's asked not to on the readme ;)

You're totally right, it's going to be bad to inherit later, I'll update my commit to keep it without the async and use the `then` to synchronize with the `updateComplete` ;D

packages/textarea/src/LionTextarea.js
mario-aleo: As I've seen, when there is the need to polyfill the shadow dom and/or the slot, the light dom that composes the elements to be slotted on beginning to be part of the dom on the next `slotchanged` after the `updateComplete`, probably duo the lack of a proper shadow dom.
So the difference between before and after the `updateComplete` is that there is no slotted element rendered, especially the `setTextareaMaxHeight` need it to have it rendered with all its dimensions to calculate the `max-height`;

mario-aleo: Here are the print screens to show this behavior on Egde.

Anotação 2019-07-27 093753

Anotação 2019-07-27 093907

Anotação 2019-07-27 094005

Anotação 2019-07-27 094023

Anotação 2019-07-27 094047

Anotação 2019-07-27 094119

mario-aleo: Just in case you want to see that it was already rendered on the `firstUpdated` here is the first two screens shots on Opera.

Anotação 2019-07-27 095017

Anotação 2019-07-27 095124

mario-aleo:
Sorry, I'm a bit later, the timezone difference got me this time :(
I just took a look at the final version and I agree with you guys, the code looks awesome :D
The only thing that makes me feel uneasy is the use of time-related APIs, even without using the time parameter. I would try to see with `requestAnimationFrame` could do the trick, since the insertion of the element on the `dom` is probably going to trigger a paint, if it works, it could even give us a better result than the `setTimeout`.
@bashmish, let me know what do you think about the `requestAnimationFrame` approach, If you think this could be a good solution, I can make some tests to see if it is going to actually work as good as I expect and open a new pull request :)

bashmish:
@mario-aleo I think we make a lot of assumptions here, while the only proper solution is to debug the polyfilll in Edge and find out when exactly it renders what we need. This is quite a time investment which I personally considered not to be worth it in this specific case. 🤷‍♂  Feel free to make investigation if you feel like we should improve the rendering timing here.

But we have a lot of other things to improve on 🙈 I'm  trying to be a little be more pragmatic. I would suggest you to check the list of issues, you might find there others interesting things which you'd like to work on.

mario-aleo:
@bashmish sorry, I've got carried away with investigations related to this issue xD
I'm going to check the list and take something to help with but feel free to mark me on anything that you would like me to take a look, I'm having a good time trying to help with the project :D

PR #209, opened 2019-07-26
[select-rich] Add index.js with class export

jorenbroekema:


jorenbroekema:
Uhh it's not prettier formatted I guess. I only have nano as my editor atm since I'm trying to get my new machine working so maybe someone else can quickly fix it :P

bashmish:
@jorenbroekema done. Also please use `fix` for such changes, because it changes the package production code.

jorenbroekema:
Aah true, good point

daKmoR:
should also have added
```
main: "index.js",
module: "index.js",
```
to `package.json` right?

jorenbroekema:
I think it's already there right?

daKmoR:
yeah true 👍

PR #211, opened 2019-07-29
[localize] Improve documentation

bashmish:
The documentation is currently very outdated, contains some very old APIs which are not supported, misleading recommendations and lacks some recent new features information.

### DIFF
--- a/packages/localize/README.md
+++ b/packages/localize/README.md
@@ -8,7 +8,7 @@ The formatting of data containing numbers and dates takes current locale into ac
 
 ## LocalizeManager
 
-The core of the system is a `LocalizeManager` instance which is responsible for data loading and working with this data.
+The core of the system is a `LocalizeManager` instance which is responsible for data loading and working with these data.
 It is exposed as a `localize` singleton instance.
 This ensures that the data can be cached in the single place and reused across different components and same component instances.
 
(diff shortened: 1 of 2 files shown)


tlouisse:
Nice improvements 👍 

Could you also maybe add:
- the rationale behind the immutability of a namespace object (once a namespace is created, we cannot add extra data)?



Some nitpicking:
- Since 'data' is plural, I think we should change all occurences of 'data is' to 'data are'
- Can we add secure addresses as examples? `http` -> `https`

tlouisse:
**Some more nitpicking with regard to the lifeycle callbacks...**

For me it's not totally clear what the difference is between `onLocaleChanged` and `onLocaleUpdated`.

```js
// LocalizeMixin

onLocaleChanged() {
  this.localizeNamespacesLoaded = localize.loadNamespaces(this.__getUniqueNamespaces());
  this.onLocaleUpdated();
  this.requestUpdate();
}

onLocaleUpdated() {}
```
When looking above code, the difference I see is that `this.requestUpdate()` is called after `this.requestUpdate()` and `localeChanged` before(keeping in mind subclassers call `super.onLocaleChanged`). Is this right and does this difference require separate callbacks?

Would it be an idea to change the current three callbacks to these two:
- `onLocaleUpdated` (it has the same as current onLocaleChanged and current onLocaleUpdated could be omitted)
-  `onLocaleFirstUpdated` (it is the same as current localeReady, but from the name its purpose is more clear)

bashmish:
> Some nitpicking:
> 
> * Since 'data' is plural, I think we should change all occurences of 'data is' to 'data are'
> * Can we add secure addresses as examples? `http` -> `https`

Good ones, fixed now 💪 

> Could you also maybe add:
> 
> the rationale behind the immutability of a namespace object (once a namespace is created, we cannot add extra data)?

Rationale will be in another MR, with a lot of extra info as well.

tlouisse:
> > Some nitpicking:
> > 
> > * Since 'data' is plural, I think we should change all occurences of 'data is' to 'data are'
> > * Can we add secure addresses as examples? `http` -> `https`
> 
> Good ones, fixed now 💪
> 
> > Could you also maybe add:
> > the rationale behind the immutability of a namespace object (once a namespace is created, we cannot add extra data)?
> 
> Rationale is in [another MR](https://github.com/ing-bank/lion/pull/216) with a lot of extra info.

Thanks!

PR #217, opened 2019-08-01
fix(calendar): disabled dates don't trigger @user-selected-date-changed

tlouisse:
fixes https://github.com/ing-bank/lion/issues/221


tlouisse:
> The commit message is somewhat confusing as well, I read it as, you fixed the bug where disabled dates don't trigger @user-selected-date-changed, but the bug was actually the opposite behavior.
> 
> Maybe `fix(calendar): stop disabled date triggering @user-selected-date-changed`? So it also complies with convention to start commit message with imperative verb.

Makes sense. I can edit the message after you approve ;)

PR #218, opened 2019-08-01
fix(overlays): global overlays have fixed positioning context

tlouisse:
fixes: https://github.com/ing-bank/lion/issues/220


jorenbroekema:
Does this fix an existing issue? What's the context?

tlouisse:
Added an issue describing the probem: https://github.com/ing-bank/lion/issues/220

Issue #215, opened 2019-07-31
[field] form registry system breaks when element never connects and creates a memory leak


LarsDenBakker:
The `FormRegistrarMixin` registers itself in the constructor. This creates two issues:

If the element never connects to the page, all forms on the page break because `all-forms-open-for-registration` is never fired. This can be observed simply by creating a lion form without connecting it: https://stackblitz.com/edit/lit-html-btb2i8

If you uncomment `document.createElement('lion-form');` the form will work.

The other problem is that the manager is creating a memory leak by holding forms (and their subtrees) on the page in memory. A WeakMap would be best to use here.

This is breaking all our tests because for some reason it's creating forms without connecting them. It works fine in the demo, so it seems to be something in the tests. Looking into why that's happening.

LarsDenBakker:
Turns out this is also affecting non-test code. We have a component with multiple steps, if you switch steps very quickly elements are created before they reach `firstUpdated`. This will break all forms on the page, as they're all waiting for that one element to reach `firstUpdated`.

daKmoR:
makes sense - we originally put it on constructor to minimize the delay... but yeah for these cases that's a no go 🙈 
Weakmap also sounds like a good idea 👍 

do you already have a fix?

tomherni:
Submitted a PR, hope it helps https://github.com/ing-bank/lion/pull/223

PR #223, opened 2019-08-01
fix(form): improve managing elements in Form Registrar Manager

tomherni:
Fixes https://github.com/ing-bank/lion/issues/215

Credit to @LarsDenBakker, this was pretty much his quick fix after his analysis. I applied it locally and tested my use case, and this resolves the issue for me. The form system is pretty complex so I hope this is on the right track.

I covered the memory leak in a test. In case you would like to see a test for registering elements in the connected callback also, I'd love some feedback on how/where you'd like to see this (I don't see related tests).

### DIFF
--- a/packages/field/src/FormRegistrarMixin.js
+++ b/packages/field/src/FormRegistrarMixin.js
@@ -34,6 +34,13 @@ export const FormRegistrarMixin = dedupeMixin(
         this.addEventListener('form-element-register', this._onRequestToAddFormElement);
       }
 
+      disconnectedCallback() {
+        if (super.disconnectedCallback) {
+          super.disconnectedCallback();
+        }
+        formRegistrarManager.remove(this);
+      }
+
       isRegisteredFormElement(el) {
         return this.formElementsArray.some(exitingEl => exitingEl === el);
       }
--- a/packages/field/src/FormRegistrarMixin.js
+++ b/packages/field/src/FormRegistrarMixin.js
@@ -28,10 +28,16 @@ export const FormRegistrarMixin = dedupeMixin(
         this.registrationReady = new Promise(resolve => {
           this.__resolveRegistrationReady = resolve;
         });
-        formRegistrarManager.add(this);
 
         this._onRequestToAddFormElement = this._onRequestToAddFormElement.bind(this);
         this.addEventListener('form-element-register', this._onRequestToAddFormElement);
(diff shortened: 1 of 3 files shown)


tlouisse:
Thanks @tomherni and @LarsDenBakker I'm currently looking in a more structural solution for this (the complexity of the registration and the potential bugs it causes) here: https://github.com/ing-bank/lion/pull/226

In the meantime I could merge this (probably on Monday) if this is really urgent for you at the moment.

tomherni:
@tlouisse It would be great if this issue could be resolved soon. That way, you could take your time with the restructuring. And considering the size of this PR it also won't really affect your restructuring much.

Up to you of course!

tlouisse:
@tomherni Could you have another look at the code for IE?
When I run the tests there, I get:
Screen Shot 2019-08-05 at 10 26 07

It probably conflicts with https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove.
We could consider calling it `_unregister` and then rename 'add' to `_register`?
(underscores because I think they should not be public methods)

tomherni:
@tlouisse the error was because I used [remove()](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove) in the test scenario. I didn't know you also run your tests in IE11. I adjusted the test to pass in IE. So no method renaming necessary in this case.

As for renaming the methods to make them private -- they don't seem private to me. The methods are called directly on the FormRegistrarManager instance (by FormRegistrarMixin). Also, the Mixin doesn't further expose the Manager. What do you think?

tlouisse:
> @tlouisse the error was because I used [remove()](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove) in the test scenario. I didn't know you also run your tests in IE11. I adjusted the test to pass in IE. So no method renaming necessary in this case.
> 
> ie11
> 
> As for renaming the methods to make them private -- they don't seem private to me. The methods are called directly on the FormRegistrarManager instance (by FormRegistrarMixin). Also, the Mixin doesn't further expose the Manager. What do you think?

Ah wait, we're talking about a Singleton service here and not a DOM element.
In that case a public interface makes sense indeed.
In other cases, we go for: https://github.com/ing-bank/lion/blob/develop/docs/Definitions.md
Also the `remove` method is still available then.

Thanks for the update

PR #225, opened 2019-08-02
fix: remove deprecated CssClassMixin

gerjanvangeest:


packages/form/test/lion-form.test.js
```diff
@@ -7,7 +7,7 @@ import '@lion/fieldset/lion-fieldset.js';
 import '../lion-form.js';
 
 describe('<lion-form>', () => {
-  it.skip('has a custom reset that gets triggered by native reset', async () => {
+  it('has a custom reset that gets triggered by native reset', async () => {
     const withDefaults = await fixture(html`
       <lion-form
```
daKmoR: the submit and reset events are a public api and should be tested with something like this
```js
   it('dispatches [submit, reset] events', async () => {
      <${tag}></${tag}>
    `);
    setTimeout(() => el.submit());
    const submitEv = await oneEvent(el, 'submit');
    expect(submitEv).to.be.instanceOf(Event);
    expect(submitEv.type).to.equal('submit');
    expect(submitEv.target).to.equal(el);
    expect(submitEv.bubbles).to.be.true;
    expect(submitEv.composed).to.be.false;

    setTimeout(() => el.reset());
    const resetEv = await oneEvent(el, 'reset');
    expect(resetEv).to.be.instanceOf(Event);
    expect(submitEv.type).to.equal('reset');
    expect(resetEv.target).to.equal(el);
    expect(resetEv.bubbles).to.be.true;
    expect(resetEv.composed).to.be.false;
  });
```

packages/form/src/LionForm.js
```diff
@@ -8,43 +7,53 @@ import { LionFieldset } from '@lion/fieldset';
...
+  reset() {
+    this.formElement.reset();
+  }
+
   /**
    * As we use a native form there is no need for a role
    */
   _setRole() {} // eslint-disable-line class-methods-use-this
 
+  _formElementEvents(ev) {
```
daKmoR: just a suggestion maybe do it in a similar manner like
https://github.com/ing-bank/lion/pull/228/commits/97b958fffe47e144c5d8e3d129bf5518f9d82a1c#diff-0ecc97735e01830c554cf791bf2f1ebaR76
?

tlouisse:
Closing this one, since all of it is covered here: https://github.com/ing-bank/lion/pull/228

PR #226, opened 2019-08-02
[wip] fix: form registration

tlouisse:
Also see: https://github.com/ing-bank/lion/issues/215
Fixes https://github.com/ing-bank/lion/issues/231

Alternative approach to form registration mechanism that makes the registration code simple again and forwards compatible after this issue is handled: https://github.com/Polymer/lit-element/issues/658

### How it works
It patches LitElement, that creates a shadowRoot and then renders async, which causes all events in between those (shadowRoot creation and rendering) to get lost in polyfilled browsers

Until https://github.com/webcomponents/polyfills/issues/95 is released, we need to keep this patch for LitElement. After that, we can just remove the patch mixin and everything should work as before.

TODO:
- remove legacy code with form-element-unregister events
- split up code in separate files
- make the registering logic (including LitElement patch) less obtrusive in our form core(=field package) by putting it in a separate folder
- document migration strategy when shadydom polyfill is fixed
- clean up karma changes and/or make separate commits

### DIFF
--- a/karma.bs.config.js
+++ b/karma.bs.config.js
@@ -12,12 +12,12 @@ module.exports = config => {
         project: 'lion',
       },
       browsers: [
-        'bs_win10_chrome_latest',
+        // 'bs_win10_chrome_latest',
         // Only chrome for now
         // 'bs_win10_firefox_latest',
         // 'bs_win10_edge_latest',
         // 'bs_osxmojave_safari_latest',
-        // 'bs_win10_ie_11',
+        'bs_win10_ie_11',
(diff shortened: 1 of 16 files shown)


tlouisse:
Closing this one, since the patch for LitElement creates breaking changes within our own codebase (this is fixable by moving logic from `connectedCallback` to `firstUpdated`, but it doesn't make it a scalable solution.
This [mr](https://github.com/ing-bank/lion/pull/241) is used in favor of this one

PR #228, opened 2019-08-06
fix: sync field properties instead of delegation

daKmoR:
- placed on top of https://github.com/ing-bank/lion/pull/225/ (will need to be rebased once merged)

ToDo List:
- [x] FocusMixin make focused a lit-element property with sync up/down
- [x] FocusMixin redispatch focus/blur event add focus/blur methods
- [x] FocusMixin redispatch focus/blur event add focus/blur methods
- [x] add getter/setter for selectionStart, selectionEnd
- [x] Make disabled a lit property (reflect: true) with sync down only
- [x] Make type a lit property (reflect: true) with sync down only
- [x] Make name lit property and find a mechanic that makes sense for forms
- [x] Adjust other code so all tests succeed
- [x] Integration tests with all fields, all passing
- [x] Cleanup commit history (after reviewing)

Post Cleanup:
- [ ] input-datepicker Remove workaround for disabled (remove property and delegation filter)


packages/button/src/LionButton.js
```diff
@@ -129,9 +122,14 @@ export class LionButton extends DisabledWithTabIndexMixin(
     };
   }
 
+  get _nativeButton() {
```
tlouisse: Can we make this _nativeButtonNode for consistency in Subclasser apis?
Think we agreed on this a few weeks ago :)

daKmoR: changed 👍

packages/field/src/LionField.js
```diff
@@ -82,20 +95,18 @@ export class LionField extends FormControlMixin(
...
-      _setDisabledClass: ['disabled'],
-    };
+  constructor() {
+    super();
+    this.name = '';
+    this.submitted = false;
   }
 
   connectedCallback() {
+    this._delegateInitialValueAttr();
```
tlouisse: @jorenbroekema could you add a comment here about why we want to add this before the connectedCallback? (I know it had something to do with testing the FormatMixin integrated with the LionField)?

packages/field/src/LionField.js
```diff
@@ -112,8 +123,22 @@ export class LionField extends FormControlMixin(
...
 
-  _setDisabledClass() {
-    this.classList[this.disabled ? 'add' : 'remove']('state-disabled');
+  updated(changedProps) {
+    super.updated(changedProps);
+
+    if (changedProps.has('disabled')) {
+      if (this.disabled) {
+        this.inputElement.disabled = true;
+        this.classList.add('state-disabled'); // eslint-disable-line wc/no-self-class
```
tlouisse: Can we add here that these classes are deprecated? Or else maybe a generic statement on top of the file?

packages/field/test-suites/FormatMixin.suite.js
```diff
@@ -0,0 +1,394 @@
...
+    suffix: '',
+    ...customConfig,
+  };
+
+  /**
+   * Mocks a value for you based on the data type
+   * Optionally toggles you a different value
+   * for needing to mimic a value-change.
+   */
+  function generateValueBasedOnType(opts = {}) {
```
tlouisse: Nice solution :)
Maybe we later elaborate on it in such a way that it doesn't know anything about our custom inputs.
That would mean it knows about platform types, but not about iban, email etc.

packages/field/test-suites/FormatMixin.suite.js
```diff
@@ -0,0 +1,394 @@
...
+
+      let counter = 0;
+      formatEl.addEventListener('model-value-changed', () => {
+        counter += 1;
+      });
+
+      mimicUserInput(formatEl, generateValueBasedOnType());
+      expect(counter).to.equal(1);
+
+      // Counter offset +1 for Date because parseDate created a new Date object
```
tlouisse: Maybe add a TODO here that the ideal fix would be to fix this in the hasChanged of the date inputs?

packages/validate/test/ValidateMixin.test.js
```diff
@@ -19,7 +19,6 @@ const tagString = defineCE(
   class extends ValidateMixin(LionLitElement) {
```
tlouisse: Any reason in particular we didn't use LitElement directly here?

packages/input-iban/src/formatters.js
```diff
@@ -8,7 +8,7 @@ import { friendlyFormatIBAN } from '@bundled-es-modules/ibantools';
  */
 export function formatIBAN(modelValue) {
   // defensive code because of ibantools
-  if (modelValue === '') {
+  if (!modelValue) {
```
tlouisse: Ideally, this should be a fix instead of a chore

packages/field/test-suites/FormatMixin.suite.js
```diff
@@ -0,0 +1,394 @@
...
+import { Unparseable } from '@lion/validate';
+import { FormatMixin } from '../src/FormatMixin.js';
+
+function mimicUserInput(formControl, newViewValue) {
+  formControl.value = newViewValue; // eslint-disable-line no-param-reassign
+  formControl.inputElement.dispatchEvent(new CustomEvent('input', { bubbles: true }));
+}
+
+export function runFormatMixinSuite(customConfig) {
+  // TODO: Maybe remove suffix
```
tlouisse: Suffix is meant to make your test output a bit more readable. Wrapping a suite in a describe at least makes the tests 'traceable', so omitting the suffix would be fine by me if it makes our suites simpler.

packages/validate/test/ValidateMixin.test.js
jorenbroekema: No idea, should probably change to LitElement ^^

packages/input-iban/src/formatters.js
jorenbroekema: Agreed

packages/field/src/LionField.js
jorenbroekema: `// Initially delegating the value attribute should happen before it is evaluated by the FormatMixin.` maybe?
I'm not entirely sure what test failed but it had to do with this delegateInitialValueAttr triggering a second FormatMixin loop due to given `value attribute that is delegated` being different from the `formatted value` that is synced back to the user

Issue #229, opened 2019-08-06
[localize] failing tests because of Chrome update


tlouisse:
Since latest Chrome Windows update, we have errors in our tests running in CI:

```
    at Context.equal (packages/localize/test/LocalizeMixin.test.js:253:49)
  formatNumber
    normalization
      bg-BG
        ✖ supports basics
          Chrome 76.0.3809 (Windows 10.0.0)
        AssertionError: expected '1230456,79 EUR' to equal '123 456,79 EUR'

      + expected - actual

      -1230456,79 EUR
      +123 456,79 EUR
      
    at Context.equal (packages/localize/test/number/formatNumber.test.js:282:66)    at Test.Runnable.run (node_modules/karma-mocha-snapshot/lib/adapter.js:14:17)
        ✖ normalizes group separator
          Chrome 76.0.3809 (Windows 10.0.0)
        AssertionError: expected '10234,57 EUR' to equal '1 234,57 EUR'

```

See: https://circleci.com/gh/ing-bank/lion/1217?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #230, opened 2019-08-07
fix(localize): for bg-BG locale, correct Intl output for group separator

tlouisse:
fixes https://github.com/ing-bank/lion/issues/229


packages/localize/src/number/convertZeroToSpaceForGroup.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @desc Intl uses 0 as group separator for bg-BG locale.
+ * This should be a ' '
+ * @param {{type,value}[]} formattedParts
+ * @returns {{type,value}[]} corrected formatted parts
+ */
+export function convertZeroToSpaceForGroup(formattedParts) {
```
daKmoR: can we rename it to forceSpaceInsteadOfZeroForGroup to keep the "normalize" function names similar 🙏

Issue #232, opened 2019-08-07
[input-amount] parser doesn't ignore non number values when spaces are present


tlouisse:
Currently,  a value like `foo` is intercepted, but `foo foo` is not, leading to incorrect modelValues.

PR #233, opened 2019-08-07
fix(input-amount): return undefined for non numbers with spaces

tlouisse:
fixes https://github.com/ing-bank/lion/issues/232

### DIFF
--- a/packages/input-amount/src/parsers.js
+++ b/packages/input-amount/src/parsers.js
@@ -110,6 +110,10 @@ function parseHeuristic(value) {
  * @param {object} options Locale Options
  */
 export function parseAmount(value, options) {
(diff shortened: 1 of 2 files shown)


packages/input-amount/src/parsers.js
```diff
@@ -110,6 +110,10 @@ function parseHeuristic(value) {
  * @param {object} options Locale Options
  */
 export function parseAmount(value, options) {
+  const containsNumbers = value.match(/\d/g);
```
jorenbroekema: I think this regex is readable enough that we don't need to store it in a variable with a descriptive name.

tlouisse: I considered that indeed, but the intentions (stripping non numbers) are just a bit more clear like this.

Issue #239, opened 2019-08-12
[lion-form] resetModelValue gets the last updated Modelvalue when additional inputs are added dynamically


zanfirandra:
### Context
Have a from with initially 3 inputs
1. modelValue changes dynamically for one of them
2. inputs in form are also added dynamically

I want to reset the modelValue of the entire form group so I call resetGroup function

###  What happens
Always getting the previous modelValue instead of initial values

### What should happen
When adding another input field dynamically and changing the modelValue, when the resetGroup function is called, modelValue should reset to the initial values.

### Bug sample reproduction 
https://codesandbox.io/s/3yeyj

tlouisse:
@zanfirandra : thanks for the bug report 👍 It should be fixed here: https://github.com/ing-bank/lion/pull/241/commits/4cae67955ba4c0d964d288b4a7f88025c0197748#diff-ede1fc987c4e048330c6e79d1dba5324R689

Issue #235, opened 2019-08-08
[lion-form] resetModelValue doesn't update if number of inputs changes


mihaisavezi:
## What happens

1.  I have a form with 10 input controls, set dynamically based on a config.
2. On changing the config and providing 9 input controls, it renders correctly,
3. I need to reset the modelValue for every control in the form, calling ```resetGroup``` on formElement
4. But the resetModelValue properties do not update to reflect the fact that I have 9 input controls now.
5. This triggers an error, and the form tries to set a modelValue on a non-existing property.

## What I've tried

1. Tried first triggering the ```updateComplete``` lifecycle method to update the form and then calling resetGroup. But the result is the same.

## Bug sample reproduction

https://codesandbox.io/s/lion-form-dynamic-controls-ppfog

tlouisse:
@mihaisavezi thanks for the bug report. It should be fixed here: https://github.com/ing-bank/lion/pull/241/commits/4cae67955ba4c0d964d288b4a7f88025c0197748#diff-ede1fc987c4e048330c6e79d1dba5324R675

Issue #231, opened 2019-08-07
[form] decrease technical debt


tlouisse:
- remove obsolete `form-unregister` events from previous system
- make registration code simpler
- make registration code forwards compatible with situation in which shadydom bug is fixed

A proposal can be found here: https://github.com/ing-bank/lion/pull/226

PR #241, opened 2019-08-13
Fix/form reg issues

tlouisse:
fixes https://github.com/ing-bank/lion/issues/239
fixes https://github.com/ing-bank/lion/issues/235
fixes https://github.com/ing-bank/lion/issues/231

The proposed approach (patching LitElement because of shadydom bug) caused breaking changes, so it's left out

### DIFF
--- a/packages/fieldset/src/LionFieldset.js
+++ b/packages/fieldset/src/LionFieldset.js
@@ -11,7 +11,7 @@ const pascalCase = str => str.charAt(0).toUpperCase() + str.slice(1);
 /**
  * LionFieldset: fieldset wrapper providing extra features and integration with lion-field elements.
  *
- * @customElement
+ * @customElement lion-fieldset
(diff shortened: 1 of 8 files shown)


packages/fieldset/src/LionFieldset.js
```diff
@@ -159,7 +161,15 @@ export class LionFieldset extends FormRegistrarMixin(
...
-    this.modelValue = this.resetModelValue;
+    this.formElementsArray.forEach(child => {
+      if (typeof child.resetGroup === 'function') {
+        child.resetGroup();
+      } else if (typeof child.reset === 'function') {
+        child.reset();
+      }
+    });
+
+    this.modelValue = this._initialModelValue;
```
tlouisse: Since modelValue is a getter, this line is most likely not needed.
TODO: delete line and see if all functionality still works

PR #245, opened 2019-08-13
fix(localize): update the way to obtain the group separator for a number

jcperea:
four digit numbers can be written without separator, so it would be
safer to obtain the separator by formatting a five digit number. The bug
was found with the locale 'es-ES', where the format returned for '1000'
was '1000' instead of the expected '1.000'. We add the specific expect
clause on the test for it.

 On branch fix/numberGroupSeparator
 Changes to be committed:
	modified:   packages/localize/src/number/getGroupSeparator.js
	modified:   packages/localize/test/number/getGroupSeparator.test.js

### DIFF
--- a/packages/localize/src/number/getGroupSeparator.js
+++ b/packages/localize/src/number/getGroupSeparator.js
@@ -12,6 +12,6 @@ export function getGroupSeparator(locale) {
   const formattedNumber = Intl.NumberFormat(computedLocale, {
     style: 'decimal',
     minimumFractionDigits: 0,
-  }).format('1000');
-  return normalSpaces(formattedNumber[1]);
+  }).format('10000');
+  return normalSpaces(formattedNumber[2]);
 }
--- a/packages/localize/test/number/getGroupSeparator.test.js
+++ b/packages/localize/test/number/getGroupSeparator.test.js
@@ -7,5 +7,6 @@ describe('getGroupSeparator', () => {
     expect(getGroupSeparator('en-GB')).to.equal(',');
     expect(getGroupSeparator('nl-NL')).to.equal('.');
     expect(getGroupSeparator('fr-FR')).to.equal(' ');
+    expect(getGroupSeparator('es-ES')).to.equal('.');
   });
 });


packages/localize/src/number/getGroupSeparator.js
```diff
@@ -12,6 +12,6 @@ export function getGroupSeparator(locale) {
   const formattedNumber = Intl.NumberFormat(computedLocale, {
     style: 'decimal',
     minimumFractionDigits: 0,
-  }).format('1000');
-  return normalSpaces(formattedNumber[1]);
+  }).format('10000');
```
tlouisse: Why is the extra 0 needed if I may ask?

jcperea: sure you can :) it is on the commit message. The norm says (you can check it on this [wikipedia](https://en.wikipedia.org/wiki/Decimal_separator#Exceptions_to_digit_grouping) link) you can omit the separator group when you have a four digit number, and for some reason in the last update of V8, they have decided to follow this optional rule, at least for es-ES. What always returns the separator is a more than four digit number, so the extra 0 makes always the job.

jcperea: Instead of returning the second position of the formatted four digit number, you can return the third position of a five digit number, and you skip the possible customization. 5 digit numbers must be formatted with the separator.

jorenbroekema:
Aahh so basically 1000 - 9999 which would normally become 1.000 - 9.999, now just stays 1000-9999, because no need for a group separator when there's only a single digit to be separated. 

We use this logic to determine what the separator character is for the locale, in getGroupSeperator. We format the number 1000 with Intl, relying on that there is always some separator in the returned value as the second character. But now that is no longer something we can rely on, so we format the number 10000 which does always have a separator, at the third character. 

It's clear to me now what this fix is :) thanks for contributing it!

gerjanvangeest:
@jorenbroekema can you check if this: https://github.com/ing-bank/lion/pull/230/files is still needed or not?

PR #248, opened 2019-08-16
Localize: add expect to test consistent bg-BG group separator

jorenbroekema:
~We added a specific bg-BG rule, which now turns out to be:~
- ~Unnecessary, because Intl does it correctly if you get the group separator by formatting a 5 digit number, rather than a 4 digit number (1000-9999). This was fixed in https://github.com/ing-bank/lion/pull/245~
- ~Incorrect. You should not always force the space. For example, if the type is currency, for bg-BG, no spaces are ever added as group separator according to what Intl does at least..~

~This change fixes a bug where bg-BG currencies with 4 or more digits would display with space group separators, and it gets rid of quite some technical debt that we added to 'normalize' it.~


tlouisse:
Good find. Where can I find docs on this? Wikipedia? 

What I wonder: do we need normalization for browsers other than Chrome now?

PR #251, opened 2019-08-19
wip: [ready making] changes needed for rich select combined with global overlay

tlouisse:
Contains a list of changes (mostly quick and dirty) that will need to be made to have a select rich that works with global overlays.

All these changes need to be done in separate tasks with proper tests. Goal of this mr is to easily see the changes on top of master

### Proposal: api normalisation between Local and Global OverlayController

Goal is to create an `OverlayController` that can dynamically switch between types of overlays. We need this for Design system layers that present as a dropdown(`LocalOverlayController`) on desktop and as a dialog (via `GlobalOverlayController`) on mobile / smaller screens.

Subclassers only need to override a method that returns an `OverlayController`. In the rest of the code there is just one reference to this controller.

For instance, in an extension layer of the datepicker / select / action-menu:
```js
  /**
   * @desc Allows Subclassers to dynamically customize overlay controllers
   * @returns {OverlayController}
   */
  _defineOverlay(contentNode, invokerNode) {
    const ctrl = new OverlayController({
        contentNode,
        invokerNode,
      },
    );

    ctrl.addEventListener('before-open', () => {
      if (window.innerWidth >= 600) {
        ctrl.setConfig(withDropdownConfig());
      } else {
        ctrl.setConfig(withModalDialogConfig());
      }
    });

    return ctrl;
  }
```

**New features added by** [`overlayController`](https://github.com/ing-bank/lion/blob/feat/overlaySwitch/packages/overlays/src/OverlayController.js):
    - Creates a predictable, normalized api that abstracts away GlobalOverlayController and LocalOverlayController. Without it, extending a component would demand several small changes at different (private) spots in the code. See: https://github.com/ing-bank/lion/pull/251/commits/c8a3c514c5e5b716bbe8d54db4f70de93928e6b2
    - Synchronizes focus between local and global
    - Exposes `before-show` event
    - Allows to switch nodes from global to local (we need to remember insertion point to reattach locally)

For an overview of all props that needs to be normalized: https://github.com/ing-bank/lion/blob/feat/overlaySwitch/packages/overlays/src/OverlayController.js#L23 (this list mentions whenever an option is either unique to Global- or unique to LocalOverlayController).


packages/overlays/src/OverlayController.js
```diff
@@ -0,0 +1,192 @@
...
+   * @property {boolean} [hidesOnEsc=false] - hides the overlay when pressing [ esc ]
+   * @property {boolean} [hidesOnOutsideClick=false] - hides the overlay when clicking next to it,
+   * exluding invoker. (currently exclusive to localOverlayController)
+   * https://github.com/ing-bank/lion/pull/61
+   * @property {TemplateResult} [contentTemplate]
+   * @property {TemplateResult} [invokerTemplate] (currently exclusive to LocalOverlayController)
+   * @property {HTMLElement} invokerNode
+   * @property {HTMLElement} contentNode
+   * @property {boolean} [isModal=false] - sets aria-modal and/or aria-hidden="true" on siblings
+   * @property {boolean} [isGlobal=false] - determines the connection point in DOM (body vs next
```
daKmoR: `renderToBody` sounds way better 👍

packages/overlays/src/OverlayController.js
```diff
@@ -0,0 +1,192 @@
...
+   * to invoker). This is what other libraries often refer to as 'portal'. TODO: rename to renderToBody?
+   * @property {boolean} [isTooltip=false] - has a totally different interaction- and accessibility pattern from all other overlays, so needed for internals.
```
daKmoR: hmmm doesn't `isTooltip` "forces" it to be a local overlay?

packages/overlays/src/OverlayController.js
daKmoR: or even better?
```js
type="global" // ['global', 'local']
```

packages/overlays/src/OverlayController.js
daKmoR: `isToolTip` is a very confusing flag?
this seems like promoting a single mega class that has all the features and we enable different parts via options... 

not so sure if that is a smart idea?

tlouisse: We already use it like this in the LocalOverlayController. I agree that that one (already) feels heavy for a tooltip, but the reason it is needed is that accessibility is just way different than in other overlays. Ergo: if we want a solution that is accessible out of the box, the local overlay **should** know we are dealing with `[role=tooltip]` 
If local means 'attached as sibling of invoker' then I think indeed that this one forces that (that is when we want to use aria-describedby and aria-labelledby). This a11y constraint doesn't hold for dropdowns (just like Popper.js allows you to render your dropdowns to the body)

But indeed, some flags 'trigger' other flags. If we deliver the right presets this should be very straightforward for the developer. There are several loose ends in the total Overlay system still, I think;  **for simplicity, we can leave the isTooltip out now.** 
Maybe we can have a separate discussion about improvements below?

### Improvements controllers

- Could be more balanced (especially local vs global) 
    - We totally position local overlays (even deliver ‘pointing frames’), but we don’t deliver any positioning for global overlays **(already being worked on atm)**
    - We only partly deliver accessibility (we should either do 100% or not at all)
    - We don’t distinguish between tooltips and others (a11y)
    - We only partly do user interaction. If we want to handle accessibility out of the box, we  should be fully aware of all interactions and not leave clicks and focuses up to the user.
- Viewport paddings are not customisable on constructor level, but do heavily depend on Design Systems
- It should be possible to deliver own ‘portals’, default one is not customisable and highly opinionated (backdrop colour, opacity, animation etc.)
- The default portal doesn’t use shadow DOM and has potential style leaks
- Support for sr-only as hiding mechanism: think of a navigation menu at the top of your page that should be fully indexed by screen readers 

### Improvements components (normalization of apis for Subclassers)

Universal interface mixing for elements wrapping overlay system (select, autocomplete, tooltip, contextmenu, popover)
- Create predictability and consistency for App Developers:
    - opened state pointing to isShown
    - Forward public api ’showOverlay’, ‘hideOverlay’, ’toggleOverlay’
- Create predictability for Subclassers:
    - _overlay holds overlay controller. A subclasser can always rely on this interface, regardless whether he has an overlay relative to viewport or the invoker and regardless of whether the content is placed in a portal or not.
    - _defineOverlay In this method the overlay controller is returned. This can be a dropdown for a select that transforms into a bottom sheet on mobile. The overlay controller allows to hook into events(like ‘before-show’) to change to the desired overlay type. When a portal switch takes place, the overlay controller takes care of keeping focus in sync.

tlouisse: But it comes down a bit to the question I think: do we want our system to handle accesibility or not :)
Currently it does, but not in all cases. This feels a bit arbitrary to me. Think we should either do it fully (aligns with our mission to make solutions that are accessible out of the box) or we shouldn't do it at all.
For me, extracting all different 'accessibility scenarios' based on w3c specs beforehand is what brings the extra added value in this system. (there is just a limited set of accessible widgets mappable to our overlays and by baling them into the system we create predictable user experiences for impaired users).

We can also say that we extract all accessibility in a different layer? Like an OverlayA11yMixin?
Then you can still always use the barebone layer when user interaction might not be involved.

packages/overlays/src/OverlayController.js
tlouisse: Yeah. Problem with global for me that I don't know if it means:
- render to body?
- position relative to viewport?
- both?

We could also introduce smth like the term 'portal' (think other libs use this term and might have seen it in a spec proposal before as well if I'm right)

tlouisse: In our current system 'global' means both. But I think we should allow local overlays to be rendered to the body as well. Here an example of a use case that needs this: https://github.com/Hacker0x01/react-datepicker/issues/1366

For me it would make sense if it works like:
- renderToBody: renders to body
- viewportPlacement: determines whether it should position relative to viewport

daKmoR:
```js
  /**
   * @desc Allows Subclassers to dynamically customize overlay controllers
   * @returns {OverlayController}
   */
  _defineOverlay(contentNode, invokerNode) {
    const config = {
        contentNode,
        invokerNode,
      };
    const ctrl = new OverlayManager(config);
    const localController = new LocalOverlayController({ ...config, { extras: }});

    const globalController = new GlobalOverlayController({ ...config, { extras: }});

    ctrl.addEventListener('before-open', () => {
      if (window.innerWidth >= 600) {
        ctrl.setController(localController);
      } else {
       ctrl.setController(globalController);
      }
    });

    return ctrl;
  }
```

tlouisse:
> ```js
>   /**
>    * @desc Allows Subclassers to dynamically customize overlay controllers
>    * @returns {OverlayController}
>    */
>   _defineOverlay(contentNode, invokerNode) {
>     const config = {
>         contentNode,
>         invokerNode,
>       };
>     const ctrl = new OverlayManager(config);
>     const localController = new LocalOverlayController({ ...config, { extras: }});
> 
>     const globalController = new GlobalOverlayController({ ...config, { extras: }});
> 
>     ctrl.addEventListener('before-open', () => {
>       if (window.innerWidth >= 600) {
>         ctrl.setController(localController);
>       } else {
>        ctrl.setController(globalController);
>       }
>     });
> 
>     return ctrl;
>   }
> ```

So OverlayManager ~= OverlayController. 
If your goal is not to write a common ancestor for both, but have it control different controllers, then this is a better name indeed 👍 
Doing it like this is (with controllers) is less invasive with regards to changes needed in our codebase. If we want to allow local (meaning positioned relatively to invoker) rendered to the body, we can always add this functionality to the LocalOverlayController, because Popper already supports it (this will then solve situations in which people cannot control surroundings (like stacking order of parent hierarchy, `overflow` settings of parent containers, transforms of parent containers, etc.)). 


What do we do with this btw: https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlaysManager.js? 
Do we give it a new purpose?

daKmoR:
hmm this overlay manager is a global singleton... where as this "new" Manager is one per instance 

so maybe call it `ControllerManager`?

tlouisse:
> hmm this overlay manager is a global singleton... where as this "new" Manager is one per instance
> 
> so maybe call it `ControllerManager`?

`DynamicOverlayController` maybe ? :)

daKmoR:
this has been addressed via other merge requests 👍 

PS: if something is still missing please feel free to reopen or create another issue or merge request

Issue #187, opened 2019-07-19
Only allow full locales for translations


LarsDenBakker:
Moved here from an internally created issue.

Currently inside ing-web components there are switch statements like this:

```js
switch (locale) {
  case 'en-GB':
    return import('./en-GB.js');
  case 'en-US':
    return import('./en-US.js');
  case 'en': 
    return import('./en.js');
}
``` 

Where both language-only ('en') and locales ('en-GB', 'en-US') are supported.

I think we should not support only 'language' specifiers within the localization system for a few reasons:
- It adds extra boilerplate code, and adds extra lazy loaded chunks for tools to manage.
- It makes tools based off of translations harder to write (we have a tool for managing translations for copy writers and a webpack plugin for optimizing translation bundling)
- It picks an implicit locale for localizing numbers and dates, possibly creating unexpected results for the user and potentially different across browsers.

bashmish:
We had some time to digest this already and collect some knowledge.

First of all, this is not a theoretical problem, we already saw bugs because of it, e.g. [this one](https://github.com/ing-bank/lion/issues/90).

Second, we found out that there is [another problem with Google Chrome translate feature](https://github.com/ing-bank/lion/issues/152) which makes it difficult to fully forbid the usage of only language specifiers. Google Chrome itself uses language only (you translate the page to English, Google Chrome sets `en`), thus interfering with our system and forcing us to support them to a certain degree, or at least not throw, because it interrupts the Google Chrome injected JS code.

I think the approach here should be practical: yes, we don't want app developers to use language only specifiers, because they will not work correctly with dates and numbers, but we don't want to break other tools like Google Chrome translate. And how to technically implement such a fix I'm not so sure yet. We can detect Google Chrome injected scripts on a page and then temporarily allow language specifiers, but this is a fragile hack on our side. And who knows how other tools/plugins are going to use `<html lang>`. Maybe we should just write a linter which will detect the wrong `lang` attribute and we can add a `console.warning` instead of an exception for developers to notice the problem in the runtime. And this will not be visible that much in the apps which follow the rule of using full locales, unless users activate Google Chrome translate on them.

What do you think?

jorenbroekema:
Now that we no longer conflict with third party translation tools, I am not really sure what to do about this one. I think we will keep allowing language only locales just because of the fact that we want to support tools like Google Translate. @tlouisse  opinion?

jorenbroekema:
[...]

PR #253, opened 2019-08-20
chore(localize): deprecate language-only locales

jorenbroekema:
https://github.com/ing-bank/lion/issues/187

We still want to keep the fallback locale, the rationale of which is explained here: https://github.com/ing-bank/lion/blob/master/packages/localize/RATIONALE.md#fallbacks
It is a good idea that if there is a case where the locale is not a full language locale, we make a best guess using the fallback locale. And for writing translations, it is still good to structure it like this to prevent duplicate translations. 

The only thing that changes is that we heavily discourage using language-only locales on purpose, shown by sending a console warning if the locale changes to a single language locale.

### DIFF
--- a/packages/input-email/stories/index.stories.js
+++ b/packages/input-email/stories/index.stories.js
@@ -19,7 +19,7 @@ storiesOf('Forms|Input Email', module)
   )
   .add('Use own validator', () => {
     const gmailOnly = modelValue => ({ gmailOnly: modelValue.indexOf('gmail.com') !== -1 });
-    localize.locale = 'en';
+    localize.locale = 'en-GB';
 
     try {
       localize.addData('en', 'lion-validate+gmailOnly', {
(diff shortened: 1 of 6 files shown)


daKmoR:
that will show the warning also when chromes translation feature is used right? (as chrome only translates with for example 'en')
could be a valid choice to show it in this case...

jorenbroekema:
Does the chrome translation tool switch your html lang attr to a single language locale? :/

daKmoR:
jup... some more details are here: https://github.com/ing-bank/lion/issues/152

jorenbroekema:
I changed it to only trigger a warning if locale is set through the locale setter. If it's set by using the html lang attribute, we won't know who is responsible for it (e.g. Google Chrome translate tool) and it will not do a console warning

daKmoR:
uh I like that 👍

PR #254, opened 2019-08-21
feat(overlays): align Overlays API + add DynamicOverlay

daKmoR:
The basis for a discussion about a dynamic way of switching between different overlay contexts


tlouisse:
> The basis for a discussion about a dynamic way of switching between different overlay contexts

Yesterday morning we said we all agreed upon this poc, right? https://github.com/ing-bank/lion/pull/251. From my understanding, this discussion was already settled.
For next time, can you please try to keep things in the same issue/branch, so we have the overview of the full discussion? Now it looks if the amount of issues is larger then in reality, something you are not a fan of ;)

As a basis for a discussion and in order to assess the validity of your BDDs, I would expect some context/rationale behind the idea of a DynamicOverlayController. So the **why** instead of the **how**. 
So from a Subclassers perspective, an answer to the question "Why do I need a DynamicOverlayController in the first place?"
This will serve as a basis for documentation and helps Subclassers understand the code (just like there lately was a need to explain form registration or form internals in general).

### Things that I currently miss in the tests
- the main added benefit: having a **normalized api** for all overlays that allows subclassers to use this abstraction layer in every context. So all props, events, methods are available
- the ability to **synchronize** focus between different types of overlays
- the ability to switch between local and global context (this involves remembering the parentNode of a contentNode to properly reattach it

These high level requirements would be enough to start a story. The implementing developer (this can be anyone) then gets the freedom to fill in the details when the story gets picked up.

_As a side node: we planned officially to pick up BDDs next sprint?_

daKmoR:
best to only review the commits 3+ as the first 2 commits are "maintainance" changes

gerjanvangeest:
All these changes are merged to master via: https://github.com/ing-bank/lion/pull/288

PR #255, opened 2019-08-22
fix(lion-radio-group): handle unchecked state

bcromijn:
Before the serializedValues and formattedValues method call failed if there weren't any selections made

fixes https://github.com/ing-bank/lion/issues/266

### DIFF
--- a/packages/radio-group/src/LionRadioGroup.js
+++ b/packages/radio-group/src/LionRadioGroup.js
@@ -35,7 +35,8 @@ export class LionRadioGroup extends LionFieldset {
   }
 
   get serializedValue() {
-    return this._getCheckedRadioElement().serializedValue;
+    const el = this._getCheckedRadioElement();
+    return el ? el.serializedValue : '';
   }
 
   set serializedValue(value) {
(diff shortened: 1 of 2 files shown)


bcromijn:
Any update on this yet?

jorenbroekema:
Seems reasonable to me :)

bcromijn:
Tests added.

packages/radio-group/src/LionRadioGroup.js
```diff
@@ -36,7 +36,7 @@ export class LionRadioGroup extends LionFieldset {
 
   get serializedValue() {
     const el = this._getCheckedRadioElement();
-    return el ? el.serializedValue : '';
+    return el ? el.serializedValue : {};
```
jorenbroekema: I think it would make sense that this is an empty string. The serialized value of a radio group should return the checked radio element's serializedValue, which should be a string, so in case there is no selected radio element, it should return empty string.

Also could you amend your last commit where you add the tests to a chore? Otherwise we get 2 bugfix entries in the changelog ;).

PR #259, opened 2019-08-22
fix(field): implement FormRegistrarPortalMixin

daKmoR:
Proposal for `FormRegistrationMixins.js`.
the `FormRegistrarPortalMixin` forwards the event to a specific target node so Registering Child does not need to be in the dom tree of the Registrar Parent.

### DIFF
--- a/packages/field/src/FormRegistrarPortalMixin.js
+++ b/packages/field/src/FormRegistrarPortalMixin.js
@@ -1,0 +1,55 @@
+import { dedupeMixin } from '@lion/core';
+import { formRegistrarManager } from './formRegistrarManager.js';
+
+/**
+ * This will forward
+ */
+export const FormRegistrarPortalMixin = dedupeMixin(
+  superclass =>
+    // eslint-disable-next-line no-shadow, no-unused-vars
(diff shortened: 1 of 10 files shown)


packages/field/src/FormRegistrarPortalMixin.js
```diff
@@ -0,0 +1,65 @@
+import { dedupeMixin } from '@lion/core';
+import { formRegistrarManager } from './formRegistrarManager.js';
+
+/**
+ * This will forward
```
gerjanvangeest: Please, explain a bit more what this Mixin does.

gerjanvangeest:
You should add an export in `./packages/field/index.js`

gerjanvangeest:
We have multiple form Registering/Registrar files.. might be time to create a documentation page about it.

packages/field/src/FormRegistrarPortalMixin.js
daKmoR: added

daKmoR:
> You should add an export in `./packages/field/index.js`

good point added 👍

PR #261, opened 2019-08-26
feat(http): implement fetch-based http client

LarsDenBakker:
This adds a new http/ajax/xhr package based on fetch instead of axios.

I know the intention was for the current ajax package to be an adapter, where the axios implementation could be exchanged for fetch underwater. I tried this, but unfortunately it didn't work out. 

The current ajax implementation leaks axios internals in the interceptors, request arguments and response object. Recreating that on top of fetch would take quite some time, result in a lot of extra code and be very error-prone.

Instead I opted for a small lightweight wrapper around fetch, adding only a few of the necessary things. This can exist side by side with ajax, so that people could already start migrating. Internally with the next breaking change, we could then remove ajax.

The benefits are:
- A saving of 11.4kb / 4.1kb minified
- Devs learn a web standard, instead of a proprietary library
- We can reuse existing fetch documentation
- Access to fetch-only features, such as streaming responses, keep-alive and others in the future

### DIFF
--- a/packages/http/README.md
+++ b/packages/http/README.md
@@ -1,0 +1,92 @@
+# HTTP
+
+[//]: # 'AUTO INSERT HEADER PREPUBLISH'
+
+`http` is a small wrapper around `fetch` which:
+
+- Allows globally registering request and response transformers
+- Throws on 4xx and 5xx status codes
+- Supports a JSON request which automatically encodes/decodes body request and response payload as JSON
+- Adds accept-language header to requests based on application language
+- Adds XSRF header to request if the cookie is present
+
+## How to use
+
+### Installation
+
+```sh
+npm i --save @lion/http
+```
+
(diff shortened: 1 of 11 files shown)


LarsDenBakker:
@jorenbroekema Sorry I missed your comment here. I had a discussion with @daKmoR about this, I didn't get around to updating this PR based on our discussions until now. It's updated now, it's overall a bit simpler because there isn't a separate `HttpJsonClient` anymore, it's merged into one.

How would you guys like to proceed further? Have a meeting IRL? It would be great to get this moving forward sooner rather than later, so that people can start using it.

jorenbroekema:
Closing this for now since the idea at the moment is to use redaxios, to keep axios-like API, but it uses fetch under the hood. It will help make migration smoother.

Issue #258, opened 2019-08-22
[localize] Unexpected behavior when editing input-amount value


vicdata4:
Hi guys,

We are using input-amount component with this configuration: .filterOptions="{minimumFractionDigits: 0, maximumFractionDigits: 0, roundMode: 'floor'}" in order to get this format: 1.000 / 1.000.000 (without decimals).

We have detected an unexpected behaviour when trying to edit input value after introducing it the first time.(example: introduce 5000(formatted as 5.000) and then edit only one digit and press enter) result is 5.

We have checked that first return from getLocale() function located in https://github.com/ing-bank/lion/blob/master/packages/localize/src/number/getLocale.js, return locale as object when should return a string.

maybe it should be this way:

export function getLocale(locale) {
  if (locale && locale.locale) {
    return locale.locale;
  }

  if (localize && localize.locale) {
    return localize.locale;
  }

  return "en-GB";
}

carlosmmdiaz:
Hi there,

We do have the same issue in our team, however we think the problem remains in the input-amount component, in the parsers.js file.

This function, "getDecimalSeparator(locale)" used in "parseWithLocale(...)" needs the parameter locale as a string not as an object.

So instead of this:

  const separator = getDecimalSeparator(options)

Should be changed to this:

  const separator = getDecimalSeparator(options.locale)

If options.locale is undefined, the "getLocale" function of Localize will use the the current localize or the default one "gb-GB"

```
function parseWithLocale(value, options) {
  const separator = getDecimalSeparator(options.locale); // Needs a string, not an object
  const regexNumberAndLocaleSeparator = new RegExp(`[0-9${separator}-]`, "g");
  let numberAndLocaleSeparator = value
    .match(regexNumberAndLocaleSeparator)
    .join("");
  if (separator === ",") {
    numberAndLocaleSeparator = numberAndLocaleSeparator.replace(",", ".");
  }
  return parseFloat(numberAndLocaleSeparator);
}
```

What do you think?

Cheers

jorenbroekema:
@carlosmmdiaz I haven't had the time yet to reproduce, but your solution sounds sensible indeed, the JSDocs clearly expects a String and not an object, so we shouldn't pass it an object ;). 

Could you create a Pull Request to fix it, and add a test that fails before the fix, and succeeds after the fix :)? Then it would be an easy approve & merge from my side

vicdata4:
[...]

PR #262, opened 2019-08-26
Fix: pass string instead of object in input-amount parser

vicdata4:
Fix parseWithLocale function: Pass a string instead of an object to getDecimalSeparator and getLocale. Test added that fails before the fix.

fixes https://github.com/ing-bank/lion/issues/258


jorenbroekema:
Thanks for the MR, looks great, awaiting CLA to be signed by you though ;)

vicdata4:
> Thanks for the MR, looks great, awaiting CLA to be signed by you though ;)

Already signed, the branch is ready to merge.

PR #263, opened 2019-08-28
chore(button): add known issue to documentation

erikkroes:


daKmoR:
we found a workaround and it got implemented in https://github.com/ing-bank/lion/pull/264

PR #270, opened 2019-09-05
wip: Validation refactor

tlouisse:
Place to discuss changes in validation system...

### DIFF
--- a/packages/validate/test/ValidateMixin.test.js
+++ b/packages/validate/test/ValidateMixin.test.js
@@ -7,6 +7,7 @@ import { localize } from '@lion/localize';
 
 import { ValidateMixin } from '../src/ValidateMixin.js';
 import { Unparseable } from '../src/Unparseable.js';
+import { Validator } from '../src/Validator.js';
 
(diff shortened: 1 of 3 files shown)


packages/validate/test/ValidateMixin.test.js
```diff
@@ -1285,8 +1286,46 @@ describe('ValidateMixin', () => {
...
+        <${tag}
+          .modelValue=${'dog'}
+          .errorValidators=${[new DelayedCatValidator()]}
+        >
+        ${lightDom}
+        </${tag}>
+      `);
+      const validator = el.errorValidators[0];
+      expect(validator instanceof Validator).to.equal(true);
+      expect(el.errorState).to.equal(false);
```
daKmoR: don't we need another errorState then? e.g. something like `pending` or so?

tlouisse: Oh yeah, definitely. Not finished by a long shot :)
This was only to make sure the concept (having async and sync in the same array) works. I think the pending state was already described in an it below...

tlouisse:
Closing this one. Please see https://github.com/ing-bank/lion/pull/271, which extends this concept

Issue #249, opened 2019-08-16
[lion-button] IE11 - submit event not called on enter keypress when there are multiple inputs


zanfirandra:
### Context
Browser: IE11.
Have a simple form with initially one input and submit event on the form.
I want to submit the form by pressing the enter key.

### Steps to reproduce
1. For the inital form, press the enter key to submit form.
2. Add another type of field, e.g one radio group.
3. Submit form by pressing the enter key.
2. Add another input in the form.
3. Submit form by pressing the enter key.

### What happens
Submit event is working for a form with one input and another type of field, but when adding another input, submit event listener is not called.

jorenbroekema:
I cannot reproduce this in the same way you have it.

What version of @lion are you running? Did you try it on the latest version?
If so, can you show me your code that reproduces the issue?

jorenbroekema:
Update: seems to be related to lion-button

jorenbroekema:
Update: 

Can be reproduced by just:
```js
const submitForm = () => console.log('hey');
return html`
  <form id="form" @submit=${submitForm}>
    <h3>Details</h3>
    <input name="input1" />
    <input name="input2" />
    <input name="input3" />
    <lion-button>Submit</lion-button>
  </form>
`;
```
Changing it to a regular button works fine on Chrome at least

jorenbroekema:
Aaand another update. This is not related to lion-button, because I can reproduce this issue with just:

```js
const submitForm = () => console.log('hey');
return html`
  <form id="form" @submit=${submitForm}>
    <h3>Details</h3>
    <input name="input1" />
    <input name="input2" />
  </form>
`;
```
Will investigate further, see if this is related to lit-html in IE11.

Issue #240, opened 2019-08-13
[lion-form]/[lion-button] Submit on `enter` keypress is triggered twice


dancarlocanlas:
## Steps to reproduce
1. Create a form using `lion-form`
2. Add a submit button using `lion-button`
3. Submit form by pressing `enter`
4. The `submit` event is dispatched twice

I did some checking and it seems that the root cause lies within the `lion-button` element, specifically in the click handler / redispatching of the click event. The default click event does not stop propagating, or might have reached the `form` element before the propagation was stopped.

jorenbroekema:
Confirmed, have reproduced this in Google Chrome

PR #281, opened 2019-09-19
Fix button implicit submission on enter keypress

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/240
fixes https://github.com/ing-bank/lion/issues/249

- [x] check if it helps with https://github.com/ing-bank/lion/issues/249 (fixed now in this PR)
- [x] add test for form submission on enter keypress
- [x] remove stories used for debugging

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -71,6 +71,12 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
 
         :host .btn ::slotted(button) {
           position: absolute;
+          clip: rect(0 0 0 0);
+          clip-path: inset(50%);
+          overflow: hidden;
+          white-space: nowrap;
+          height: 1px;
+          width: 1px;
         }
 
         .click-area {
@@ -110,22 +116,6 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
           fill: #adadad;
         }
       `,
-      ...(this.__isIE11() /* visibility hidden in IE11 breaks native button functionality */
-        ? [
-            /* TODO: Make SR-only css mixin? */
-            css`
-              :host .btn ::slotted(button) {
-                clip: rect(0, 0, 0, 0);
-              }
-            `,
-          ]
-        : [
-            css`
-              :host .btn ::slotted(button) {
-                visibility: hidden;
-              }
-            `,
-          ]),
     ];
   }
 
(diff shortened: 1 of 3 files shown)


packages/button/src/LionButton.js
```diff
@@ -120,7 +135,6 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
       _button: () => {
         if (!this.constructor._button) {
           this.constructor._button = document.createElement('button');
-          this.constructor._button.setAttribute('slot', '_button');
```
jorenbroekema: Dont think this is necessary, slotmixin does this right?

packages/button/src/LionButton.js
```diff
@@ -111,6 +110,22 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
           fill: #adadad;
         }
       `,
+      ...(this.__isIE11() /* visibility hidden in IE11 breaks native button functionality */
+        ? [
+            /* TODO: Make SR-only css mixin? */
+            css`
+              :host .btn ::slotted(button) {
+                clip: rect(0, 0, 0, 0);
```
erikkroes: Please expand this a little further following this pattern: https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html#hiding-content-visually

```suggestion
                clip: rect(0 0 0 0); 
                clip-path: inset(50%);
                height: 1px;
                overflow: hidden;
                position: absolute;
                white-space: nowrap; 
                width: 1px;
```

jorenbroekema: Yeah to be honest, in retrospect, I think we should create an srOnlyMixin that returns this css :)

packages/button/src/LionButton.js
```diff
@@ -167,31 +171,13 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-    try {
-      // this is for IE11 (and works in others), because `Object.defineProperty` does not give any effect there
-      event.__defineGetter__('target', () => this); // eslint-disable-line no-restricted-properties
-    } catch (error) {
-      // in case `__defineGetter__` is removed from the platform
-      Object.defineProperty(event, 'target', { writable: false, value: this });
+  __clickDelegationHandler() {
+    if (this.type === 'submit' && this._nativeButtonNode.form) {
+      this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
+      this._nativeButtonNode.form.submit();
```
tlouisse: Did we need both in the end?

packages/button/src/LionButton.js
```diff
@@ -167,31 +171,13 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-    const newEvent = new MouseEvent(oldEvent.type, oldEvent);
-    newEvent.__isRedispatchedOnNativeButton = true;
-    this.__enforceHostEventTarget(newEvent);
-    this._nativeButtonNode.dispatchEvent(newEvent);
-  }
-
   /**
-   * Prevent normal click and redispatch click on the native button unless already redispatched.
+   * Prevent normal click and redispatch submit on the native button unless already redispatched.
    */
```
tlouisse: We also don't prevent normal click anymore, so that event delegation works as expected c.q. points to the right target...
And we dispatch a submit event (not re) to make it follow the exact same behaviour of a native button

jorenbroekema: Yes, we should change the comment to something like "Dispatch submit event and invoke submit on the native form when clicked"

packages/button/src/LionButton.js
jorenbroekema: Yes ;) submit() wouldn't trigger @submit handler, and dispatchEvent wouldn't cause the actual form to submit and cause page refresh, so both are necessary.

packages/button/src/LionButton.js
```diff
@@ -167,31 +171,13 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-    // replacing `MouseEvent` with `oldEvent.constructor` breaks IE
-    const newEvent = new MouseEvent(oldEvent.type, oldEvent);
-    newEvent.__isRedispatchedOnNativeButton = true;
-    this.__enforceHostEventTarget(newEvent);
-    this._nativeButtonNode.dispatchEvent(newEvent);
-  }
-
   /**
-   * Prevent normal click and redispatch click on the native button unless already redispatched.
+   * Prevent normal click and redispatch submit on the native button unless already redispatched.
```
erikkroes:
```suggestion
   * Dispatch submit event and invoke submit on the native form when clicked.
```

PR #282, opened 2019-09-20
Feat/bottomsheet rebased

gerjanvangeest:


packages/overlays/src/GlobalOverlayController.js
```diff
@@ -69,6 +73,8 @@ export class GlobalOverlayController extends BaseOverlayController {
       return;
     }
     if (!this.content.isConnected) {
+      this.content.classList.add('global-overlays__overlay-container');
```
jorenbroekema: `global-overlays__overlay-container` is hardcoded twice, so let's abstract it to a variable?

packages/overlays/test/GlobalOverlayController.test.js
```diff
@@ -29,6 +29,7 @@ function getTopContainer() {
   return getRenderedContainers().find(container => {
     const rect = container.getBoundingClientRect();
     const topElement = document.elementFromPoint(Math.ceil(rect.left), Math.ceil(rect.top));
+    console.log('test topElement:', topElement, container);
```
jorenbroekema: Don't forget to remove this one :)

packages/overlays/test/GlobalOverlayController.test.js
```diff
@@ -111,34 +115,6 @@ describe('GlobalOverlayController', () => {
       expect(ctrl.isShown).to.equal(false);
     });
 
-    it('puts the latest shown overlay always on top', async () => {
```
jorenbroekema: Why did this one get removed?

gerjanvangeest: This test is a duplicate of: https://github.com/ing-bank/lion/blob/feat/bottomsheetRebased/packages/overlays/test/OverlaysManager.test.js#L84 , but was failing. I double checked the code, which worked. I couldn't get the test working, within a certain time frame, and since it's a duplicate i removed it.

packages/overlays/test/GlobalOverlayController.test.js
gerjanvangeest: :)

packages/overlays/src/GlobalOverlayController.js
gerjanvangeest: Where else do you see it?

gerjanvangeest:
> * Does dynamically/imperatively changing the viewportConfig placement option correctly re-render the overlay? From what I can see from the demo, this doesn't work yet. Should this be included in the scope here? It's a feature that works for LocalOverlay popper config, so would be nice to have it here as well.

For a LocalOverlay since it should react on the viewport, but i can't think of any use-case for GlobalOverlays.

> * Do we know what happens if the content inside a bottomsheet goes beyond the viewheight? Should we add a maximum BottomSheet height so there's always x amount of space between the edge of the viewport and the sheet? Also, does the content become scrollable and does that work both vertically / horizontally?
 
All styling you mention can be set on the bottomsheet (frame) itself and considered out of scope for the controller, since it is more use-case dependent.

packages/overlays/src/GlobalOverlayController.js
gerjanvangeest: Where else do you see it?

PR #285, opened 2019-09-24
feat(overlays): add element property

michellangeveld:
Add element property to retrieve element in the overlay.
This property can be used for example in animations.

### DIFF
--- a/packages/overlays/src/GlobalOverlayController.js
+++ b/packages/overlays/src/GlobalOverlayController.js
@@ -82,6 +82,13 @@ export class GlobalOverlayController {
   }
 
   /**
+   * @type {HTMLElement | null} the first rendered element of the template
+   */
+  get contentElement() {
+    return this._container && this._container.firstElementChild;
+  }
+
+  /**
    * Updates an overlay's template. Creates a render container and appends it to the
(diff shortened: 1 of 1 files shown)


jorenbroekema:
Hey, thanks for contributing! Can you explain your use case a little further? How would it be used in animations? Knowing this is helpful for us to understand how our users want to use this.

As for the name, element is a bit ambiguous because it could be the content as well as invoker nodes. So perhaps `contentElement` instead :)?

Rudolphous:
If a contentElement animates and fires an event when finished  - with this property it's is possible to set an eventListener and catch that event.

Another example is when function needs to be called on the contentElement.

About the name: will rename the property.

daKmoR:
exposed as `contentNode` within https://github.com/ing-bank/lion/pull/254 it's maybe better to wait for that one to be merged?

LarsDenBakker:
Any ETA on when that will be merged?

gerjanvangeest:
> Any ETA on when that will be merged?

It is already merged via another MR: https://github.com/ing-bank/lion/pull/288

michellangeveld:
@gerjanvangeest : works correct, will close this merge request

PR #286, opened 2019-09-24
fix(calendar): render the calendar when min and max date are same date

jcperea:
It is an edge case, but when trying to render a calendar which has `minDate` and `maxDate` with the same date, and this date is not today, the calendar can't render since tries to select a `centralDate` outside the valid date range.

When the date is "today", everything goes as expected, but when it is a future or a past date, the calendar throws an error at `__findBestEnabledDateFor`.

I'm providing a test to this feature, and I did change two values within two tests, which I think they make sense. `centralDate` was calculated as the next day from the `minDate` value, or the day before from the `maxDate` value, EXCEPT when `minDate` is lower than today, or `maxDate` is greater (and we only provide one validator), or when the range between `minDate` and `maxDate` contains today.


gerjanvangeest:
Hi, sounds like a good fix, we will look at it later this/next week.

PR #288, opened 2019-09-25
merge next overlays to master

tlouisse:


packages/overlays/src/GlobalOverlayController.js
```diff
@@ -243,4 +251,12 @@ export class GlobalOverlayController extends BaseOverlayController {
 
     this.__hasActiveBackdrop = false;
   }
+
+  // TODO: this method has to be removed when EventTarget polyfill is available on IE11
+  __fakeExtendsEventTarget() {
```
daKmoR: already part of the BaseOverlayController? e.g. probably not needed here?

PR #289, opened 2019-09-26
Feat/rich select allow dynamic overlays

tlouisse:


jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/298

PR #290, opened 2019-09-26
fix: Change CN translation for date format validation

gafortuno:


gerjanvangeest:
Thanks for the pull request :)

PR #291, opened 2019-09-27
fix contentNodes for local and global overlays

jorenbroekema:


packages/overlays/test/LocalOverlayController.test.js
```diff
@@ -122,6 +122,54 @@ describe('LocalOverlayController', () => {
...
+      const node = document.createElement('div');
+      node.innerHTML = '<div id="content">Content</div>';
+
+      const ctrl = new LocalOverlayController({
+        contentNode: node,
+        invokerNode,
+      });
+      const el = await fixture(html`
+        <div>
+          ${ctrl.invoker} ${ctrl.content}
```
daKmoR: `ctrl.invoker` is deprecated? should be `invokerNode`?

packages/overlays/test/GlobalOverlayController.test.js
```diff
@@ -154,6 +154,50 @@ describe('GlobalOverlayController', () => {
...
+      it('accepts HTML Element (contentNode) as content', async () => {
+        const contentNode = await fixture(
+          html`
+            <p>my content</p>
+          `,
+        );
+
+        const ctrl = overlays.add(
+          new GlobalOverlayController({
+            contentNode,
```
daKmoR: what's the use-case? we should never do this as it would require to move user-provided light dom to move to the body... (which will mix up a lot of stuff... like the parents' lit-html template)?

jorenbroekema: I agree it's not ideal to do this, but I was under the impression we should allow this as it is public API. And we kind of need it to combine rich select with global overlays, where the rich select needs to have the DOM (template isn't enough) to function properly. So therefore we figured let's make it work with nodes first and see if we can make it work for templates later.

packages/overlays/test/LocalOverlayController.test.js
jorenbroekema: Aahhhh I didn't catch that memo, I kinda copied the template from some other test :P

Issue #272, opened 2019-09-10
[localize] formatNumber function returns 'č' when the currency is 'CZK'


revoul:
Hi guys!

I'm using the formatNumber function of @lion/localize and I think it has an unexpected behavior.

```html
<html lang="cs-CZ">
```
```js
const myNumber = formatNumber(1000, {
    style: 'currency',
    currencyDisplay: 'symbol',
    currency: 'CZK',
});
```
The return value is 1 000,00 `č`, when I hope to receive 1 000,00 `Kč`.

Example: https://codesandbox.io/embed/magical-banzai-70smm

jorenbroekema:
Thanks for the codesandbox, makes it pretty clear the difference between Intl and our own formatter.

From first glance I'm not sure what the cause is.

Perhaps you can do a PR for a fix?

PR #292, opened 2019-09-27
Fix/currency symbols

gerjanvangeest:
fix: https://github.com/ing-bank/lion/issues/272


packages/localize/test/number/formatNumber.test.js
```diff
@@ -259,19 +259,19 @@ describe('formatNumber', () => {
         expect(formatNumber(123456.789, currencyCode('USD'))).to.equal('123 456,79 USD');
         expect(formatNumber(123456.789, currencyCode('JPY'))).to.equal('123 457 JPY');
         expect(formatNumber(123456.789, currencySymbol('EUR'))).to.equal('123 456,79 €');
-        expect(formatNumber(123456.789, currencySymbol('USD'))).to.equal('123 456,79 $');
+        expect(formatNumber(123456.789, currencySymbol('USD'))).to.equal('123 456,79 $US');
         expect(formatNumber(123456.789, currencySymbol('JPY'))).to.equal('123 457 ¥');
```
gerjanvangeest: For `fr-FR` and `fr-BE` we did a forceYenSymbol, i dont know if it should be `¥`, `JP¥` (like other countries) or `¥JP` (just like the France do `$US`). Left it as is for now.

PR #294, opened 2019-09-30
[WIP] test: trying to make a select work with global overlay using nodes

jorenbroekema:
Highly WIP. 

After a fair bit of trial and error I figured nodes were easier than templates, so perhaps ok to start there. Kind of have a working demo, but not yet working with dynamic overlay controller because of multiple global controllers using the same contentNode reference and conflicting each other.

### DIFF
--- a/packages/overlays/src/DynamicOverlayController.js
+++ b/packages/overlays/src/DynamicOverlayController.js
@@ -31,9 +31,6 @@ export class DynamicOverlayController {
     this.__list = [];
     this.__active = undefined;
     this.nextOpen = undefined;
-    if (!this.content) {
-      this.content = document.createElement('div');
-    }
     this.__fakeExtendsEventTarget();
     this.__delegateEvent = this.__delegateEvent.bind(this);
   }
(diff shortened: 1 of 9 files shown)


jorenbroekema:
Ah shit I meant to make a draft PR, this is not really ready for review but feel free to comment I suppose 🙈

packages/overlays/src/DynamicOverlayController.js
```diff
@@ -106,6 +103,20 @@ export class DynamicOverlayController {
     return this.active.invokerNode;
   }
 
+  get contentNode() {
+    return this.active ? this.active.contentNode : document.createElement('div');
```
jorenbroekema: probably should return undefined and not empty div

packages/overlays/src/GlobalOverlayController.js
```diff
@@ -57,6 +57,14 @@ export class GlobalOverlayController extends BaseOverlayController {
     }
   }
 
+  _connectContent() {
```
jorenbroekema: this should probably be called when a contentNode is passed to a new controller instance, because it means 'css' should be show hide mode, and if that's true, the content should be connected and hidden. Right now I'm doing it in the account select

PR #295, opened 2019-10-01
fix(form-registrar-mixin): invoke becomesReady on connectedCallback

JaySunSyn:
When using `ing-radio-group` in a component and caching the component with the `cache` directive, the formRegisteringMixin does not register elements because the manager is not ready (https://github.com/ing-bank/lion/blob/master/packages/field/src/FormRegisteringMixin.js#L31).

Therefore `formElements` is empty when returning from the cache.

**Underlying problem**

When `disconnectedCallback` is called in components with the `FormRegistrarMixin` applied, the form elements are removed via `formRegistrarManager.remove(this)`;

When the component comes back from the cache, `connectedCallback` gets called but **not** `firstUpdated`.  Currently, we call `formRegistrarManager.becomesReady(this);` only in the `firstUpdated` callback. This is why `formRegistrarManager` never gets ready when coming back from the cache and `formElements` stay empty.


JaySunSyn:
I sthere any update on this?

packages/field/src/FormRegistrarMixin.js
```diff
@@ -38,6 +38,7 @@ export const FormRegistrarMixin = dedupeMixin(
           super.connectedCallback();
         }
         formRegistrarManager.add(this);
+        formRegistrarManager.becomesReady(this);
```
daKmoR: you will probably need to have a 
```js
if (this.__hasBeenRenedered) {
  formRegistrarManager.becomesReady(this);
}

  // ...
  this.__hasBeenRenedered = true; // also pls init it with fall in the constructor
  // or is there a LitElement flag for that state already?
}
```

in there to make sure it does not become ready before it has been rendered at least once (to not break polyfilled browsers)

can you test it that would still work in your use case?

a test would be highly appreciated 🤗

packages/field/src/FormRegistrarMixin.js
```diff
@@ -25,6 +25,7 @@ export const FormRegistrarMixin = dedupeMixin(
         super();
         this.formElements = [];
         this.__readyForRegistration = false;
+        this.__hasBeenRenedered = false;
```
gerjanvangeest: Typo: `__hasBeenRendered `

packages/field/src/FormRegistrarMixin.js
```diff
@@ -38,6 +39,9 @@ export const FormRegistrarMixin = dedupeMixin(
           super.connectedCallback();
         }
         formRegistrarManager.add(this);
+        if (this.__hasBeenRenedered) {
+          formRegistrarManager.becomesReady(this);
```
gerjanvangeest: `becomesReady()` doesn't need a parameter.

JaySunSyn: Thanks

packages/field/src/FormRegistrarMixin.js
JaySunSyn: Thanks

packages/field/test/FormRegistrarMixin.test.js
```diff
@@ -0,0 +1,71 @@
+import { fixture, html, expect } from '@open-wc/testing';
+import { cache, LitElement } from '@lion/core';
+import { LionFieldset } from '../../fieldset/src/LionFieldset.js';
+import { FormRegistrarMixin } from '../src/FormRegistrarMixin.js';
+
+describe.only('FormRegistrarMixin', () => {
```
gerjanvangeest: `only` is very handy to run your unit tests, but should not end up in the pull request :)

PR #300, opened 2019-10-03
fix(form-registering-mixin): remove 'all-forms-open-for-registration' on disconnected

AnthonyKot:
**Problem**
Listener 'all-forms-open-for-registration' from FormRegisteringMixin.js line 34 is never removed, causing memory leaks and errors in multiple component registration

I think it could be removed on `disconnectedCallback()` like many others.

### DIFF
--- a/packages/field/src/FormRegisteringMixin.js
+++ b/packages/field/src/FormRegisteringMixin.js
@@ -26,7 +26,7 @@ export const FormRegisteringMixin = dedupeMixin(
         }
         formRegistrarManager.removeEventListener(
           'all-forms-open-for-registration',
-          this._dispatchRegistration(),
+          this._dispatchRegistration,
         );
         this._unregisterFormElement();
       }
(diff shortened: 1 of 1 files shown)


packages/field/src/FormRegisteringMixin.js
```diff
@@ -24,16 +24,21 @@ export const FormRegisteringMixin = dedupeMixin(
...
       __setupRegistrationHook() {
         if (formRegistrarManager.ready) {
           this._dispatchRegistration();
         } else {
-          formRegistrarManager.addEventListener('all-forms-open-for-registration', () => {
-            this._dispatchRegistration();
-          });
+          formRegistrarManager.addEventListener(
+            'all-forms-open-for-registration',
+            this._dispatchRegistration(),
```
JaySunSyn: Hey, just stumbled upon the PR and noticed that you invoke a function here rather than passing in a function. (I'm not an official reviewer of this PR)

Issue #190, opened 2019-07-22
[field] Feedback slots lack aria-live


erikkroes:
The inputs have feedback slots that have a relationship to the input with `aria-describedby`. This makes it that a user of assistive technology is able to perceive the feedback when focusing on the input. However, sometimes the feedback updates while the user is on the input. For example, when a user is typing, it can get validated instantly. 
`aria-live="politie"`should be added to the feedback slot to have these updates communicated to assistive technology.

gerjanvangeest:
It did, so closing this issue!

PR #301, opened 2019-10-03
[field] Add aria-live=polite attribute to feedback slot

noventadev:
Closes [field] Feedback slots lack aria-live #190


noventadev:
Happy to help!

PR #304, opened 2019-10-04
Overlay controller refactors

jorenbroekema:
- Merge global, local and base into OverlayController
- Deprecate old controllers
- Tests
  - Fix broken tests
  - New tests
  - Skip tests that we can fix later (Popper, and rich-select which we broke with these changes)

This MR is based on the poc/accountSelect branch, but takes only the changes necessary for accountSelect, and does not bother to apply changes to demos, documentation, or components that use the overlay system like datepicker, etc. etc. Also does not move the tests to a new folder.
We can do all that later.


daKmoR:
superseded by https://github.com/ing-bank/lion/pull/315

PR #307, opened 2019-10-07
Feat/overlay controller

CubLion:
content wrapper has display inline-block when shown


daKmoR:
superseded by https://github.com/ing-bank/lion/pull/315

PR #314, opened 2019-10-09
[select-rich] Add left and right button for navigating the options

jorenbroekema:
In retrospect, not sure if we want this. But I did the work already so I guess I just put this here.

### DIFF
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -401,6 +401,14 @@ export class LionSelectRich extends FormRegistrarMixin(
         ev.preventDefault();
         this.activeIndex = this.__getNextEnabledOption(this.activeIndex);
         break;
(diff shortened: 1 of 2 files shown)


tlouisse:
Looks great.

I will add it to our extension layer and put it on your name:) 
Since here we followed the official aria specs and took the native `<select>` as reference.

See:
- https://www.w3.org/TR/wai-aria-practices/#listbox_kbd_interaction
- https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

tlouisse:
Looks great.

Will add this in our extending layer and put it on your name :)
Since here we followed the official aria design patterns and took the native select as a reference.

See:
- https://www.w3.org/TR/wai-aria-practices/#listbox_kbd_interaction
- https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

Issue #296, opened 2019-10-02
[localize] date formatting forced


LarsDenBakker:
Formatting options for year, month and day are forcibly set. It's not possible to format a date without a year right now. See: https://github.com/ing-bank/lion/blob/master/packages/localize/src/date/formatDate.js#L22

For example with native Intl this prints the date of today without the year:
```
new Intl.DateTimeFormat('en-GB', { day: '2-digit', month: 'long' }).format(new Date());
```

Passing the same options to `parseDate` forcibly adds the year.

daKmoR:
how would you like to do it? with `false`?
```js
formatDate(new Date(), { year: false }); // return 10/30 // depending on locale and date
```

LarsDenBakker:
I expect `formatDate` to accept the same options as `Intl.DateTimeFormat`. So to leave out the year we don't pass an option.

r4zorgeek:
@LarsDenBakker Hey I am new to open source. Can I take this issue?

gerjanvangeest:
> @LarsDenBakker Hey I am new to open source. Can I take this issue?

I know @LarsDenBakker is busy with a lot of other issues as well. So I am quite sure he doesn't mind if you add this feature.

LarsDenBakker:
Yeah go aheae

r4zorgeek:
@LarsDenBakker @gerjanvangeest  Hey can you explain the issue in detail. Sorry for the inconvenience new to open source.
According to my understanding of the problem, we want a date without year for that to happen we will not pass the year in options. Correct me if I am wrong.

So basically to solve I removed line no. 23 from formatDate.js

r4zorgeek:
@LarsDenBakker @gerjanvangeest 
also if we remove these lines i think dates formatting would not be forced.
if (options && !(options && options.year)) {

  }

gerjanvangeest:
> @LarsDenBakker @gerjanvangeest
> also if we remove these lines i think dates formatting would not be forced.
> if (options && !(options && options.year)) {
> 
> }

Yes, and also the lines for `day` and `month`.

Please also add some tests and fix where needed. And add an example to the storybook demo.

r4zorgeek:
@LarsDenBakker @gerjanvangeest Fixed the issue. Please review it. Also added test cases and demo in storybook.

PR #321, opened 2019-10-11
fix: [localize] date formatting forced #296

r4zorgeek:
also added test cases and storybook demo.

### DIFF
--- a/packages/localize/src/date/formatDate.js
+++ b/packages/localize/src/date/formatDate.js
@@ -13,22 +13,16 @@ export function formatDate(date, options) {
     return '';
   }
   const formatOptions = options || {};
-  // make sure months and days are always 2-digits
-  if (!options) {
+  /**
+   * Set smart defaults if:
+   * 1) no options object is passed
+   * 2) options object is passed, but none of the following props on it: day, month, year.
+   */
+  if (!options || (options && !options.day && !options.month && !options.year)) {
     formatOptions.year = 'numeric';
     formatOptions.month = '2-digit';
     formatOptions.day = '2-digit';
   }
-  if (options && !(options && options.year)) {
-    formatOptions.year = 'numeric';
-  }
-  if (options && !(options && options.month)) {
-    formatOptions.month = '2-digit';
-  }
-  if (options && !(options && options.day)) {
-    formatOptions.day = '2-digit';
-  }
-
   const computedLocale = getLocale(formatOptions && formatOptions.locale);
   let formattedDate = '';
   try {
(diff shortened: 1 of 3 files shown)


r4zorgeek:
@LarsDenBakker Hey its been a couple of days since I committed can you please review it?

jorenbroekema:
Hey @r4zorgeek , is this tied to an issue? Because I am not entirely sure about this PR. 

Basically, PR does two things:
- Allow people to format without a certain part (day, month, year), this is intended.
- Remove our defaults in a breaking way (not intended, see example below).

Why is it breaking? The failing test shows it actually.
For example:
```js
formatDate(new Date('2017/06/15'), { locale: 'en-US' });
```
outputs 6/15/2017 instead of 06/15/2017 now, and this is breaking behavior.

So basically, I think there should be some discussion / research.

In my opinion:
I agree we should allow people to format without certain parts (day, month, year). The way we set "smart defaults" overrides this Intl feature and that's in my opinion a "bug" on our part, I think it was unintended and even if it was intended, I think it's too opinionated that we do this. However, as I said before, it is a breaking change and requires people extending lion to change their code if they want the same output as they had before. So we should definitely be mindful about that and release this with the information that this is breaking. I don't think backwards compatibility is necessary.

If it was up to me, we change the tests to `6/15/2017` and change commit to `fix(localize): remove defaults that prevented excluding parts (BREAKING)`. Then on our extension layer we can add these defaults back in if we still feel like we should enforce it, but then we should add an API that allows people to exclude those parts, as a feature. 

Proposal for API on our extension layer could be:
```js
formatDate(new Date(), { 
  weekday: 'long',
  year: 'numeric',
  month: 'long',
  day: 'none'
});
```
Where under the hood we take 'none' and disable the defaults we set for that specific part (day, month, year), so that way it gets excluded entirely.

note: semver/conventional commits allow for specifying breaking changes with a commit: `BREAKING CHANGE(localize): remove defaults that prevented excluding parts` but im not sure what lerna would do with this, usually this is a major bump so I'm not sure if it would release v1, that would be very bad :P since we are still in alpha. @daKmoR ?

Sorry for the long message, the PR seems simple at the start, but has side effects which are breaking, so we have to think carefully about it :)

gerjanvangeest:
We can do the following to prevent the failing tests:

```js
  if (!options.day && !options.month && !options.year) {
    formatOptions.year = 'numeric';
    formatOptions.month = '2-digit';
    formatOptions.day = '2-digit';
  }
```

But I agree with @jorenbroekema that the change is breaking, but we will fix that on our extension layer.

r4zorgeek:
@jorenbroekema @gerjanvangeest Thanks for your great comments. Can you please tell how should I proceed further. Should I add the code which is suggested by @gerjanvangeest

gerjanvangeest:
> @jorenbroekema @gerjanvangeest Thanks for your great comments. Can you please tell how should I proceed further. Should I add the code which is suggested by @gerjanvangeest

Yes please

Issue #316, opened 2019-10-10
[button] always submit the form 


igomezal:
With the new change in lion-button, it always submits the form even if you try to catch the event and prevent the default behavior of it (with `ev.preventDefault()`).

It seems to be caused by https://github.com/ing-bank/lion/blob/master/packages/button/src/LionButton.js#L177

```javascript
  __clickDelegationHandler() {
    if (this.type === 'submit' && this._nativeButtonNode && this._nativeButtonNode.form) {
      this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
      this._nativeButtonNode.form.submit();
    }
  }
```

Although it tries to replicate the behavior of a native form, it is not working correctly.

Here you have an example of two forms using lion-button and another two using a native button, each example is listening to different events to show how `ev.preventDefault()` should work.

https://stackblitz.com/edit/submit-lion-button?file=index.js

igomezal:
It seems tests are not really checking the functionality correctly, as they are overwriting the submit method from the form HTMLElement: https://github.com/ing-bank/lion/blob/master/packages/button/test/lion-button.test.js#L236

It shouldn't be needed if you are already doing the standard event.preventDefault in the submit or click event: https://github.com/ing-bank/lion/blob/master/packages/button/test/lion-button.test.js#L229

For example:
```javascript
it('behaves like native `button` when clicked', async () => {
      const formSubmitSpy = sinon.spy(e => e.preventDefault()); // You should only need to do this
      const form = await fixture(html`
        <form @submit="${formSubmitSpy}">
          <lion-button type="submit">foo</lion-button>
        </form>
      `);
      // Prevent page refresh
      form.submit = () => {}; // You shouldn't need to do this

      const button = form.querySelector('lion-button');
      getTopElement(button).click();

      expect(formSubmitSpy.called).to.be.true;
    });
```

Instead this test should be something like:

```javascript
it('behaves like native `button` when clicked', async () => {
      const formSubmitSpy = sinon.spy(e => e.preventDefault());
      const form = await fixture(html`
        <form @submit="${formSubmitSpy}">
          <lion-button type="submit">foo</lion-button>
        </form>
      `);

      const button = form.querySelector('lion-button');
      getTopElement(button).click();

      expect(formSubmitSpy.called).to.be.true;
    });
```

jorenbroekema:
Did you try if that test runs now that the page refresh is not explicitly prevented by setting the submit method to an empty function? 

[...]

PR #326, opened 2019-10-14
WIP: [lion-button] fix submit of lion-buttons

igomezal:
Failing tests to help solving https://github.com/ing-bank/lion/issues/316

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -176,8 +176,7 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    */
   __clickDelegationHandler() {
     if (this.type === 'submit' && this._nativeButtonNode && this._nativeButtonNode.form) {
-      this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
-      this._nativeButtonNode.form.submit();
+      this._nativeButtonNode.click();
     }
   }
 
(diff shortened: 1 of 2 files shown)


gerjanvangeest:
Thanks for helping with better tests, I believe i have fixed the code: https://github.com/ing-bank/lion/pull/328 , will use your tests to confirm.

packages/button/src/LionButton.js
```diff
@@ -176,8 +176,7 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    */
   __clickDelegationHandler() {
     if (this.type === 'submit' && this._nativeButtonNode && this._nativeButtonNode.form) {
-      this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
-      this._nativeButtonNode.form.submit();
+      this._nativeButtonNode.click();
```
tlouisse: I think this would now fire two events that would both bubble to parents. Not really ideal I think. What fails when we only have `this._nativeButtonNode.form.dispatchEvent(new Event('submit'));`?

packages/button/src/LionButton.js
```diff
@@ -229,6 +230,11 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
     }
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  __nativeButtonHandler(event) {
+    event.stopPropagation();
```
tlouisse: So implicit form submission still works like this?

igomezal: Yeah, because it is only stopping the propagation of the event, it does not prevent the default action (in this case the submit action).

> The Event interface's preventDefault() method tells the user agent that if the event does not get explicitly handled, its default action should not be taken as it normally would be. [MDN preventDefault docs](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

gerjanvangeest:
We solved the issue in: https://github.com/ing-bank/lion/pull/328 , thanks for helping out, it gave us some good insights.

Issue #322, opened 2019-10-12
lion-form is submitted 2 times


lfusiara:
In code like that, if you push enter on doc field form is submitted 2 times

```
    return html`
      <lion-form id="client-search-form" @submit=${e => console.log(e)}>
        <form>
          <lion-input name="doc"></lion-input>
          <lion-button type="submit" name="sub"></lion-button>
        </form>
      </lion-form>
    `;
```

gerjanvangeest:
Hi @lfusiara, we had this bug a few weeks ago too, and fixed it in this pull request: https://github.com/ing-bank/lion/issues/240

Did you use the latest version, or could it be that the issue is already resolved in the latest version of `lion`?

SantiEspada:
I think it is actually caused by that pull request, there is another related issue #316

gerjanvangeest:
Should be fixed at: https://github.com/ing-bank/lion/pull/328

PR #328, opened 2019-10-15
fix(button): remove submit() on click

gerjanvangeest:
fixes https://github.com/ing-bank/lion/issues/322
fixes https://github.com/ing-bank/lion/issues/316

### DIFF
--- a/packages/button/README.md
+++ b/packages/button/README.md
@@ -28,15 +28,38 @@ import '@lion/button/lion-button.js';
 <lion-button>Button Text</lion-button>
 ```
 
-- Don't use a button when you want a user to navigate. Use a link instead.
-- Not all color and font size combinations are available because some do not meet accessibility contrast requirements
-
 ## Considerations
 
-### Why a webcomponent?
+### Why a Web Component?
 
-There are multiple reasons why we used a web component as opposed to a CSS component.
+There are multiple reasons why we used a Web Component as opposed to a CSS component.
 
 - **Target size**: The minimum target size is 40 pixels, which makes even the small buttons easy to activate. A container element was needed to make this size possible.
 - **Accessibility**: Our button is accessible because it uses the native button element. Having this native button element available in the light dom, preserves all platform accessibility features, like having it recognized by a native form.
 - **Advanced styling**: There are advanced styling options regarding icons in buttons, where it is a lot more maintainable to handle icons in our button using slots. An example is that a sticky icon-only buttons may looks different from buttons which have both icons and text.
+
+### Event target
+
+We want to ensure that the event target returned to the user is `lion-button`, not `button`. Therefore, simply delegating the click to the native button immediately, is not desired. Instead, we catch the click event in the `lion-button`, and ensure delegation inside of there.
+
+### Flashing a native button click as a direct child of form
+
+By delegating the `click()` to the native button, it will bubble back up to `lion-button` which would cause duplicate actions. We have to simulate the full `.click()` however, otherwise form submission is not triggered. So this bubbling cannot be prevented.
+Therefore, on click, we flash a `<button>` to the form as a direct child and fire the click on that button. We then immediately remove that button. This is a fully synchronous process; users or developers will not notice this, it should not cause problems.
+
+### Native button & implicit form submission
+
+Flashing the button in the way we do solves almost all issues except for one.
+One of the specs of W3C is that when you have a form with multiple inputs, pressing enter while inside one of the inputs only triggers a form submit if that form has a button of type submit.
+
+To get this particular implicit form submission to work, having a native button in our `lion-button` is a hard requirement. Therefore, not only do we flash a native button on the form to delegate `lion-button` trigger to `button` and thereby trigger form submission, we **also** add a native `button` inside the `lion-button` which `type` property is synchronized with the type of the `lion-button`.
+
+### Preventing full page reloads
+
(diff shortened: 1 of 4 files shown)


packages/button/src/LionButton.js
```diff
@@ -172,12 +172,25 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
+      this.type === 'submit' &&
+      this._nativeButtonNode &&
+      this._nativeButtonNode.form &&
+      e.target === this
+    ) {
+      // this submit cannot be prevented in FF
       this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
-      this._nativeButtonNode.form.submit();
     }
   }
```
jorenbroekema: I would suggest cleaning it a bit, hard to understand what's going on.
```suggestion
__clickDelegationHandler(e) {
const isNativeSubmit = this.type === 'submit' && this._nativeButtonNode && e.target === this;

if (this.constructor.__isIE11()) {
  // no need to do anything for IE11
  e.stopPropagation();
  return;
}

if (this.constructor.__isFF() && isNativeSubmit) {
  // this click renders multiple submits in other browsers
  this._nativeButtonNode.click();
  return;
}

if (isNativeSubmit && this._nativeButtonNode.form) {
  // this submit cannot be prevented in FF
  this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
}
```

jorenbroekema: Hmm multi-line comment suggestions not entirely working as I expected 🙈

packages/button/test/lion-button.test.js
```diff
@@ -225,74 +225,132 @@ describe('lion-button', () => {
...
+        `);
+
+        pressEnter(form.querySelector('lion-button'));
+        await aTimeout();
+        await aTimeout();
+
+        expect(formButtonClickedSpy.callCount).to.equal(1);
+      });
+
+      // input "enter" keypress mock doesn't seem to work right now, but should be tested in the future (maybe with Selenium)
```
jorenbroekema: Duplicate test

packages/button/src/LionButton.js
```diff
@@ -172,12 +179,22 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-  __clickDelegationHandler() {
-    if (this.type === 'submit' && this._nativeButtonNode && this._nativeButtonNode.form) {
-      this._nativeButtonNode.form.dispatchEvent(new Event('submit'));
-      this._nativeButtonNode.form.submit();
+  __clickDelegationHandler(e) {
+    if (this.constructor.__isIE11()) {
+      e.stopPropagation();
+    }
+
+    if (this.type === 'submit' && e.target === this) {
```
igomezal: Maybe it could be a good idea to also add a check for the type `reset`, something like
```javascript
if((this.type === 'submit' || this.type === 'reset') && e.target === this) { // Here you check if it is reset type
   if (this._form) {
        const nativeButton = document.createElement('button');
        nativeButton.type = this.type; // The button is created with the correct type
        this._form.appendChild(nativeButton);
        nativeButton.click();
        this._form.removeChild(nativeButton);
      }
    }
}
```
This way `lion-button` will also be able to clear all the inputs inside a form

LarsDenBakker:
This kind of trickery might set us up for more bugs in the future. 

Have you considered if it's not a better idea now to move back to just regular buttons, or rendering the button in the shadow dom? We can't always meet all requirements.

packages/button/README.md
```diff
@@ -28,15 +28,28 @@ import '@lion/button/lion-button.js';
...
+Therefore, on click, we flash a `<button>` to the form as a direct child and fire the click on that button. We then immediately remove that button. This is a fully synchronous process; users or developers will not notice this, it should not cause problems.
+
+### Native button & implicit form submission
+
+Flashing the button in the way we do solves almost all issues except for one.
+One of the specs of W3C is that when you have a form with multiple inputs, pressing enter while inside one of the inputs only triggers a form submit if that form has a button of type submit.
+
+To get this particular implicit form submission to work, having a native button in our `lion-button` is a hard requirement. Therefore, not only do we flash a native button on the form to delegate `lion-button` trigger to `button` and thereby trigger form submission, we **also** add a native `button` inside the `lion-button` which `type` property is synchronized with the type of the `lion-button`.
```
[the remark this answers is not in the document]
jorenbroekema: Good idea, I have added it :)

jorenbroekema:
@LarsDenBakker  I agree with that it's trickery. We did actually have some discussions regarding this. The requirement that the custom element button should support implicit form submission similar to the native way, stems from the polymer era. In the end we did not feel comfortable removing that. If anything I would call removing it, pretty 'breaking'.

I'd also be down with a button-simple component that is just.. a LOT simpler, without wrapping native, where we just make it work with lion-form and let that be good enough. Then at least we can check to see which button people prefer using. If it turns out 99% of people use the simple button and don't actually need the features of native, we could hypothetically deprecate the complicated-native-button-wrapping button in favour of the simple more stable one. Just an idea.

packages/button/src/LionButton.js
jorenbroekema: Aah yeah good catch, we supported that (but did not test) it in the past so we should support it now. I added a commit on top for it and included a test in it.

PR #330, opened 2019-10-16
Feat/switch

CubLion:


packages/switch/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@lion/switch",
+  "version": "0.0.1",
+  "descriotion": "TODO: add",
```
CubLion: description

packages/switch/stories/index.stories.js
```diff
@@ -0,0 +1,90 @@
...
+    'Default',
+    () => html`
+      <lion-input-switch></lion-input-switch>
+    `,
+  )
+  .add(
+    'All text slots',
+    () => html`
+      <lion-input-switch label="Label" help-text="Help text">
+        <div slot="after">After</div>
```
CubLion: remove the "after" slotted div

packages/switch/stories/index.stories.js
```diff
@@ -0,0 +1,90 @@
...
+              ...super.localizeNamespaces,
+            ];
+            return result;
+          }
+
+          render() {
+            return html`
+              <lion-form id="postsForm" @submit="${this.submit}">
+                <form>
+                  <lion-input-switch name="emailAddress" label="Share email address">
```
CubLion: consistent indentation with the "validated" switch below

packages/switch/stories/index.stories.js
```diff
@@ -0,0 +1,90 @@
...
+                    Submit
+                  </button>
+                </form>
+              </lion-form>
+            `;
+          }
+
+          submit() {
+            const form = this.shadowRoot.querySelector('#postsForm');
+            const validateElement = this.shadowRoot.querySelector('[name=emailAddress]');
```
CubLion: remove "validateElement" block as it no longer makes sense

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,95 @@
+import { LionButton } from '@lion/button';
+import { html, css } from '@lion/core';
+
+export class LionButtonSwitch extends LionButton {
+  static get properties() {
+    return {
+      readOnly: {
```
tlouisse: Any reason we redefine this attr here? It's already added by LionButton (which shouldn't add it, because it's already part of HTMLElement ;))

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,95 @@
...
+
+  constructor() {
+    super();
+    this.type = 'button';
+    this.checked = false;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this._clickHandler = () => {
```
tlouisse: Small remark: we usually do double underscores for handlers are not intended to be overwritten by Subclassers. Maybe a less generic name would also prevent chances of someone extending this class accidentally overriding it.

packages/switch/src/LionInputSwitch.js
```diff
@@ -0,0 +1,63 @@
...
+    super.disconnectedCallback();
+    this._cleanButtonSwitch();
+  }
+
+  updated(changedProperties) {
+    super.updated(changedProperties);
+    this._syncButtonSwitch();
+  }
+
+  get _buttonSwitch() {
```
tlouisse: As a convention: getters like these are usually suffixed with `Node`. 
Kind of means they refer functionally important nodes in your component and Subclassers are allowed to either call or override those getters.

Something we should definitely document somewhere ;)

packages/switch/src/LionButtonSwitch.js
erikkroes: Should we even have a readonly? A switch is meant to immediately activate something. I would say it's either enabled or disabled. Readonly is for something that should be sent with a form, but you can't change. I'll check with UX.

packages/switch/stories/index.stories.js
erikkroes: I think the "all text slots" can be the default demo. It should never be used without a label anyway.

packages/switch/stories/index.stories.js
```diff
@@ -0,0 +1,90 @@
...
+            validateElement.validate();
+            if (form.errorState === false) {
+              console.log(form.serializeGroup());
+            }
+          }
+        },
+      );
+    }
```
erikkroes: An added demo that shows a disabled switch would be nice

packages/switch/src/LionButtonSwitch.js
daKmoR: native button does not have readonly? not as attribute nor as property
also I assume readOnly is NOT part of HTMLElement?

did you maybe mean the FieldMixin or so? (which is not applied to a button?)

packages/switch/stories/index.stories.js
CubLion: added, please confirm and "Resolve conversation"

packages/switch/src/LionInputSwitch.js
CubLion: updated, please confirm and "Resolve conversation"

packages/switch/src/LionButtonSwitch.js
CubLion: updated, please confirm and "Resolve conversation"

packages/switch/src/LionInputSwitch.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  _setupButtonSwitch() {
+    this._buttonCheckedHandler = e => {
+      this.checked = e.detail;
+    };
+    this._buttonSwitchNode.addEventListener('checked-changed', this._buttonCheckedHandler);
+  }
+
+  _cleanButtonSwitch() {
```
jorenbroekema: if the input node is disconnected, so is the switch button right? Since it is a child. 

Then I don't think we need the setup / teardown pattern and we can put the handler as a direct method on the class. Would save quite some boilerplate code.

packages/switch/src/LionInputSwitch.js
```diff
@@ -0,0 +1,86 @@
...
+    super.disconnectedCallback();
+    this._cleanButtonSwitch();
+  }
+
+  updated(changedProperties) {
+    super.updated(changedProperties);
+    this._syncButtonSwitch();
+  }
+
+  // TODO: should be replaced by "_inputNode"
```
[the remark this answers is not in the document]
jorenbroekema: Link to the part of docs that show this breaking change / todo

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    el = await fixture(html`
+      <lion-button-switch></lion-button-switch>
+    `);
+  });
+
+  it('should behave as a button', () => {
+    expect(el instanceof LionButton).to.be.true;
+  });
+
+  it('should have tabindex="0"', () => {
```
jorenbroekema: maybe `should be focusable` to make it read more like english and to communicate better to non-devs (which is one of the strives of BDDs)

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    expect(el.tabIndex).to.equal(0);
+    expect(el.getAttribute('tabindex')).to.equal('0');
+  });
+
+  it('should have type="button"', () => {
+    expect(el.type).to.equal('button');
+    expect(el.getAttribute('type')).to.equal('button');
+  });
+
+  it('should have disabled=false by default', () => {
```
jorenbroekema: I think it's better to test that it can be disabled. An attribute or property that doesn't exist is by default false, so you're actually not testing anything here that is specific to implementation.

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    expect(el.disabled).to.equal(false);
+    expect(el.hasAttribute('disabled')).to.be.false;
+  });
+
+  it('should have checked=false by default', () => {
```
jorenbroekema: I think this one makes sense to test :), because it's a core feature that is toggleable so default state here is kinda important to show as "behavior"

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    expect(el.hasAttribute('checked')).to.be.true;
+    // when
+    el.click();
+    await el.updateComplete;
+    // then
+    expect(el.checked).to.be.false;
+    expect(el.hasAttribute('checked')).to.be.false;
+  });
+
+  it('should not toggle the valoue of "checked" on click if disabled', async () => {
```
jorenbroekema: `can be disabled` should be sufficient

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    el.checked = true;
+    // when
+    el.click();
+    await el.updateComplete;
+    // then
+    expect(el.checked).to.be.true;
+    expect(el.hasAttribute('checked')).to.be.true;
+  });
+
+  it('should dispatch "checked-changed" event', () => {
```
jorenbroekema: maybe add 'when toggled' to make the condition clear.

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,146 @@
...
+    el.disabled = true;
+    el.addEventListener('checked-changed', handlerSpy);
+    // when
+    el.click();
+    // then
+    expect(handlerSpy.called).to.be.false;
+  });
+
+  describe('a11y', () => {
+    describe('aria-pressed', () => {
```
jorenbroekema: It's pretty verbose. I like that it's thorough though. I wonder if it would be nicer in a single it
```js
// expect that checks the default
el.click()
// expect that checks attr
el.checked = false;
// expect that checks attr
el.setAttribute('checked', true);
// expect that checks attr
```

packages/switch/test/lion-input-switch.test.js
```diff
@@ -0,0 +1,72 @@
...
+    const el = await fixture(html`
+      <lion-input-switch></lion-input-switch>
+    `);
+    expect(el.slots).to.have.a.property('input');
+    expect(el).to.have.a.property('inputElement');
+    expect(el.slots.input()).to.be.an.instanceof(LionButton);
+    expect(el.inputElement).to.be.an.instanceof(LionButton);
+  });
+
+  it('should trickle "disabled" state to child button', async () => {
```
jorenbroekema: trickle sounds like trickle down which sounds like delegate, but it's more like a sync-up right? So it should sync its disabled state to its child button

packages/switch/test/lion-input-switch.test.js
```diff
@@ -0,0 +1,72 @@
...
+import { LionButton } from '@lion/button';
+
+import '../lion-input-switch.js';
+
+describe('lion-input-switch', () => {
+  it('should have default "input" element', async () => {
+    const el = await fixture(html`
+      <lion-input-switch></lion-input-switch>
+    `);
+    expect(el.slots).to.have.a.property('input');
```
jorenbroekema: These expects seem wrong to me :P but the test is good

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,98 @@
...
+    return html`
+      <div class="button-switch__track"></div>
+      <div class="button-switch__thumb"></div>
+    `;
+  }
+
+  constructor() {
+    super();
+    this.type = 'button';
+    this.disabled = false;
```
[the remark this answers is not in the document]
CubLion: You are right about "disabled". I see that it is declared and a value is assigned to it in "DisabledMixin". I think that "checked", since this is the first place where it appears, needs to be assigned a non-default (undefined) value.

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,98 @@
...
+import { html, css } from '@lion/core';
+
+export class LionButtonSwitch extends LionButton {
+  static get properties() {
+    return {
+      disabled: {
+        type: Boolean,
+        reflect: true,
+      },
+      checked: {
```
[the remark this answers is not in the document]
CubLion: The reason was to keep it consistent with other check exhibiting components (radio & checkbox). It also keeps the API consistent if we do indeed decide later on to go for a "extend checkbox" approach.

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('aria-checked', `${this.checked}`);
+  }
+
+  firstUpdated(changedProperties) {
+    super.firstUpdated(changedProperties);
+    this.removeAttribute('type');
```
jorenbroekema: Can't we do this in constructor?

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+      }),
+    );
+  }
+
+  /**
+   * We synchronously update aria-checked to support voice over on safari.
+   *
+   * @override
+   */
+  _requestUpdate(name, oldValue) {
```
jorenbroekema: This is a private lifecycle method of lit-element, it would be safer to use `updated()` here instead, and it doesn't really matter performance wise because we're only setting an attribute, so no re-render is necessary.

packages/switch/src/LionButtonSwitch.js
```diff
@@ -0,0 +1,94 @@
...
+    super._requestUpdate(name, oldValue);
+    if (this.isConnected && name === 'checked') {
```
jorenbroekema: if we use updated() then the `this.isConnected` guard is not necessary. Because at the updated callback you can be sure it is connected.

packages/switch/src/LionInputSwitch.js
```diff
@@ -0,0 +1,65 @@
...
+  get slots() {
+    return {
+      ...super.slots,
+      input: () => document.createElement('lion-button-switch'),
+    };
+  }
+
+  render() {
+    return html`
+      <div class="input-switch__container">
```
jorenbroekema: Liking the new css class! :)

packages/switch/src/LionInputSwitch.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  updated(changedProperties) {
+    super.updated(changedProperties);
+    this._syncButtonSwitch();
+  }
+
+  // TODO: should be replaced by "_inputNode" after the next breaking change
+  // https://github.com/ing-bank/lion/blob/master/packages/field/src/FormControlMixin.js#L78
+  get _buttonSwitchNode() {
```
jorenbroekema: This getter can be completely removed because you already have access to `this.inputElement` through the LionField class (which uses FormControlMixin). 

Then maybe add the TODO note to the place where you use `this.inputElement` (in the `__handleButtonSwitchCheckedChanged()` method maybe)

packages/switch/src/LionButtonSwitch.js
jorenbroekema: Also, this sets the type to null which is also not intended I think, this is an invalid value for type attribute.

I think just setting to type = button is fine right?

packages/switch/test/lion-button-switch.test.js
```diff
@@ -0,0 +1,114 @@
...
+    await el.updateComplete;
+    expect(el.checked).to.be.true;
+    expect(el.hasAttribute('checked')).to.be.true;
+    el.click();
+    await el.updateComplete;
+    expect(el.checked).to.be.false;
+    expect(el.hasAttribute('checked')).to.be.false;
+  });
+
+  it('could be disabled', async () => {
```
jorenbroekema: `can` present tense :)

PR #339, opened 2019-10-22
fix(lion-select-rich): keep showing selected item dynamic options bug

### LINKED ISSUE #265: [select-rich] Shows incorrect option as selected when new option is added
## Steps to reproduce

1. Create a select using `select-rich`
2. Select any option (i.e option with index=1)
2. Add new option at selectedItem index
3. Open & close select
4. Newly added option is auto selected

I have created a new story book, to reproduce it.

https://github.com/littletv/lion/tree/select-rich-issue

rodrigo-garcia-leon:
Fixes #265

### DIFF
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -360,8 +360,10 @@ export class LionSelectRich extends OverlayMixin(
     const foundChecked = this.modelValue.find(subModelValue => subModelValue.checked);
     if (foundChecked && foundChecked.value !== this.checkedValue) {
       this.checkedValue = foundChecked.value;
+    }
 
-      // sync to invoker
+    // sync to invoker
+    if (this._invokerNode) {
       this._invokerNode.selectedElement = this.formElements[this.checkedIndex];
     }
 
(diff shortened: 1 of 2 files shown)


jorenbroekema:
I kind of went down a bit of a rabbit hole on this one and discovered that in general `formElements`, which is a property of the FormRegistrarMixin, does not take into account the order of the form elements if you add them at different times. 

E.g. inserting an `option` into the `select` after initial connected callback will push that option to the end of the `formElements` array. :(  

## Long explanation:
I was looking at your PR and found the solution a little weird / inexplicable, at a conceptual level. The invoker's selectedElement should be in sync with the checkedIndex at all times. So I started by trying to fix it by moving this out of the if statement:
```js
// sync to invoker
if (this._invokerNode) {
  this._invokerNode.selectedElement = this.formElements[this.checkedIndex];
}
```
So it always syncs when the modelValue changes, whether the checked value has changed or not.
This made your test pass except for 1 expect --> the checkedIndex was now wrong, even though the value was correct. So something was wrong with the formElements array. The new item was pushed to the end of it.. So it is a deeper issue with the form registration system, not with the rich select itself.

I published a fix for the registration system issue here: https://github.com/ing-bank/lion/pull/339 . I believe that should fix your issue too. I'll have to make sure though, since this is a monorepo and the registration system is in a different package, it'll have to be released first before it's easy for me to confirm that your failing test in select-rich also passes.

Will get back to this soon! :)

rodrigo-garcia-leon:
@jorenbroekema ok perfect! is that's the case then we can just add the test of the MR then

daKmoR:
thaaaankkksss 🤗

PR #340, opened 2019-10-22
feat(core): export Part classes from lit-html

LarsDenBakker:
I need to use lit-html Parts for directives I'm writing internally, but they're not exported by lion.

We don't need this if we decide to go for https://github.com/ing-bank/lion/pull/341 instead


daKmoR:
thxx 🤗 

went with the named exports as rollup and vscode auto complete had problems with star exports and I'm not sure if they are resolved yet 🙈

PR #341, opened 2019-10-22
feat(core): use export star for lit-html and lit-element

LarsDenBakker:
A followup from https://github.com/ing-bank/lion/pull/340

I think we can just proxy all exports from `lit-html` and `lit-element`. Surprisingly the browser is able to dedupe the conflicting `html` exports from both packages because they eventually point to the same variable.

For some reason eslint doesn't like this setup when importing from `index.js` inside core, while it runs fine in the browser and for other packages importing from core it's fine as well.

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -4,17 +4,7 @@
  */
 
 // lit-html
-export {
-  html,
-  svg,
-  render,
-  noChange,
-  nothing,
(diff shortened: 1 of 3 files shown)


daKmoR:
resolved via https://github.com/ing-bank/lion/pull/340

PR #342, opened 2019-10-23
fix(form-system/stories): add label in demo umbrella form

NioniosSpeed:


jorenbroekema:
Hey @NioniosSpeed , nice one, could you change your commit to `chore(form-system)` instead, so it does not show up in the release notes? Because this only changes the demo, so it's not a bug with the actual package source code, but more a mistake in the demo :).

NioniosSpeed:
> Hey @NioniosSpeed , nice one, could you change your commit to `chore(form-system)` instead, so it does not show up in the release notes? Because this only changes the demo, so it's not a bug with the actual package source code, but more a mistake in the demo :).

NioniosSpeed:
sure, done!

PR #343, opened 2019-10-23
fix(overlays): remove setting inheritsReferenceWidth by default

jorenbroekema:


packages/overlays/src/OverlayController.js
```diff
@@ -498,8 +498,10 @@ export class OverlayController {
       case 'full':
         this._contentNodeWrapper.style.width = referenceWidth;
         break;
-      default:
+      case 'min':
         this._contentNodeWrapper.style.minWidth = referenceWidth;
+        break;
+      default:
```
daKmoR:
```suggestion
      /* no default */
```
will work for eslint 👍

jorenbroekema: Done!

Issue #168, opened 2019-07-15
[input-date] IE11 incorrectly reformats dates


erikkroes:
Use IE11 and enter a date like 15/06/17, with the year consisting of 2 digits. Tab out of the field. The information in the field becomes 06/05/1916, instead of the expected 15/06/2017.

- 12/12/01 becomes 12/01/1912. 
- 12/01/12 becomes 01/12/1912. 
- 01/12/12 becomes 12/12/1901.

bashmish:
We expect a full date. I don't think we ever supported or wanted to support such a shortcut, did we? cc @tlouisse 

The other thing is that it does not mark it as invalid, but tries to reformat and present some wrong date. Probably we should think about fixing that.

gerjanvangeest:
Input-date says dates with a year of 2 digits, are already not valid. So closing this issue.

PR #344, opened 2019-10-23
[WIP] fix(localize): Fixed the parse of dates like "D-M-Y" and allow only "D-M-YYYY"

paulcodiny:
The issue is that when I type "1-1-1" to a `date-input` it parses the date as "01-01-01", which is correct date.
Then I continue "1-1-19" and this date is incorrect now.
Then "1-1-199" which is incorrect.   
Then "1-1-1990" which is correct again.  
That makes the error message appear and disappear on every user input

### DIFF
--- a/packages/localize/src/date/parseDate.js
+++ b/packages/localize/src/date/parseDate.js
@@ -26,32 +26,40 @@ const memoizedGetDateFormatBasedOnLocale = memoize(getDateFormatBasedOnLocale, l
 export function parseDate(date) {
   const stringToParse = addLeadingZero(date);
   let parsedString;
+  let isFullYear = false;
   switch (memoizedGetDateFormatBasedOnLocale()) {
     case 'day-month-year':
-      parsedString = `${stringToParse.slice(6, 10)}/${stringToParse.slice(
-        3,
-        5,
-      )}/${stringToParse.slice(0, 2)}`;
+      isFullYear = stringToParse.length === 10;
+      parsedString = [
+        stringToParse.slice(6, 10),
+        stringToParse.slice(3, 5),
+        stringToParse.slice(0, 2),
+      ].join('/');
       break;
     case 'month-day-year':
-      parsedString = `${stringToParse.slice(6, 10)}/${stringToParse.slice(
-        0,
-        2,
-      )}/${stringToParse.slice(3, 5)}`;
+      isFullYear = stringToParse.length === 10;
+      parsedString = [
+        stringToParse.slice(6, 10),
+        stringToParse.slice(0, 2),
+        stringToParse.slice(3, 5),
+      ].join('/');
       break;
     case 'year-month-day':
-      parsedString = `${stringToParse.slice(0, 4)}/${stringToParse.slice(
-        5,
-        7,
-      )}/${stringToParse.slice(8, 10)}`;
+      isFullYear = /\d+/.test(stringToParse.slice(0, 4));
+      parsedString = [
+        stringToParse.slice(0, 4),
+        stringToParse.slice(5, 7),
+        stringToParse.slice(8, 10),
+      ].join('/');
       break;
     default:
       parsedString = '0000/00/00';
   }
+
   const parsedDate = new Date(parsedString);
   // Check if parsedDate is not `Invalid Date`
   // eslint-disable-next-line no-restricted-globals
-  if (!isNaN(parsedDate)) {
+  if (!isNaN(parsedDate) && isFullYear) {
     return parsedDate;
   }
   return undefined;


jorenbroekema:
So if I understand this PR correctly, parsedDate should always return undefined if the year is not given as a full year (4 digits)? 

So `01-01-19` would be invalid, even though using the user's locale, we should be able to make a good guess on what they intend to input. 

So back to your scenarios:
1-1-1 --> correct, yes I would expect that parseDate will do a "best guess" based on user locale
1-1-19 --> incorrect, I find this to be a bug, because `new Date('1-1-19')` --> year parsed as 2019, which is indeed what I would expect as a "best guess".
1-1-199 --> correct, when I try this, I get a valid date parse. So not sure why you are getting invalid one. I expect this, because now it's clear 199 can only mean 1 thing --> the year 199. 
1-1-1999 --> correct, same as above, clear that we mean the year 1999.

So I agree with you there is a bug for the case of providing a double digit year. It should be smart enough to guess that the year is probably in the current millenium. I disagree with the proposed solution (if I understand it correctly) though because it makes a lot of inputs invalid (everything that doesn't have YYYY, fullYear). 

The reason why 1-1-19 goes wrong is that our parseDate checks locale for the format, and then switches substrings around to normalize to year-month-day format. So 1-1-19 becomes 01-01-19 first, then becomes 19-01-01 (year-month-date format) and then it's put into the Date object, which returns Invalid Date on this. So because we normalize all inputted dates to year-month-date, for 01-01-19 you get an invalid date. 

I would propose that we can fix this by simply adding a check where, no matter the format, if the year is only 2 digits, we prepend `20` (current millenium). What do you think?

paulcodiny:
I like the idea to make `1-1-19` a valid date, but perhaps it's not that straightforward:
1. Adding `20` to the year
Because if it's a date of birth input, like in our case, users will input `1-1-88` or `1-1-90` more often, which will result in invalid date for them, because the year is in the future.
2. Adding `20` or `19` depending on the current year
In this case, the most challenging part is to guess whether it's `20` or `19`. If we compare the input with the current year (everything that below goes to the same century, above - to the previous), then expire passport dates will mess this logic (25 will go to 1925). If we add a gap for the future, then very old users will mess the logic (with 5 years gap when 20 will go to 2020 instead of 1920)
3. Adding `20` or `19` depending on the predefined set of values, when the year can be determined reliably
For instance, `60-99` must have `19` in the beginning, when the other input can not definitely determine. This will create a confusion for users that sometimes you can input `1-1-88` but not `1-1-25`

That's why it seems to me that by allowing only four digits for a year we get rid of all potential confusion for users and for potential mistakes.

What do you think?

jorenbroekema:
Hmm I see what you mean. Native behavior of Date() seems to agree with your third point, and I think we should mimic that to be predictable to our users.

Screenshot 2019-10-25 at 14 28 17

So, 00-49 --> 2000-2049, 50-99 --> 1950-1999.
With that, I'm happy for you to create tests and finalize this so we can be accepting to double digit years :)

paulcodiny:
No sure, that the native browser `Date` behavior is the best relevance point for this. Users don't really check the native behavior to feel comfortable (developers do though :)  
In the native Date the problem stays - a user inputs `14-10-45` as a birth date. Validator says, that the date cannot be bigger than the current which is a very vague error message for this case...  

Closer example, but still not the best, is native `input[type=date]` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date. If in the example you change `max` to `2025-01-01` and then try to enter `1` to year, it will stay `1`, without any guessing. 

The best examples would be the other services, but I did not manage to find any with date picker quickly.

jorenbroekema:
I think we both agree option 1 and 2 are worse than option 3 right?

So if 3 is better, and the only other alternative is only accepting 4-digit years, I would pick the first. Because making a guess is better than throwing an error message in my opinion. If the user sees it's not what they wanted, they can always edit.

gerjanvangeest:
Related to: https://github.com/ing-bank/lion/issues/168

CubLion:
We acknowledge the issue, which is essentially how to deal with multiple ways of interpreting user date input. We need to wait on some UX alignment on how to tackle this properly (feedback message, user controls to select the right interpretation, etc.).

Closing the PR for now since its unclear on how to move forward.

ruphin:
May I suggest the solution used by Material Design. They follow the guidelines outlined in this post: https://uxplanet.org/date-picker-design-best-practices-41bd522f10a

- Always show placeholders for the input
- Automatically fill in/remove separators

That way the user can keep typing digits and it is always clear what is expected.

You can try an example implementation here: https://material-ui-pickers.dev/demo/datepicker#keyboard-input

jorenbroekema:
I personally really like that, but it's a very opinionated UX force the user into a path. You can always extend the lion date input and add that behavior, that's not a problem!

However, as "white label" library we need to try to make the best out of what the user gives us when it's not double digits for months / days or quadruple digits for years. And for that I say we follow the platform as much as possible and just go with what Intl would do. Anything added on top I would say does not belong in lion, because there are too many equally strong approaches to give the user more validation/hints to help them along.

tlouisse:
I would be quite in favor of a guided approach as proposed by @ruphin, because this is also what the platform has by default (`<input type="date">` ). Comparing to `input type=date` seems more logical to me than comparing to `Intl` (end user vs developer interfaces).

The main difference between us and that material design example would then be that we don't show direct validation feedback (red border + message), but that's how our validation feedback is configured by default.
If we do that, then the material design example behaves the same as the platform (tested on Chrome).

PR #337, opened 2019-10-18
Adds the readOnly attribute to LionTextarea

ktnl97:

Issue #333, opened 2019-10-16
[LionTextarea] dose not support readonly attribute


lfusiara:
`LionTextarea `dose not support `readonly` attribute, After adding readonly attribute `textarea `is still editable.
Example:

`<lion-textarea readonly></lion-textarea>`

gerjanvangeest:
Take a look at [input](https://github.com/ing-bank/lion/blob/master/packages/input/src/LionInput.js)  what is needed to fix this.

ktnl97:
Hi @lfusiara, Can I pick this up?

lfusiara:
Thanks, PR seems ok

jorenbroekema:
Fixed here: https://github.com/ing-bank/lion/pull/348

PR #348, opened 2019-10-25
fix(textarea): delegate readonly to the native element

jorenbroekema:
Also a small fix for readOnly prop on LionField that didn't have reflect set to true.

@ktnl97 tagging you here, I co-authored you for the fix.

closes https://github.com/ing-bank/lion/pull/337
fixes https://github.com/ing-bank/lion/issues/333

### DIFF
--- a/packages/input/src/LionInput.js
+++ b/packages/input/src/LionInput.js
@@ -20,6 +20,7 @@ export class LionInput extends LionField {
       readOnly: {
         type: Boolean,
         attribute: 'readonly',
+        reflect: true,
       },
       type: {
         type: String,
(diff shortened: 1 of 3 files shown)


packages/input/src/LionInput.js
```diff
@@ -20,6 +20,7 @@ export class LionInput extends LionField {
       readOnly: {
         type: Boolean,
         attribute: 'readonly',
+        reflect: true,
```
daKmoR: should be a `fix(field)` right?

jorenbroekema: Hmm yeah I hesistated because it's not linked to any existing issue, but I suppose we can be quite sure this is a fix and needs a release. Amended it just now

PR #351, opened 2019-10-25
fix(field): add form elements in the correct order

jorenbroekema:
if you add a form elements after first render, your form element will always be pushed (appended) to the array of formElements. 

This PR makes sure we take care of the correct order, so that if you have

```
<parent>
  <child></child>
  <child></child>
</parent>
```

and you want to insert something after the first child, it should end up in the formElements array as the 2nd item as well (index 1). 

It may also deal with https://github.com/ing-bank/lion/pull/339 but should be confirmed

### DIFF
--- a/packages/field/src/FormRegistrarMixin.js
+++ b/packages/field/src/FormRegistrarMixin.js
@@ -63,12 +63,16 @@ export const FormRegistrarMixin = dedupeMixin(
         this.__hasBeenRendered = true;
       }
 
-      addFormElement(child) {
+      addFormElement(child, index) {
         // This is a way to let the child element (a lion-fieldset or lion-field) know, about its parent
         // eslint-disable-next-line no-param-reassign
         child.__parentFormGroup = this;
 
-        this.formElements.push(child);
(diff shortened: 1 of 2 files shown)


packages/field/src/FormRegistrarMixin.js
```diff
@@ -89,7 +93,14 @@ export const FormRegistrarMixin = dedupeMixin(
           return;
         }
         ev.stopPropagation();
-        this.addFormElement(child);
+
+        // Check for siblings to determine the right order to insert into formElements
+        // If there is no next sibling, index is -1
+        let indexToInsertAt = -1;
+        if (this.formElements && this.formElements instanceof Array) {
```
daKmoR:
```suggestion
        if (this.formElements && Array.isArray(this.formElements)) {
```

instanceof may fail for arrays coming from an iframe

jorenbroekema: Changed!

daKmoR: nice 🤗

PR #352, opened 2019-10-25
fix(switch): make it toggle on IE

CubLion:


packages/switch/test/lion-button-switch.test.js
```diff
@@ -13,7 +13,7 @@ describe('lion-button-switch', () => {
     `);
   });
 
-  it('should behave like a button', () => {
+  it.skip('should behave like a button', () => {
```
gerjanvangeest: I guess this one can be fully gone.

packages/switch/stories/index.stories.js
```diff
@@ -83,4 +84,10 @@ storiesOf('Forms|Switch', module)
     return html`
       <lion-switch-validation-demo></lion-switch-validation-demo>
     `;
-  });
+  })
```
gerjanvangeest: What is the benefit of only showing the button?

PR #353, opened 2019-10-25
fix(field): delegate autocomplete to inputElement for security

jorenbroekema:


packages/field/src/LionField.js
```diff
@@ -49,6 +49,10 @@ export class LionField extends FormControlMixin(
         type: String,
         reflect: true,
       },
+      autocomplete: {
```
daKmoR: shouldn't that be on the `lion-input` a field like select/checkbox does NOT have `autocomplete` right?

jorenbroekema:
```
The HTML autocomplete attribute is available on <input> elements that take a text or numeric value as input, <textarea> elements, <select> elements, and <form> elements.
```
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

daKmoR: fair enough 👍

PR #354, opened 2019-10-28
feat: added placeholder attribute to lion-input

SwyserDev:
feat: added placeholder attribute to lion-input

I added placeholder text to the lion-input. It is a feature that is missing, that even the native input provides.


daKmoR:
hey there 🤗 

to my knowledge placeholder are considered "bad" UX right? so I'm not sure if we want to enable users to do it?

what would be your use case?

jorenbroekema:
Hmm according to our UX'ers it's a bad practice. Not all design systems agree with that

For example https://material.io/components/text-fields/#usage . They use placeholders as labels interchangeably. Another example is bootstrap: https://getbootstrap.com/docs/4.0/components/forms/

SwyserDev:
I am busy creating a home automations app using the wonderful lion-web! I am a use case where I would like to show placeholder text instead of a label due to space restrictions.

packages/input/src/LionInput.js
```diff
@@ -43,6 +47,9 @@ export class LionInput extends LionField {
         } else if (this.hasAttribute('value')) {
           native.setAttribute('value', this.getAttribute('value'));
         }
+        if (this.placeholder) {
```
jorenbroekema: I didn't notice this earlier, but putting the logic in the slots getter means this only is synced when the SlotMixin is connected, which happens on connectedCallback of the host element. If you change it later, it is not synced.

Usually for syncing attributes to the native node, we do it in the `updated` callback, you can follow the example of `type` and `step` in LionInput for this. Perhaps good to add to your test that it should sync the placeholder attribute to the native node when changing it after initialisation of the component.

SwyserDev: @jorenbroekema,

Ohhh, that is so cool! Well spotted!
Learned something new today :)

Oh, I also fixed it...

daKmoR:
sounds good 👍 

commit message needs to be a `fix` as it would otherwise trigger a breaking release...

also we will need to actively disallow it on our extension layer then...

packages/input/test/lion-input.test.js
```diff
@@ -38,4 +38,10 @@ describe('<lion-input>', () => {
     expect(el.getAttribute('type')).to.equal('foo');
     expect(el.inputElement.getAttribute('type')).to.equal('foo');
   });
+
+  it('has an attribute that can be used to set the placeholder text of the input', async () => {
+    const el = await fixture(`<lion-input placeholder="text"></lion-input>`);
+    expect(el.getAttribute('placeholder')).to.equal('text');
+    expect(el.inputElement.getAttribute('placeholder')).to.equal('text');
```
jorenbroekema: I saw you resolved my earlier remark but I would like to suggest adding this:
```suggestion
    el.placeholder = 'foo';
    await el.updateComplete;
    expect(el.getAttribute('placeholder')).to.equal('foo');
    expect(el.inputElement.getAttribute('placeholder')).to.equal('foo');
```
To verify that it syncs to native input on-change. Similar to the test above this one.

PR #356, opened 2019-10-28
feat(localize): added MessageFormatLite

tlouisse:
Proposal for a lightweight MessageFormat that only does variable replacement and will be a performance gain for most translations (for all translations within the @lion package).

This code is the outcome of a poc I did, we will pick it up as a story later....
TODOs:
- tests
- decide on final api
- decide on dynamic/lazy load of MessageFormat


gerjanvangeest:
Closing the MR, it is planned for next big breaking change.

Issue #305, opened 2019-10-04
fieldset do not show feedback if the first validation fails


lfusiara:
Fieldset does not show validation message of its validator if first validation of child component fails

How to reproduce the error:
```
let fakeValidator = (...factoryParams) => [
  (...params) => ({'validator': isFakeValidator(...params)}),
  ...factoryParams
];

function isFakeValidator() {
  return false;
}

return html`
  <lion-fieldset
    id="test-fieldset1"
    name="test-fieldset1"
    .errorValidators=${[fakeValidator()]}
  >
    <lion-input-amount name="input1"></lion-input-amount>
  </lion-fieldset>
`;
```
If you click on input1 and enter ther for example letter 'b' validator of input-amount display error. Then fakeValidator will run but fieldset thase not show error feedback becaouse of roule: "touched && dirty && !prefilled". Input1 will be touched when you click outside component and blure event will be produced. In this case, the only chance to display the error for fieldset is to set the correct value in input1 and re-enter the error.

If at first you 'touch' input and then put the wrong value everything is fine

Is this a bug or a feature?

gerjanvangeest:
I don't fully follow your question or explanation, so correct me if i'm wrong.

1. For `fieldsets` we hide the validation messages for now: https://github.com/ing-bank/lion/blob/master/packages/fieldset/src/LionFieldset.js#L204
2. On the `input-amount` we already added a `isNumberValidator` that gets triggered as you described above.

I hope this answers your question, if so please close the issue.

lfusiara:
The `fakeValidator()` does not have hidden attribute so it should be presented. And it will be as described above but only when `input-amount` field will be at first focused then blured and after that you will put wrong number to `input-amout`

gerjanvangeest:
I can see the error you are talking about, by setting the same validator on a normal input it was easier to see what goes wrong.

```js
  let isFakeValidator = (...factoryParams) => [
    (...params) => ({ 'isFake' : isFake(...params) }),
     ...factoryParams
  ];

  function isFake() {
    return false;
  } 

  return html`
    <lion-fieldset
      label="fieldset"
      name="fieldset1"
      .errorValidators=${[isFakeValidator()]}
    >
      <lion-input 
        label="input"
        name="input1"
        .errorValidators=${[isFakeValidator()]}
      ></lion-input>
    </lion-fieldset>
  `;
```

[...]

PR #357, opened 2019-10-29
fix(fieldset): manage when to show error messages

jorenbroekema:
for fieldsets/forms/radiogroup/checkboxgroup

- show error messages only once you leave the group
- leave fieldset/form/checkboxgroup by click outside or tab on last element
- leave radiogroup by any change of on a child radio
- some cleanup of ObserverMixin in FormatMixin and ValidateMixin

closes: #305


packages/fieldset/src/LionFieldset.js
```diff
@@ -84,28 +93,46 @@ export class LionFieldset extends FormRegistrarMixin(
...
-    this.addEventListener('focused-changed', this._updateFocusedClass);
-    this.addEventListener('touched-changed', this._updateTouchedClass);
-    this.addEventListener('dirty-changed', this._updateDirtyClass);
+    this.addEventListener('dirty-changed', this._syncDirty);
+
     this._setRole();
+    document.addEventListener('click', this._checkForOutsideClick);
+  }
+
+  _checkForOutsideClick(event) {
```
CubLion: what if you have 2 fieldsets, then both will be touched?

packages/fieldset/src/LionFieldset.js
```diff
@@ -263,16 +317,20 @@ export class LionFieldset extends FormRegistrarMixin(
     }
   }
 
-  _updateFocusedClass() {
-    this.classList[this.touched ? 'add' : 'remove']('state-focused');
+  _syncFocused() {
+    this.focused = this._anyFormElementHas('focused');
```
CubLion:
```suggestion
    this.focused = true;
````

File: packages/fieldset/test/lion-fieldset.test.js [hunk 3]
```diff
@@ -342,66 +332,151 @@ describe('<lion-fieldset>', () => {
...
       );
-
-      fieldset.formElements.color.dirty = true;
-      await fieldset.updateComplete;
-      expect(fieldset.classList.contains('state-dirty')).to.equal(true, 'has class "state-dirty"');
     });
 
-    it('sets prefilled when field left and value non-empty', async () => {
-      const fieldset = await fixture(html`<${tag}>${inputSlots}</${tag}>`);
+    it(`becomes "touched" once the group as a whole becomes blurred via mouse interaction after
```
CubLion: improve this by testing multiple fields not just one.

packages/validate/src/ValidateMixin.js
```diff
@@ -211,30 +214,79 @@ export const ValidateMixin = dedupeMixin(
...
-          _onInfoStateChanged: ['infoState'],
-          _onSuccessStateChanged: ['successState'],
-        };
+          ].some(key => name === key)
+        ) {
+          this.validate();
+        }
+
+        // @deprecated adding css classes for backwards compatibility
+        this.constructor.validationTypes.forEach(type => {
```
CubLion: move the classes to async

packages/fieldset/src/LionFieldset.js
daKmoR: yeah won't work as the focusin event has `composed: true` 😭 
so we need this extra check here

packages/fieldset/test/lion-fieldset.test.js
daKmoR: done

packages/fieldset/src/LionFieldset.js
daKmoR: done

packages/validate/src/ValidateMixin.js
daKmoR: tried to move but fails a test in a strange way 😭 
gonna leave if for now like this... we can't save the whole world in one PR 😬

PR #358, opened 2019-10-29
WIP: Feat/tabs

CubLion:


packages/tabs/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+```sh
+npm i --save @lion/tabs;
+```
+
+### Usage
+
+```js
+import '@lion/tabs/lion-tabs.js';
+import '@lion/tabs/lion-tab.js';
```
daKmoR: there is no `lion-tab` web component?

packages/tabs/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+  See Note at <https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-19>
+
+  > It is recommended that tabs activate automatically when they receive focus as long as their
+  > associated tab panels are displayed without noticeable latency. This typically requires tab
+  > panel content to be preloaded.
+
+- **Panels are not focusable**
+
+  Focusable elements should have a mean to interact with them. Tab panels itself do not offer any interactive.
```
tlouisse: `means`, `themselves`, `interactiveness`?

packages/tabs/src/LionTabs.js
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  __setupStore() {
+    this.__store = [];
+    const buttons = this.querySelectorAll('[slot="tab"]');
+    const panels = this.querySelectorAll('[slot="panel"]');
+    if (buttons.length !== panels.length) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'The amount of tabs (',
```
daKmoR: not blocking at all... but code style wise wouldn't
```js
console.log(`The amount of tabs (${buttons.length}) does not match the amount of panels(${panels.length}).`);
```
be way more readable?

packages/tabs/src/LionTabs.js
```diff
@@ -0,0 +1,248 @@
...
+    element.setAttribute('id', `button-${uid}`);
+    element.setAttribute('role', 'tab');
+    element.setAttribute('aria-controls', `panel-${uid}`);
+    element.addEventListener('click', clickHandler);
+    element.addEventListener('keyup', keydownHandler);
+    this.__deselectButton(element);
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  __cleanButton(element, clickHandler, keydownHandler) {
```
daKmoR: also not blocking - I'm just curious

I assume you put these pure function in the class so you can override them at some point?
however `__foo` sort of means never override me... 

so if they are pure and truly private (e.g. extending classes should not be able to override them) then you could put them as actual separate functions [which would enforce it]

or you could make them "protected" `_foo` so it's ok for extending classes to modify them

packages/tabs/src/LionTabs.js
```diff
@@ -0,0 +1,248 @@
...
+    element.removeAttribute('selected');
+    element.setAttribute('aria-selected', false);
+    element.setAttribute('tabindex', -1);
+  }
+
+  set selectedIndex(value) {
+    const stale = this.__selectedIndex;
+    this.__selectedIndex = value;
+    this.__updateSelected();
+    // TODO: discuss before or after request update?!
```
daKmoR: doesn't matter in this case as for setter/getters the value is already updated by the time you fire the event (which is what we want)

`this.requestUpdate` is async so if anyone needs it to be done they need to await updateComplete

Issue #277, opened 2019-09-19
[checkbox-group] checkedValues API feature specs


jorenbroekema:
For radio-group, it is possible to get the property `.checkedValue` which will return a string: either empty if nothing is selected, or the string value of the checked radio element within the radio group.

We have discussed already that we want the same thing for checkbox-group, but the API will be slightly different because you can select multiple options:
`checkboxGroup.checkedValues` which returns an array of strings (or an empty array if nothing is selected).

We also want a setter for .checkedValues where you can do something like: `checkboxGroup.checkedValues = ['diplodocus', 't-rex']` where the strings need to correspond to actual values of existing checkboxes in the checkbox-group. 

This has been on our backlog for a bit but I'm opening this issue now to ask if someone feels up for making this feature. Needs to be tested of course as well :).

NioniosSpeed:
Hey @jorenbroekema can I pick this up? :-)

gerjanvangeest:
> Hey @jorenbroekema can I pick this up? :-)

Yes that would be great!!

jorenbroekema:
Hey Dio, just closing this issue as per our new way of handling feature requests. We will still of course review your pull request for this as soon as we are able!! :) sorry for taking long on it

jorenbroekema:
https://github.com/ing-bank/lion/pull/507 will tackle the need for getting / setting the array of values that are checked, through the modelValue directly.

PR #359, opened 2019-10-29
feat(checkbox-group): added checkedValues API

NioniosSpeed:
From issue: checkedValues API feature specs #277

WIP, since no tests are added.
Any comments wrt to the approach are welcome :-)


jorenbroekema:
Really liking the tests, it's what I was expecting in terms of API, so that's good! I'm only wondering why the `await nextFrame()` is necessary.

I would also add some tests about what happens if you set the checkedValues to non-existing choiceValues, and maybe test also with complex (array or object type) values like:
`.choiceValue=${{ key: 'value', anotherKey: 'anotherValue' }}` so that way it is a bit more complete.

My suggestion for if you try to set checkedValues to an array where an item is not existing as a choiceValue in any of the checkboxes, it should throw an error, in my opinion. And if that's the case, we should test for that error throw as well.

Do you want a review on the implementation code as well already?

NioniosSpeed:
Thanks for the comments, yes please share your comments for the implementation too!

jorenbroekema:
Hey Dio, just an update, from these changes and feedback we got from Lars, we have decided to make some changes to the modelValue / serializedValue of choice groups like the checkbox-group.

They change things quite a bit and it will mean that we will not need a checkedValue(s) anymore, because the modelValue of the group will reflect the checkedValue(s) already :) this simplifies our inputs quite a bit. Progress can be tracked on this https://github.com/ing-bank/lion/pull/507

Thanks for contributing though, this PR, among other things. triggered us to make these fundamental changes, so in that sense the work is definitely not in vain 👍

Issue #336, opened 2019-10-18
resetInteractionState  part of resetGroup doesn't work outside of form


RickdeGroot:
When the resetGroup funtion is called outside the form it clears the fields but doesnt reset the interactionstate(they remain dirty)

'Form Submit/Reset', () => {
    const submit = () => {
      const form = document.querySelector('#form');
      if (form.errorState === false) {
        console.log(form.serializeGroup());
        form.resetGroup();
      }
    };

by just adding  form.resetGroup(); to the Form storybook you can see the behavior,
if you click "Reset" it works fine. When you click "Submit" the fields are cleared as expected but the validations are triggered because the fields are still dirty.

the resetInteractionState call by ResetGroup should remove dirty/state-dirty

gerjanvangeest:
The issue is that the `submit` is still in progress while you `reset`, which causes the following:

- serialize
- reset
- submit (with empty values, so giving warnings)

Calling the `reset` in a separate function works.

Is the above example your actual use case, or just a way to demonstrate?

palash2601:
@gerjanvangeest We do have a scenario as below:

1. User fills data in form and Submits it
2. Filled data is displayed in a table on same page
3. User gets the empty form again to fill new data

So to have the above flow we need to reset the form on submit

gerjanvangeest:
Ok, i labelled it as a bug, so we can pick it up soon. If anybody has a good idea and time please share or make a pull request.

RickdeGroot:
Had a quick look, the issue is not the actual state-dirty, it's the state-error-show.

palash2601:
[This](https://github.com/ing-bank/lion/issues/336#issuecomment-544382113) input from @gerjanvangeest is correct

So to have `reset` wait for `serialize` and `submit` tasks to be completed `updateComplete` hook of LitElement can be used.

```js
'Form Submit/Reset', () => {
  const submit = async () => {
  const form = document.querySelector('#form');
  if (form.errorState === false) {
    console.log(form.serializeGroup());
    await form.updateComplete;
    form.resetGroup();
  }
};
```

Please try once.

Although it would be good if this is handled internally

RickdeGroot:
created pull request https://github.com/ing-bank/lion/pull/369 with the code proposed by @palash2601

jorenbroekema:
I need some help understanding Palash's comment.

As I understand:
```js
console.log(form.serializeGroup());
await form.updateComplete;
[...]

PR #369, opened 2019-11-04
fix: resetgroup

RickdeGroot:
Pull request with fix for issue https://github.com/ing-bank/lion/issues/336 as provided by @palash2601

> check of `updateComplete` can be also put up in resetGroup()
> 
> ```js
> async resetGroup() {
>     if (!this.submitted) {
>       await this.updateComplete;
>     }
>     this.formElementsArray.forEach(child => {
>       if (typeof child.resetGroup === 'function') {
>         child.resetGroup();
>       } else if (typeof child.reset === 'function') {
>         child.reset();
>       }
>     });
> 
>     this.resetInteractionState();
>   }
> ```

### DIFF
--- a/packages/fieldset/src/LionFieldset.js
+++ b/packages/fieldset/src/LionFieldset.js
@@ -223,9 +223,8 @@ export class LionFieldset extends FormRegistrarMixin(
   }
 
   async resetGroup() {
-    if (!this.submitted) {
-      await this.updateComplete;
-    }
+    // To ensure resetting happens after submission process, in case it was fired right before reset
+    await this.updateComplete;
     this.formElementsArray.forEach(child => {
       if (typeof child.resetGroup === 'function') {
         child.resetGroup();
--- a/packages/fieldset/src/LionFieldset.js
+++ b/packages/fieldset/src/LionFieldset.js
@@ -222,9 +222,7 @@ export class LionFieldset extends FormRegistrarMixin(
     return serializedValues;
   }
 
-  async resetGroup() {
-    // To ensure resetting happens after submission process, in case it was fired right before reset
-    await this.updateComplete;
+  resetGroup() {
     this.formElementsArray.forEach(child => {
       if (typeof child.resetGroup === 'function') {
         child.resetGroup();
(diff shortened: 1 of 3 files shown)


packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -646,6 +646,7 @@ describe('<lion-fieldset>', () => {
       expect(input.modelValue).to.equal('Bar');
 
       el.resetGroup();
+      await nextFrame();
```
jorenbroekema: Would `await el.updateComplete;` suffice? Is a bit cleaner / reliable.

packages/fieldset/src/LionFieldset.js
```diff
@@ -222,7 +222,10 @@ export class LionFieldset extends FormRegistrarMixin(
     return serializedValues;
   }
 
-  resetGroup() {
+  async resetGroup() {
+    if (!this.submitted) {
```
jorenbroekema: I'm not sure if this guard makes sense, or is even necessary. If no update is pending, `await this.updateComplete` should resolve immediately anyway.

So perhaps remove the guard and just add a comment: "To ensure resetting happens after submission process, in case it was fired right before reset"

packages/fieldset/test/lion-fieldset.test.js
RickdeGroot: decided to use an ` await el.resetGroup();` where possible, seemed even cleaner to me.

packages/fieldset/src/LionFieldset.js
RickdeGroot: Fixed

packages/fieldset/test/lion-fieldset.test.js
jorenbroekema: This definitely makes a lot of sense if you're testing submit followed by reset. 

I see a lot of awaits added to the tests that don't submit prior to the reset, and I'm now wondering why we need to await it in those tests? 

Just because it's now an async method and you can await it, doesn't mean you have to right? Do the tests fail without the await keyword? Because if they do, I would like to understand why they fail, because for most of the tests there should not be a lit update in queue when calling resetGroup.

jorenbroekema:
Okay I looked into this a bit further and I think we need to tackle this differently, at least do more investigation of the root cause.  Please discuss on the issue https://github.com/ing-bank/lion/issues/336

tlouisse:
I will close this, since this code could be breaking like Joren suggested and it also seems to have some specific knowledge of what happens in the particular `@submit` handler that was executed before the internal handler of LionForm. Please see: https://github.com/ing-bank/lion/pull/427

PR #371, opened 2019-11-06
feat: remove all deprecations from lion

jorenbroekema:
Almost done 👍 few more deprecations to go


CubLion:
Good job! A lot of patience went into this.

There still is a failing test:
Screenshot 2019-11-11 at 13 16 38

As far as you are concerned, is there anything else to be addressed in this repo?

jorenbroekema:
Yess that test failed because I accidentally removed a line of code that wasn't deprecated 🙈  

It's done now from my side, the only deprecations left are ValidationMixin related which will be addressed when the new Validate system gets released.

CubLion:
indeed. after pulling the test passed!

packages/localize/src/LocalizeManager.js
```diff
@@ -43,16 +43,21 @@ export class LocalizeManager extends LionSingleton {
...
-        Language only locales are deprecated, please use the full language locale e.g. 'en-GB' instead of 'en'.
-        See https://github.com/ing-bank/lion/issues/187 for more information.
-      `);
+      this.__handleLanguageOnly(value);
     }
 
     this._onLocaleChanged(value, oldLocale);
   }
 
+  // eslint-disable-next-line class-methods-use-this
```
jorenbroekema: Created its own method for handling this, so we can override it back to warning on our extension layer to prevent breaking change. Even so, it is private because you should not override this.

packages/choice-input/src/ChoiceInputMixin.js
```diff
@@ -217,23 +215,4 @@ export const ChoiceInputMixin = superclass =>
      * Synchronization from user input is already arranged in this Mixin.
      */
     _syncValueUpwards() {}
-
-    /**
-     * @deprecated use .checked
-     */
-    get choiceChecked() {
```
daKmoR: ah it will be so nice to work in a clean base 👍

packages/field/src/FormControlMixin.js
```diff
@@ -65,18 +64,31 @@ export const FormControlMixin = dedupeMixin(
...
-        return {
-          ...super.asyncObservers,
-          _onAriaLabelledbyChanged: ['_ariaLabelledby'],
-          _onAriaDescribedbyChanged: ['_ariaDescribedby'],
-          _onLabelChanged: ['label'],
-          _onHelpTextChanged: ['helpText'],
-        };
+      _requestUpdate(name, oldVal) {
+        super._requestUpdate(name, oldVal);
+      }
```
daKmoR: only super call so could be removed?

packages/localize/src/LocalizeManager.js
```diff
@@ -43,16 +43,21 @@ export class LocalizeManager extends LionSingleton {
...
+  __handleLanguageOnly(value) {
+    console.error(`
```
daKmoR: it should throw as well right?

packages/localize/test/LocalizeManager.test.js
```diff
@@ -303,23 +303,23 @@ describe('LocalizeManager', () => {
         throw new Error('did not throw');
       });
 
-      it('throws a warning if the locale set by the user is not a full language locale', async () => {
-        const spy = sinon.spy(console, 'warn');
+      it('throws an error if the locale set by the user is not a full language locale', async () => {
+        const spy = sinon.spy(console, 'error');
```
daKmoR: throw is not the same as console.error... let's throw and test throwing here with

```
expect(() => manager.locale = 'nl').to.throw('You can not...');
```

no need to "tinker" with console 👍

jorenbroekema: Aaaaaaah im not actually throwing, thats why it didnt work when I was trying the approach you're proposing. Makes sense now... will change ^^

daKmoR:
> ...test/lion-input-iban-integrations.test.js → ...a/test/lion-textarea-integrations.test.js

file moved?

daKmoR:
aahh I love this MR 😍 
it will be sooo nice to work on a clean code base 💪

jorenbroekema:
Test file `packages/textarea/test/lion-input-iban-integrations.test.js` was renamed to `packages/textarea/test/lion-textarea-integrations.test.js` ;) hope that makes sense. Just a small cleanup

jorenbroekema:
I resolved your comments I'm pretty sure. Also rebased and squashed and made it a `feat`.

💪

PR #375, opened 2019-11-10
Fix input date picker

ovidiu1:
I think is a confusion between _**selectedDate**_ and  **_centralDate_**  from LionCalendar: 
image

In order to show the current selected date we have to use the _**this._calendarElement.focusSelectedDate()**_  in the _**_onCalendarOverlayOpened()**_ method  to be able to use _**selectedDate**_ property.

Tests: 
  -  i remove  **_'focuses interactable date on opening of calendar'_** because the **_selectedDate_** 
 property don't use focus on dayDate;

-  i'm not sure if we need any more **_'closes the calendar overlay on "user-selected-date-changed'_**

Working well on local.(Chrome & IE11)

Note: **_Also fix ' New Issue: Even if you change the day/month/year in the input is not reflected in the pop-up.'_**

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -284,6 +284,7 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
   _onCalendarOverlayOpened() {
     if (this._focusCentralDateOnCalendarOpen) {
       this._calendarElement.focusCentralDate();
+      this._calendarElement.focusDate(this.modelValue);
     }
   }
 
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -283,7 +283,6 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
    */
   _onCalendarOverlayOpened() {
     if (this._focusCentralDateOnCalendarOpen) {
-      // this._calendarElement.focusCentralDate();
       this._calendarElement.focusSelectedDate();
     }
   }
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -283,8 +283,8 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
    */
   _onCalendarOverlayOpened() {
     if (this._focusCentralDateOnCalendarOpen) {
-      this._calendarElement.focusCentralDate();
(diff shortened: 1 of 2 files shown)


jorenbroekema:
Hmm, I think the point of CentralDate is that this is the date set by the developer of the application/feature. It's the "central date" that he expects to be a friendly offset for the user to start selecting a date. Usually the central date is "today" for most use cases. 

It could be that a date is (pre-)selected during closed state. Then if you open it, question is whether the calendar should open with the selected date in focus, or still the central date which was set by the developer. 

So in that sense, I think this comes down to "what is better UX?". Perhaps this means we should challenge UX or at least get a rationale and drop that in our readme so it's clear why we opt for focusing the central date on every open.

I see three possible solutions.

1) App developer extends the datepicker and overrides:
```js
_onCalendarOverlayOpened() {
    if (this._focusCentralDateOnCalendarOpen) {
      this._calendarElement.focusSelectedDate(); // <-- override
    }
}
```

2) We provide a friendly API on the datepicker component, e.g. an enum property/attribute: "focus-on-open"/".focusOnOpen" --> can either be "selected" or "central" (keeping the default to what UX feels is best)
```js
_onCalendarOverlayOpened() {
    if (this._focusCentralDateOnCalendarOpen) {
      if (this.focusOnOpen === 'selected') {
        this._calendarElement.focusSelectedDate();
      } else {
        this._calendarElement.focusCentralDate();
      }
    }
}
```

3) We change the default behavior as proposed in this PR, where it focuses the selected date if there is one, if not, focus the central date. From my personal viewpoint I think this makes the most sense, but I don't know what the UX considerations are.

ovidiu1:
Good points, I think also the 3'th point is the best choice for this case(how is now in the storybook).

Also agree this is come to  "what is better UX?", let's wait for a proper answer from the UX and after than we can fix this issues.

tlouisse:
Hi @ovidiu1, does UX already provided you with an answer?

ovidiu1:
@tlouisse “I don’t know anyone from the UX.

tlouisse:
> Hi @tlouisse “ No.
> I don’t know anyone from the UX. Can you tag somebody from the UX here?

I sent a mail to our UX and cc-ed you ;)

ovidiu1:
Any update on this? :-)
I don't receive any email.

gerjanvangeest:
> Any update on this? :-) I don't receive any email.

They approved your suggestion of "focussing the `selectedDate` on reopen".

gerjanvangeest:
fixed in https://github.com/ing-bank/lion/pull/475

PR #376, opened 2019-11-12
fix: added placeholder attribute for input

SwyserDev:
Added the option to provide the input component with placeholder text

Previous PR before branch name change:
https://github.com/ing-bank/lion/pull/354


jorenbroekema:
Thanks 🎉 !

PR #378, opened 2019-11-12
Add a11y addon to storybook and fix all violations

jorenbroekema:
Add a11y knob to storybook and resolve all a11y issues 🎉  Some of them are actual a11y bugfixes

I have removed the <lion-option> demo's, because if they are not in a parent that sets the right aria stuff, the whole thing is just not accessible due to `role="option"` missing the expected parent aria elements.
I figured, lion-option by itself is not a sensible component. It should be reused at some point in a listbox component for example, so it's fine to have it as a package. But we should not demo it standalone.


erikkroes:
Fixes #289

PR #386, opened 2019-11-18
fix(datepicker): close with click

erikkroes:
Makes the datepicker closable by clicking outside of it


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -265,6 +265,7 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
       contentNode,
       invokerNode,
       elementToFocusAfterHide: invokerNode,
+      hidesOnOutsideClick: true,
```
gerjanvangeest: I believe this should be done on the `modalDialog` directly, it applies to all overlays that can be closed on ESC.

daKmoR:
UX agrees for datepicker but not for all - so let's merge this 👍

PR #393, opened 2019-11-22
fix(select-rich): keyboard navigation should handle scrolling

daKmoR:
Old:
old

New:
Kapture 2019-11-22 at 16 29 38

unfortunately, I could not add a test for it as apparently the scrolling is taking quite some time 😭

### DIFF
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -19,6 +19,26 @@ function detectInteractionMode() {
   return 'windows/linux';
 }
 
+function isInView(container, element, partial = false) {
+  const cTop = container.scrollTop;
+  const cBottom = cTop + container.clientHeight;
+  const eTop = element.offsetTop;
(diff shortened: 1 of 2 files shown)


packages/select-rich/src/LionSelectRich.js
```diff
@@ -129,9 +149,18 @@ export class LionSelectRich extends OverlayMixin(
...
+  }
+
   set activeIndex(index) {
     if (this.formElements[index]) {
-      this.formElements[index].active = true;
+      const el = this.formElements[index];
+      el.active = true;
+
+      if (!isInView(this.scrollTarget, el)) {
+        el.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
```
CubLion: The options have limited support:

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

daKmoR: yeah it's not as smooth but it gives the same end result 👍

PR #394, opened 2019-11-24
feat(lion-tabs): added event details with the selected tab index

Ronrock:
The selected-changed event now also has event details indicating to which tab the selection has
changed to. The event details now contains the selected-tab-index value. In some cases the content of a tab might be interested in which tab context it is loaded.

### DIFF
--- a/packages/tabs/src/LionTabs.js
+++ b/packages/tabs/src/LionTabs.js
@@ -200,7 +200,15 @@ export class LionTabs extends LitElement {
     const stale = this.__selectedIndex;
     this.__selectedIndex = value;
     this.__updateSelected();
-    this.dispatchEvent(new Event('selected-changed'));
+    this.dispatchEvent(
+      new CustomEvent('selected-changed', {
+        bubbles: true,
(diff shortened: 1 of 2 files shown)


daKmoR:
how about doing it like so?
```js
myTabs.addEventListener('selected-changed', ev => {
  console.log(ev.target.selectedIndex);
});
```

or what would be your use case for it?

Ronrock:
A use case could be where you have a component outside of the <lion-tabs> component that does need the context of the currently selected tab. event.target.selectedIndex would work, however adding it to the event details is a bit of a nicer pattern in my option.  Providing the id in the selected event details is something I think users would expect? In any case both are fine and workable :)

daKmoR:
hey so if using the event target is a workable solution then we prefer it - we do not want to put redundant information into the event especially as they can get out of sync and also often you need more information then one property and then we would need to add more and more info to the event...

so unless there is a use case that would require it let's not do it 🤗 

if there are any more question feel free to reopen or create an issue where we can discuss further 🤗

PR #397, opened 2019-11-26
fix(input-amount): a11y currency

tlouisse:


packages/localize/test/number/formatNumberToParts.test.js
```diff
@@ -72,6 +72,44 @@ describe('formatNumberToParts', () => {
     });
   });
 
+  // describe.only("style: 'currency name'", () => {
```
tlouisse: will remove again.... :)

PR #398, opened 2019-11-26
fix(input-amount): handle user pasting of amounts heuristically

daKmoR:
we want to handle user pasting special however our field reformats on input AND change event... 

so it respects it when you paste but it overrides it when you tab out (as change event is also triggered on lost of focus)

=> we probably should not use input AND change event?

### DIFF
--- a/packages/input-amount/src/LionInputAmount.js
+++ b/packages/input-amount/src/LionInputAmount.js
@@ -46,8 +46,26 @@ export class LionInputAmount extends FieldCustomMixin(LocalizeMixin(LionInput)) 
     super();
     this.parser = parseAmount;
     this.formatter = formatAmount;
+    this.__isPasting = false;
+
+    this.addEventListener('paste', () => {
+      this.__isPasting = true;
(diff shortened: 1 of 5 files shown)


packages/input-amount/src/LionInputAmount.js
```diff
@@ -46,10 +46,28 @@ export class LionInputAmount extends FieldCustomMixin(LocalizeMixin(LionInput))
...
+    this.addEventListener('paste', () => {
+      this.__isPasting = true;
+      this.__parserCallcountSincePaste = 0;
+    });
 
     this.defaultValidators.push(new IsNumber());
   }
 
+  __callParser(value = this.formattedValue) {
+    // TODO: input and change events both trigger parsing therefore we need to handle the second parse
```
tlouisse: Not sure if this explanation would be clear for everyone. You might consider the elaborate version here, but I'm fine with both.
Wrt the 'TODO:' it doesn't really explain what still needs to be done... :)

PR #403, opened 2019-11-27
fix: make success message disappear

erikkroes:


packages/validate/test/lion-validation-feedback.test.js
```diff
@@ -30,4 +31,28 @@ describe('lion-validation-feedback', () => {
     await el.updateComplete;
     expect(el.getAttribute('type')).to.equal('warning');
   });
+
+  it.only('success message clears after 3s', async () => {
```
gerjanvangeest: `only`

gerjanvangeest: Also test the `clearTimeout`?

erikkroes: Good one, fixed both.

PR #409, opened 2019-11-28
feat(tooltip): add arrow component

CubLion:


packages/tooltip/stories/tail.stories.js
```diff
@@ -0,0 +1,133 @@
...
+        height: 0;
+
+        border-color: red;
+        border-style: solid;
+      }
+
+      [x-placement^='top'] .arrow {
+        border-width: ${bw}px ${bw}px 0 ${bw}px;
+        border-color: ${bc} transparent transparent transparent;
+        bottom: -${bw}px;
```
jorenbroekema: I wonder if we should do this here, or simply compensate for the arrow height/width in the popperConfig offset modifier. Probably the latter if you ask me, but it's an implementation detail so don't worry about it too much for now :).

jorenbroekema:
Looks great! Love the simplicity of the POC, yet it shows a lot.

It may be worth checking if it's easy to enforce that the arrow is always x amount of pixels from the edge of the popper element (content element) since I believe that is still a requirement for the context menu component. I remember POC'ing that before and it turned out pretty easy back then. But since this is purely for tooltip, maybe we should leave it out of the scope :) especially since I am already confident it's possible.

I also want to discuss whether we want to provide the arrow ourselves "hardcoded" (we provide our own node but make it easy to override stying) or allow for an arrow slot in OverlayController/Mixin that you can provide your own arrow node.

packages/tooltip/stories/tail.stories.js
jorenbroekema: ^ was wrong about this 🙈

packages/tooltip/test/lion-tooltip.test.js
```diff
@@ -103,6 +102,40 @@ describe('lion-tooltip', () => {
     });
   });
 
+  describe('Positioning', () => {
+    it('updates popper positioning correctly, without overriding other modifiers', async () => {
```
jorenbroekema: Here I solved headless vs actual browser inconsistency by using a div instead of button for invoker, and also using position: absolute and then offsetting it with top, to get it to pass in both browser and headless.

Might be possible to re-enable some of the skipped test with this.. but didn't have time for it.

Issue #368, opened 2019-11-04
[tooltip] Tooltip should be dismissable on Hover


erikkroes:
## Description

A tooltip can easily get in the way to users who zoom in a lot. Being able to dismiss the tooltip without moving the mouse can aid in this.

WCAG 1.4.13 states: 
> A mechanism is available to dismiss the additional content without moving pointer hover or keyboard focus.
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html

The method to satisfy this criterion is also listed:
> Provide a mechanism to easily dismiss the additional content, such as by pressing Escape. 

## To Reproduce

1. Open the demo of the button tooltip
2. Hover over the button with your mouse
3. A tooltip appears
4. (Expected but missing behavior) Pressing escape dismisses the tooltip

daKmoR:
did you try in a separate preview window (e.g. without the storybook UI around)?

as Esc generally does not seem to work there 🙈

erikkroes:
Good point, but it still doesn't work I'm afraid.
https://lion-web-components.netlify.com/iframe.html?id=local-overlay-system-tooltip--button-tooltip

malhar12:
@erikkroes @daKmoR I'd like to work on this issue if you guys are still accepting PRs for this issue.

daKmoR:
I think that should be already fixed 🤔

can someone verify? 🙏

gerjanvangeest:
It is not fixed for mouseover, ESC works for keyboard navigation. 

@malhar12 if you have time, please help us fixing it.

New URL: https://lion-web-components.netlify.com/?path=/story/overlays-specific-wc-tooltip--test

daKmoR:
as hovering does not change the element selected it will result in something like this

#### how it works for keyboard:
- open page
- \> body selected
- tab, tab, tab, ...
- \> tooltip invoker selected
- tooltip appears
- hit esc
- \> event fired on invoker (as currently selected)

#### how it works with mouse
- open page
- \> body selected
- use mouse to hover over tooltip
- tooltip appears
- hit esc
- \> event fired on body (as currently selected)

## Possible solution

- add Event (capture phase) to body
- check if I (the tooltip) am open => close

daKmoR:
issue is being worked on 💪

daKmoR:
solved via https://github.com/ing-bank/lion/pull/542

PR #413, opened 2019-11-28
fix(#368): add accessible behaviors for tooltip

fcarrascosa:


gerjanvangeest:
hey thank you for picking this up.

We fixed it here https://github.com/ing-bank/lion/pull/388

and you can test it here
https://lion-web-components.netlify.com/iframe.html?id=overlays-specific-wc-tooltip--button-tooltip

fcarrascosa:
Hi, @gerjanvangeest I checked the behavior you provide and it seems to work only if the element is focused... Otherwise, the tooltip won't disappear. Is this an expected behavior?

PR #426, opened 2019-12-05
Fix bug in serializeGroup and serializeElements

danychi:
0 (number) is considered as falsy by JavaScript. That means the previous check was ignoring the 0 when defining the `serializedValues`  in `serializeGroup`and `__serializeElements` methods.


tlouisse:
Hi, looks good :)

Few nitpicks about the commits. Could you:
- use `chore:` or `chore(fieldset):` instead of `test:` ?
- squash the two latest commits into one?
- use `fix(fieldset):` in this squashed commit

danychi:
@tlouisse I updated the PR, let me know if I need to adjust something else.

PR #436, opened 2019-12-09
feat(overlay): close/hide events on dom (OverlayMixin) level

tlouisse:
@LarsDenBakker requested a way to synchronize and intercept open/close state in an api similar to:

```javascript
render() {
  return html`
    <lion-dialog .opened=${this.opened} @close="${() => this.opened = false;}" @before-close=${(e) => {
       if (shouldIntercept) {
          e.preventDefault();
       } 
     }}></ing-dialog>
  `;
}
```

### DIFF
--- a/packages/overlays/src/OverlayMixin.js
+++ b/packages/overlays/src/OverlayMixin.js
@@ -36,6 +36,13 @@ export const OverlayMixin = dedupeMixin(
         this.__config = value;
       }
 
+      _requestUpdate(name, oldValue) {
+        super._requestUpdate(name, oldValue);
+        if (name === 'opened') {
+          this.dispatchEvent(new Event('opened-changed'));
+        }
+      }
+
       /**
        * @overridable method `_defineOverlay`
        * @desc returns an instance of a (dynamic) overlay controller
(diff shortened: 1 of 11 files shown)


packages/overlays/src/OverlayMixin.js
```diff
@@ -150,23 +169,41 @@ export const OverlayMixin = dedupeMixin(
       __setupSyncFromOverlayController() {
         this.__onOverlayCtrlShow = () => {
           this.opened = true;
+          this.dispatchEvent(new Event('opened'));
```
daKmoR: maybe move it to `_requestUpdate`?
ala
```js
_requestUpdate(name, oldValue) {
  super._requestUpdate(name, oldValue);
  if (name === 'opened) {
    this.dispatchEvent(new Event('opened-changed'));
  }
}
```

OR
use `this.dispatchEvent(new Event('opened-changed'));` in `__onOverlayCtrlShow ` and `__onOverlayCtrlHide`

daKmoR:
code looks good 👍 (only change to opened-changed event name)

cleanup history and then it should be good to go 👍

Issue #382, opened 2019-11-15
[OverlayMixin] Shadow outlet slot wrapper breaks content slot


jorenbroekema:
## Describe the bug

The problem is https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayMixin.js#L120

It adds a wrapper in between the host element and the content node. The content node happens to be a slot (slot="content") of the host element, and should always be a direct child. Due to us adding a wrapper in between, the content slot is no longer a direct child of its host element, and slots are not spec'd to be that way.

DOM:
<img width="405" alt="Screenshot 2019-11-15 at 14 14 04" src="https://user-images.githubusercontent.com/36734656/68946664-d0b0a780-07b3-11ea-95ca-d66920e0b93b.png">

https://www.w3.org/TR/dom41/#finding-slots-and-slotables  --> see second point "Let shadow be slotable’s parent’s shadow root.". If the slotable's parent is not the host element, it won't have a shadow root nor a shadow.

https://github.com/ing-bank/lion/pull/383/ I introduced some FIXME's related to using querySelector for slots instead of only checking direct children in this PR so please resolve those once this is fixed :).

jorenbroekema:
Soooo.. 

Basically, the slottable DOM element gets moved to the global root node when it's a global overlay. This means that the slot no longer works, for example you could try to style the slottable with the :slotted CSS selector from the ShadowDOM, but since the slottable element is now gone from the host element, it wouldn't work.

The problem is not the outlet. The problem I think is that we still have the slot attribute on the outlet and the content after it's moved away from the host scope. So we should be transparent here by removing the slot attribute --> it doesn't work anyway 🤷‍♂ and that's normal because we're moving LightDOM (with slottables in it) around, which we can't really avoid for global overlays to work properly.

@tlouisse  Maybe you can give your 2 cents here, I'm tempted to actually remove the bug label because it's not a "bug", but we should clarify / clean it up.

tlouisse:
Yeah indeed, the element with **[slot="content"]**  will be moved into **[slot="_overlay-shadow-outlet"]** and therefore the shadowRoot of your host element will not regard **[slot="content"]** as a slot anymore after moving.

Functionally this is not a problem, unless you want to style the element **[slot="content"]** from your host, which I agree we should support...

So it should be possible to put `._contentNodeWrapper`/ **[slot="_overlay-shadow-outlet"]** in the shadow root.

[...]

PR #438, opened 2019-12-11
feat(tooltip): arrow

CubLion:


packages/tooltip/test/lion-tooltip-arrow.test.js
```diff
@@ -1,87 +1,71 @@
 import { aTimeout, expect, fixture, html } from '@open-wc/testing';
+import '../../button/lion-button.js';
```
gerjanvangeest: Do we really need this dependency, or would a normal button be enough?

tlouisse: this is because it's a div under the hood, because a native button has different dimensions on different browsers

packages/tooltip/src/LionTooltip.js
```diff
@@ -36,63 +18,84 @@ export class LionTooltip extends OverlayMixin(LitElement) {
...
+  }
 
-    this.__showKey = () => {
-      if (!this.mouseActive) {
-        this.keyActive = true;
-        this.opened = true;
-      }
-    };
+  __setupPopperStateSyncPromise() {
+    this.syncFromPopperStateComplete = new Promise((resolve) => {
```
tlouisse: maybe leave out 'Popper' and call `repositionComplete` ?

packages/tooltip/src/LionTooltip.js
```diff
@@ -36,63 +18,84 @@ export class LionTooltip extends OverlayMixin(LitElement) {
...
     `;
   }
 
-  _setupOpenCloseListeners() {
-    this.__resetActive = () => {
-      this.mouseActive = false;
-      this.keyActive = false;
-    };
+  __setupArrowElement() {
+    this.__arrowElement = this.querySelector('[slot~=arrow');
```
tlouisse: I think we should only allow `[slot=arrow]`

packages/tooltip/src/LionTooltip.js
```diff
@@ -103,4 +106,35 @@ export class LionTooltip extends OverlayMixin(LitElement) {
     this._overlayInvokerNode.removeEventListener('focusin', this._showKey);
     this._overlayInvokerNode.removeEventListener('focusout', this._hideKey);
   }
+
+  __resetActive() {
+    this.mouseActive = false;
```
tlouisse: maybe all those internal states should be private vars?

tlouisse:
Looks good, few small remarks. Maybe we could later still consider https://github.com/ing-bank/lion/issues/382#issuecomment-564930695

packages/tooltip/src/LionTooltipArrow.js
```diff
@@ -0,0 +1,46 @@
+import { LitElement, html, css, UpdateStylesMixin } from '@lion/core';
+
+export class LionTooltipArrow extends UpdateStylesMixin(LitElement) {
```
CubLion: Does not need `UpdateStylesMixin`

packages/overlays/test/local-positioning.test.js
```diff
@@ -214,8 +214,8 @@ describe('Local Positioning', () => {
       );
     });
 
-    // TODO: dom get's removed when hidden so no dom node to update placement
-    it('updates placement properly even during hidden state', async () => {
+    // TODO: Refactor to the new system, last expect is broken at the moment (the others pass still)
+    it.skip('updates placement properly even during hidden state', async () => {
```
CubLion: why is this skipped.

packages/overlays/test/local-positioning.test.js
```diff
@@ -263,7 +265,8 @@ describe('Local Positioning', () => {
       );
     });
 
-    it('updates positioning correctly during shown state when config gets updated', async () => {
+    // TODO: Refactor to the new system, last expect is broken at the moment (the others pass still)
+    it.skip('updates positioning correctly during shown state when config gets updated', async () => {
```
CubLion: why is this skipped?

PR #445, opened 2019-12-18
Convert calendar to new storybook

jorenbroekema:


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "scripts": {
     "start": "npm run storybook",
-    "storybook": "start-storybook -p 9001 --stories \"./packages/*/stories/*.stories.{js,mdx}\" --node-resolve --watch --open",
+    "storybook": "start-storybook -p 9001 --stories \"./packages/calendar/stories/*.stories.{js,mdx}\" --node-resolve --watch --open",
```
CubLion: Should we merge with `calendar` being hardcoded or should `*` be put back?

packages/calendar/stories/index.stories.mdx
```diff
@@ -0,0 +1,257 @@
...
+      <lion-calendar class="demo-calendar" .minDate="${minDate}"></lion-calendar>
+    `;
+  }}
+</Story>
+
+```html
+<lion-calendar .minDate="${new Date()}"></lion-calendar>
+```
+
+### Provide a higher limit
```
CubLion:
```suggestion
### Provide a upper limit
```

packages/calendar/stories/index.stories.mdx
```diff
@@ -0,0 +1,257 @@
...
+To give a lower limit you can bind a date to the `minDate` property.
+
+<Story name="minDate">
+  {() => {
+    const minDate = new Date();
+    return html`
+      <style>
+        ${calendarDemoStyle}
+      </style>
+      <lion-calendar class="demo-calendar" .minDate="${minDate}"></lion-calendar>
```
CubLion: Do you think it would make sense to specify if `minDate` or `maxDate` specify the bounds of a closed or open interval? 
Given a `date X` and setting `minDate to Y` would dates be disabled up to `Y - 1` or up to `Y`?
Given a `date X` and setting `maxDate to Y` would dates be disable starting from `Y + 1` or from `Y`?

packages/calendar/stories/index.stories.mdx
```diff
@@ -0,0 +1,257 @@
...
+In some cases a specific date or day of the week needs to be disabled, supply those days to the `disableDates` property.
+
+<Story name="disableDates">
+  {html`
+    <style>
+      ${calendarDemoStyle}
+    </style>
+    <lion-calendar
+      class="demo-calendar"
+      .disableDates=${day => day.getDay() === 6 || day.getDay() === 0}
```
CubLion:
```suggestion
      .disableDates=${cellDate=> cellDate.getDay() === 6 || cellDate.getDay() === 0}
```

CubLion:
After addressing https://github.com/ing-bank/lion/pull/445/files#r360287087 LGTM.

package.json
jorenbroekema: We're merging these to next/storybook. Once everything is there, we can set it to * again. If we do it before, storybook doesn't work because of combining .js with old format, with .mdx story files. So for reviewing purposes I left it at calendar so you can just run npm start and review the storybook ;)

PR #446, opened 2019-12-18
chore(ajax): convert demos to new storybook

jorenbroekema:


packages/ajax/stories/index.stories.mdx
```diff
@@ -0,0 +1,164 @@
...
+    console.log(response.data);
+  })
+  .catch(error => {
+    console.log(error.message);
+  });
+```
+
+## Considerations
+
+Due to a [bug in axios](https://github.com/axios/axios/issues/385) options may leak in to other instances.
```
CubLion: Links are rendered weird:

```html
<a href="http://localhost:9002https://github.com/axios/axios/issues/385" class="sbdocs sbdocs-a css-1cvpre6">bug in axios</a>
```

jorenbroekema: Odd.. will need to look into how to prevent that in storybook .mdx files

jorenbroekema: https://github.com/storybookjs/storybook/issues/9005 I think this is related

CubLion:
After https://github.com/ing-bank/lion/pull/446/files#r360317791 LGTM

PR #450, opened 2019-12-23
fix(select-rich): add aria-invalid when invalid

gerjanvangeest:


packages/select-rich/test/lion-select-rich-interaction.test.js
```diff
@@ -667,5 +667,31 @@ describe('lion-select-rich interactions', () => {
...
+          </lion-options>
+        </lion-select-rich>
+      `);
+      const invokerNode = el._invokerNode;
+      const options = el.querySelectorAll('lion-option');
+      await el.feedbackComplete;
+      await el.updateComplete;
+      expect(invokerNode.getAttribute('aria-invalid')).to.equal('false');
+
+      options[0].checked = true;
```
jorenbroekema: Shouldn't this be false? I would expect this test to fail now

gerjanvangeest: Options[0] has a choiceValue of `null`, which means that nothing is selected.

PR #452, opened 2019-12-23
chore(dialog): convert demos to new storybook

jorenbroekema:


packages/overlays/docs/OverlayController.md
```diff
@@ -30,9 +30,9 @@ In total, we should end up with configuration options as depicted below, for all
 All boolean flags default to 'false'.
```
gerjanvangeest: All documents should be turned into mdx files.

PR #459, opened 2019-12-30
chore(input-amount): convert demos to new storybook

jorenbroekema:


tlouisse:
Looks good 👍  Nice addition about the paste behavior. Maybe in the future we can make a formatting (explaining the heursitic based on ux best practices) rationale in general and make this a part of it.

PR #474, opened 2020-01-06
chore(localize): readme to mdx

CubLion:


tlouisse:
Looks good. Be sure to add all demos (stories) as well

daKmoR:
until all `storiesOf` stories are gone you need to be careful with `stories.js`

https://github.com/ing-bank/lion/pull/474/files#diff-fd81539648ed9dbc635047fc3de2e779R5

For this PR I added
```js
stories: './packages/localize/stories/*.stories.{js,mdx}',
```
e.g. so localize works fine.

=> all stories are converted and I improved some docs... still more would probably be nice... but it's quite a good start I would say 👍 

=> uses separate js stories for the messages and "imports" them into the mdx stories (as you can not use dynamic imports in mdx 😭 )

packages/localize/README.md
```diff
@@ -3,187 +3,42 @@
...
-Async method `loadNamespace()` can be used to load these data.
+```sh
+npm i --save @lion/localize
+```
 
 ```js
-localize.loadNamespace(namespace).then(() => {
-  // do smth when data are loaded
-});
+import @lion/form/lion-form.js;
```
jorenbroekema: Where is this used in the example?

packages/localize/README.md
```diff
@@ -3,187 +3,42 @@
...
-In addition to that the fallback is a necessary mechanism to allow some features of the browsers like Google Chrome translate to work and use the same original data for translations into all not officially supported languages.
-
-> We highly discourage using fallback locales as a user, because it is bug-prone with things like date and number formatting. Please use full locales whenever possible.
-
-### Rendering data
-
-When all necessary data are loaded and you want to show localized content on the page you need to format the data.
-`localize.msg` comes into play here.
-It expects a key in the format of `namespace:name` and can also receive variables as a second argument.
+Rendering data:
```
jorenbroekema: Maybe makes sense to show at least a quick `litMsg` example

packages/localize/stories/10-overview.stories.mdx
```diff
@@ -0,0 +1,119 @@
+import { Story, Meta, html } from '@open-wc/demoing-storybook';
+import { render } from '@lion/core';
+import { localize, formatNumber, formatDate } from '../index.js'
+
+<Meta title="Localize System/Localize" />
+
+<Story name="Default"></Story>
```
jorenbroekema: Empty story

packages/localize/stories/10-overview.stories.mdx
```diff
@@ -0,0 +1,119 @@
...
+npm i --save @lion/localize
+```
+
+```js
+import { localize } from '@lion/localize';
+```
+
+## Web Component
+
+For use in a web component we have `LocalizeMixin` that let's you define namespaces and awaits loading of those translations.
```
jorenbroekema: `lets`

jorenbroekema:
Aah also maybe add https://lion-web-components.netlify.com/?path=/story/forms-input-localize--localize where we show localized feedback message on an input.

packages/localize/README.md
daKmoR: imho `litMsg` should only be used if you really know what you are doing... 
can be added to deep dive or so...

packages/localize/README.md
daKmoR: updated readme

packages/localize/stories/10-overview.stories.mdx
daKmoR: right now needed to show a menu point 🙈 
don't know if there is another way... gonna ask in storybook (and give this as an example)
e.g. is a workaround to get the stories.js <> stories.mdx somewhat working

daKmoR:
yeah there are definitely more things to clean up... will merge for now and we can then adjust in next/storybook

PR #475, opened 2020-01-06
fix(datepicker): focus on selected date on opening

gerjanvangeest:
Improved MR of https://github.com/ing-bank/lion/pull/375

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -287,7 +287,11 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
    */
   _onCalendarOverlayOpened() {
     if (this._focusCentralDateOnCalendarOpen) {
-      this._calendarElement.focusCentralDate();
+      if (this._calendarElement.selectedDate) {
+        this._calendarElement.focusSelectedDate();
+      } else {
+        this._calendarElement.focusCentralDate();
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -286,7 +286,9 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
    * Lifecycle callback for subclassers
    */
   _onCalendarOverlayOpened() {
-    if (this._focusCentralDateOnCalendarOpen) {
```
tlouisse: Dove a bit into it. It's actually the duty of the LionCalendar to do the synchronization of selectedDate to focusedDate on init.

It does this via `__calculateInitialCentralDate`: https://github.com/ing-bank/lion/blob/master/packages/calendar/src/LionCalendar.js#L248
In my opinion it would be the nicest to leave all this logic in the calendar and expose this method and call it like this.

About the prop `_focusCentralDateOnCalendarOpen`: it is mainly meant to prevent automatic synchronization of the centralDate, which can be needed in certain Design Systems (in this case Material Design) In our case `centralDate === selectedDate`.

So in then it should become:

```js
if (this._focusCentralDateOnCalendarOpen) {
  if (this._calendarElement.selectedDate) {
      this._calendarElement.focusSelectedDate();
    } else {
       this._calendarElement.focusCentralDate();
    }
}
```

Ideally, in my opinion, it would be:
```js
if (this._focusCentralDateOnCalendarOpen) {
   this._calendarElement.calculateInitialCentralDate();
   this._calendarElement.focusCentralDate();
}
```

PR #476, opened 2020-01-07
fix(localize): allow negative bulgarian numbers

gerjanvangeest:


packages/localize/src/number/forceAddGroupSeparators.js
```diff
@@ -7,38 +7,44 @@
...
-        numberOfGroups += 1;
-        numberArray.push({ type: 'integer', value: numberPart });
-        if (numberOfGroups !== groups) {
-          numberArray.push({ type: 'group', value: groupSeparator });
-        }
-        numberPart = '';
+  let firstPart;
+  let integerPart;
+
+  for (let i = 0; i < formattedParts.length; i += 1) {
```
jorenbroekema: for (let i = 0; i < formattedParts.length; i += 1) {
    if (formattedParts[i].type === 'integer') {
      firstPart = formattedParts.splice(0, i);
      integerPart = formattedParts.splice(0, 1);
    } else {
      return concatArray;
    }
  }

gerjanvangeest: This suggestion does not work, but i found another way.

packages/localize/src/number/forceAddGroupSeparators.js
```diff
@@ -7,9 +7,17 @@
...
+    if (formattedParts[i].type === 'integer') {
+      firstPart = formattedParts.splice(0, i);
+      integerPart = formattedParts.splice(0, 1);
+    }
+  }
+  if (integerPart[0].value) {
```
jorenbroekema: Will still throw..

Uncaught TypeError: Cannot read property '0' of undefined
    at <anonymous>:1:24

I'm curious why my suggestion doesn't work. In the old diff the same thing happens: it returns concatArray if type of the part is not 'integer'

jorenbroekema: Ahh I know why, because in my suggestion it returns already on the first part, even though the second part might be the integer, which is the case for negative numbers. So then try:
`if (integerPart !== undefined)`

jorenbroekema: Important to strict check for specifically undefined (rather than falsy so that you don't accidentally match with `0` 🙈 )

PR #479, opened 2020-01-07
WIP: feat(sb-action-logger): add action logger component for storybook docs

jorenbroekema:


packages/helpers/sb-action-logger/sb-action-logger.js
```diff
@@ -0,0 +1,3 @@
+import { SbActionLogger } from './src/SbActionLogger.js';
```
daKmoR: let's move this to `packages/helpers/sb-action-logger.js`

and also add `packages/helpers/index.js`
```js
export { SbActionLogger } from './sb-action-logger/src/SbActionLogger.js';
```

e.g. make it sort of a fake mono repo but with a root entry point...
and a way to import like `import @lion/helpers/sb-action-logger.js;`

make sense?

jorenbroekema: Done

PR #488, opened 2020-01-09
chore: render-lit-as-node moved to helpers

CubLion:


packages/helpers/render-lit-as-node/test/render-lit-as-node.test.js
```diff
@@ -1,6 +1,6 @@
 import { expect } from '@open-wc/testing';
```
jorenbroekema: I would call the package `renderLitAsNode` (same as function name) otherwise the import could be quite confusing. We only use the kebab case for web components where the hyphens are mandatory.

PR #496, opened 2020-01-10
chore: remove button icon dependency

CubLion:


packages/button/stories/index.stories.mdx
```diff
@@ -62,26 +60,6 @@ import '@lion/button/lion-button.js';
...
-</Story>
-
-```html
-<lion-button><lion-icon .svg="${bug12}"></lion-icon>Bug</lion-button>
-```
-
-### Icon only button
-
-<Story name="Icon-only">
-  {html`<lion-button aria-label="Bug"><lion-icon .svg="${bug12}"></lion-icon></lion-button>`}
```
daKmoR: Story is still valid but copy paste the svg content in there

PR #498, opened 2020-01-10
Chore/storysort

tlouisse:


.storybook/preview.js
```diff
@@ -24,6 +26,12 @@ async function run() {
     docs: {
       iframeHeight: '200px',
     },
+    options: {
+      // showRoots: true,
+      storySort: storySort({
```
daKmoR: let's just copy the 
```
(a, b) => {
  if (a[1].kind.includes('Intro|')) {
```

directly in here

tlouisse: 👍

PR #499, opened 2020-01-11
Fix calendar todos

ovidiu1:


packages/calendar/src/LionCalendar.js
```diff
@@ -277,7 +277,12 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     return html`
       <div class="calendar__header">
         ${this.__renderPreviousButton(previousMonth, previousYear)}
-        <h2 class="calendar__month-heading" id="month_and_year" aria-atomic="true">
+        <h2
+          class="calendar__month-heading"
+          id="month_and_year"
+          aria-label="${month} ${year}"
```
CubLion: According to MDN:

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is **not visible on the screen**.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

PR #318, opened 2019-10-11
[WIP] feat: dynamically add name attr to child checkboxes/radios

jorenbroekema:
- [ ] Move it from connectedCallback to addFormElements (so it's actually dynamic)
- [ ] Throw if user manually puts wrong name on checkbox/radio when it has a parent group that sets it already

PR #507, opened 2020-01-16
Feat/simplify model values

jorenbroekema:
- Simplify modelValue for choice inputs
- Automatically delegate/copy name attribute to its children as property, to make API for user simpler.

### DIFF
--- a/packages/checkbox-group/stories/index.stories.mdx
+++ b/packages/checkbox-group/stories/index.stories.mdx
@@ -16,24 +16,24 @@ Its purpose is to provide a way for users to check **multiple** options amongst 
 <Story name="Default">
   {html`
     <lion-checkbox-group
-      name="scientistsGroup"
+      name="scientists[]"
       label="Favorite scientists"
     >
-      <lion-checkbox name="scientists[]" label="Archimedes" .choiceValue=${'Archimedes'}></lion-checkbox>
-      <lion-checkbox name="scientists[]" label="Francis Bacon" .choiceValue=${'Francis Bacon'}></lion-checkbox>
-      <lion-checkbox name="scientists[]" label="Marie Curie" .choiceValue=${'Marie Curie'}></lion-checkbox>
+      <lion-checkbox label="Archimedes" .choiceValue=${'Archimedes'}></lion-checkbox>
+      <lion-checkbox label="Francis Bacon" .choiceValue=${'Francis Bacon'}></lion-checkbox>
+      <lion-checkbox label="Marie Curie" .choiceValue=${'Marie Curie'}></lion-checkbox>
     </lion-checkbox-group>
   `}
 </Story>
 
 ```html
 <lion-checkbox-group
-  name="scientistsGroup"
+  name="scientists[]"
   label="Favourite scientists"
 >
-  <lion-checkbox name="scientists[]" label="Archimedes" .choiceValue=${'Archimedes'}></lion-checkbox>
-  <lion-checkbox name="scientists[]" label="Francis Bacon" .choiceValue=${'Francis Bacon'}></lion-checkbox>
-  <lion-checkbox name="scientists[]" label="Marie Curie" .choiceValue=${'Marie Curie'}></lion-checkbox>
+  <lion-checkbox label="Archimedes" .choiceValue=${'Archimedes'}></lion-checkbox>
+  <lion-checkbox label="Francis Bacon" .choiceValue=${'Francis Bacon'}></lion-checkbox>
+  <lion-checkbox label="Marie Curie" .choiceValue=${'Marie Curie'}></lion-checkbox>
 </lion-checkbox-group>
 ```
 
@@ -62,20 +62,17 @@ You can pre-select options by targeting the `modelValue` object of the option an
 
 <Story name="Pre-select">
   {html`
-    <lion-checkbox-group name="scientistsGroup" label="Favorite scientists">
+    <lion-checkbox-group name="scientists" label="Favorite scientists">
       <lion-checkbox
-        name="scientists[]"
         label="Archimedes"
         .choiceValue=${'Archimedes'}
       ></lion-checkbox>
       <lion-checkbox
-        name="scientists[]"
         label="Francis Bacon"
         .choiceValue=${'Francis Bacon'}
         checked
       ></lion-checkbox>
       <lion-checkbox
-        name="scientists[]"
         label="Marie Curie"
         .modelValue=${{ value: 'Marie Curie', checked: true }}
       ></lion-checkbox>
@@ -84,20 +81,17 @@ You can pre-select options by targeting the `modelValue` object of the option an
 </Story>
 
 ```html
-<lion-checkbox-group name="scientistsGroup" label="Favorite scientists">
+<lion-checkbox-group name="scientists[]" label="Favorite scientists">
   <lion-checkbox
-    name="scientists[]"
     label="Archimedes"
     .choiceValue=${'Archimedes'}
   ></lion-checkbox>
   <lion-checkbox
-    name="scientists[]"
     label="Francis Bacon"
     .choiceValue=${'Francis Bacon'}
     checked
   ></lion-checkbox>
   <lion-checkbox
-    name="scientists[]"
     label="Marie Curie"
     .modelValue=${{ value: 'Marie Curie', checked: true }}
   ></lion-checkbox>
@@ -110,19 +104,16 @@ You can disable the entire group by setting the `disabled` attribute on the `<li
 
 <Story name="Disabled">
   {html`
-    <lion-checkbox-group name="scientistsGroup" label="Favorite scientists" disabled>
+    <lion-checkbox-group name="scientists[]" label="Favorite scientists" disabled>
       <lion-checkbox
-        name="scientists[]"
         label="Archimedes"
         .choiceValue=${'Archimedes'}
       ></lion-checkbox>
       <lion-checkbox
-        name="scientists[]"
         label="Francis Bacon"
         .choiceValue=${'Francis Bacon'}
       ></lion-checkbox>
       <lion-checkbox
-        name="scientists[]"
         label="Marie Curie"
         .modelValue=${{ value: 'Marie Curie', checked: true }}
       ></lion-checkbox>
@@ -131,19 +122,16 @@ You can disable the entire group by setting the `disabled` attribute on the `<li
 </Story>
 
 ```html
-<lion-checkbox-group name="scientistsGroup" label="Favorite scientists" disabled>
+<lion-checkbox-group name="scientists[]" label="Favorite scientists" disabled>
   <lion-checkbox
-    name="scientists[]"
     label="Archimedes"
     .choiceValue=${'Archimedes'}
   ></lion-checkbox>
   <lion-checkbox
-    name="scientists[]"
     label="Francis Bacon"
     .choiceValue=${'Francis Bacon'}
   ></lion-checkbox>
   <lion-checkbox
-    name="scientists[]"
     label="Marie Curie"
     .modelValue=${{ value: 'Marie Curie', checked: true }}
   ></lion-checkbox>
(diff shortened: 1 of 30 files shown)


jorenbroekema:
https://github.com/ing-bank/lion/pull/318 this one already started implementation for having the name attribute be dynamically added to child choice input fields.

packages/checkbox-group/test/lion-checkbox-group.test.js
```diff
@@ -1,46 +1,127 @@
...
+        <lion-checkbox .choiceValue=${'male'}></lion-checkbox>
+        <lion-checkbox .choiceValue=${'female'} checked></lion-checkbox>
+        <lion-checkbox .choiceValue=${'other'}></lion-checkbox>
+      </lion-checkbox-group>
+    `);
+    await nextFrame();
+    const invalidChild = await fixture(html`
+      <lion-checkbox name="first-name" .modelValue=${'Lara'}></lion-checkbox>
+    `);
+
```
thatianamg: add another invalid child that is the wrong element (input)

packages/radio-group/test/lion-radio-group.test.js
```diff
@@ -1,144 +1,157 @@
...
-    await el.updateComplete;
-
-    expect(el.checkedValue).to.equal('alien');
-    expect(el.formElementsArray[2].checked).to.be.true;
+    expect(el.formElementsArray[0].name).to.equal('gender');
+    expect(el.formElementsArray[1].name).to.equal('gender');
+    expect(el.formElementsArray[2].name).to.equal('gender');
   });
 
-  it('can handle complex data via choiceValue', async () => {
```
thatianamg: Add it back or explain why you delete

packages/radio-group/test/lion-radio-group.test.js
```diff
@@ -168,27 +181,24 @@ describe('<lion-radio-group>', () => {
...
 
     // not found values trigger no event
-    el.checkedValue = 'foo';
+    el.modelValue = 'foo';
     expect(counter).to.equal(4);
 
-    el.checkedValue = 'male';
+    el.modelValue = 'male';
     expect(counter).to.equal(6); // male becomes checked, alien becomes unchecked
   });
```
thatianamg: remove aliens

packages/radio-group/test/lion-radio-group.test.js
jorenbroekema: Ah, I see now it was moved down the file for organisation purposes but git did not catch the change properly

packages/radio-group/test/lion-radio-group.test.js
```diff
@@ -1,157 +1,157 @@
...
+    const validChild = await fixture(html`
+      <lion-radio .choiceValue=${'male'}></lion-radio>
     `);
-    await nextFrame();
+    el.appendChild(validChild);
 
-    expect(el.checkedValue).to.equal(0);
-    el.formElementsArray[1].checked = true;
-    expect(el.checkedValue).to.equal('');
+    // expect(el.formElementsArray[2].name).to.equal('gender');
```
CubLion: This expect should be un-commented.

packages/select-rich/src/LionSelectRich.js
```diff
@@ -132,16 +107,40 @@ export class LionSelectRich extends OverlayMixin(
...
     }
-    return -1;
+
+    this.__onModelValueChanged();
+    this.requestUpdate('modelValue');
+  }
+
+  get checkedIndex() {
+    let checkedIndex = -1;
+    Array.from(this._listboxNode.children).forEach((option, i) => {
```
jorenbroekema: Note, using listboxNode directly , because formElements array is not great with timing and keeping the right order of the elements... it only appends to the end of the array even if you prepend a new <lion-option> node to the listbox. Caused quite some bugs / weird behavior. See advanced use case tests for a bit more context

packages/select-rich/src/LionSelectRich.js
```diff
@@ -200,31 +197,17 @@ export class LionSelectRich extends OverlayMixin(
     this.__setupInvokerNode();
     this.__setupListboxNode();
 
-    this._invokerNode.selectedElement = this.formElements[this.checkedIndex];
+    // Now that we have rendered our listbox, try setting the user defined modelValue again
```
jorenbroekema: Bit hacky, I tried using updateComplete but then you combine setting child.checked with updateComplete which causes infinite loops. This is because a model-value-changed event is fired due to toggling child.checked and that bubbles back to your rich select causing another update, etc. etc. until infinity.. So just caching the not found modelValue and syncing down after firstUpdated is pretty clean in comparison 🙈

packages/select-rich/src/LionSelectRich.js
```diff
@@ -200,31 +197,17 @@ export class LionSelectRich extends OverlayMixin(
...
-    ) {
-      if (this.checkedIndex) {
-        // Necessary to sync the checkedIndex through the getter/setter explicitly
-        // eslint-disable-next-line no-self-assign
-        this.checkedIndex = this.checkedIndex;
-      }
-    }
-
-    if (name === 'modelValue') {
-      this.dispatchEvent(new CustomEvent('select-model-value-changed'));
```
jorenbroekema: Untested, and just generally very nasty and bad practice. Better to not support this and force ourselves to clean up model-value-changed events to not be composed / bubble..

CubLion:
There are quite a few similarities between `LionCheckboxGroup` and `LionRadioGroup`. I believe we should look at generalizing them to keep things DRY. What do you think?

packages/radio-group/test/lion-radio-group.test.js
jorenbroekema: uncommented ;)

packages/choice-input/src/ChoiceGroupMixin.js
```diff
@@ -0,0 +1,169 @@
...
+          currentElIsCheckedEl = value.includes(this.formElementsArray[i].value);
+        } else {
+          // Allows checking against custom values e.g. formattedValue or serializedValue
+          currentElIsCheckedEl = check(this.formElementsArray[i], value);
+        }
+
+        if (currentElIsCheckedEl) {
+          this.formElementsArray[i].checked = true;
+        } else if (this.multipleChoice) {
+          this.formElementsArray[i].checked = false;
```
CubLion: Regarding what happens in this loop, could we use the: `this.formElementsArray[i].checked = currentElIsCheckedEl` or find a way to write the code that makes it easier to read?

packages/choice-input/src/ChoiceGroupMixin.js
```diff
@@ -0,0 +1,169 @@
...
+        }
+      }
+    }
+
+    __triggerCheckedValueChanged() {
+      const value = this.modelValue;
+      if (value != null && value !== this.__previousCheckedValue) {
+        this.touched = true;
```
CubLion: Where does this property come from? Could we make the relationships between mixins a bit more explicit in the current file?

packages/radio-group/src/LionRadioGroup.js
```diff
@@ -24,102 +25,26 @@ import { LionFieldset } from '@lion/fieldset';
...
-  }
-
+export class LionRadioGroup extends ChoiceGroupMixin(LionFieldset) {
   get formattedValue() {
-    const el = this._getCheckedRadioElement();
+    const el = this._getCheckedElements();
     return el ? el.formattedValue : '';
   }
 
   set formattedValue(value) {
```
CubLion: Why does `formattedValue` exist only on `LionRadioGroup`? Should we implement it for `LionCheckboxGroup` as well? Does it even make sense for this property to exist?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -338,8 +327,19 @@ export class LionSelectRich extends OverlayMixin(
...
+      child.name = this.name;
+    } else {
+      throw new Error(
+        `The lion-select-rich name="${
+          this.name
+        }" does not allow to register ${child.tagName.toLowerCase()} with custom names (name="${
+          child.name
+        }" given)`,
+      );
+    }
```
CubLion: `__delegateNameAttribute` also exists in `ChoiceGroupMixin` with almost the same body. Can we remove it from `LionSelectRich`?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -660,7 +651,7 @@ export class LionSelectRich extends OverlayMixin(
   }
 
   _isEmpty() {
```
CubLion: `_isEmpty` exists in `ChoiceGroupMixin` with almost the same implementation. Should we remove it from here.

packages/choice-input/src/ChoiceGroupMixin.js
jorenbroekema: I improved it :) had to make sure for single choice it only sets checked true if modelValue matches something (and don't do the reverse). Otherwise if we set modelValue to non existing value it will uncheck all the other radios and that's not intended.

packages/choice-input/src/ChoiceGroupMixin.js
jorenbroekema: It's the same for FormRegistrarMixin. The ChoiceGroupMixin overrides / makes references to the FormRegistrarMixin. I know Thijs agrees with wrapping the mixin in the mixins it depends on and using the dedupemixin to dedupe. Feel free to experiment with that and see if it all works when you do it :). 'touched' property comes from InteractionStateMixin if I recall correctly.

packages/radio-group/src/LionRadioGroup.js
jorenbroekema: The setter will set checked to the element of which the formattedValue is the same as the group's formattedValue, but to me this makes no sense for choice inputs/groups, because you don't actually format the user input, you simply respond to the user selection with a visual element (the radio / checkbox). So I'm personally fine with removing it, since it's also untested feature..

packages/select-rich/src/LionSelectRich.js
jorenbroekema: Yes I would be in favor :) I probably missed it

packages/select-rich/src/LionSelectRich.js
jorenbroekema: Yep, if possible :)!

packages/radio-group/src/LionRadioGroup.js
```diff
@@ -24,102 +25,26 @@ import { LionFieldset } from '@lion/fieldset';
...
-    }
-  }
-
+  /**
+   * @override from LionFieldset
+   */
   _onFocusOut() {
     this.touched = true;
     this.focused = false;
   }
```
CubLion: Do you see any point on having this method here? It seems like a duplicate of whats going on in `LionFieldset`.

packages/choice-input/index.js
```diff
@@ -1 +1,2 @@
+export { ChoiceGroupMixin } from './src/ChoiceGroupMixin.js';
```
gerjanvangeest: For `checkbox` or `radio` we have different packages for the `choice` and the `group`. So why is it the same package for these?

packages/checkbox-group/stories/index.stories.mdx
```diff
@@ -62,20 +62,17 @@ You can pre-select options by targeting the `modelValue` object of the option an
 
 <Story name="Pre-select">
   {html`
-    <lion-checkbox-group name="scientistsGroup" label="Favorite scientists">
+    <lion-checkbox-group name="scientists" label="Favorite scientists">
```
daKmoR: shouldn't that be `scientists[]`?

we should throw an error if you try to use a name without an ending `[]` for a checkbox-group

jorenbroekema: Yes by convention it should. To my knowledge it was not spec'd that we should throw an error if people don't use this convention for multiple choice though.

PR #512, opened 2020-01-18
chore: cleanup testing configuration

LarsDenBakker:
There are some old testing commands and configurations that are no longer necessary. Compatibility are now handled automatically.


jorenbroekema:
Thanks :D!

PR #514, opened 2020-01-20
WIP: chore(fieldset): remove serializedGroup specs

jorenbroekema:
Notes for implementation:
- __serializeElements / __serializeElement / serializeGroup methods should be removed, that's the purpose of this. Just the serializedValue getter in LionFieldset should be enough.
- The modelValue of choice groups changes in https://github.com/ing-bank/lion/pull/507 and will affect the serializedValue as well.

### DIFF
--- a/packages/fieldset/test/lion-fieldset.test.js
+++ b/packages/fieldset/test/lion-fieldset.test.js
@@ -1,18 +1,18 @@
+import { LionField } from '@lion/field';
+import '@lion/field/lion-field.js';
+import { localizeTearDown } from '@lion/localize/test-helpers.js';
+import { IsNumber, Validator } from '@lion/validate';
 import {
+  defineCE,
   expect,
   fixture,
   html,
-  unsafeStatic,
-  triggerFocusFor,
   nextFrame,
-  defineCE,
+  triggerFocusFor,
+  unsafeStatic,
 } from '@open-wc/testing';
 import sinon from 'sinon';
-import { Validator, IsNumber } from '@lion/validate';
-import { localizeTearDown } from '@lion/localize/test-helpers.js';
 import '../lion-fieldset.js';
-import { LionField } from '@lion/field';
-import '@lion/field/lion-field.js';
 
 const childTagString = defineCE(
   class extends LionField {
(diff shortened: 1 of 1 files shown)


packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -614,24 +614,20 @@ describe('<lion-fieldset>', () => {
...
       <${tag}>
         <${childTag} name="price"></${childTag}>
+        <${childTag} name="firstName"></${childTag}>
       </${tag}>`);
       await nextFrame();
       fieldset.formElements.price.modelValue = 0;
-      expect(fieldset.serializeGroup()).to.deep.equal({ price: 0 });
-    });
-
-    it('__serializeElements serializes 0 as a valid value', async () => {
```
jorenbroekema: Unnecessary test, and this method will probably be removed as it's only used by serializeGroup and not necessary for serializedValue

packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -614,24 +614,32 @@ describe('<lion-fieldset>', () => {
...
+      const fieldset = await fixture(html`
+      <${tag}>
+        <${childTag} name="price"></${childTag}>
+        <${childTag} name="firstName"></${childTag}>
+      </${tag}>`);
       await nextFrame();
-      const elements = [{ serializedValue: 0 }];
-      expect(fieldset.__serializeElements(elements)).to.deep.equal([0]);
+      fieldset.formElements.price.modelValue = null;
+      fieldset.formElements.firstName.modelValue = undefined;
```
thatianamg: Re-evaluate undefined, see what the platform does, might make sense to include this.

thatianamg: See also separate Unparseable story

daKmoR:
maybe that was a profound misunderstanding but do we want to the `serializedValue` to be able to sync from/to a database?
```
<form>
  <fielset disabled name="fullName">
    <input name="firstName">
    <input name="lastName">
  </fieldset>
  <input name="age">
```

now the modelValue will be
```
{
  fullName: {
    firstName: '...',
    lastName: '...',
  },
  age: 10
}
```

but if we combine serializeGroup and serializedValue e.g. by not doing including disabled dates it would just be
```
{
  age: 10
}
```

e.g. enabling if at some point the fieldset gets enabled... it will mean the firstName/lastName will not have the database values...

but then again - we do not have a disabled state anywhere in the modelValue, serializedValue so maybe it's fine... and you need to manage anything disabled separately?

daKmoR:
will be addressed by https://github.com/ing-bank/lion/pull/583

PR #515, opened 2020-01-22
Chore storybook localize

gerjanvangeest:


packages/localize/stories/20-web-components.stories.mdx
```diff
@@ -1,87 +0,0 @@
-import { Story, Meta, html } from '@open-wc/demoing-storybook';
-import { render } from '@lion/core';
-import { localize, formatNumber, formatDate } from '../index.js';
-
-<Meta title="Localize/Web Components" />
-
-## Web Component
```
daKmoR: this should not be deleted but be filled with information about how the LocalizeMixin works

gerjanvangeest: I just moved it around, all under section `Translate Text`

PR #520, opened 2020-01-24
fix: issue 170

CubLion:


packages/fieldset/package.json
```diff
@@ -37,6 +37,7 @@
     "@lion/validate": "0.6.4"
   },
   "devDependencies": {
+    "@lion/button": "0.5.2",
```
daKmoR: this will mean that an update of button will trigger a release of fieldset - which will trigger, form, checkbox-group, radio-group, select, ...

could we move the test into the `form-system` package?

daKmoR:
what happens if you have
```
<fieldset>
  <button disabled>
  <button>
</fieldset>
```

now if you disable and then enable the fieldset both buttons will be enabled right? 😭

CubLion:
> what happens if you have
> 
> ```
> <fieldset>
>   <button disabled>
>   <button>
> </fieldset>
> ```
> 
> now if you disable and then enable the fieldset both buttons will be enabled right? 😭

Yes, that is correct, both buttons will have their "disabled" property synced with the fieldset.
Do you see it working in a different way?

daKmoR:
we have it behave like native for the inputs
https://lion-web-components.netlify.com/?path=/docs/forms-fieldset-examples--disabled

interestingly for buttons it behaves the same on native fieldsets... but NOT on a form
https://webcomponents.dev/edit/14CvqnHq3vucH4j6LuHQ 

so we might want to integrate buttons into the form registration system? 🤔

CubLion:
> we have it behave like native for the inputs
> https://lion-web-components.netlify.com/?path=/docs/forms-fieldset-examples--disabled
> 
> interestingly for buttons it behaves the same on native fieldsets... but NOT on a form
> https://webcomponents.dev/edit/14CvqnHq3vucH4j6LuHQ
> 
> so we might want to integrate buttons into the form registration system? 🤔

When you say `integrate buttons into the form registration system`, I assume you're talking about `lion-button` right? 
If that is the case, what would happen with a `native button`?

daKmoR:
the issue is that we went ahead and didn't use a native fieldset within our lion-fieldset 😭 
otherwise, this would just work for the native button... (as native disabled fieldsets would disable native buttons)

so either we would need to reconsider how our fieldset works and if it might make sense to use a native fieldset under the hood? (this would probably also reduce the complexity we have for disabled on the input fields - however given that a button within a disabled fieldset has $0.disabled === false... even though it clearly is visually disabled will make it quite tough 🙈 )

or we would say native buttons do not get the disabled applied within our fieldsets...

imho the latter - 
you can integrate into the form system by the Form Registation System and we can especially handle it like for the inputs... but there is no reason to map it to a native button... as it will always be "different"

packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -1152,4 +1153,58 @@ describe('<lion-fieldset>', () => {
       });
     });
   });
+
+  describe('issue 170 - https://github.com/ing-bank/lion/issues/170', () => {
```
erikkroes: Do we really want permanent github links in our code?

erikkroes:
> what happens if you have
> 
> ```
> <fieldset>
>   <button disabled>
>   <button>
> </fieldset>
> ```
> 
> now if you disable and then enable the fieldset both buttons will be enabled right? 😭

If this is not an issue for us, or not part of the requirements. Then I think you can merge if you like.

erikkroes:
Oh, and it would be great to make native buttons "first class citizens" just like our components. Makes it easier to create a "button-light" instead of this web component.

gerjanvangeest:
Closing this MR since it is a breaking change and causes other issues, which will take too much time ATM.

PR #531, opened 2020-01-29
fix(button): using space on button should not scroll page

### LINKED ISSUE #525: [button] space bar scrolls the page
## Describe the bug

When using the spacebar to click a `lion-button`, the page scrolls down. When using spacebar on a native `button` the page does not scroll down. This happens because in `lion-button` we're not interacting with the native button.

## To Reproduce

- Go the storybook
- Use spacebar on a button
- See the page scrolled down

## Expected behavior

The page was not scrolled down

## Additional context

I'm not sure if it's possible to cancel this browser behavior. 

This could be solved by making the native button inside `lion-button` the interactable node, and not the `lion-button` node itself. The main reason I think this wasn't done is because of the automagic 40px click area. Perhaps it's possible to solve this in a different way, like an absolute positioned element or something using `::before` or `::after`. Or let go of this requirement. Given the amount of cross-browser bugs with the button setup, this might be a better approach. It will avoid a lot of headaches in the future.

noventa-dev:
Fix #525: Add preventDefault event as @erikkroes suggested

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -2,6 +2,8 @@ import { css, html, SlotMixin, DisabledWithTabIndexMixin, LitElement } from '@li
 
 // eslint-disable-next-line class-methods-use-this
 const isKeyboardClickEvent = e => e.keyCode === 32 /* space */ || e.keyCode === 13; /* enter */
+// eslint-disable-next-line class-methods-use-this
+const isSpaceKeyboardClickEvent = e => e.keyCode === 32; /* space */
 
 export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement)) {
   static get properties() {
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Does this work cross browsers/different scenarios? These kind of changes have caused a lot of trouble in the past.

erikkroes:
It was an untested remark by me. So yes, this certainly would need cross browser testing.

noventa-dev:
Yeah, I will test it today in different browsers!

noventa-dev:
It seems to be working cross-browser. I manually tested in:
- Chrome Win and Mac
- Edge Mac
- Firefox Mac
- Safari
- IE11

I tested in storybook: [Forms > Feature overview > Example](http://localhost:9001/?path=/docs/forms-features-overview--example)*
And in [Overlays > Dialog](http://localhost:9001/?path=/docs/overlays-dialog--default-story)* changing native buttons for lion-buttons

\* Links to localhost

daKmoR:
thaaankkk you 🤗

PR #539, opened 2020-02-03
Fix/field label as html

gerjanvangeest:


tlouisse:
LGTM 👍  Should we actually do the same for helpText?

packages/field/src/FormControlMixin.js
```diff
@@ -53,6 +53,34 @@ export const FormControlMixin = dedupeMixin(
...
+        return (
+          (this.querySelector('[slot="label"]') &&
+            this.querySelector('[slot="label"]').textContent) ||
+          this.__label
+        );
+      }
+
+      set label(newValue) {
+        const oldValue = this.label;
+        this.__label = newValue;
```
jorenbroekema: we have protected _labelNode but private __label? Shouldn't they have the same access?

jorenbroekema:
What I am missing is an initially failing test that now succeeds, where we test that label with specifically html inside of it, doesn't work, and that now it does work. It's a bit unclear to me how this PR does anything to the "html inside label" feature.

packages/field/src/FormControlMixin.js
gerjanvangeest: `label` is the public api, `__label` is just needed for internal use.

Issue #431, opened 2019-12-06
[Field] Remove `state-*` css classes


daKmoR:
## Describe the bug

LionField still has `state-disabled` and `state-filled` as css classes instead of attribute (disabled & filled)

daKmoR:
fixed via https://github.com/ing-bank/lion/pull/541

PR #541, opened 2020-02-03
fix(field): replace state-disabled/filled css classes with attributes

AymenBenAmor:
Fixing issue : https://github.com/ing-bank/lion/issues/431
Removing state-disabled and state-filled classes and adapting tests and CSS.


daKmoR:
awesome thxxx 🤗

PR #542, opened 2020-02-04
fix(tooltip): dismissable on escape keypress

### LINKED ISSUE #368: [tooltip] Tooltip should be dismissable on Hover

noventa-dev:
Fixes #368

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -32,6 +32,7 @@ export class OverlayController {
       preventsScroll: false,
       trapsKeyboardFocus: false,
       hidesOnEsc: false,
+      hidesOnOutsideEsc: false,
       hidesOnOutsideClick: false,
       isTooltip: false,
       handlesUserInteraction: false,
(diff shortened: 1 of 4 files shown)


daKmoR:
test looks good 👍 
and it fails as expected 💪 

could do add another commit to make it work 🙏 🤗

noventa-dev:
What do you think about this approach?

daKmoR:
woop woop - that was the last a11y issue for tooltip 👍

Issue #544, opened 2020-02-04
[validate] getMessage config value should throw error if value not of type 'function'


T-Motion:
## Describe the bug

When passing a custom getMessage config value to a validator constructor, he should throw an error if it is not of type 'function'.

## To Reproduce

Pass for example a string to the validator's constructor:
`<lion-input .validators=${[new Required({},{ getMessage: 'test' })]}></lion-input>`

Expected: an error is thrown saying the value of getMessage should be of type function.

Actual behavior: he ignores the configuration value.

daKmoR:
solved via https://github.com/ing-bank/lion/pull/545

PR #545, opened 2020-02-04
fix(validator): throw error if getMessage not of type Function

T-Motion:
#544 fix

### DIFF
--- a/packages/validate/test/Validator.test.js
+++ b/packages/validate/test/Validator.test.js
@@ -1,8 +1,21 @@
 import { expect, fixture, html, unsafeStatic, defineCE } from '@open-wc/testing';
-import { LitElement } from '@lion/core';
 import sinon from 'sinon';
+import { LitElement } from '../../core/index.js';
 import { ValidateMixin } from '../src/ValidateMixin.js';
 import { Validator } from '../src/Validator.js';
+
+async function expectThrowsAsync(method, errorMessage) {
+  let error = null;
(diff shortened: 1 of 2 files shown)


packages/validate/test/Validator.test.js
```diff
@@ -1,9 +1,22 @@
 import { expect, fixture, html, unsafeStatic, defineCE } from '@open-wc/testing';
-import { LitElement } from '@lion/core';
 import sinon from 'sinon';
+import { LitElement } from '../../core/index.js';
```
daKmoR: that needs to stay `@lion/core` otherwise it only works within the monorepo and not for our users 🙈

T-Motion: Ofcourse, new commit pushed 😊

daKmoR: tests are not real user code - still, from an npm package point of view it is an "external" package so let's be consistent 👍

daKmoR: nice 👌

T-Motion: Yes ofcourse, i agree. I updated it because i first did "npm i" instead of "yarn install" 😊

daKmoR:
fyi this particular test is flaky so the ci is fine 👍

daKmoR:
nice 👌  this will help a lot of people using it correctly 💪

PR #549, opened 2020-02-05
WIP: POC removal shadow outlet for local

jorenbroekema:
So far I was able to get rid of the shadow outlet in local overlays, and I could not find issues from a first investigation. @tlouisse  you should take a look since you know more what the shadow outlet is useful for ;).

For global overlays, currently we use flexbox which requires the use of at least a parent elem and a child elem (the elem to be positioned), therefore the outlet is quite useful for global overlays.

There is also a proposal to make the global-overlays container div a custom element and to append global overlay content nodes to its shadow root, so we can easily style it without leakage.

Very very pseudo
```html
<lion-global-overlays>
  #shadow-root
    <div class="backdrop"></div>
    <div slot="_shadow-outlet" class="overlay--center"> // display: flex, justify-content: center; align-items: center;
      <div class="overlay-content-node">
      </div>
    </div>
</lion-global-overlays>
```

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -68,7 +68,9 @@ export class OverlayController {
 
     this.manager.add(this);
 
-    this._contentNodeWrapper = document.createElement('div');
+    this._contentNodeWrapper =
+      config.placementMode === 'global' ? document.createElement('div') : config.contentNode;
+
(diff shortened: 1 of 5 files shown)


tlouisse:
fixed by https://github.com/ing-bank/lion/pull/695

Issue #546, opened 2020-02-04
[button] executes regexp on each render


LarsDenBakker:
## Describe the bug

`LionButton` executes a regexp on the user agent string on each render: https://github.com/ing-bank/lion/blob/master/packages/button/src/LionButton.js#L30 this is unnecessary work, executing regexp can be expensive

This should just be constant on top of the file.

daKmoR:
I am reserving this issue 🤗

daKmoR:
solved via https://github.com/ing-bank/lion/pull/550

PR #550, opened 2020-02-05
fix(button): run regexp once instead of every render cycle

wsloth:
Resolves https://github.com/ing-bank/lion/issues/546

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -1,8 +1,13 @@
-import { css, html, SlotMixin, DisabledWithTabIndexMixin, LitElement } from '@lion/core';
+import {
+  css,
+  html,
+  browserDetection,
+  SlotMixin,
+  DisabledWithTabIndexMixin,
+  LitElement,
+} from '@lion/core';
 
-// eslint-disable-next-line class-methods-use-this
 const isKeyboardClickEvent = e => e.keyCode === 32 /* space */ || e.keyCode === 13; /* enter */
-// eslint-disable-next-line class-methods-use-this
 const isSpaceKeyboardClickEvent = e => e.keyCode === 32; /* space */
 
 export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement)) {
@@ -27,7 +32,7 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
     return html`
       <div class="btn">
         ${this._renderBefore()}
-        ${this.constructor.__isIE11()
+        ${browserDetection.isIE11
           ? html`
               <div id="${this._buttonId}"><slot></slot></div>
             `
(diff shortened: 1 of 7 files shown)


packages/button/src/LionButton.js
```diff
@@ -4,6 +4,8 @@ import { css, html, SlotMixin, DisabledWithTabIndexMixin, LitElement } from '@li
 const isKeyboardClickEvent = e => e.keyCode === 32 /* space */ || e.keyCode === 13; /* enter */
 // eslint-disable-next-line class-methods-use-this
 const isSpaceKeyboardClickEvent = e => e.keyCode === 32; /* space */
+// eslint-disable-next-line class-methods-use-this
```
daKmoR: is that needed?
it's outside of the class right?

maybe we can even remove it from the other outside functions as well 🙈

LarsDenBakker:
Now its still run for each component :) im not sure hoe much impact that is though.. less than render at least

wsloth:
Maybe it's worth adding checks like this to `@lion/core`, so we can reuse the check in multiple files. It could even be a singleton export so it's only run the first time you import the file

wsloth:
Perhaps something like:
```js
const isIE11 = /Trident/.test(window.navigator.userAgent);

export const browserDetection = {
  isIE11,
}
```

The reason I would export an object instead of just the constant is so you can more easily test this: just stub the `browserDetection` object, because you can not stub/spy the `userAgent` property on `window.navigator`.

What do you think?

packages/button/src/LionButton.js
wsloth: Removed it

daKmoR:
sounds like a good plan to me 👍 

`isIE11` is also used here `packages/field/src/utils/getAriaElementsInRightDomOrder.js` can you adjust it as well 🙏

wsloth:
I'm not sure if it's possible at all to test the actual `browserDetection` utility class, as you cannot easily stub the `userAgent` property.

Also, should we document this somewhere or leave it for internal usage for now?
I'm also not sure if you would want a more complete browser detection utility, where you can detect (almost) any type of browser.

daKmoR:
we looked into multiple solutions but all felt quite heavy for what we needed - so let's leave it as an internal thing 👍

wsloth:
Is the current solution sufficient or are there some changes needed? 😁

daKmoR:
that looks really good 🤗 

may I nit pic? 😬 

all commits should be fixes (as we are in < 1.x and all changes are released it's nice to have something in the changelog)

and if we do a change in core anyways (which will release everything then let's add another small change) e.g. change the version of lit-element to `^2.2.1`
https://github.com/ing-bank/lion/blob/master/packages/core/package.json#L34
please in a separate commit like `fix(core): change the version of lit-element to latest ^2.2.1`

PS: all this is nitpicking 🙈 feel free to ignore and I will change it tomorrow before I merge 👍

wsloth:
I'll do it 👍

daKmoR:
awesome thxxx 🤗

PR #559, opened 2020-02-09
chore: update to storybook 1.10.x

LarsDenBakker:
Updates `@open-wc/demoing-storybook` to 1.10.x.
- Adds reload when renaming story files
- Regenerate list of stories on reload, so you don't need to restart the server when adding stories
- Uses static imports instead of dynamic imports for stories, drastically reducing filecount after build

I made this PR to do a confidence check the new release works correctly in a big project, but useful to merge this in master as well :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "devDependencies": {
     "@commitlint/cli": "^7.0.0",
     "@commitlint/config-conventional": "^7.0.0",
-    "@open-wc/demoing-storybook": "^1.8.3",
+    "@open-wc/demoing-storybook": "^1.10.1",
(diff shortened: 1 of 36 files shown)


jorenbroekema:
@tlouisse  your current work on serializedValue / serializeGroup will also cause lots of releases so maybe we can combine this PR with yours (e.g. merge to dev branch first before merge to master) so we reduce the amount of releases by half 🙈 ?

tlouisse:
Good idea. I could rebase this 'under my branch' later

daKmoR:
we will go directly to `1.10.1` 🤗 

thx for the hint to update 💪

PR #564, opened 2020-02-11
chore(validation): improve validate example

T-Motion:
Improve validation code example.

### DIFF
--- a/packages/validate/README.md
+++ b/packages/validate/README.md
@@ -48,11 +48,11 @@ class IsInitialsExample extends Validator {
   }
 
   execute(value) {
-    let hasError = false;
+    let hasFeedback = false;
     if (!IsString || !isInitialsRegex.test(value.toLowerCase())) {
-      hasError = true;
+      hasFeedback = true;
     }
-    return hasError;
+    return hasFeedback;
   }
 
   static getMessage({ fieldName }) {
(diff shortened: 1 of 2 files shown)


packages/validate/README.md
```diff
@@ -48,11 +48,11 @@ class IsInitialsExample extends Validator {
   }
 
   execute(value) {
```
jorenbroekema: Maybe we could simplify this to:

```js
execute(value) {
  return (!IsString(value) || !isInitialsRegex.test(value.toLowerCase()));
}
```

packages/validate/stories/Overview.stories.mdx
```diff
@@ -52,11 +52,11 @@ class MyValidator extends Validator {
   }
 
   execute(modelValue, param) {
```
jorenbroekema: Same here, I think

```js
execute() {
  return modelValue === param;
}
```

Is clear enough.

T-Motion: It should be clear the execute function will show any feedback if it returns true. The way I understood it at first is "when validator returns false, there is an error". This should be clarified imo. The example shouldn't be written efficient I think. It should just explain how it work. What do you think?

jorenbroekema: For me an example should showcase usage. 

Documentation / types (jsdocs in our case) should annotate what the return value is and what that means in my opinion.

I get your point of view now though and maybe it is indeed nice that the example provides clarity on top of docs. this PR is an improvement for sure. I'm fine with going with your changes ^^ I will edit my review

daKmoR:
nice 👌

PR #565, opened 2020-02-12
fix(validate): reset validation result on validators change

T-Motion:
When removing a validator from an element, the results of the validation should be reset.
Otherwise the old validation result will stay and the validation will always fail.

### DIFF
--- a/packages/validate/test-suites/ValidateMixin.suite.js
+++ b/packages/validate/test-suites/ValidateMixin.suite.js
@@ -812,6 +812,28 @@ export function runValidateMixinSuite(customConfig) {
         expect(el.validationStates.error).to.eql({});
       });
 
+      it('clears current validation results when validators array updated', async () => {
+        const validators = [new Required()];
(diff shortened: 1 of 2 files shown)


packages/validate/test-suites/ValidateMixin.suite.js
```diff
@@ -812,6 +812,28 @@ export function runValidateMixinSuite(customConfig) {
...
       });
 
+      it('clears current validation results when validators array updated', async () => {
+        const validators = [new Required()];
+        const el = await fixture(html`
+          <${tag}
+            .validators=${validators}
+          >${lightDom}</${tag}>
+        `);
+        el.modelValue = undefined;
```
tlouisse: Thanks for the fix :)
Since we should test whether a new validation is triggered when we change the `.validators`,
it would be better to leave out all the lines with `el.modelValue = undefined;`?
(Especially since a modelValue change also creates a recomputation of the validity state.)

PR #568, opened 2020-02-13
chore(overlays): fix demos

jorenbroekema:


packages/overlays/stories/20-index.stories.mdx
```diff
@@ -7,7 +7,7 @@ import {
 } from '../index.js';
 
 import './demo-overlay-system.js';
-import './applyDemoOverlayStyles.js';1
```
jorenbroekema: I've seen this a lot of times already and it breaks the import on our extension layer 🙈 .. dunno why it happens that suddenly a 1 is added there sometimes.

Issue #533, opened 2020-01-31
[button] sets attribute in constructor which breaks IE11


palash2601:
## Describe the bug

When using `lion-button` in IE11, I get this error

```
Failed to construct 'CustomElement': The result must not have attributes
```
This is because of attributes being set in constructor as can be seen [here](https://github.com/ing-bank/lion/blob/master/packages/button/src/LionButton.js#L160)

Removing this `setAttribute` or waiting with `updateComplete` solves the problem.

As I understand from discussion with @daKmoR , setting attributes is not allowed in Constructor.but Button does so only for IE11 which breaks.

PR #571, opened 2020-02-13
fix(button): ie11 error when setting attribute in constructor

wsloth:
Potentially resolves https://github.com/ing-bank/lion/issues/533


daKmoR:
niiiceee 👍

PR #578, opened 2020-02-14
fix(textarea): support placeholder

amielucha:
Added the missing textarea placeholder attribute to match the input's placeholder functionality.


daKmoR:
we generally don't recommend to use placeholders (as it's a bad UX)
still the platform supports it so we should do

thxx 🤗

PR #582, opened 2020-02-17
WIP: POC shadow outlet and Popper2

tlouisse:
ContentNodeWrapper is now part of shadow dom, so that:
- slot=content can be styled from shadow dom
- we can create tooltip and an arrow all in one component

Approach is as described here: https://github.com/ing-bank/lion/issues/382#issuecomment-564930695

Popper seems to work, except for the offset modifiers.

TODO: fInd out if the problem (offset) is on our side or on Popper side...
- Check all Popper migration guidelines and see if everything is handled
- Make sure to depend on latest version of Popper to see if this makes a difference

TODO: rebase with master (ignore all form changes, they are already on master)


packages/overlays/stories/20-index.stories.mdx
```diff
@@ -473,8 +473,9 @@ class MyOverlayWC extends OverlayMixin(LitElement) {
   render() {
     return html`
       <slot name="invoker"></slot>
-      <slot name="content"></slot>
-      <slot name="_overlay-shadow-outlet"></slot>
+      <div id="overlay-content-node-wrapper">
```
jorenbroekema, line 476: ❤️ this change

Issue #399, opened 2019-11-27
[form] Serialize form with name


LarsDenBakker:
## Describe the bug

When I serialize a form, I expect it to respect the `name` attribute of the form since that is a fieldset in itself.

## To Reproduce

1. Go to https://lion-web-components.netlify.com/?path=/story/forms-form--form-submit-reset
2. Add a `name` attribute on the form with value `myForm`
3. Serialize the form

The form should look like:
```js
 <lion-form id="form" name="myForm" @submit="${submit}"
        ><form>
          <lion-fieldset label="Name" name="name">
            <lion-input name="firstName" label="First Name" .validators=${[new Required(), new MaxLength(15)]}>
            </lion-input>
            <lion-input name="lastName" label="Last Name" .validators=${[new Required(), new MaxLength(15)]}>
            </lion-input>
          </lion-fieldset>
          <button type="submit">Submit</button>
          <button type="button" @click=${() => document.querySelector('#form').resetGroup()}>
            Reset
          </button>
          <p>
            A reset button should never be offered to users. This button is only used here to demonstrate the
            functionality.
          </p>
        </form></lion-form
      >
```

The returned object is: 
```js
{
  name: { 
    firstName: '...', '
    lastName: '..' 
  }
}
```

## Expected behavior

The returned object is:

```js
{
  myForm: {
    name: { 
      firstName: '...', '
      lastName: '..' 
    }
  }
}
```

## Additional context

I can make a PR for this if this, but I need some guidance on what's the best place to implement this.

We could add a `serializeForm` method to the form, or respect the `name` attribute whenever you try to try to grab the `serializedValue`. This could be on `LionFieldset` too.

gerjanvangeest:
If you get the `serializedValue` off an input or a fieldset, you wont get the name either. You don't want to pollute these.

So adding a `serializeForm` method sounds like a better solution to me.

LarsDenBakker:
@gerjanvangeest shall I make a PR for `serializeForm` ?

I see you guys are planning on overhauling `modelValue`, what's the intended use case for `modelValue` and `serializedValue`? It seems like both may be useful in different cases, also when serializing a whole form.

gerjanvangeest:
We are gonna work on the `serializedValue` next sprint. We will take this issue into consideration. 

Please be patient.

tlouisse:
[...]

PR #583, opened 2020-02-17
feat: api normalisation formElements and values

tlouisse:
## Changes

**Simplify / normalize api**
- Discriminate between just 3 types of form controls:
   1. **field**: regular (text) inputs like `lion-input`, `lion-textarea`, `lion-input-range`, `lion-input-(email|amount|date|datepicker)` etc. 
    -> _Children: none_
   2. **choice group**: like `radio-group`, `checkbox-group`, `select-rich`, (and todo: `lion-listbox`, `lion-combobox`)
    -> _Children: array of options_  
   3. **(sub) form**: those are `lion-form` and `lion-fieldset` and can be accessed both like an object and an array (just like platform). A (sub) form can be serialized as a group
    -> _Children: array of controls_, which will be either of category 1 or 2. Those children can be accessed by name
- formElements work like the platform `form.elements`. See: https://github.com/ing-bank/lion/blob/72fcd3768f9f59bfc21cc9ba90b21ddcf600cf74/packages/field/src/registration/FormControlsCollection.js
- modelValue/serializedValue are derived from formElements structure. The only exception is the ChoiceGroup, where only checked values are filtered out and a `multipleChoice=false` results in a single value
- `.serializedValue` and `serializeGroup()` have same outcome (except for disabled values being ignored). 
- `serializeGroup()` has two options: `keepDisabled` and `includeSelf`:  https://github.com/ing-bank/lion/blob/72fcd3768f9f59bfc21cc9ba90b21ddcf600cf74/packages/fieldset/test/lion-fieldset.test.js#L789. The latter addresses: https://github.com/ing-bank/lion/issues/399

**Fixes:**
- compute interaction state after registration: https://github.com/ing-bank/lion/blob/fix/serializedValues/packages/fieldset/test/lion-fieldset.test.js#L601

Cleanup:
- Removed FieldCustomMixin. It doesn't belong in `packages/field` which is actually the form core package. It has specific logic for our extension layer and should be placed there


daKmoR:
Submit & Reset Demo no longer works :(

https://deploy-preview-583--lion-web-components.netlify.com/?path=/docs/forms-form-examples--default-story

packages/choice-input/src/ChoiceGroupMixin.js
```diff
@@ -119,24 +143,25 @@ export const ChoiceGroupMixin = dedupeMixin(
...
 
       _getCheckedElements() {
-        const filtered = this.formElementsArray.filter(el => el.checked === true);
-        if (this.multipleChoice) {
-          return filtered;
-        }
-        return filtered.length > 0 ? filtered[0] : undefined;
+        // We want to filter out disabled values out by default
+        return this.formElements.filter(el => el.checked && !el.disabled);
+        // if (this.multipleChoice) {
```
daKmoR: why leave commented out code in there? imho should be removed

tlouisse: Definitely :) Alex pointed this already out as well. I'm on it

tlouisse:
> Submit & Reset Demo no longer works :(
> 
> https://deploy-preview-583--lion-web-components.netlify.com/?path=/docs/forms-form-examples--default-story

I reverted the demo functionality to how it was before

PR #584, opened 2020-02-17
docs: added line to explain how to demo locally

soelen:
It is not clear for people which do not work with yarn workspaces that installing depencies with yarn is necessary.

Looked into README.md and expected something about it in the demo section so... here is a PR to explain short why and what to do.


daKmoR:
thxxxx 🤗

PR #585, opened 2020-02-18
WIP: Fix/serialized values wip disabled

CubLion:


packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -1191,4 +1191,83 @@ describe('<lion-fieldset>', () => {
...
+        // when
+        el.restore({
+          full_name: {
+            first_name: 'John',
+            last_name: 'Doe',
+          },
+          city: 'Amsterdam',
+        });
+        // then
+        expect(el.modelValue).to.eql({
```
tlouisse: Looks good. 

I think this would actually give back nothing, since modelValue should not give back disabled values.
But if you replace this with a capture() or do modelValue with (at least some) none disabled on top, it would be great

PR #586, opened 2020-02-18
fix(field): do not validate disabled fields

AymenBenAmor:
disable validation of disabled fields

### DIFF
--- a/packages/validate/src/ValidateMixin.js
+++ b/packages/validate/src/ValidateMixin.js
@@ -261,6 +261,12 @@ export const ValidateMixin = dedupeMixin(
        * call. Situation B will occur after every call.
        */
       async validate({ clearCurrentResult } = {}) {
+        if (this.disabled) {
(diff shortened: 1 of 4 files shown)


packages/validate/src/ValidateMixin.js
```diff
@@ -261,6 +261,12 @@ export const ValidateMixin = dedupeMixin(
        * call. Situation B will occur after every call.
        */
       async validate({ clearCurrentResult } = {}) {
+        if (this.hasAttribute('disabled')) {
```
daKmoR: could we use `this.disabled` here? as attributes are updated async

AymenBenAmor: Changed to this.disabled as suggested.

daKmoR: nice tests are still green 💪

daKmoR:
awesome thxxx 💪

PR #588, opened 2020-02-19
chore(textarea): document readonly

gserb:
improve documentation by adding details and example about the `readonly` attribute


daKmoR:
thaaaankkks 🤗

PR #590, opened 2020-02-19
chore: fix readme validate

jorenbroekema:


packages/validate/README.md
```diff
@@ -49,7 +49,7 @@ class IsInitialsExample extends Validator {
 
   execute(value) {
     let hasError = false;
-    if (!IsString || !isInitialsRegex.test(value.toLowerCase())) {
+    if (!IsString(value) || !isInitialsRegex.test(value.toLowerCase())) {
```
daKmoR: I think that will not work either right?
it is a class so it will need something like this?

```js
const validator = new IsString();
if (!validator.execute(value) ...)
```

PR #592, opened 2020-02-19
fix(tab): remove scroll while navigating the tabs using up/down arrows 

### LINKED ISSUE #589: [tab] when navigating the tabs with the up/down keys it also scrolls the page
## Describe the bug

When navigating the tabs with the up/down keys it also scrolls the page

## To Reproduce

1. visit https://lion-web-components.netlify.com/?path=/docs/navigation-tabs--default-story
2. select a tab and use the arrow up/down
3. the page will scroll a little each time

## Expected behavior

When navigating the tabs with the up/down keys the page should NOT scroll

## Additional context

possible solution: prevent default action on up/down key while tabs are focused

gserb:
closes #589


gserb:
Fix for issue #589

daKmoR:
nice - thxxx 🤗

PR #598, opened 2020-02-20
chore(lion-switch): show info type validator

Hzunax:
Info type is not supported for this component. Throws error in browser:
image


jorenbroekema:

Issue #591, opened 2020-02-19
Make templates more extendable


CubLion:
## Decouple styling and logic

Let's make it easier for Sub Classers that want to extend a certain component to adhere to their Design System.

In most cases, extending a component comes down to:
1. overriding/extending the `_xTemplate` (for instance `_inputTemplate`) functions that define markup
2. overriding/extending the `static get styles` with Design System specific styles 

Point 2 is straightforward. 
For point 1 though, we currently have a lot of places where we mix logic and markup/styling. An example (slightly adjusted) currently in https://github.com/ing-bank/lion/blob/master/packages/calendar/src/LionCalendar.js:

### Current implementation
```js
_previousButtonTemplate(previousMonth, previousYear) {
    const previousButtonTitle = `${this.msgLit(
      'lion-calendar:previousMonth',
    )}, ${previousMonth} ${previousYear}`;

    return html`
      <button
        class="calendar__previous-month-button"
        aria-label=${previousButtonTitle}
        title=${previousButtonTitle}
        @click=${this.__myPrivateHandler}
        ?disabled=${this.isPreviousMonthDisabled}
      >
        &lt;
      </button>
    `;
}
``` 

If we can make the following assumptions:
- Lion layer takes care of logic like:
  - what should happen when we click on this button?
- Lion layer takes care of accessibility (including translations of purely functional aspects, like an alternative text for navigating to previous month), similar to how native ui elements in the platform would do this.
- Subclassers should not use private methods (__myPrivateHandler above)
- Classes on elements should never be used as functional hooks and can be interchanged for styling purposes. When our component code needs a reference to a certain element, it's best to give this element a 'data-' attribute that the extending party is required to implement

We can approach this using the spread directive: https://open-wc.org/developing/lit-helpers.html#spread-directives.

### Decoupled implementation
```js
_previousButtonTemplate({ prevButtonProps }) {
  return html`
    <button
      data-ref="prevButton"
      class="calendar__previous-month-button"
       ...=${spread(prevButtonProps)}>
      &lt;
   </button>
`;
}
```

### An extension implementation

Let's say we want to create our own Material Design calendar. Our template function would look like:

```js
_previousButtonTemplate({ prevButtonProps }) {
  return html`
    <md-button
      data-ref="prevButton"
      class="md-calendar-next-button"
       ...=${spread(prevButtonProps)}>
      &lt;
    </md-button>
  `;
}
```

[...]

PR #599, opened 2020-02-20
WIP: Breaking/scoped elements

jorenbroekema:
TODO:
- Documentation / rationale on why/when to use scoped elements.

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -3,7 +3,6 @@ import { html, ifDefined, render } from '@lion/core';
 import { LionInputDate } from '@lion/input-date';
 import { OverlayMixin, withModalDialogConfig } from '@lion/overlays';
 import { ScopedElementsMixin } from '@open-wc/scoped-elements';
-import { LionValidationFeedback } from '@lion/validate';
 import { LionCalendarOverlayFrame } from './LionCalendarOverlayFrame.js';
 
 /**
@@ -14,9 +13,6 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
   static get scopedElements() {
     return {
       ...super.scopedElements,
-      // TODO (@CubLion): see if registering "lion-validation-feedback"
-      // is indeed needed once a new release of  "@lion/validate" happens.
-      'lion-validation-feedback': LionValidationFeedback,
       'lion-calendar': LionCalendar,
       'lion-calendar-overlay-frame': LionCalendarOverlayFrame,
     };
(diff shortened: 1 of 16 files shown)


packages/validate/src/ValidateMixin.js
```diff
@@ -24,7 +26,15 @@ function arrayDiff(array1 = [], array2 = []) {
 export const ValidateMixin = dedupeMixin(
   superclass =>
     // eslint-disable-next-line no-unused-vars, no-shadow
-    class ValidateMixin extends SyncUpdatableMixin(SlotMixin(superclass)) {
+    class ValidateMixin extends ScopedElementsMixin(SyncUpdatableMixin(SlotMixin(superclass))) {
+      static get scopedElements() {
+        return {
+          ...super.scopedElements,
+          // TODO: Remove once we can load it asynchronously, see todo below in loadFeedbackComponent method
```
daKmoR: we no longer need to load it asynchronously as we can override it when extending

jorenbroekema: True! Good catch :D

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -350,7 +359,7 @@ export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
     if (this._cachedOverlayContentNode) {
       return this._cachedOverlayContentNode;
     }
-    this._cachedOverlayContentNode = this.shadowRoot.querySelector('lion-calendar-overlay-frame');
+    this._cachedOverlayContentNode = this.shadowRoot.querySelector('.calendar__overlay-frame');
```
tlouisse: I would argue to not use class selectors (styles) for functional purposes (element references): https://css-tricks.com/stop-using-css-selectors-non-css/

Especially in our Lion layer, they can be swapped out by Sub classers.
For a practical example: see 'An extension implementation' here: https://github.com/ing-bank/lion/issues/591

jorenbroekema: I didn't see an alternative, I am aware classes are as a best practice for styling only, but as with all best practices, you only follow them if you can. If you can't you need to go with the lesser of multiple evils, which is either id or class, or do you propose something else?

Soon, scoped-elements will release a helper to do query selection on scoped elements by allowing to get the tagname from the scoped registry so we may use that, but I had to patch this somehow in the meantime

tlouisse: Let's go for the tag name from scoped-elements indeed. In other contexts, I would use a data attr (or alternatively a class with 'js' prefix) like the article suggests

CubLion:
Still to be done:
- select-rich
- switch

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -1,15 +1,23 @@
...
 
 /**
  * @customElement lion-input-datepicker
  * @extends {LionInputDate}
  */
-export class LionInputDatepicker extends OverlayMixin(LionInputDate) {
+export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionInputDate)) {
+  static get scopedElements() {
+    return {
+      ...super.scopedElements,
```
CubLion: `scopedElements` is static. Does this super call make sense?

jorenbroekema: Yes, super refers to the parent constructor which can access static methods and properties.

daKmoR:
implemented via https://github.com/ing-bank/lion/pull/629

Issue #597, opened 2020-02-19
[validate] when using a mangler the default error message becomes useless


daKmoR:
## Describe the bug

The default error message is using the `static name` of a class however it class names are mangled to optimize performance the error message becomes gibberish like.

```
Please configure an error message for "Ot" by overriding "static async getMessage()"
```
while it is a "Required" Validator.

## To Reproduce

1. Open demo [Form Features Overview](https://lion-web-components.netlify.com/?path=/story/forms-features-overview--example)
2. Fill in something into for example the textarea
3. The gibberish error message will appear

## Additional context

The reason for that is that we use `this.name` in a static function... which can not access the `this.name` in the constructor

```js
export class Validator {
  constructor(param, config) {
    fakeExtendsEventTarget(this);

    this.name = '';
    // ...
  }
  static async getMessage(/* data */) {
    return `Please configure an error message for "${this.name}" by overriding "static async getMessage()"`;
  }
```
https://github.com/ing-bank/lion/blob/master/packages/validate/src/Validator.js#L82

**Possible solution**:

Use `static get validatorName() { return 'Required'; }` to set the name... e.g. to make it available statically as well.

A validator probably should never change it's name... hence static should work perfectly...

tlouisse:
Yeah, makes sense. Is indeed good  to be unable to change the name of an instance

daKmoR:
So let's make these "static" settings like name, async actually static...

```js
// from
constructor(param, config) {
  this.name = 'Required';
  this.async = true;
}

// to
static get validatorName() { return 'Required'; }
static get async() { return true; }
```

daKmoR:
fixed via https://github.com/ing-bank/lion/pull/600

PR #600, opened 2020-02-20
feat(validate): change to use static validatorName instead of instance name

gserb:
Fix for issue #597

### DIFF
--- a/packages/field/test-suites/FormatMixin.suite.js
+++ b/packages/field/test-suites/FormatMixin.suite.js
@@ -344,9 +344,13 @@ export function runFormatMixinSuite(customConfig) {
         // that set hasError back to false when the user input is mimicked.
 
         const AlwaysInvalid = class extends Validator {
+          // eslint-disable-next-line no-useless-constructor
           constructor(...args) {
             super(...args);
-            this.name = 'AlwaysInvalid';
+          }
+
+          static get validatorName() {
+            return 'AlwaysInvalid';
           }
 
           execute() {
--- a/packages/field/test-suites/FormatMixin.suite.js
+++ b/packages/field/test-suites/FormatMixin.suite.js
@@ -344,11 +344,6 @@ export function runFormatMixinSuite(customConfig) {
         // that set hasError back to false when the user input is mimicked.
 
         const AlwaysInvalid = class extends Validator {
-          // eslint-disable-next-line no-useless-constructor
-          constructor(...args) {
-            super(...args);
-          }
-
           static get validatorName() {
             return 'AlwaysInvalid';
           }
--- a/packages/input-iban/src/validators.js
+++ b/packages/input-iban/src/validators.js
@@ -19,11 +19,6 @@ const loadTranslations = async () => {
 };
 
 export class IsIBAN extends Validator {
-  // eslint-disable-next-line no-useless-constructor
-  constructor(...args) {
-    super(...args);
-  }
-
   static get validatorName() {
     return 'IsIBAN';
   }
--- a/packages/input-iban/src/validators.js
+++ b/packages/input-iban/src/validators.js
@@ -19,9 +19,13 @@ const loadTranslations = async () => {
 };
 
 export class IsIBAN extends Validator {
+  // eslint-disable-next-line no-useless-constructor
   constructor(...args) {
     super(...args);
-    this.name = 'IsIBAN';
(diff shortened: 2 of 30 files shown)


daKmoR:
`static get name` is already used as the actual name/class given by javascript itself and we don't wanna override that

we don't need to store it as we only allow it to be changed by extending and overriding the actual static getter
so in the base class it should look like this

```js
static get validatorName() {
   return ''
}
```

same for async
```js
static get async() {
   return false;
}
```

this means to remove `name` and `async` from the constructor (to remove it from the instance)

Also, we would need to adjust it everywhere where it is used. e.g. for the 4 Validators we have here https://github.com/ing-bank/lion/tree/master/packages/validate/src/validators

this is a breaking change e.g. a lot of tests should/will fail and those will need to be fixed as well.

also, documentation will need to be adjusted 🙈

packages/field/test-suites/FormatMixin.suite.js
```diff
@@ -344,9 +344,13 @@ export function runFormatMixinSuite(customConfig) {
         // that set hasError back to false when the user input is mimicked.
 
         const AlwaysInvalid = class extends Validator {
+          // eslint-disable-next-line no-useless-constructor
           constructor(...args) {
```
daKmoR: eslint is right... let's just remove the constructor here?

packages/validate/src/Validator.js
```diff
@@ -4,22 +4,30 @@ export class Validator {
...
    * @param {string|Date|Number|object} modelValue
    * @param {object} param
    * @returns {Boolean|Promise<Boolean>}
    */
   execute(/* modelValue, param */) {
-    if (!this.name) {
-      throw new Error('You must provide a name like "this.name = \'IsCat\'" for your Validator');
+    if (!this.validatorName) {
+      throw new Error(
+        'You must provide a name like "this.validatorName = \'IsCat\'" for your Validator',
```
daKmoR:
```suggestion
        A validator needs to have a name! Please set it via "static get validatorName() { return 'IsCat'; }"
```

packages/validate/test/DateValidators.test.js
```diff
@@ -13,7 +13,7 @@ describe('Date Validation', () => {
   it('provides new isDate() to allow only dates', () => {
     let isEnabled;
     const validator = new IsDate();
-    expect(validator.name).to.equal('IsDate');
+    expect(IsDate.validatorName).to.equal('IsDate');
```
daKmoR: nit pick 😬 
maybe use `validator.constructor.name`? might be a more realistic use-case 🙈

packages/validate/src/Validator.js
gserb: updated 😊

packages/validate/test/DateValidators.test.js
gserb: done 😊

packages/field/test-suites/FormatMixin.suite.js
gserb: 👌

packages/validate/test-suites/ValidateMixin.suite.js
```diff
@@ -218,9 +221,13 @@ export function runValidateMixinSuite(customConfig) {
 
       it('finally checks for ResultValidators: creates TotalValidationResult', async () => {
         class MyResult extends ResultValidator {
+          // eslint-disable-next-line no-useless-constructor
```
jorenbroekema: Remove the constructor

packages/validate/test-suites/ValidateMixin.suite.js
```diff
@@ -391,10 +404,17 @@ export function runValidateMixinSuite(customConfig) {
       });
 
       class IsAsyncCat extends Validator {
+        // eslint-disable-next-line no-useless-constructor
```
jorenbroekema: same here

packages/input-iban/src/validators.js
```diff
@@ -36,9 +35,13 @@ export class IsIBAN extends Validator {
 }
 
 export class IsCountryIBAN extends IsIBAN {
+  // eslint-disable-next-line no-useless-constructor
```
jorenbroekema: same here

jorenbroekema:

Issue #505, opened 2020-01-16
model-value-changed lacks consistency


jorenbroekema:
## Describe the bug

`model-value-changed` event is fired when the modelValue of a field changes. There are a few problems. It has `composed: true` which is pretty overkill, since we don't have a `render` method with a template for most of our fields. They usually spawn native elements in the LightDOM, and I don't think we usually recommend having shadow boundaries if you extend our (simple) fields.

Another issue is the bubbling. Usually this is useful, because if a field changes model-value, the parent fieldsets all change model-value as well, so to make it bubble makes a lot of sense.

However, there's a few fields where we get a weird number of `model-value-changed` when listening to it on the wrapping form:

- If you have a `checkbox-group` or `radio-group`, you do not get any `model-value-changed` on initialization. For all other fields in the umbrella form, you do get it.
- If you have a `select-rich`, you get `model-value-changed` multiple times. I tested it with 3 choices, and got it 3 times, then another 2 times, on init. Probably has to do with the options  nodes being moved?
- In both `radio-group` and `select-rich`, if an option is selected, and you then change the selection, you get a `model-value-changed` from both the option that was checked, and from the option that was unchecked as a result (because it's only single-select). 

Reproduce it by listening to `model-value-changed` on a form that has radio-group, checkbox-group, rich-select in it.

## Expected behavior

- A single model-value-changed event to reach the form when we change the selection in a field that has options.
- Either 1 model-value-changed event on init for all fields, or none. Right now it is 0, 1 or .. 5
- model-value-changed to not go through shadow boundaries

gerjanvangeest:
Fixed in: https://github.com/ing-bank/lion/pull/647

PR #602, opened 2020-02-21
WIP: fix model value inconsistency

CubLion:
Addresses:
- https://github.com/ing-bank/lion/issues/505
- TODO: add more issues

### DIFF
--- a/packages/fieldset/src/FormGroupMixin.js
+++ b/packages/fieldset/src/FormGroupMixin.js
@@ -106,6 +106,7 @@ export const FormGroupMixin = dedupeMixin(
         this.dirty = false;
         this.touched = false;
         this.focused = false;
+        this._repropagateRole = 'form-group'; // configures FormControlMixin
         this.__addedSubValidators = false;
 
         this._checkForOutsideClick = this._checkForOutsideClick.bind(this);
@@ -135,8 +136,8 @@ export const FormGroupMixin = dedupeMixin(
       }
 
       async __initInteractionStates() {
-        if (!this.__readyForRegistration) {
-          await this.registrationReady;
+        if (!this.registrationHasCompleted) {
+          await this.registrationComplete;
         }
         this.formElements.forEach(el => {
           if (typeof el.initInteractionState === 'function') {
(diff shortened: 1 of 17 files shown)


jorenbroekema:
The test is a good start to talk about what the behavior of this event should be. Eventually I would make this a test suite (or maybe two test suites, one for fieldsets, one for fields) that can be tested with various fieldsets and fields (e.g. radio-group, select-rich and radio, input-amount), so basically integration tests. See interaction state mixin integration tests.

As far as I remember a summary of the proposal was:
- The event never is composed: true.
- The event does not bubble from fieldsets
- The event bubbles from all fields
- For single-choice we need to put something in place to prevent double model value changed when selection changes, because changing selection means 2 options' modelValues change because the `checked` flips for both the thing that got selected, and the thing that got unselected as a direct result. We could prevent the bubble on the model-value-changed that got unselected. Then people can still listen to it on the radio itself if they really want, but it no longer bubbles up.
- All model-value-changed events stop propagating at <lion-form> level, where a form-model-value-changed event is dispatched? I forgot what the purpose of this was, I think with this proposal it is not necessary, because for any change to the form and its sub-fieldsets and sub-fields, only 1 event should bubble to the top level now.

packages/fieldset/src/FormGroupMixin.js
```diff
@@ -394,5 +395,21 @@ export const FormGroupMixin = dedupeMixin(
         super.removeFormElement(...args);
         this.validate();
       }
+
+      __onModelValueChanged(e) {
+        // TODO (@tlouisse): explain why this early return is needed.
```
tlouisse: that would catch the event below

packages/fieldset/src/FormGroupMixin.js
```diff
@@ -394,5 +395,21 @@ export const FormGroupMixin = dedupeMixin(
...
+        if (e.target === this) {
+          return;
+        }
+        e.stopImmediatePropagation();
```
tlouisse: Can we get without the exception for a choice group or is that a todo?

Issue #587, opened 2020-02-19
[tab] tab button gain forced focus at loading


georges-gomes:
## Describe the bug

The selected tab button gains focus.

## To Reproduce

If the tab is down in a page, the page will scroll to reach the button - at load.

## Expected behavior

Button should be selected but not gain focus

daKmoR:
hey there 🤗 

I am not sure what you mean? e.g. you load the page and the tab is focused? or do you change the `activeIndex` and then it becomes focused?

I can't seem to reproduce it 🙈 do you maybe have a gif or so?

georges-gomes:
I did used "selectedIndex" attribute but I quickly tested without it and got the same result. Maybe badly tested.

I fixed it quick and dirty with 
```
const selectButton = element => {
  //element.focus();
  element.setAttribute('selected', true);
  element.setAttribute('aria-selected', true);
  element.setAttribute('tabindex', 0);
};
```

Because I think the `selectedIndex` attribute does call `__updateSelected()` which calls `selectButton`.

My fix is aggressive and works for my case but it should more subtle like only focusing on 
`set selectedIndex(value)` but not when first landed on the page.

Hope this helps

jorenbroekema:
Should be fixed now, please let us know if something is still broken regarding this! Thanks for raising it ;)

PR #603, opened 2020-02-21
fix(tabs): prevent scrolling when initially focussing a tab

wsloth:
Resolves https://github.com/ing-bank/lion/issues/587

## Input needed
Is it desired that focussing on a tab after a selectedIndex change does _not_ scroll to the element anymore? I am not sure from an accessibility point of view how this should behave.

## How to reproduce the issue before the fix

Here's a hacky way to see the issue. In the selected index tabs storybook, just a while bunch of whitespace on top of the tabs component. Then, refresh the page and you will see it scrolls down to the selected tab. After the fix, it does not behave like that anymore.

```
### Selected Index

You can set the `selectedIndex` to select a certain tab.

<Preview>
  <Story name="Selected index">
    {html`
    <br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/><br/>
      <lea-tabs .selectedIndex=${1}>
        <lea-tab slot="tab">Info</lea-tab>
        <lea-tab-panel slot="panel">
          Info page with lots of information about us.
        </lea-tab-panel>
        <lea-tab slot="tab">Work</lea-tab>
        <lea-tab-panel slot="panel">
          Work page that showcases our work.
        </lea-tab-panel>
      </lea-tabs>
    `}
  </Story>
</Preview>
```

### DIFF
--- a/packages/tabs/src/LionTabs.js
+++ b/packages/tabs/src/LionTabs.js
@@ -37,8 +37,12 @@ const cleanButton = (element, clickHandler, keydownHandler) => {
   element.removeEventListener('keydown', e => e.preventDefault());
 };
 
-const selectButton = element => {
-  element.focus();
+const selectButton = (element, firstUpdate = false) => {
+  // Don't focus on first update, as the component might be lower on the page
+  if (!firstUpdate) {
+    element.focus();
+  }
+
   element.setAttribute('selected', true);
   element.setAttribute('aria-selected', true);
   element.setAttribute('tabindex', 0);
@@ -109,6 +113,7 @@ export class LionTabs extends LitElement {
 
   firstUpdated() {
     super.firstUpdated();
+    this.__firstUpdate = true;
     this.__setupSlots();
   }
 
@@ -198,6 +203,7 @@ export class LionTabs extends LitElement {
   }
 
   set selectedIndex(value) {
+    this.__firstUpdate = false;
     const stale = this.__selectedIndex;
     this.__selectedIndex = value;
     this.__updateSelected();
@@ -231,7 +237,7 @@ export class LionTabs extends LitElement {
     }
     const { button: currentButton, panel: currentPanel } = this.__store[this.selectedIndex];
     if (currentButton) {
-      selectButton(currentButton);
+      selectButton(currentButton, this.__firstUpdate);
     }
     if (currentPanel) {
       selectPanel(currentPanel);
--- a/packages/tabs/test/lion-tabs.test.js
+++ b/packages/tabs/test/lion-tabs.test.js
@@ -243,6 +243,50 @@ describe('<lion-tabs>', () => {
     });
   });
 
+  describe('Initializing without Focus', () => {
+    it('keeps track of when the component is updated', async () => {
+      const el = await fixture(html`
+        <lion-tabs>
+          <button slot="tab">tab 1</button>
+          <div slot="panel">panel 1</div>
+          <button slot="tab">tab 2</button>
+          <div slot="panel">panel 2</div>
+        </lion-tabs>
+      `);
+
+      expect(el.__firstUpdate).to.be.true;
+      el.selectedIndex = 1;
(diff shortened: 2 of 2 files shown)


daKmoR:
hmm maybe we want to do this only while initializing? e.g. don't scroll when the first load happens... (maybe a this.isConnected check is good enough?) or an extra flag that get's set on `firstUpdated` and we only scroll afterwards?

later if you tab and navigate with keyboard/mouse it probably makes sense to scroll?

hmmm also tough to test 😅

wsloth:
That is a good one. I changed it to skip focus on first update, because to me it does not make sense that the tab would grab focus when it's first rendered. I know of a place in my application where we want to show tabs, and select one off-index by default. But it's down somewhere on the page and we definitely don't want it to grab focus when first loading the application.

Let me know what you think.

daKmoR:
the more I think about this the more I believe that it should not set any focus when you change `activeIndex`.

Only if you interact with the keyboard (tab in, switch via arrows( then it should set the focus...

sounds like a plan?

wsloth:
I'm not sure. To me it makes sense that the focus is set to the tab you're making active, regardless of whether you used the keyboard or the mouse. 

If you make a distinction between the two, you can end up in awkward situations:
- When tabbing through the page, you land on the first tab
- You click on the second one
- The first one is visibly focused, but the second tab is active

What is the best desired behavior? As the MR currently shows, is that the behavior is the same as before creating this MR. The only difference is that on initialising the component, the focus is not set to the active tab immediately.

wsloth:
I'm not sure. To me it makes sense that the focus is set to the tab you're making active, regardless of whether you used the keyboard or the mouse. 

If you make a distinction between the two, you can end up in awkward situations:
- When tabbing through the page, you land on the first tab
- You click on the second one
- The first one is visibly focused, but the second tab is active

What is the best desired behavior? As the MR currently shows, is that the behavior is the same as before creating this MR. The only difference is that on initialising the component, the focus is not set to the active tab immediately.

jorenbroekema:
To me this solution seems fine. The only way `.focus` is called is by changing the `selectedIndex`, and this usually only happens on click, on keyboard interaction, or by programmatically using the `selectedIndex` setter yourself, and in all those cases I believe the focus should be set to the button, with the exception of firstUpdated. 

From Thomas' comment it seems like he wants to also exempt the programmatic usage of the `selectedIndex` setter from setting focus, but it is unclear to me why that is. If the user of the tabs sets the index programmatically after first update, to me that seems like the use case for that is that he wants the user to be directed to the tab that the selectedIndex was changed to. If that's not the case, then I agree it's really annoying for the flow. Example: maybe you have tabs far down the page, but depending on interaction that happens earlier on the page at the top, you want to change the default shown tab. So then as an app developer I'd like to be able to prevent the .focus(), because I just want to change the default shown tab, not direct the user to it as well.

If you guys agree, it makes sense for me to add a feature for the setter to provide it with a flag 'delegateFocusToTab' that is by default false? Then it doesn't do the .focus thing on first render, we will set it to true for keyboard + mouse interactions, and a user can set it to true if it wants the focus delegation when he is using the setter programmatically himself.

Let me know ;)

daKmoR:
let's move forward with this as it is definitely an improvement 💪  if we need to improve even more later we can do so 👍 

thxxxx 🤗

PR #605, opened 2020-02-23
refactor(localizenamespaces): remove switch statement

armujahid:
This makes `localizeNamespaces` implementation consistent. Switch statement was added to make it compatible with polymer build

Reference thread: https://polymer.slack.com/archives/CJGFWJN9J/p1581678489008400


packages/calendar/src/LionCalendar.js
```diff
@@ -22,50 +22,7 @@ import { createDay } from './utils/createDay.js';
...
-            case 'cs-CZ':
-              return import('../translations/cs.js');
-            case 'de-AT':
-            case 'de-DE':
-              return import('../translations/de.js');
-            case 'en-AU':
-            case 'en-GB':
-            case 'en-PH':
-            case 'en-US':
-              return import('../translations/en.js');
```
armujahid: Also note that this isn't exact refactoring because currently multiple locales are targeting same file in some cases.

LarsDenBakker:
Note that rollup doesn't support this out of the box, rollup users will need to add special logic to do magic imports like this.

armujahid:
Yes I think 
- `babel-plugin-syntax-dynamic-import` is required for `import()` function support
- `es-module-shims` is required for dynamic import expressions

We cannot assume that everyone is using these (although I think we should force people to use latest tooling :P ) `locale` variable can be sanitized so that we can eliminate `../` or any other dangerous string.

LarsDenBakker:
You don't want to evaluate the expression at runtime, then a build tool cannot do any optimizations and you cannot compile the es modules for older browsers. 

You want the imports to become part of the regular module graph at compile time, but in order to do that a build tool needs to be able to statically analyze your code and know what you're going to import. Using expressions makes this more complicated, and doesn't work out of the box with some tools. That's what I mean.

armujahid:
Yes I agree.

LarsDenBakker:
I didnt mean to close the PR entirely, im not a core maintainer for this project so thats not my decision in the end :). 

You may still be able to do if you really want it for code maintainability, for example by compiling away the expressions before publishing.

tlouisse:
I think we ultimately want this as an 'npm prepublish' script: https://github.com/ing-bank/lion/blob/master/scripts/npm-prepublish.js

Here is a setup of how this could be done via Babel:
https://astexplorer.net/#/gist/cda9cd6974c8c60d87d1c2eb7a61757b/5fc17975819b256f8f93f291926830c5956a30e9
(note that the outputted 'codeToReplace' should be replaced with the string that is currently logged to the console. In a nodejs context this string could be parsed with `@babel/parser`).

Ideally we should also have some tests for this and make it 100% bullet proof, so that it cannot create corrupt code and halts the publishing process when unexpected code is found.

daKmoR:
we ultimately went on putting the full list "everywhere" as tools like unpkg.com would not work otherwise 😅

closing this as we sadly will not be able to proceed with it... 🙈

Issue #607, opened 2020-02-24
[lion-input-date, lion-input-datepicker] serializedValue's day is always 1 day earlier


T-Motion:
Value returned from lion-input-date and lion-input-datepicker is always 1 day earlier than the date that was filled in.

1. Enter "01/02/2020" into a lion-input-date.
2. Get serializedValue from the input: { date: "2020-01-31" }
 
OS: Windows 10
Browser: Chrome 79.0.3945.130
lion-input-date version: 0.5.13
lion-input-datepicker version: 0.8.17

PR #608, opened 2020-02-25
fix(input-date): use timezone offset in serializedValue

T-Motion:
fix bug #607

### DIFF
--- a/packages/input-date/test/lion-input-date.test.js
+++ b/packages/input-date/test/lion-input-date.test.js
@@ -141,7 +141,7 @@ describe('<lion-input-date>', () => {
   it('serializes to iso format', async () => {
     const el = await fixture(
       html`
-        <lion-input-date .modelValue="${new Date('2000-12-15T00:00:00')}"></lion-input-date>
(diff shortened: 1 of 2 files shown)


packages/input-date/test/lion-input-date.test.js
```diff
@@ -141,9 +141,9 @@ describe('<lion-input-date>', () => {
   it('serializes to iso format', async () => {
     const el = await fixture(
       html`
-        <lion-input-date .modelValue="${new Date('2000-12-12')}"></lion-input-date>
+        <lion-input-date .modelValue="${new Date('2000-12-15T00:00:00')}"></lion-input-date>
```
daKmoR:
```suggestion
        <lion-input-date .modelValue="${new Date('2000/12/15')}"></lion-input-date>
```

I think that is better-supported cross-browser... can we test if that works as well?

T-Motion: Also works, updated in new commit.

jorenbroekema:

PR #611, opened 2020-02-26
chore(switch): add docu for checked-changed handler

Hzunax:
1. Added docs for checked-changed handler
2. Fixed error showing when switching from Canvas to Docs (or vice versa) for the Validation story of the Switch:

Screenshot 2020-02-26 at 16 08 58


packages/switch/stories/index.stories.mdx
```diff
@@ -81,7 +81,8 @@ Simple example that illustrates where validation feedback will be displayed.
         return [...super.validationTypes, 'info'];
       }
     }
-    customElements.define('custom-switch', CustomSwitch);
+    if (!customElements.get('custom-switch'))
```
jorenbroekema: I guess the linter doesn't lint javascript inside mdx but we try to use { } for if statement bodies even if they are single statements and can do without it. Mostly for readability.

And usually I would say this guard is a bad practice because not getting the error when a 2nd custom-switch is trying to define is potentially more harmful than getting the error, but for a custom-switch inside a demo, purely for demo purposes, I think it's fine :).

Hzunax: Thanks for the feedback! I added the `{ }` for the if statement.

daKmoR:
very nice 💪  thaannkk you 🤗

PR #618, opened 2020-03-01
fix(input-amount): normalize added for currency label

demeteren:
I created a method that would normalize currency label. It fixes a Turkish currency problem, that would show TRY for locale TR which is unknown for Turkey. It is an override method.


demeteren:
@gerjanvangeest hi, not sure if you recieve notification after a pull request is made. CI changed since my last contribution. Not sure about this one :)
Thanks

packages/input-amount/src/formatters.js
```diff
@@ -23,3 +23,10 @@ export function formatAmount(modelValue, givenOptions) {
 
   return formatNumber(modelValue, options);
 }
+
+export function normalizeCurrencyLabel(currency, locale) {
```
tlouisse: Good fix. Maybe it's an idea to put this function in the `localize` package as a file inside `src/number`? There we keep all corrections on top of `Intl.Numberformat`

packages/input-amount/src/LionInputAmount.js
```diff
@@ -127,4 +128,12 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
     // sufficient, it should also contain the abbreviation.
     this._currencyDisplayNode.setAttribute('aria-label', getCurrencyName(this.currency));
   }
+
+  __CurrencyLabel() {
```
tlouisse: Maybe make this `get __currencyLabel() `?

demeteren: Thanks @tlouisse . Renamed the method name.

packages/input-amount/src/formatters.js
demeteren: @tlouisse Thanks, exactly did that. Is it good now ?

packages/input-amount/src/LionInputAmount.js
```diff
@@ -127,4 +128,12 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
     // sufficient, it should also contain the abbreviation.
     this._currencyDisplayNode.setAttribute('aria-label', getCurrencyName(this.currency));
   }
+
+  __getCurrencyLabel() {
```
jorenbroekema:
```suggestion
  get __currencyLabel() {
```

I think this is what @tlouisse  meant

jorenbroekema:
Thanks for your contribution 🤗

Issue #432, opened 2019-12-06
[localize] formatNumber for en-AU is broken with currency sign


daKmoR:
## Describe the bug

these tests fail 

```js
expect(formatNumber(123456.789, currencySymbol('EUR'))).to.equal('€123,456.79');
expect(formatNumber(123456.789, currencySymbol('USD'))).to.equal('$123,456.79');
expect(formatNumber(123456.789, currencySymbol('JPY'))).to.equal('¥123,457');
```

Guddu327:
Hi..!! I am new to open source and I would like to work on this, if no one else is currently working on this..

Guddu327:
I tried looking into the issue and found out the formatted number returned by `Intl.NumberFormat`, inside the function `formatNumberToParts`, is inconsistent for the mentioned test cases of `en-AU` locale, as compared to similar test cases used with other locales. I'm unsure of how to proceed with this now, since `Intl` is a web api.

daKmoR:
hey, there and welcome to open source 🤗 

web apis are made by people too and mistakes happen 🙈 

that's why we have localize so we are able to normalize browser bugs/inconsistencies...
we already have multiple such "fixes":
- https://github.com/ing-bank/lion/blob/master/packages/localize/src/number/forceCurrencyToEnd.js
- https://github.com/ing-bank/lion/blob/master/packages/localize/src/number/forceAddGroupSeparators.js

basically every file in that starts with `force` is a workaround for some "bad" responses from `Intl`.

So best would be to add these tests, let them fail and then start adding a `forceCurrencyForEnAu` or something in the likes.

does that sound like a plan?

Guddu327:
Hey thank you for the suggestions.
I was able to add some changes and added some new scripts which modifies the `normalizeInt` script.

I am having some difficulties in adding the pull requests. I have done till the last step by following `Contribution.md`, but while adding the pull request, it's showing 'There isn’t anything to compare.'.

daKmoR:
fixed via https://github.com/ing-bank/lion/pull/619

PR #619, opened 2020-03-02
fix: force symbols for locale 'en-AU'

Guddu327:
See the issue for details on the symbols fixed for the particular locale 'en-AU'

fixes issue #432

### DIFF
--- a/packages/localize/src/number/forceENAUSymbols.js
+++ b/packages/localize/src/number/forceENAUSymbols.js
@@ -1,18 +1,21 @@
 export function forceENAUSymbols(formattedParts, options) {
   const result = formattedParts;
   const numberOfParts = result.length;
-  // Change the symbol from JPY to ¥, due to bug in Chrome
+  // Change the symbols for locale 'en-AU', due to bug in Chrome
   if (numberOfParts > 1 && options && options.currencyDisplay === 'symbol') {
-    if (options.currency === 'EUR') {
-      result[0].value = '€';
-      result[1].value = '';
-    } else if (options.currency === 'USD') {
-      result[0].value = '$';
-      result[1].value = '';
-    } else if (options.currency === 'JPY') {
-      result[0].value = '¥';
-      result[1].value = '';
+    switch (options.currency) {
+      case 'EUR':
+        result[0].value = '€';
+        break;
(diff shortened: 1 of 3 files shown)


packages/localize/src/number/forceENAUSymbols.js
```diff
@@ -0,0 +1,18 @@
+export function forceENAUSymbols(formattedParts, options) {
+  const result = formattedParts;
+  const numberOfParts = result.length;
+  // Change the symbol from JPY to ¥, due to bug in Chrome
```
jorenbroekema: Can you explain this comment? It's a little confusing to me because it sounds like this is a bug unrelated to what this function solves (issue: unsupported en-AU locale).

packages/localize/src/number/forceENAUSymbols.js
```diff
@@ -0,0 +1,18 @@
...
+    if (options.currency === 'EUR') {
+      result[0].value = '€';
+      result[1].value = '';
+    } else if (options.currency === 'USD') {
+      result[0].value = '$';
+      result[1].value = '';
+    } else if (options.currency === 'JPY') {
+      result[0].value = '¥';
+      result[1].value = '';
+    }
```
jorenbroekema:
```suggestion
    switch(options.currency) {
      case 'EUR':
        result[0].value = '€';
        break;
      case 'USD':
        result[0].value = '$';
        break;
      case 'JPY':
        result[0].value = '¥';
        break;
        /* no default */
    }
    result[1].value= '';
```

Guddu327:
Hey, thanks for the suggestions, I was able to fix the nested if (with the switch) and changed the comment on the file. Kindly let me know if the fixes are okay.

packages/localize/src/number/forceENAUSymbols.js
Guddu327: Hey, thanks for the suggestions, I was able to fix the nested if (with the switch) and changed the comment on the file. Kindly let me know if the fixes are okay.

jorenbroekema:
Thanks for the contribution 🤗! It's merged to dev for now (we're bundling a bunch of PRs and will release later this week to reduce the amount of releases)

Issue #580, opened 2020-02-14
Consistent naming convention: changedProps vs changedProperties


amielucha:
It would be beneficial for consistency and ease of comparison between components' code, to refactor the update methods to use the same variable naming convention for properties.

- `changedProps` is being used in 7 files
- `changedProperties` is being used in 11 files

Is there a strong preference towards one or the other?

daKmoR:
`changedProperties` seems like a more descriptive name 👍

jorenbroekema:
We also use `c` in some places I think so you may want to search for `update(c` and maybe even , for me both proposed names are fine. Also maybe check `firstUpdated(c` , same thing in that lifecycle method

PR #620, opened 2020-03-02
chore: unify the naming of changedProperties param

amielucha:
Renaming `changedProps` and `c` to `changedProperties` as per #580


jorenbroekema:

PR #623, opened 2020-03-03
reduce package count: move radio to radio-group and checkbox to checkbox-group

CubLion:


daKmoR:
has been merged via https://github.com/ing-bank/lion/pull/625

Issue #631, opened 2020-03-09
[dialog] close-overlay doesn't work from web component


T-Motion:
## Describe the bug

When you create a custom web component and load this in the content slot of a lion-dialog, the close-overlay event does not work. When you dispatch this, the dialog is not closed.

Dialog:
```
<lion-dialog .config=${{ hidesOnOutsideClick: true, hidesOnEsc: true }}>
    <button slot="invoker">Styled dialog</button>
    <styled-dialog-content slot="content"></styled-dialog-content>
</lion-dialog>
```

styled-dialog-content:
```
render() {
    return html`
      <div><p>First part nice looking dialog</p></div>
      <div><p>Look how nice this <span class="nice">dialog</span> looks!</p><p>This is also content.</p></div>
      <button
        class="close-button"
        @click=${e => e.target.dispatchEvent(new Event('close-overlay', { bubbles: true }))}
      >⨯
      </button>
    `;
  }
```

jorenbroekema:
I assume this is because of the shadow boundary. There's two solutions:

1. Set the option `composed: true` next to the option `bubbles`, so it bubbles past shadow boundaries. A brute force solution that usually we would discourage.
3. Render to LightDOM with `createRenderRoot() { return this; }`, probably not what you want since you want encapsulated styling on your content.
3. Change the @click to not dispatch the event on the `<button>` (through `e.target.dispatchEvent`), but to dispatch it on the host element instead (which is not in the ShadowDOM) `this.dispatchEvent`. What you could do is set the event target to your `<button>` if you'd like to, but for me it makes sense that, since you're explicitly redispatching, the target should be your Web Component, which it is by default.

Let me know if that helps ;)!

jorenbroekema:
Also, if you go for option 3 and it works perfectly, would maybe be nice to add a demo for it since using a web component for the content slot will be VERY common, because it will be the default way people will style content in their dialogs in an encapsulated way.

jorenbroekema:
Now documented on how to get passed this :)

PR #633, opened 2020-03-10
chore(dialog): improve styled content docs

T-Motion:
Improvement of documentation for styled content in a lion-dialog.
#631


packages/dialog/stories/styled-dialog-content.js
```diff
@@ -0,0 +1,38 @@
+import { LitElement, html, css } from '../../core/index.js';
```
daKmoR:
```suggestion
import { LitElement, html, css } from '@lion/core';
```

packages/dialog/stories/index.stories.mdx
```diff
@@ -76,6 +77,64 @@ The documentation of the full config object can be found in the `lion/overlay` p
...
+  static get styles() {
+    return [
+      css`
+        :host {
+          background-color: #fff;
+        }
+        .nice {
+          font-weight: bold;
+          color: green;
+        }
```
jorenbroekema: .close-button styling is missing here in the code snippet

packages/dialog/stories/index.stories.mdx
```diff
@@ -76,6 +77,64 @@ The documentation of the full config object can be found in the `lion/overlay` p
 
 The `config` property uses a setter to merge the passed configuration with the current, so you only **overwrite what you pass** when updating `config`.
 
+### Styling content
+
+Because the content of a dialog is slotted, it can only be styled one level deep. If you need more, you should create a custom component that passes the styles along with the html.
```
jorenbroekema: It's actually impossible to style it from the `dialog` component class, because the content slot is moved to the global root node. This is why you should create a custom component and slot it in as the content, this ensures style encapsulation on your dialog content.

packages/dialog/stories/styled-dialog-content.js
```diff
@@ -0,0 +1,38 @@
...
+          color: black;
+          font-size: 28px;
+          line-height: 28px;
+        }
+      `,
+    ];
+  }
+
+  _closeOverlay() {
+    this.dispatchEvent(new Event('close-overlay', { bubbles: true }));
```
jorenbroekema: I think this is a great way of solving the issue you had, which is that the close-overlay does not cross the shadow boundary of your styled component, which means you cannot close your dialog. So then you make it a method on your styled component and (re)dispatch the close-overlay event from the component rather than the <button> in your template and putting `composed: true` which I would consider the "bad" solution. 

I think this by itself deserves documentation as well :) because otherwise lots of people would do this incorrectly I think.

T-Motion: I added this documentation in a new commit. I'm using the same example though.

jorenbroekema:
Thanks for the docs improvement :D!

Issue #569, opened 2020-02-13
[lion-radio-group] can't get keyboard focus when untouched/not prefilled


gerjanvangeest:
## Describe the bug

When a `radio-group` is untouched/ not prefilled it can't receive keyboard focus.

## To Reproduce

Try to focus via keyboard the `radio-group` in the first story:

https://lion-web-components.netlify.com/?path=/docs/forms-radio-group--default-story

Once an option is checked it can receive focus.

## Additional context

It looks like this behaviour appears since the latest `modelValue/name` refactoring

tlouisse:
I think this s not a bug, but a flaw in our docs (after we moved to Storybook docs mode). Since all demos have the name `dinos`, the platform combines them into one single group.
The second demo has the checked value and therefore interacts with tab.

See a small simulation here: https://jsfiddle.net/pqt2hdam/2/ 
Tabstop 2 can only be reached once you set a checked value there.

**Solution**
Give all radio groups in our demos a unique name.

PR #636, opened 2020-03-12
chore(radio-group): fix demo tab selection due to duplicate name

AymenBenAmor:
fix issue with tab for documentation. fixes https://github.com/ing-bank/lion/issues/569


packages/radio-group/stories/radio-group.stories.mdx
```diff
@@ -21,7 +21,7 @@ You should use `<lion-radio>`s inside this element.
 </Story>
 
 ```html
-<lion-radio-group name="dinos" label="What are your favourite dinosaurs?">
```
jorenbroekema: This is just the code snippet that the user can copy right? So here, I wouldn't change it. I would only change it in the <Story> sections.

packages/radio-group/stories/radio-group.stories.mdx
```diff
@@ -176,8 +176,8 @@ const validate = () => {
 
 ```html
 <lion-radio-group
-  id="dinos"
-  name="dinos"
+  id="dinosGroup"
```
jorenbroekema: Reason for this change? We kinda changed the way we handle `name` attributes and in most places I think we changes `{x}Group` to just `{x}s`

AymenBenAmor: I saw that we were using document.querySelector('#dinosGroup'); in line 172, so i changed the id and since i had to change the name to something in order to fix this issue I kept the name same as id

packages/radio-group/stories/radio-group.stories.mdx
AymenBenAmor: For me, we should keep the same. As there is an understanding of the reader that the snippet of code will result in exactly what he sees. But i can change it if you think it's better for readability.

jorenbroekema: I think for most developers they will just (blindly) copy paste the code to get a working radio group and then change the name to their own context so I guess it doesn't matter either way. I see your point that it's more 1 to 1 to keep it like this 👍 .

packages/radio-group/stories/radio-group.stories.mdx
jorenbroekema: Ah fair enough, fine with me then, as long as it works

jorenbroekema:
There's a merge conflict, may have to rebase and solve it ;)

PR #639, opened 2020-03-16
fix(overlays): support backdrop with local overlay

maxevilmind:
I think what is appropriate to do in the case when we want to pass controls over styling and animation to the developer is to avoid all the style and animation manipulations in OverlayController.
Only dispatch events for 'init' 'show' 'hide' and 'teardown' and pass backdropNode as a parameter for an overlay controller.


jorenbroekema:
Bit hard to find the demo in storybook 🙈  http://localhost:9001/?path=/docs/overlays-system-overview--local-overlay at the bottom in case anyone wants to look

maxevilmind:
@tlouisse please take a look and leave your comments

daKmoR:
thank you for your contribution 🤗

maxevilmind:
always glad to contribute here 😇

PR #641, opened 2020-03-16
fix(localization): for tr currencycode must be on end

demeteren:
It is uncommon for Turkish people to use currency code at the start of a number. This commit fixes that.


demeteren:
Hello @jorenbroekema ,

Could you review this commit as well ? Thanks :)

jorenbroekema:
Hey, do you have a source or something that shows currency ISO codes should be forced to the end for Turkish locale? I think in general for overrides of Intl / default formatting it would be a nice idea to put a source inside the code as a comment in case people check the source code and wonder why a certain override is done.

jorenbroekema:
@demeteren Hey, did you see my message?

PR #643, opened 2020-03-18
chore: small storybook updates

gerjanvangeest:


packages/form-system/stories/17-Validation-Examples.stories.mdx
```diff
@@ -421,9 +415,11 @@ class MyValidator extends Validator {
 ></lion-input>
 ```
 
-## Override default messages
+## Default messages
+
+To get default validation messages you need to import and call the `loadDefaultFeedbackMessages` function once in your application.
 
-Oftern
+Sometimes the default messages doesn't make sense for your input field. In that case you want to override it by adding a `getMessage` function to your validator.
```
jorenbroekema: `messages don't make sense`
```suggestion
Sometimes the default messages don't make sense for your input field. In that case you want to override it by adding a `getMessage` function to your validator.
```

packages/form-system/stories/17-Validation-Examples.stories.mdx
```diff
@@ -469,6 +465,8 @@ Oftern
 
 ## Override fieldName
 
+In other cases the default messages are correct, but you only want to change the `fieldName`. This can be done for only one specific validator or for all at once.
```
jorenbroekema:
```suggestion
In the scenario that the default messages are correct, but you only want to change the `fieldName`, this can both be done for a single specific validator or for all at once.
```

PR #644, opened 2020-03-18
feat(fieldset): expose filter on serializedValue

marvin89:
Hi,

I noticed that there is a default filter of non-disabled elements when returning the serialized value, one of my use cases is that I would need to also filter by non-readonly elements. For this purpose I propose the exposure of the filter under a new method `filterSerializedValue(filter)`  to allow more flexibility in handling serializedValue:

```js
form.filterSerializedValue(el => !el.disabled && !el.readOnly);
```

### DIFF
--- a/packages/fieldset/src/FormGroupMixin.js
+++ b/packages/fieldset/src/FormGroupMixin.js
@@ -203,6 +203,13 @@ export const FormGroupMixin = dedupeMixin(
       }
 
       /**
+       * @param {Function} filter Filter for form elements (default: el => !el.disabled)
+       */
+      filterSerializedValue(filter) {
+        return this._getFromAllFormElements('serializedValue', filter);
+      }
+
+      /**
        * @desc Handles interaction state 'submitted'.
        * This allows children to enable visibility of validation feedback
        */
--- a/packages/fieldset/test/lion-fieldset.test.js
+++ b/packages/fieldset/test/lion-fieldset.test.js
@@ -757,6 +757,38 @@ describe('<lion-fieldset>', () => {
       });
     });
 
+    it('does not serialize disabled values when no filter is given to `filterSerializedValue`', async () => {
+      const fieldset = await fixture(html`
+        <${tag}>
+          <${childTag} name="custom[]"></${childTag}>
+          <${childTag} name="custom[]"></${childTag}>
+        </${tag}>
+      `);
+      await nextFrame();
+      fieldset.formElements['custom[]'][0].modelValue = 'custom 1';
+      fieldset.formElements['custom[]'][1].disabled = true;
+
+      expect(fieldset.filterSerializedValue()).to.deep.equal({
+        'custom[]': ['custom 1'],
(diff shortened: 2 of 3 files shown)


jorenbroekema:
Hi! Can I ask what your use case is? Are you extending lion-fieldset or some other lion component? 

Because the filter feature is there right now, though only exposed as a protected method (_getFromAllFormElements), the reason is that we only see a use case for "subclassers" (people extending our components into their own) to be able to override. We don't have a use case for people overriding it on the LionFieldset as a public method, mostly because we don't expect people to use use our components inside their templates directly.

If you are extending, I would suggest something like:

```js
class WolfFieldset extends LionFieldset { 
  /**
   * @override LionFieldset
   */
  _getFromAllFormElements(property, filterCondition = el => !el.disabled && !el.readOnly) {
    super._getFromAllFormElements(property, filterCondition);
  }
}
```
Then if you query `myFieldset.serializedValue` it should filter out the read-only fields

Or if you're extending form:
```js
class WolfForm extends LionForm { 
  /**
   * @override LionFieldset
   */
  _getFromAllFormElements(property, filterCondition = el => !el.disabled && !el.readOnly) {
    super._getFromAllFormElements(property, filterCondition);
  }
}
```
it should work the same way I believe. 

If you are extending many of these fieldset extensions (e.g. radio-group and checkbox-group as well), and you are perhaps also doing other overrides of our protected methods, you could also make it a mixin to avoid code duplication and have it nicely packaged.

```js
import { dedupeMixin } from '@lion/core';

export const WolfFieldsetMixin = dedupeMixin(
  superclass =>
    // eslint-disable-next-line no-unused-vars, no-shadow
    class WolfFieldsetMixin extends superclass {
      /**
       * @override LionFieldset
       */
      _getFromAllFormElements(property, filterCondition = el => !el.disabled && !el.readOnly) {
        super._getFromAllFormElements(property, filterCondition);
      }
```
And apply it:
```js
class WolfForm extends WolfFieldsetMixin(LionForm) { 
 
}
```

Hopefully this helps, again, let us know what your use case is so we can see if a publicly exposed method as you have proposed makes sense 🤗

jorenbroekema:
Been 17 days and we won't merge this as is I think. So I will close this. If you wanna further discuss we are always open to it, please create an issue for it :)!

marvin89:
Sorry, totally missed your initial comment.
My actual use case is that I am using ` ing-web`, the `ing-form` component to be more specific. 

Extending is an option, but I prefer to leave it as a last option, especially when there are enough layers already present.

If you think the filter method doesn't provide enough value of does not fit enough use cases for other users, I can also make the proposal in `ing-web`.

jorenbroekema:
Yeah I think it's up to an extension layer to provide a default filter (could be different from lion), and also up the extension layer to allow their users to have the flexibility to pass their own, so far we haven't had a use case in our ING extension layer, so it would be a good idea to create an issue for it there to see what we can do :)

Issue #645, opened 2020-03-18
[lion-input] State attribute filled only update in onblur event


carlosesteban55:
## Describe the bug

lion input don´t update the state "filled" attribute when the user is typing . Only when user throws onblur event.

https://github.com/ing-bank/lion/blob/master/packages/field/src/LionField.js#L187 this is the line that toggles the filled attribute and onValueChanged gets triggered by the value setter which gets triggered by FormatMixin.. so in FormatMixin may lie the root cause

when the user is typing (focused inside the input), the formatted value is not reflected back to the user yet, so the value setter isn't called, as a consequence, the filled attribute is only toggled when the formatted value is reflected back to the user.

> _reflectBackFormattedValueToUser() {
>   if (this._reflectBackOn()) {
>     // Text 'undefined' should not end up in <input>
>     this.value = typeof this.formattedValue !== 'undefined' ? this.formattedValue : '';
>   }
> }
> ​
> _reflectBackOn() {
>   return !this.__isHandlingUserInput;
> }

## To Reproduce

1. type on the lion input. If you have the filled attribute active then delete all characters and see that filled attribute don´t update

2. If the input is empty, type some characters and see that not update filled attribute.

## Expected behavior

Filled attribute updates during user types.

## Screenshots

![image](https://user-images.githubusercontent.com/49064780/76963964-2cd23680-6922-11ea-80aa-1f5ef2021284.png)

jorenbroekema:
Yeah can confirm, flow is indeed:
- modelValue is formatted (only when user is not handling the input, e.g. on blur)
- formatted value changed and therefore has to be reflected back to the user
- value setter is invoked by setting `this.value`
- value setter in LionField calls the method that toggles the `filled` attribute

I have opened a PR to change this behavior to be dependent on the modelValue (and onModelValueChanged event), this will ensure `filled` is toggled also when the model value is non-empty during user input, but hasn't been reflected back as a formatted version to the user.

jorenbroekema:
https://github.com/ing-bank/lion/pull/646 fixed here

PR #646, opened 2020-03-18
Field isEmpty / filled / prefilled aligning

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/645

### DIFF
--- a/packages/field/src/LionField.js
+++ b/packages/field/src/LionField.js
@@ -1,10 +1,10 @@
-import { SlotMixin, LitElement } from '@lion/core';
+import { LitElement, SlotMixin } from '@lion/core';
 import { DisabledMixin } from '@lion/core/src/DisabledMixin.js';
 import { ValidateMixin } from '@lion/validate';
+import { FocusMixin } from './FocusMixin.js';
+import { FormatMixin } from './FormatMixin.js';
 import { FormControlMixin } from './FormControlMixin.js';
 import { InteractionStateMixin } from './InteractionStateMixin.js'; // applies FocusMixin
-import { FormatMixin } from './FormatMixin.js';
-import { FocusMixin } from './FocusMixin.js';
 
 /* eslint-disable wc/guard-super-call */
 
@@ -80,7 +80,6 @@ export class LionField extends FormControlMixin(
     if (this._inputNode) {
       this._setValueAndPreserveCaret(value);
     }
-    this._onValueChanged({ value });
   }
 
   get value() {
@@ -107,7 +106,6 @@ export class LionField extends FormControlMixin(
     // FormatMixin
     this._delegateInitialValueAttr();
     super.connectedCallback();
-
     this._onChange = this._onChange.bind(this);
     this._inputNode.addEventListener('change', this._onChange);
     this.classList.add('form-field'); // eslint-disable-line
@@ -176,16 +174,6 @@ export class LionField extends FormControlMixin(
         bubbles: true,
       }),
     );
-  }
-
-  _onValueChanged({ value }) {
-    if (super._onValueChanged) {
-      super._onValueChanged();
-    }
-    // For styling purposes, make it known the input field is not empty
-    if (this._inputNode) {
-      this[value ? 'setAttribute' : 'removeAttribute']('filled', '');
-    }
   }
 
   /**
--- a/packages/field/test-suites/InteractionStateMixin.suite.js
+++ b/packages/field/test-suites/InteractionStateMixin.suite.js
@@ -31,9 +31,8 @@ export function runInteractionStateMixinSuite(customConfig) {
 
             set modelValue(v) {
               this._modelValue = v;
-              this.dispatchEvent(
-                new CustomEvent('model-value-changed', { bubbles: true, composed: true }),
-              );
+              this.dispatchEvent(new CustomEvent('model-value-changed', { bubbles: true }));
+              this.requestUpdate('modelValue');
             }
 
             get modelValue() {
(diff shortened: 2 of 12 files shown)


jorenbroekema:
https://github.com/ing-bank/lion/issues/645 this issue is related as well, I proposed a fix in the issue where we also use modelValue as a reference point instead of viewValue (value). 

@tlouisse  I'm just gonna assign you, since I think you have worked on this the most recently right?

packages/field/src/LionField.js
```diff
@@ -178,14 +177,12 @@ export class LionField extends FormControlMixin(
...
-      super._onValueChanged();
+  _onModelValueChanged({ modelValue }) {
+    if (super._onModelValueChanged) {
+      super._onModelValueChanged();
     }
     // For styling purposes, make it known the input field is not empty
-    if (this._inputNode) {
-      this[value ? 'setAttribute' : 'removeAttribute']('filled', '');
-    }
+    this[modelValue !== '' ? 'setAttribute' : 'removeAttribute']('filled', '');
```
tlouisse: Totally agree we should base derived state always on. modelValue, since this is our internal SSOT.

This might actually be cleaner (and better for js engine optimalization(v8 etc)) to do it in a regular if/else.

Or we could maybe just make it a property with `reflect:true`?

tlouisse: We could also consider putting this logic in interactionState.

Also, ideally an empty modelValue should be `undefined` instead of ''. (I think there was a mr for this a while ago, but maybe we never followed through)

tlouisse:
In general: totally agree we should base derived state always on. modelValue, since this is our internal SSOT 👍

packages/field/src/LionField.js
jorenbroekema: Empty string would still parse to a modelValue -> '' right? It doesn't seem right to me that a parser ever returns undefined, unless the literal input you give it is JavaScript's primitive: `undefined`. `null` would be more correct than `undefined` in my opinion, but I consider `null` to also not be as appropriate as empty string.

And yes I think it makes sense to make it a property with reflect: true and putting it inside InteractionStateMixin :) I can do that

tlouisse: Ah, didn't see your response. I changed my initial reaction a bit in the meantime... :)
'' assumes modelValue is a String, which we can't do.

I was also in favor of `null` before, but it was a team decision to go for `undefined` in the end. The nice thing is that it is similar to an initally empty value.

jorenbroekema: Alright makes sense, and I agree with "The nicest solution", and then filled is only an attribute, just as a public styling hook. `_isEmpty` should be used for logic by subclassers, and it is used to set the filled styling hook by doing in InteractionStateMixin on model value changed:
```js
if (this._isEmpty) {
  this.removeAttribute('filled');
} else {
  this.setAttribute('filled');
}
``` 
Seems ok?

The reason I don't think we need filled to be a lit property is because 1) it's not a complex value so attr is fine and 2) we won't need to re-render the component since it's just meant as a styling hook (using css attr selector) which doesn't require a re-render.

tlouisse: Cool that we agree 😎
About the lit prop: It won't trigger a rerender of the lit template, but the attribute is used as a styling hook for the outside world. Whenever there is styling for `[filled]`, a new paint will be triggered by the browser. It's a better practice to batch all those attrs in a batch I think (like LitElement does for you).
On top of that, I think it's nice to have an overview of all InteractionState properties that, when our api tables are in place, will automatically be documented.

jorenbroekema: Ah I will make it a lit property. I never considered that making a style hook attr a lit property gives the benefit of ending up in a batched paint, that's a really cool feature (maybe side effect) of lit actually :O

jorenbroekema:
If I remove the scope from the commit message, will the commit show up in all the published packages' release notes? Because in that case, we should remove "(field)" from commit msg, otherwise it only shows in the field changelog and this also affects validate

packages/field/test-suites/InteractionStateMixin.suite.js
```diff
@@ -34,6 +34,7 @@ export function runInteractionStateMixinSuite(customConfig) {
               this.dispatchEvent(
                 new CustomEvent('model-value-changed', { bubbles: true, composed: true }),
```
tlouisse: It;s just a test. But without `composed:true` would be the right way now :)

packages/fieldset/src/FormGroupMixin.js
```diff
@@ -392,7 +392,7 @@ export const FormGroupMixin = dedupeMixin(
        */
       removeFormElement(...args) {
         super.removeFormElement(...args);
-        this.validate();
+        this.validate({ clearCurrentResult: true });
```
tlouisse: We should also clear in `addFormElement ` I think

packages/validate/src/ValidateMixin.js
```diff
@@ -479,15 +480,19 @@ export const ValidateMixin = dedupeMixin(
         return JSON.stringify(result) !== JSON.stringify(prevResult);
       }
 
-      __isEmpty(v) {
+      __isEmpty() {
         if (typeof this._isEmpty === 'function') {
-          return this._isEmpty(v);
+          return this._isEmpty();
```
tlouisse: Maybe we should pass the modelValue for backw compatibility?

packages/validate/src/ValidateMixin.js
```diff
@@ -479,15 +480,19 @@ export const ValidateMixin = dedupeMixin(
...
         }
         // // TODO: move to compat layer. Be sure to keep this, because people use this a lot
         // if (typeof this.__isRequired === 'function') {
         //   return !this.__isRequired(v);
         // }
-        return v === null || typeof v === 'undefined' || v === '';
+        return (
```
tlouisse: This code is probably never reached because of the fact that we depend on FormControlMixin and that one defines _isEmpty?

Maybe we can get rid of `__isEmpty` and call `_isEmpty` directly inside `validate()`?

packages/validate/src/ValidateMixin.js
jorenbroekema: Done 👍

packages/fieldset/src/FormGroupMixin.js
jorenbroekema: done :)

packages/field/test-suites/InteractionStateMixin.suite.js
jorenbroekema: done :)

packages/validate/src/ValidateMixin.js
jorenbroekema: Added a TODO for removing this by the time that we have a form package, so we can more easily rely on that (currently implicit) dependency

PR #647, opened 2020-03-18
Fix/model value event consistent

tlouisse:
Replacement for https://github.com/ing-bank/lion/pull/602

### DIFF
--- a/packages/choice-input/src/ChoiceGroupMixin.js
+++ b/packages/choice-input/src/ChoiceGroupMixin.js
@@ -55,20 +55,7 @@ export const ChoiceGroupMixin = dedupeMixin(
       constructor() {
         super();
         this.multipleChoice = false;
-      }
-
-      connectedCallback() {
-        super.connectedCallback();
-        if (!this.multipleChoice) {
-          this.addEventListener('model-value-changed', this._checkSingleChoiceElements);
-        }
-      }
-
-      disconnectedCallback() {
-        super.disconnectedCallback();
-        if (!this.multipleChoice) {
-          this.removeEventListener('model-value-changed', this._checkSingleChoiceElements);
-        }
+        this._repropagationRole = 'choice-group'; // configures event propagation logic of FormControlMixin
       }
 
       /**
@@ -157,9 +144,10 @@ export const ChoiceGroupMixin = dedupeMixin(
         }
       }
 
-      __triggerCheckedValueChanged() {
+      __setChoiceGroupTouched() {
         const value = this.modelValue;
         if (value != null && value !== this.__previousCheckedValue) {
+          // TODO: what happens here exactly? Needs to be based on user interaction (?)
           this.touched = true;
           this.__previousCheckedValue = value;
         }
@@ -178,6 +166,25 @@ export const ChoiceGroupMixin = dedupeMixin(
             }" given)`,
           );
         }
+      }
+
+      /**
+       * @override FormControlMixin
+       */
+      _onBeforeRepropagateChildrenValues(ev) {
+        // Normalize target, since we might receive 'portal events' (from children in a modal,
+        // see select-rich)
+        const target = (ev.detail && ev.detail.element) || ev.target;
+        if (this.multipleChoice || !target.checked) {
+          return;
+        }
+        this.formElements.forEach(option => {
+          if (target.choiceValue !== option.choiceValue) {
+            option.checked = false; // eslint-disable-line no-param-reassign
+          }
+        });
+        this.__setChoiceGroupTouched();
+        this.requestUpdate('modelValue');
       }
     },
 );
--- a/packages/choice-input/test/ChoiceGroupMixin.test.js
+++ b/packages/choice-input/test/ChoiceGroupMixin.test.js
@@ -2,7 +2,8 @@ import { html, LitElement } from '@lion/core';
 import { FormGroupMixin } from '@lion/fieldset';
 import { LionInput } from '@lion/input';
 import { Required } from '@lion/validate';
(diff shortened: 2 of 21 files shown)


packages/choice-input/test/ChoiceGroupMixin.test.js
```diff
@@ -2,7 +2,8 @@ import { html, LitElement } from '@lion/core';
 import { FormGroupMixin } from '@lion/fieldset';
 import { LionInput } from '@lion/input';
 import { Required } from '@lion/validate';
-import { expect, fixture, nextFrame } from '@open-wc/testing';
+import { expect, nextFrame } from '@open-wc/testing';
+import { formFixture as fixture } from '@lion/field/test-helpers/formFixture.js';
```
daKmoR: can we do?
```suggestion
import { formFixture as fixture } from '@lion/field/test-helpers.js';
```

daKmoR: e.g. export all test helper from a "root" entry point

packages/choice-input/src/ChoiceGroupMixin.js
```diff
@@ -179,5 +167,22 @@ export const ChoiceGroupMixin = dedupeMixin(
           );
         }
       }
+
+      /**
+       * @override FormGroupMixin
+       */
```
jorenbroekema: This was moved from the RadioGroup? to ensure that only a single radio can get checked at once?

packages/field/src/FormControlMixin.js
```diff
@@ -55,6 +55,11 @@ export const FormControlMixin = dedupeMixin(
            * Contains all elements that should end up in aria-describedby of `._inputNode`
            */
           _ariaDescribedNodes: Array,
+          /**
+           * Based on the role, details of handling model-value-changed repropagation
+           * @type {'child'|'form-group'|'choice-group'}
```
jorenbroekema: form-group? why not fieldset?

jorenbroekema: same with child, why not field?

packages/field/src/FormControlMixin.js
```diff
@@ -553,5 +571,96 @@ export const FormControlMixin = dedupeMixin(
...
+      }
+
+      __dispatchInitialModelValueChangedEvent() {
+        // When we are not a form-group / choice-group, we don't need to wait for our children
+        // to send a unified event
+        if (this._repropagationRole === 'child') {
+          return;
+        }
+
+        this.registrationComplete.then(() => {
```
jorenbroekema: Nitpick but it would be clearer perhaps if this was its own async function called something like `async _repropagateAfterChildrenRegistered() {}`

Nice comment btw, helped me understand what this does 👍

packages/field/src/FormControlMixin.js
```diff
@@ -553,5 +571,96 @@ export const FormControlMixin = dedupeMixin(
...
+            }),
+          );
+        });
+      }
+
+      // eslint-disable-next-line class-methods-use-this, no-unused-vars
+      _onBeforeRepropagateChildrenValues(ev) {}
+
+      __repropagateChildrenValues(ev) {
+        // Allows parent classes to internally listen to the children change events
```
jorenbroekema: Shouldn't that be "child classes"?

packages/choice-input/src/ChoiceGroupMixin.js
```diff
@@ -179,5 +167,22 @@ export const ChoiceGroupMixin = dedupeMixin(
           );
         }
       }
+
+      /**
+       * @override FormControlMixin
+       */
+      _onBeforeRepropagateChildrenValues(ev) {
+        const target = (ev.detail && ev.detail.element) || ev.target;
```
jorenbroekema: add comment here as well about portals requiring this normalization?

packages/field/src/FormControlMixin.js
```diff
@@ -553,5 +571,96 @@ export const FormControlMixin = dedupeMixin(
...
+        const target = (ev.detail && ev.detail.element) || ev.target;
+        const isEndpoint =
+          this._isRepropagationEndpoint || this._repropagationRole === 'choice-group';
+
+        // Prevent eternal loops after we sent the event below.
+        if (target === this) {
+          return;
+        }
+
+        // A. Stop sibling handlers
```
jorenbroekema: Well documented 👍  would be very hard to understand otherwise :)

packages/choice-input/src/ChoiceGroupMixin.js
tlouisse: Before it added its own model-value-changed event listener, now it hooks into this callback provided by FormControlMixin (best way to control timing issues)

packages/field/src/FormControlMixin.js
tlouisse: `fieldset` is indeed a better fit 👍 
 `child` because it can either be an option or a field.

packages/field/src/FormControlMixin.js
tlouisse: I already put it in a distinct function, before everything was in firstUpdated :)
Will change to `await this.registrationComplete;`

packages/field/src/FormControlMixin.js
tlouisse: Yep :)

packages/field/src/FormControlMixin.js
tlouisse: Hmm, will be a bit weird if I do, because then The formGroupMixin either wouldn't have a role(would inherit 'child' then) or would have the role `fieldset` (and there is a subtle difference between form-group and fieldset, the latter allows children with different names).

PR #649, opened 2020-03-19
fix(form): remove correct reset event listener on teardown

wsloth:
Spotted this small typo while looking into for LionForm works.

Doesn't cause big issues, but long-term it might allocate too much memory to unused event listeners :)


jorenbroekema:
Haha thanks for this one! 👍

PR #652, opened 2020-03-21
WIP chore(overlays): refactor backdrop code, provide default backdrop styles

jorenbroekema:
I was looking at the demos, and wanted to give an example where you don't pass your own node.

Then I saw we don't give any default styling like we do for the global overlay backdrop, so I implemented that. And then I refactored some of the code in OverlayController to be cleaner / more DRY :)

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -1,4 +1,5 @@
 import '@lion/core/src/differentKeyEventNamesShimIE.js';
+import { localOverlaysStyle } from './localOverlaysStyle.js';
 import { overlays } from './overlays.js';
 import { containFocus } from './utils/contain-focus.js';
 import './utils/typedef.js';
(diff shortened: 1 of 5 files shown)


packages/overlays/src/OverlayController.js
```diff
@@ -423,76 +424,60 @@ export class OverlayController {
...
       case 'init':
-        this.backdropNode = document.createElement('div');
-        this.backdropNode.classList.add('global-overlays__backdrop');
+        // For global overlays, styles for this are added to head by the OverlayManager
+        // For local, we add it locally in DOM here
+        if (this.placementMode === 'local') {
+          const styleTag = document.createElement('style');
+          styleTag.setAttribute('data-local-overlays', '');
+          styleTag.textContent = localOverlaysStyle.cssText;
+          this._contentNodeWrapper.parentElement.insertBefore(styleTag, this._contentNodeWrapper);
```
tlouisse: Can we assume here that we add this in shadow dom? Even if so, not conflicting with your host is not 100% guaruanteed.
If a dev purely uses OverlayController, this would add the same style tag multiple times. (think of a page full with tooltips or popovers).

packages/overlays/src/localOverlaysStyle.js
```diff
@@ -0,0 +1,41 @@
+import { css } from '@lion/core';
```
tlouisse: Would it also be possible to split up between functional (always applied, but overridable) and theme styling?

We do this basically for every component. (In this case the backdrop would be just a webcomponent that you can extend and supply yourself). The backdrop would then also become it's own BEM component `.backdrop` that can be shared for global and local (would also be convenient if you switch from local to overlay based on viewport)

Functional would be:
- positioning and dimensions
- display
- default z-index
- no animjation

Theme styles would involve
- background
- animations can be added in either js or css, up to developer

packages/overlays/src/OverlayController.js
jorenbroekema: I'm fine with going with `<lion-overlays-backdrop></lion-overlays-backdrop>` and encapsulating the styles, but then I think we should do that for both local and global placement backdrops right? And if we do that, it's a somewhat substantial breaking change if people were relying on being able to style the backdrop from outside for modal dialogs for example.

Maybe an intermediate solution before reworking the overlay system to use ShadowDOM more, would be to move this style tag logic inside the `if (!this.backdrop) {} ` block, that way the style tag does not get added if the user passes their own backdrop node, meaning they are then fully able to use their own backdrop without any leftover style tags they don't need/want. Thoughts?

packages/overlays/src/localOverlaysStyle.js
jorenbroekema: I like that, but see my other comment, I think we should tackle this in a separate PR, so we have a `<lion-overlays-backdrop>` for both global and local, and then we can implement this. But we need to think about backwards compatibility with our extension layer, where people probably rely right now on overriding styles of global overlays from the document, because it's not encapsulated in a ShadowDOM at the moment :\

tlouisse: If only we had a tool to find this out... :)

gerjanvangeest:
In storybook the `backdrop` is above all `copy` buttons before the story, but the `copy` buttons after the story are still visible/clickable.

image

packages/overlays/src/OverlayController.js
```diff
@@ -17,6 +18,7 @@ export class OverlayController {
    * when `.updateConfig()` is called.
    */
   constructor(config = {}, manager = overlays) {
+    console.log(config.backdropNode);
```
gerjanvangeest: Please remove!

jorenbroekema:
As discussed with Thijs, adding the style tag in the local overlay like this is actually quite bad, and should be a blocker for this PR. E.g. imagine 100 tooltips, causing 100 style tags with the same css that leaks

I will continue this PR later and create <lion-overlay-backdrop> in a commit on top of this to be able to properly do default tbackdrop styling for both local/global overlays :)

jorenbroekema:
Way too many conflicts so I started over:

https://github.com/ing-bank/lion/pull/1008

PR #656, opened 2020-03-24
Select Rich fixes

jorenbroekema:
Some fixes cherry-picked from our extension layer where we had to do these fixes in order for our lion rich select extension to work properly.

### DIFF
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -1,18 +1,17 @@
 import { ChoiceGroupMixin } from '@lion/choice-input';
 import {
   css,
+  getScopedTagName,
   html,
   LitElement,
-  SlotMixin,
   ScopedElementsMixin,
-  getScopedTagName,
+  SlotMixin,
 } from '@lion/core';
 import { FormControlMixin, FormRegistrarMixin, InteractionStateMixin } from '@lion/field';
 import { formRegistrarManager } from '@lion/field/src/registration/formRegistrarManager.js';
 import { OverlayMixin, withDropdownConfig } from '@lion/overlays';
 import { ValidateMixin } from '@lion/validate';
 import './differentKeyNamesShimIE.js';
-
 import { LionSelectInvoker } from './LionSelectInvoker.js';
 
 function uuid() {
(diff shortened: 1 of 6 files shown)


packages/select-rich/src/LionSelectRich.js
```diff
@@ -90,6 +89,13 @@ export class LionSelectRich extends ScopedElementsMixin(
 
       name: {
         type: String,
+        reflect: true,
+      },
+
+      // Copied from LionField
+      submitted: {
```
tlouisse: That's why it should be put in InteractionStsteMixin :)

packages/select-rich/src/LionSelectRich.js
```diff
@@ -90,6 +89,13 @@ export class LionSelectRich extends ScopedElementsMixin(
 
       name: {
         type: String,
+        reflect: true,
```
tlouisse: I think we already inherit this: https://github.com/ing-bank/lion/blob/master/packages/field/src/FormControlMixin.js#L37
We can omit name prop here

packages/select-rich/src/LionSelectRich.js
jorenbroekema: True, I guess I'll just fix that in the other PR? :P

packages/select-rich/src/LionSelectRich.js
jorenbroekema: Ah just checked, in old lion it was bugged because the form control mixin didn't have it. So I can remove it yea

packages/select-rich/src/LionSelectRich.js
jorenbroekema: Actually I'll fix it here

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -189,9 +188,12 @@ describe('lion-select-rich', () => {
     expect(el.hasFeedbackFor.includes('error')).to.be.true;
     expect(el.showsFeedbackFor.includes('error')).to.be.false;
 
+    // test submitted prop explicitly, since we dont extend field, we add the prop manually
+    el.submitted = true;
+    await el.updateComplete;
+    expect(el.showsFeedbackFor.includes('error')).to.be.true;
+
     el._listboxNode.children[1].checked = true;
-    // Set touched to true (needed for feedback show) because we simulate a user touching the select
```
jorenbroekema: This code didn't make sense here, probably old, and works without it.

tlouisse: Nice

PR #660, opened 2020-03-26
feat(select-rich): add has no default selection feature

jorenbroekema:


packages/select-rich/src/LionSelectRich.js
```diff
@@ -6,6 +6,17 @@ import { OverlayMixin, withDropdownConfig } from '@lion/overlays';
 import { ValidateMixin } from '@lion/validate';
 import './differentKeyNamesShimIE.js';
 import { LionSelectInvoker } from './LionSelectInvoker.js';
+<<<<<<< HEAD
```
gerjanvangeest: What is this?

jorenbroekema: Oops missed it

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -204,6 +204,22 @@ describe('lion-select-rich', () => {
...
+      <lion-select-rich id="color" name="color" label="Favorite color" has-no-default-selected>
+        <lion-options slot="input">
+          <lion-option .choiceValue=${'red'}>Red</lion-option>
+          <lion-option .choiceValue=${'hotpink'} disabled>Hotpink</lion-option>
+          <lion-option .choiceValue=${'teal'}>Teal</lion-option>
+        </lion-options>
+      </lion-select-rich>
+    `);
+
+    console.log(el.selectedElement, el.modelValue);
```
gerjanvangeest: console..

packages/select-rich/src/LionSelectRich.js
jorenbroekema: fixed

packages/select-rich/test/lion-select-rich.test.js
jorenbroekema: Having a bit of a monday-morning thing today 🙈

jorenbroekema: I removed the console logs

gerjanvangeest:
Please addd a storybook demo for this feature

gerjanvangeest:
Why did we not take the same approach we have with a native select? 
So adding an option with an empty value.

Now it is not possible to have an empty option without extending the `LionSelectRich`.

jorenbroekema:
I don't understand what you mean. Can you elaborate on why you think it is not possible to have an empty option without extending the lion select rich?

This feature is meant for ensuring that once a user selects an option, you cannot unselect by choosing <empty option>. 

We had some discussions with UX about this, but the specifically do not want users to be able to unselect to empty. It is purely a way of rendering a placeholder before any selection has taken place, and it's a common design pattern that I've seen in other design systems, where people do not prefer the native way of having an empty option that is explicitly selectable by the user.

jorenbroekema:
?path=/story/forms-select-rich--validation --> this story showcases it with empty option, it still works

gerjanvangeest:
> I don't understand what you mean. Can you elaborate on why you think it is not possible to have an empty option without extending the lion select rich?
> This feature is meant for ensuring that once a user selects an option, you cannot unselect by choosing .

Ah ok, sorry i misunderstood the meaning of the added code.

> ?path=/story/forms-select-rich--validation --> this story showcases it with empty option, it still works

But if this new feature is the preferred way of having a empty default, then i would say we show that one instead.

PR #665, opened 2020-03-30
feat: use markdown javascript (mdjs) for documentation

daKmoR:
still outstanding 
- [x] do basic sanity check on guidelines and move "not 100% clear" topics to an issue
- [x] move all devDependencies to root
- [x] final check

optional:
- [ ] introduce `form-core` (includes: field, validate, choice, ...)


daKmoR:
merged via https://github.com/ing-bank/lion/pull/734

PR #666, opened 2020-03-31
Fix reconnecting overlay nodes inside overlays

jorenbroekema:


packages/overlays/test-suites/OverlayMixin.suite.js
```diff
@@ -137,4 +140,59 @@ export function runOverlayMixinSuite({ /* tagString, */ tag, suffix = '' }) {
...
+        const outletNode = Array.from(actualNestedOverlay.children).find(
+          child => child.slot === '_overlay-shadow-outlet',
+        );
+        const contentNode = Array.from(outletNode.children).find(child => child.slot === 'content');
+
+        expect(contentNode).to.not.be.undefined;
+        expect(contentNode.innerText).to.equal('content of the nested overlay');
+      }
+
+      expect(true).to.be.true;
```
tlouisse, line 195: Not so sure about this one... :D

tlouisse: Or is it meant to make the test pass in case it's a local overlay? In that case, we can just drop this expect and it will pass as well

jorenbroekema: Oops :D

jorenbroekema: removed it

tlouisse:
Great job fixing this hard bug 👍 

I think we still agree that it feels like tech debt.
Hopefully we can mitigate this tech debt when we implement this: 
https://github.com/ing-bank/lion/pull/582

Could you maybe add a TODO about this, plus an explanation in the code why all of this is needed (or maybe mention a reference to the test at least)
Then I will approve

jorenbroekema:
Can you point out where you'd like to see another TODO? Because I added a TODO, and I added comments in the code why we have a promise->resolve thing in the connectedCallback that awaits updateComplete

Issue #653, opened 2020-03-22
Importing Radio and RadioGroup in the wrong order causes runtime exception


RobM-ADP:
## Describe the bug

When importing the radio-group and radio components in the wrong order, a runtime exception is thrown.

## To Reproduce

A simple project to reproduce the error can be found at: https://github.com/RobM-ADP/lion-components-issue

This seems to be an issue with order in which you import the components. I am able to cause the same error using the lion components directly if I import the radio before the radio-group. For example:

No error when using
```
import '@lion/radio-group/lion-radio-group.js';
import '@lion/radio-group/lion-radio.js';
```

But using:
```
import '@lion/radio-group/lion-radio.js';
import '@lion/radio-group/lion-radio-group.js';
```
Results in this error
```
npm.lion.js:2924 Uncaught (in promise) TypeError: Cannot set property 'feedbackData' of undefined
    at Array.<anonymous> (npm.lion.js:2924)
    at AsyncQueue.__run (npm.lion.js:2223)
    at AsyncQueue.add (npm.lion.js:2217)
    at HTMLElement._updateFeedbackComponent (npm.lion.js:2923)
    at HTMLElement.updated (npm.lion.js:2970)
    at HTMLElement.updated (npm.lion.js:4162)
    at HTMLElement.updated (npm.lion.js:7372)
    at HTMLElement.performUpdate (npm.lit-element.js:588)
    at HTMLElement._enqueueUpdate (npm.lit-element.js:528)
    at async HTMLElement._enqueueUpdate (npm.lit-element.js:522)
```

## Expected behavior

I would expect to be able to extend the Radio and RadioGroup in any order or have it at least be documented that the order is significant.

## Additional context

As you can see from the repo above, I am building with webpack and using Typescript so perhaps it is related to my configuration... or I'm just doing something dumb.

jorenbroekema:
I will close this now since it should be fixed. At least, on unpkg we no longer have errors anymore, so I believe whatever build setup you have should now also work. Let me know if it's still broken on latest master though, then I will reopen

RobM-ADP:
@jorenbroekema if I upgrade to the latest (or pull down master and then link) in the repo mentioned above, I still see the same issue thrown from the `ValidateMixin`
```
ValidateMixin.js:574 Uncaught (in promise) TypeError: Cannot set property 'feedbackData' of undefined
    at Array.eval (ValidateMixin.js:574)
    at AsyncQueue.__run (AsyncQueue.js:28)
[...]

Issue #396, opened 2019-11-26
[validate] loadDefaultFeedbackMessages fails to load in unpkg


LarsDenBakker:
I'm unable to demo lion using unpkg, because it throws an error when serving https://unpkg.com/@lion/validate@0.4.0/src/loadDefaultFeedbackMessages.js?module

See: https://jsbin.com/vimisivawa/edit?html,output as an example.

jorenbroekema:
@LarsDenBakker really late, but do you know what could be causing this? I don't have much experience with unpkg, is this something we need to change on our end?

jorenbroekema:
Probably due to

```js
return import(`../../${locale}.js`);
```

being hard to statically analyze. Most build tools don't provide out of the box support for it.

Action on our side could be a prepublish step where we analyze our file system of existing translation files and generate a switch case based on that, that way we avoid the dynamic statement inside the dynamic import. And we also avoid the ugly switch in our source code, it would only be in the published code. 

That should fix unpkg, and other build tools, from error'ing on it.

jorenbroekema:
Related issues that are likely caused by this:
https://github.com/ing-bank/lion/issues/654
https://github.com/ing-bank/lion/issues/653

jorenbroekema:
Should be fixed by this PR https://github.com/ing-bank/lion/pull/655 but seems like it isn't. https://unpkg.com/@lion/validate@0.9.0/src/loadDefaultFeedbackMessages.js?module

Not sure where to look, quite hard to debug unpkg :P but I can see that the same error is thrown for other files with the switch statement with dynamic imports e.g. @lion/input-iban/src/validators

jorenbroekema:
Same for https://unpkg.com/@lion/input-datepicker@0.11.1/src/LionCalendarOverlayFrame.js

daKmoR:
I did check out unpkg and it seems to be an issue with dynamic imports that have a ` in it.

like what we have in that file
```js
return import(`../translations/en.js`);
```

I opened an issue on unpkg and added a filing test
https://github.com/mjackson/unpkg/issues/253

I think however replacing the \` with a `'` will be a better solution for our issue... we use it 5 times so it should be a straight forward change

daKmoR:
oohhhh yes it finally works 💪

https://unpkg.com/@lion/validate@0.9.1/src/loadDefaultFeedbackMessages.js?module

PR #667, opened 2020-04-01
fix: convert unnecessary backticks to single quotes for unpkg

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/653
fixes https://github.com/ing-bank/lion/issues/396


jorenbroekema:
Yeah good idea, let's merge to a dev branch and bundle it

PR #672, opened 2020-04-03
feat(OverlayController): add invokerRelation option

### LINKED ISSUE #269: [tooltip/overlay] Unable to use aria-labelledby
Currently, the localOverlayController automatically sets an `aria-describedby` on the invoker, with a reference to the content. When you use the tooltip with an ℹ️icon, and provide a tooltip with additional information, `aria-describedby` is the appropriate relationship. 

You may want to use a tooltip as a visual way to show a label or accessible name. For example, when you have a button with only an icon as content. In that case you would want to set `aria-labelledby` instead.

And when you use a local overlay for something like a menu, you don't want either `aria-describedby` or `aria-labelledby`.

I suggest making the `aria-describedby` in localOverlay to be optional, with another option to set an `aria-labelledby` instead. Suggested API in #267

T-Motion:
implementation for bug #269 
fixes #269

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -248,7 +248,10 @@ export class OverlayController {
     if (this.isTooltip) {
       // TODO: (@tlouisse) this could also be labelledby.
       if (this.invokerNode) {
-        this.invokerNode.setAttribute(this.invokerRelation === 'label' ? 'aria-labelledby' : 'aria-describedby', this._contentId);
+        this.invokerNode.setAttribute(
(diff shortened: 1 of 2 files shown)


jorenbroekema:
Very nice :) can you remove the TODO in https://github.com/ing-bank/lion/pull/672/files#diff-265d92898c7da859ad5e1d717675572aL249 ? Since we now did it :D

daKmoR:
is there no documentation for this on purpose? 😅

jorenbroekema:
Admittedly, I forgot about that completely :P but we should indeed add a demo + docs, @T-Motion  you wanna do this?

Issue #469, opened 2020-01-02
[overlays] Intercepting open/close, opened still toggled


jorenbroekema:
## Describe the bug

I am doing `preventDefault()` on the "before-opened"/"before-closed" event, but that does not prevent `opened-changed` from firing. In addition, `opened` state toggles as usual. The only thing that is prevented is the visual show/hide of the overlay content.

https://github.com/ing-bank/lion/blob/master/packages/overlays/test-suites/OverlayMixin.suite.js#L103 this unit test here tests for it, however it uses the `.show()` and then tests the opened state. 

In the demo, the invoker is clicked, so there's a slight difference in the demo vs the unit test. In the demo if I use the .show() method directly on the overlay component, rather than clicking the invoker, it works indeed as expected (`opened-changed` is not fired, `opened` does not change).

## To Reproduce

https://lion-web-components.netlify.com/?path=/story/overlays-system-overview--intercept-open-close

Run the story locally, but add the following event listener
```js
<demo-overlay-system #overlay=${r(refs)} @before-closed=${intercept} @before-opened=${intercept} @opened=${(e) => console.log('opened!', e.target.opened)}>
```
With that console log you can see that 'opened' is fired during blocked, and also the 'opened' state is toggled. 

## Expected behavior

I expect preventDefault to prevent not just visually opening/closing the overlay, but also prevent opened from toggling (and as a result, prevent opened-changed from firing).

> Note: What is also a bit confusing is that logging e.target.opened like that will give you an "outdated" state, because the show method is async, by the time you log the e.target.opened, it has not updated yet to the new state... but that's a different issue 🙈

AymenBenAmor:
I made some investigations here and this is what i found out : 
In the demo, we create a demo overlay system DemoOverlaySystem. This component extend the overlayMixin. The issue here is that the component adds an EventListener that changes the opened property directly as the code below shows :
`
_setupOpenCloseListeners() {
    super._setupOpenCloseListeners();
    this.__toggle = () => {
      this.opened = !this.opened;
    };

    if (this._overlayInvokerNode) {
      this._overlayInvokerNode.addEventListener('click', this.__toggle);
    }
  }

  _teardownOpenCloseListeners() {
    super._teardownOpenCloseListeners();

    if (this._overlayInvokerNode) {
      this._overlayInvokerNode.removeEventListener('click', this.__toggle);
    }
  }
`

[...]

PR #673, opened 2020-04-06
wip: fix(overlays): synchronize intercepted opened state OverlayMixin

tlouisse:
Proposal to fix https://github.com/ing-bank/lion/issues/469

TODO:
- It currently relies on setting `this.__opened` to synchronize `this.opened`. This is an implementation detail of UpdatingElement, so preferably we add a way to take ownership of the implementation detail (for instance by overriding the getter/setter, making us in control of the private variable name.
- add a unit test


AymenBenAmor:
A new PR is ready with the full fix. https://github.com/ing-bank/lion/pull/680

PR #674, opened 2020-04-06
fix: consistent reset styles

jorgecasar:
I suggest this changes to make the button consist and to be easier to restyle.

```
:host {
  /* Avoid vertical space outside .btn */
  padding: 0; 
}
.btn {
  /* Fit with the 40px height (24+8+8) */
  /* Same padding keep icon only buttons squared */
  padding: 8px; 
}
.click-area {
  /* Same size as .btn */
  margin: 0;
}
:host .btn ::slotted(button) {
  /* Set the button in the top left corner to showcase the difference */
  top: 0;
  left: 0;
  /* reset agent style and now is 1px square */ 
  padding: 0;
  border: 0;
}
```

### DIFF
--- a/packages/button/stories/index.stories.mdx
+++ b/packages/button/stories/index.stories.mdx
@@ -1,5 +1,4 @@
 import { Story, Meta, html } from '@open-wc/demoing-storybook';
-import '@lion/input/lion-input.js';
 
 import '../lion-button.js';
 import iconSvg from './icon.svg.js';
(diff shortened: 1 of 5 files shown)


packages/button/stories/index.stories.mdx
```diff
@@ -142,6 +143,24 @@ Important notes:
 - `@click` on `<lion-button>` and `@submit` on `<form>` are triggered by these use cases. We strongly encourage you to listen to the submit handler if your goal is to do something on form-submit.
 - To prevent form submission full page reloads, add a **submit handler on the form** `@submit` with `event.preventDefault()`. Adding it on the `<lion-button>` is not enough.
 
+### Usage with lion inputs
```
jorenbroekema, line 146: We usually try to avoid that our package depend on one another, and then for demos where we combine multiple components, we usually put them inside the form-system package, so we only have a single package that depends on all the others. I think the demo is nice, but then maybe we can add it inside `form-system` package as a separate commit?

jorenbroekema:
Thanks, I merged it!

Issue #576, opened 2020-02-14
`this.children` not available for typescript compiler


jdvivar:
## Describe the bug

Typescript compiler doesn't recognise `this.children`

## To Reproduce

Check out this simple repo https://github.com/jdvivar/demo-bug-children-lion
Run `npm run compile`

### This is the failing code
```js
import {
  html,
  LitElement,
  TemplateResult,
} from '@lion/core'

export class DemoChildren extends LitElement {

  render (): TemplateResult {
    return html`
        ${[...this.children]}
    `
  }
}
```

### This is the non-failing code
```js
import {
  html,
  LitElement,
  TemplateResult,
} from 'lit-element'

export class DemoChildren extends LitElement {

  render (): TemplateResult {
    return html`
        ${[...this.children]}
    `
  }
}
```

## Expected behaviour

Both `demo-children-*.ts` files should be equally compiled by Typescript with no errors

## Actual behaviour

```
$ npm run compile

[11:10:10 AM] File change detected. Starting incremental compilation...

demo-children-lion-core.ts:11:20 - error TS2339: Property 'children' does not exist on type 'DemoChildren'.

11         ${[...this.children]}
                      ~~~~~~~~

[11:10:10 AM] Found 1 error. Watching for file changes.
```

## Additional comments
Note I'm not declaring the `lit-element` as a dependency in `package.json`. This is done on purpose so I can be sure the package loaded is the exactly the one declared as a dependency from and used by `@lion/core`

jdvivar:
I've come to the conclusion it's probably a Typescript bug instead, however I'll keep it open in case someone can give me a hint about why exactly this is happening. Although I suspect it's just Typescript messing around with typings.

jdvivar:
OK, I tried with Typescript 3.7.2 and 3.7.5, but at least I can confirm this is solved in `next`. I guess it's a freshly fixed bug. Closing as it's solved!

jdvivar:
Making a copy of `@lion/core/index.js` into `@lion/core/index.ts` makes it work 🤔

jorenbroekema:
My suspicion is that because we re-export LitElement without exporting the type (we don't export types anywhere in lion currently), the compiler doesn't understand that LitElement extends HTMLElement and has the children property.

That would also explain why making a copy of the index.js makes it work, because then you have the context of the re-export, so then when you import, the TS compiler can actually trace that it comes from lit-element and that it extends HTMLElement.

jdvivar:
[...]

PR #675, opened 2020-04-07
Add Types for @lion/core, @lion/tabs, and add type lint and types generate to build

tlouisse:
**Important**
> Will release almost every package, because publish files in package.jsons are now aligned, and to include .d.ts files

Fixes https://github.com/ing-bank/lion/issues/576

### DIFF
--- a/packages/core/src/DelegateMixin.js
+++ b/packages/core/src/DelegateMixin.js
@@ -1,203 +1,177 @@
 /* eslint-disable class-methods-use-this */
-
 import { dedupeMixin } from '@open-wc/dedupe-mixin';
 
 /**
- * # DelegateMixin
- * Forwards defined events, methods, properties and attributes to the defined target.
- *
- * @example
- * get delegations() {
- *   return {
- *     ...super.delegations,
- *     target: () => this.shadowRoot.getElementById('button1'),
- *     events: ['click'],
- *     methods: ['click'],
- *     properties: ['disabled'],
- *     attributes: ['disabled'],
- *   };
- * }
- *
- * render() {
- *   return html`
- *     <button id="button1">with delegation</button>
- *   `;
- * }
- *
- * @type {function()}
- * @polymerMixin
- * @mixinFunction
+ * @typedef {import('../types/DelegateMixinTypes').DelegateMixin} DelegateMixin
  */
-export const DelegateMixin = dedupeMixin(
-  superclass =>
-    // eslint-disable-next-line
-    class DelegateMixin extends superclass {
-      constructor() {
-        super();
-        this.__eventsQueue = [];
-        this.__propertiesQueue = {};
-        this.__setupPropertyDelegation();
-      }
 
-      /**
-       * @returns {{target: null, events: Array, methods: Array, properties: Array, attributes: Array}}
-       */
-      get delegations() {
-        return {
-          target: null,
-          events: [],
-          methods: [],
-          properties: [],
-          attributes: [],
-        };
-      }
+/**
+ * @typedef DelegateEvent
+ * @property {string} type - Type of event
+ * @property {Array<?>} args - Event arguments
+ */
 
-      connectedCallback() {
-        if (super.connectedCallback) {
-          super.connectedCallback();
-        }
-        this._connectDelegateMixin();
-      }
+/** @type {DelegateMixin} */
+const DelegateMixinImplementation = superclass =>
+  // eslint-disable-next-line
+  class DelegateMixin extends superclass {
+    constructor() {
+      super();
 
-      updated(...args) {
-        super.updated(...args);
-        this._connectDelegateMixin();
-      }
+      /** @type {DelegateEvent[]} */
+      this.__eventsQueue = [];
 
-      /**
-       * @param {string} type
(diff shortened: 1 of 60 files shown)


packages/tabs/src/LionTabs.js
```diff
@@ -108,11 +112,14 @@ export class LionTabs extends LitElement {
 
   constructor() {
     super();
+    /**
+     * An index number of the selected tab
+     */
```
tlouisse, line 117: Can you explain why we need to put it here (initialisation phase) and not in the properties list (define phase)?

packages/tabs/src/LionTabs.js
```diff
@@ -1,4 +1,4 @@
-import { LitElement, css, html } from '@lion/core';
+import { LitElement, css, html } from 'lit-element';
```
tlouisse, line 1: Can you elaborate on this choice? 
Should we maybe do it in a seperate merge request for all packages?
Since we want to have all our packages work nicely together (reducing the risk of dependency conflicts) I like a centrally orchestrated version. (not saying @lion/core is the only way to achieve this...)

daKmoR: this is to make it work for now...
as soon as you import from `@lion/core` all the typings fail as apparently the typings for lit-element are not "forwarded"

I do hope this get's fixed once `@lion/core` becomes typed as well

packages/tabs/src/LionTabs.js
daKmoR: typescript only knows about "javascript" and has no knowledge about web components/lit-element properties... if you want that info in the autocomplete then we will need to be put it with the js property

tlouisse: I was hoping it would pick 
```js
    /**
     * An index number of the selected tab
     * @property {Number} selectedIndex
     */
```
but it doesn't...

packages/tabs/src/LionTabs.js
tlouisse: Ah :)

packages/tabs/src/LionTabs.js
```diff
@@ -65,33 +63,34 @@ function cleanButton({ el, clickHandler, keydownHandler, keyupHandler }) {
...
   el.removeAttribute('selected');
-  el.setAttribute('aria-selected', false);
-  el.setAttribute('tabindex', -1);
+  el.setAttribute('aria-selected', 'false');
+  el.setAttribute('tabindex', '-1');
 }
 
 /**
- * @param {KeyboardEvent} ev
+ * @param {Event} ev
```
jorenbroekema, line 89: EventHandlerNonNull only accepts normal Events, so I'm type casting it to a KeyboardEvent in the function. Alternatively we could create KeyboardEventHandlerNonNull type definition but I'm not sure how to do this in JSDocs, so help is welcome :)

packages/tabs/src/LionTabs.js
```diff
@@ -161,26 +160,36 @@ export class LionTabs extends LitElement {
...
+      }
+    }
   }
 
   __setupStore() {
     /** @type {StoreEntry[]} */
     this.__store = [];
-    const buttons = this.querySelectorAll('[slot="tab"]');
-    const panels = this.querySelectorAll('[slot="panel"]');
+    const buttons = /** @type {HTMLElement[]} */ (Array.from(
```
jorenbroekema, line 187: Type casting to HTMLElement[] because the default is Element[], but then we get problems further down the line

packages/core/src/DelegateMixin.js
```diff
@@ -1,203 +1,179 @@
 /* eslint-disable class-methods-use-this */
-
 import { dedupeMixin } from '@open-wc/dedupe-mixin';
 
 /**
- * # DelegateMixin
```
jorenbroekema, line 6: These docs/comments are now in the manually written type definition file so you get it as a popup on the typehint

packages/core/src/UpdateStylesMixin.js
```diff
@@ -1,59 +1,80 @@
...
+
+        Object.keys(newStyles).forEach(key => {
+          if (key.indexOf('--') === -1) {
+            newStylesString += `${key}:${newStyles[key]};`;
+          } else {
+            newCssVariablesObj[key] = newStyles[key];
+          }
+        });
+        this.setAttribute('style', newStylesString);
+        // @ts-ignore not sure how to type ShadyCSS..
```
jorenbroekema, line 66: I checked the ShadyCSS lib but they don't have types, so not sure what to do with this with regards to typing it

tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "esModuleInterop": true
+  },
+  "include": ["packages/core/**/*.js", "packages/tabs/**/*.js"],
+  "exclude": [
+    "node_modules",
+    "**/node_modules/*",
+    "**/coverage/*",
+    "**/dist/**/*",
+    "packages/**/demo/**/*",
+    "packages/**/test",
```
jorenbroekema, line 23: Do we wanna type lint test files? I saw Thomas did add JSDocs types for some tests.

Issue #677, opened 2020-04-10
[lion-select-rich]: Single option support lazy loaded


palash2601:
# Feature request

When there is only a single option available in the listbox of lion-select-rich, an attribute `singleOption` should be added invoker element so that desired styling can be defined by the extended component.

Also since it is a single option will be available, it should not open overlay which displays the option

tlouisse:
@palash2601 thanks for your pull request 💪 Although the related pr is already merged, there are still a few things that need to be tackled (also see the comments on the pr):
- could you add some docs? Maybe a story that explains what it does.
- could you add a test that handles dynamically added options (support for lazily loaded options). For this, you could hook into `addFormElement` and `removeFormElement`.
- could you maybe change attr name to `single-option`?
- it's recommended to use `this.formElements.length` for checking the amount of children (since it is allowed to add more visual sibling elements)
- the rich-select was deisgned to closely mimic the native select. Maybe we can make this an opt-in flag that you enable by default on your extension layer?

gerjanvangeest:
Updated in: https://github.com/ing-bank/lion/pull/682

jorenbroekema:
Closing as enhancement for lazy loaded single-option with votes needed

daKmoR:
this has been implemented - if there is still something open please feel free to create a dedicated issue for it 🤗

"closing" this feature request

PR #678, opened 2020-04-10
feat(select-rich): added signleoption functionality

palash2601:
Closes issue
https://github.com/ing-bank/lion/issues/677

### DIFF
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -28,13 +28,22 @@ export class LionSelectInvoker extends LionButton {
         type: Object,
       },
       /**
-       * @desc When the connected LionSelectRich insteance is readOnly,
+       * @desc When the connected LionSelectRich instance is readOnly,
        * this should be reflected in the invoker as well
        */
       readOnly: {
         type: Boolean,
         reflect: true,
         attribute: 'readonly',
+      },
+      /**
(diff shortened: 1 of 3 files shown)


daKmoR:
documentation missing? 😭

packages/select-rich/src/LionSelectInvoker.js
```diff
@@ -28,14 +28,23 @@ export class LionSelectInvoker extends LionButton {
...
         attribute: 'readonly',
       },
+      /**
+       * @desc When the connected LionSelectRich instance has only one option,
+       * this should be reflected in the invoker as well
+       */
+      singleOption: {
+        type: Boolean,
+        reflect: true,
+        attribute: 'singleOption',
```
tlouisse, line 46: attribute names can not be camelCased. Should be `single-option`?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -219,6 +219,10 @@ export class LionSelectRich extends ScopedElementsMixin(
 
   firstUpdated(changedProperties) {
     super.firstUpdated(changedProperties);
+    if (this._listboxNode.childElementCount === 1) {
```
tlouisse, line 222: It's a better practice to check for `this.formElements.length`, since we also allow for other visual elements inside the listbox node (think of separators/groups etc.)

Also, do we also update accordingly when new options are removed/added dynamically afterwards?

PR #680, opened 2020-04-14
fix(overlays): synchronize intercepted opened state OverlayMixin

AymenBenAmor:
this MR fixes issue 469 
Thanks to @tlouisse for the fix. I added a test to cover the real use case (clicking on the invoker and not calling the show directly)


tlouisse:
Merging it. Thanks @AymenBenAmor 💪

PR #682, opened 2020-04-15
chore(select-rich): docs for single-option and fix review comments

palash2601:
all changes done as suggested in [this](https://github.com/ing-bank/lion/pull/678) MR

### DIFF
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -43,7 +43,7 @@ export class LionSelectInvoker extends LionButton {
       singleOption: {
         type: Boolean,
         reflect: true,
-        attribute: 'singleOption',
+        attribute: 'single-option',
       },
     };
   }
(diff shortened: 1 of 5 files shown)


jorenbroekema:
Maybe we should add a test for imperatively changing singleOption (because some logic was moved to updated, we should also verify that it works now). And perhaps another test for confirming that we don't have an 'after' slot in the DOM when using singleOption?

tlouisse:
Great improvements.

Could you maybe also make support for:

```js
// dynamic import
import('./lion-option-lazily-loaded');
```

```html
<lion-select-rich>
   <lion-options>
     <lion-option-lazily-loaded/>
   </lion-options>
</lion-select-rich>
```
You could probably do this by abstracting the logic (currently in firstUpdated) in a function and call that on every addFormElement/removeFormElement (that happens after firstUpdated).


Also, could you enable the blocking overlay behavior behind a flag `block-interaction-on-single-option` (I'm thinking of a better name :) ), so we keep the default completely aligned with the platform

palash2601:
@jorenbroekema, I have added tests which you mentioned

@tlouisse The support for lazy loading of options, I would suggest to take up that separately and merge this one

tlouisse:
> @jorenbroekema, I have added tests which you mentioned
> 
> @tlouisse The support for lazy loading of options, I would suggest to take up that separately and merge this one

Sounds fair to do that. Thanks for all the effort :muscle

Issue #234, opened 2019-08-08
[localize] use a different attribute for language


LarsDenBakker:
Currently we use `lang` on the html tag for determining the language of the localize manager. As we saw with https://github.com/ing-bank/lion/issues/152 this breaks when using some kind of translation tool, as that modifies this attribute.

It's fixed in https://github.com/ing-bank/lion/pull/164 by falling back to a default language when a language can't be found. This is good, but doesn't fix other cases where for example you use the locale to send to an API which returns localized content.

Perhaps instead we should use a different attribute like `data-ing-locale` (or something else). That way we have a constant that doesn't accidentally change. It properly fixes using a translation tool, because `data-ing-locale` will determine the language that's used to render the content while `lang` determines the language the content was translated to.

For backwards compatibility we can keep using `lang` if there is no `data-ing-locale`.

tlouisse:
I'm not so much into this topic, but if I'm right:

Let's say I want to make a Dutch site, (for which all translations are available to Localize).
I will set:
- lang attr to `nl-NL` 
- fallback locale to `nl-NL` 

Now our users want to translate this to German. The lang attr is changed by Chrome, but the fallback locale will still be used to fetch Dutch translations (which can be translated to German by Chrome (because it expects Dutch input everywhere).

Would it be a possibility then to use the fallback locale to do the merge request?

Since I wasn't too much involved in the changes described above, I might get them wrong and the above doesn't make sense ;)

LarsDenBakker:
You're right that it works that way. 

However let's say your dutch website offers `en-GB` texts and it's visited by a polish person. They set the language to `en-GB`, and they will get `en-GB` texts. If they then use google translate to translate that english text into polish, google translate will change `lang` to `pl` and change the text on the screen. The localize manager now gets an unrecognized language and will fallback to `nl-NL` which breaks this scenari.

Instead the localize manager should keep using `en-GB`, and let google translate handle it.

The essence of the problem is that we use `lang` for two things: what's locale do we use for fetching/rendering content, and what's the locale that's currently displayed on the screen. Often those two are the same, but sometimes it's not.

tlouisse:
Ah right, for some reason I thought the fallback was introduced to solve the 'Google translate problem'.

Providing an alternative attr would be a good idea, but I think it would raise accessibility problems (screen reader users need to know which language their content is in).

[...]

PR #684, opened 2020-04-15
fix localize when combined with google translate

tlouisse:
## Support for Google Translate

closes https://github.com/ing-bank/lion/issues/234

**!! update !!**
An enhanced (non breaking/backwards compatible because of automatic feature detection) version of this merge request has been implemented in our extension layer in the meantime. Whenever proceeding this mr, please make sure to use that solution as a fundament.

-------------------------------------------

All options are behind an opt-in flag `supportExternalTranslationTools`, since it's a breaking change.

## The flow explained

```
/**
 * Support for Google Translate (or other 3rd parties taking control
 * of the html[lang] attribute).
 *
 * Have the following scenario in mind:
 * 1. locale is initialized by developer via html[data-localize-lang="en-US"] and
 * html[lang="en-US"]. When localize is loaded (note that this also can be after step 2 below),
 * it will sync its initial state from html[data-localize-lang]
 * 2. Google Translate kicks in for the French language. It will set html[lang="fr"].
 * This new language is not one known by us, so we most likely don't have translations for
 * this file. Therefore, we do NOT sync this value to LocalizeManager. The manager should
 * still ask for known resources (in this case for locale 'en-US')
 * 3. locale is changed (think of a language dropdown)
 * It's a bit of a weird case, because we would not expect an end user to do this. If he/she
 * does, make sure that we do not go against Google Translate, so we maintain accessibility
 * (by not altering html[lang]). We detect this by reading _langAttrSetByTranslationTool:
 * when its value is null, we consider Google translate 'not active'.
 *
 * When Google Translate is turned off by the user (html[lang=auto]),
 * `localize.locale` will be synced to html[lang] again
 *
 *
 * Keep in mind that all of the above also works with other tools than Google Translate,
 * but this is most widely used tool and therefore used as an example.
 */
``` 

_Open question: do we enable this flag by default (this allows to disable it in our extension layer and keep it backward compatible) or do we disable it by default on Lion as well?_

Currently it's enabled and then we have 17 failing tests throughout Lion in packages that depend on `localize`.

### To do
- [ ] I seperated the the 'Google Translate' test in a distinct describe to make it easier to review. They need to be integrated into the other tests
- [ ] fix failing tests in other packages (assuming we go for `supportExternalTranslationTools` enabled by default)
- [ ] clean up commit history

### DIFF
--- a/packages/localize/test/LocalizeManager.test.js
+++ b/packages/localize/test/LocalizeManager.test.js
@@ -27,11 +27,6 @@ describe('LocalizeManager', () => {
     resetFakeImport();
   });
 
-  it('initializes locale from <html> by default', () => {
-    manager = new LocalizeManager();
-    expect(manager.locale).to.equal('en-GB');
-  });
-
   it('syncs locale back to <html> if changed', () => {
     manager = new LocalizeManager();
     manager.locale = 'nl-NL';
@@ -57,16 +52,6 @@ describe('LocalizeManager', () => {
       manager = new LocalizeManager();
       setTimeout(() => {
         manager.locale = 'en-US';
-      });
-      const event = await oneEvent(manager, 'localeChanged');
-      expect(event.detail.newLocale).to.equal('en-US');
-      expect(event.detail.oldLocale).to.equal('en-GB');
-    });
-
-    it('fires "localeChanged" event if locale was changed via <html lang> attribute', async () => {
-      manager = new LocalizeManager();
-      setTimeout(() => {
-        document.documentElement.lang = 'en-US';
       });
       const event = await oneEvent(manager, 'localeChanged');
       expect(event.detail.newLocale).to.equal('en-US');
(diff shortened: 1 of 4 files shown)


packages/localize/test/LocalizeManager.test.js
```diff
@@ -593,4 +593,31 @@ describe('LocalizeManager', () => {
...
+      expect(manager.locale).to.equal(undefined);
+    });
+
+    it(`doesn't synchronize from html[lang] attribute to LocalizeManager
+      when changed by 3rd parties`, async () => {
+        manager = new LocalizeManager();
+        manager.locale = 'nl-NL';
+
+        document.documentElement.lang = 'fr-FR'; // this is something google translate would do
+        expect(manager.locale).to.equal('nl-NL');
```
gerjanvangeest, line 620: 1. If we have the locale available, why not showing it?
2. If we don't have the locale, didn't we want to have the default locale (`en-GB`), so the translator can use that?

Btw google translate only sets the language and not the locale so `fr`

tlouisse: 1. it would break a11y, since end user gets a French site (translated by Google)
2. The fallback locale (en-GB) is still hardcoded in LocalizeManager
Thanks, I updated all code and docs to `fr` instead of `fr-FR`

tlouisse:
WIll be solved with https://github.com/ing-bank/lion/pull/758

PR #685, opened 2020-04-15
feat(babel-plugin-extends-docs): extend docs by rewriting imports/code

jorenbroekema:
- [x] Move all missing specs into here
- [x] Implement import paths transformations
- [x] Implement class names transformations
- [x] Implement tag names transformations
- [x] Check what other transforms we might need

### DIFF
--- a/packages/babel-plugin-extend-docs/src/babelPluginExtendDocs.js
+++ b/packages/babel-plugin-extend-docs/src/babelPluginExtendDocs.js
@@ -1,0 +1,63 @@
+/* eslint-disable no-param-reassign */
+const {
+  renameAndStoreImports,
+  generateImportStatements,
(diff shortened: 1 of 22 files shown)


packages/babel-plugin-extend-docs/src/babelPluginExtendDocs.js
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable no-param-reassign */
+
+// -1 because filepath is an absolute path starting with '/' and we turn it into a relative path without a '/' at the start
+const getFolderDepth = filePath => [...filePath.matchAll(new RegExp('/', 'g'))].length - 1;
```
daKmoR, line 4: string.matchAll is not supported in node LTS (10.x)

PR #692, opened 2020-04-20
fix: add display:none for hidden

davidoff31:
Add display: none rule to all UI components that apply a :host display rule as suggested https://developers.google.com/web/fundamentals/web-components/best-practices#add-a-host-display-style-that-respects-the-hidden-attribute

### DIFF
--- a/packages/form-system/CHANGELOG.md
+++ b/packages/form-system/CHANGELOG.md
@@ -3,6 +3,15 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+
+## [0.7.9](https://github.com/ing-bank/lion/compare/@lion/form-system@0.7.8...@lion/form-system@0.7.9) (2020-04-20)
+
+**Note:** Version bump only for package @lion/form-system
+
+
+
+
+
 ## [0.7.8](https://github.com/ing-bank/lion/compare/@lion/form-system@0.7.7...@lion/form-system@0.7.8) (2020-04-17)
 
 **Note:** Version bump only for package @lion/form-system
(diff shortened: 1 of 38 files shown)


packages/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -493,6 +493,9 @@ describe('<lion-input-datepicker>', () => {
         expect(myOverlayOpenedCbHandled).to.be.true;
         await myElObj.selectMonthDay(1);
         expect(myUserSelectedChangedCbHandled).to.be.true;
+        /*         myElObj.overlayEl.hidden = true;
```
jorenbroekema, line 496: I think you forgot to clean up this comment? :P

packages/form-system/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+
```
jorenbroekema, line 6: might be accidental change?

packages/input-datepicker/test/lion-input-datepicker.test.js
davidoff31: Yes, I did! Made new commit to resolve.

packages/form-system/CHANGELOG.md
davidoff31: Definitely! Removed in the last commit. Thanks!

jorenbroekema:
Thanks for the contribution! I've merged it to dev for now because this will release almost all of our packages, so I hope to bundle some other things with it to reduce the number of releases. I'll send an update once it's released so we can make sure it's fixed on the latest version of our extension layer.

PR #695, opened 2020-04-28
Overlays shadow outlet issue

AymenBenAmor:
closes https://github.com/ing-bank/lion/pull/582, https://github.com/ing-bank/lion/pull/549


packages/overlays/src/OverlayController.js
```diff
@@ -284,14 +291,30 @@ export class OverlayController {
...
+      this._contentNodeWrapper.style.display = this.placementMode === 'local' ? 'flex' : '';
+      this._contentNodeWrapper.style.flexDirection = this.getFlexDirectionAccordingToPosition();
       await this._handleFeatures({ phase: 'show' });
       await this._handlePosition({ phase: 'show' });
       this.elementToFocusAfterHide = elementToFocusAfterHide;
       this.dispatchEvent(new Event('show'));
     }
   }
 
+  getFlexDirectionAccordingToPosition() {
```
tlouisse, line 303: Can you explain why this is needed?

packages/overlays/README.md
```diff
@@ -106,7 +106,7 @@ class MyOverlayComponent extends LitElement {
     return html`
       <slot name="invoker"></slot>
       <slot name="content"></slot>
-      <slot name="_overlay-shadow-outlet"></slot>
+      <div id="overlay-content-node-wrapper"></div>
```
tlouisse, line 109: You forgot `<slot name="content"></slot>` ;)

packages/overlays/src/OverlayMixin.js
```diff
@@ -3,7 +3,7 @@ import { OverlayController } from './OverlayController.js';
 
 /**
  * @type {Function()}
- * @polymerMixin
+ * @polymerMixinOverlayMixin
```
tlouisse, line 6: Not sure which static analysis tool looks for `@polymerMixin` and if we still use that (do we need it for webcomponent-analyzer @daKmoR?), but I think we need and extra space here :)

packages/overlays/src/OverlayController.js
AymenBenAmor: By encapsulating content and arrow inside of the node_wrapper, pepper could not calculate the right position of the tooltip anymore. To fix this I had to change the CSS and use flex instead. so according to the position property the flex-direction is different.

jorenbroekema:
As a TODO, either for this PR or a second one, we can deprecate lion-tooltip-arrow and encapsulate the styles for the arrow now inside the lion-tooltip, and use the slotted selector directed at slot=arrow :)

packages/overlays/src/OverlayController.js
tlouisse: Hmm,  Popper 1 or Popper 2? 
Popper 1 seemed to work fine for me...🤔
Maybe I'm wrong ;)

AymenBenAmor: I'm using Popper 1. (did not want to upgrade at the same time.) the upgrade can be part of another PR.
In the PR you sent me you upgraded to popper2 so maybe that's why it was working for you.

tlouisse: Yeah it should, because it was breaking. When I did the change I first tried on the current version of popper and it seemed to work and then when moving to Popper2 the margin between invoker and content was gone.
Could you maybe elaborate on how those flex-direction props solve the problem?
Does it also work for global overlays?

AymenBenAmor: Actually what happens is that when we are in tooltip mode (so local) the arrow itself is not well calculated( not the tooltip). the rationale says that popper wants the arrow to be part of the contentNode. What I did here is that according to the placement you provide I change the flex-direction. So if it's top it's column reverse because the arrow needs to be in the bottom and the tooltip on top. and the opposite for placement: bottom. The same strategy is applied for right and left.

tlouisse: Well, if this is the only way to trigger Popper ;) Maybe you could make the method private and make sure it only applies those styles for loval overlays?

packages/tooltip/stories/index.stories.mdx
```diff
@@ -72,25 +71,21 @@ You can easily change the placement of the content node relative to the invoker.
...
-        <lion-tooltip-arrow slot="arrow"></lion-tooltip-arrow>
       </lion-tooltip>
-      <lion-tooltip .config=${{ popperConfig: { placement: 'left' } }}>
+      <lion-tooltip .config=${{ popperConfig: { placement: 'left' } }} has-arrow>
         <button slot="invoker">Left</button>
         <div slot="content" class="demo-tooltip-content">Its left placement</div>
-        <lion-tooltip-arrow slot="arrow"></lion-tooltip-arrow>
       </lion-tooltip>
     </div>
   `}
```
gerjanvangeest, line 91: The code-block below doesn't have the `has-arrow` attribute.

tlouisse: good catch

PR #700, opened 2020-04-29
chore: upgrade open-wc/demoing-storybook version

gerjanvangeest:


package.json
```diff
@@ -41,6 +41,7 @@
     "start": "npm run storybook",
     "storybook": "start-storybook -p 9001",
     "storybook:build": "build-storybook",
+    "storybook:build:start": "npm run storybook:build && es-dev-server --root-dir storybook-static --open",
```
jorenbroekema, line 44:
```suggestion
    "storybook:build": "build-storybook",
    "storybook:build:start": "es-dev-server --root-dir storybook-static --open",
```
I would rather have them separated, so I can build it once, and start it up whenever I want without having to rebuild. If I need to rebuild I can do that manually. For me it's a bit better dev experience but it's a small detail, I don't really mind either way, up to you :).

gerjanvangeest: I also don't mind either way

PR #701, opened 2020-05-06
poc overlays shared state

tlouisse:
Proof of concept for handling multiple overlay manager that share state


tlouisse:
closing, since we solve with singleton-manager

Issue #676, opened 2020-04-09
Child dialogs not working


gerjanvangeest:
## Child dialogs do not work

The parent dialog opens as normal, but the child dialog (so a dialog inside the parent dialog) doesn't want to open.

```html
    <lion-dialog>
      <button slot="invoker">Open parent overlay</button>
      <div slot="content">
        Parent dialog
        <lion-dialog>
          <button slot="invoker">Open child overlay</button>
          <div slot="content">
              Child dialog
          </div>
        </lion-dialog>
      </div>
    </lion-dialog>
```

Broken in latest releases of dialog/overlay.
Still worked in version dialog 0.4.12 & overlay 0.12.7.

T-Motion:
The invoker of the child dialog has two click events attached. _setupOpenCloseListeners in LionDialog is being called twice, so two events are attached. This is causing the __toggle function in LionDialog to be called twice. 
Not sure how to fix it though.

jorenbroekema:
Might be related to my recent fix for dialogs being moved in the DOM and going through init - teardown cycle twice?  It may be worth parking this issue until we've gotten rid of the shadow overlay outlet that Aymen is working on 🙈 

Thanks for the investigation @T-Motion  really helps to know the root cause of the nested dialog not (visually) opening.

jorenbroekema:
I realized that the shadow outlet stuff wasn't affecting anything about this issue. So I created a fix. I checked that it fixes things for both the current overlays/dialog, as well as on the shadow outlet branch

PR #702, opened 2020-05-06
fix(dialog): teardown open close listeners after initial updateComplete

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/676

The reason this bug was here is because open/close event listeners were added on a nested dialog, but then the nested dialog would be moved through the DOM **before** finishing setting up the overlay controller. So then disconnectedCallback would happen and not tear down the open/close event listeners (because of the overlay controller guard). 

And then you get duplicate open/close listeners on the nested dialog, resulting in a double toggle on click, which means it opens and instantly closes.

### DIFF
--- a/packages/dialog/test/lion-dialog.test.js
+++ b/packages/dialog/test/lion-dialog.test.js
@@ -1,6 +1,5 @@
-import { expect, fixture, html, unsafeStatic } from '@open-wc/testing';
 import { runOverlayMixinSuite } from '@lion/overlays/test-suites/OverlayMixin.suite.js';
-
+import { expect, fixture, html, unsafeStatic } from '@open-wc/testing';
 import '../lion-dialog.js';
 
 describe('lion-dialog', () => {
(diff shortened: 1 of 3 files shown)


packages/dialog/test/lion-dialog.test.js
```diff
@@ -30,5 +29,30 @@ describe('lion-dialog', () => {
...
+            </lion-dialog>
+          </div>
+          <button slot="invoker">invoker button</button>
+        </lion-dialog>
+      `);
+
+      el._overlayInvokerNode.click();
+      expect(el.opened).to.be.true;
+
+      const wrapperNode = Array.from(document.querySelector('.global-overlays').children)[1];
```
jorenbroekema, line 52: I had to do it this way, because if you create a node reference for the inner dialog, it renders fully before being moved. But in a real scenario where you create this template, the inner dialog doesn't get fully rendered, because it gets moved earlier. There's a timing difference, and for the node reference you don't get a failing test, whereas for a nested template like this, you do get this failing test.

packages/overlays/src/OverlayMixin.js
```diff
@@ -147,8 +147,8 @@ export const OverlayMixin = dedupeMixin(
         if (this._overlayCtrl) {
           this.__tornDown = true;
           this.__overlayContentNodeWrapperBeforeTeardown = this._overlayContentNodeWrapper;
-          this._teardownOverlayCtrl();
         }
+        this._teardownOverlayCtrl();
```
jorenbroekema, line 151: We tear down regardless of whether there is an overlay controller yet, because we setup the open and close event listeners on connectedCallback, which happens before setting up the overlay controller. So it can (and does) happen where the element gets connected, but no overlay controller is created before being disconnected.

Issue #690, opened 2020-04-16
How to handle singletons


LarsDenBakker:
By using scoped elements it is possible to use multiple variants of lion in an app, is there a plan on how to make this work for the singletons? 

Will they guard their side effects? Or defer to the first registered singleton?

tlouisse:
Good that you bring this up. We just discussed this internally in the team last week. @daKmoR  has an idea to solve this via dependency injection. I hope we can get it on our backlog soon 🙂

LarsDenBakker:
Dependency injection would work for components (though we need to make sure that spawning lots of events isn't a performance problem). But that doesn't solve it for regular code which uses singletons, and it doesn't prevent singletons from being loaded in the first place.

tmsns:
Is the goal to make singletons work across version/instance (and thus actually create “globals”). 
If yes, what happens if the interface of that singleton/global changes? Should a new singleton/global be created in that case?

jorenbroekema:
Also, what if you have a breaking change, e.g. the name of method `msg` changed to `message`. If you resolve to the latest version, you will need to proxy the breaking changes somehow.

You could have a singleton manager that does this, as well as control which version of the singleton to use, and prevent instantiating multiple instances of essentially the same singleton class?

jorenbroekema:
Problems to solve:
- Multiple majors of singleton classes (possible with ScopedElementsMixin) could mean multiple instantiations of something that should only be instantiated once: localize, overlays, ajax, etc.
- Breaking changes between major versions of singletons need backwards/forwards compatibility
- App Developer currently has no control on how problem 1 and 2 are resolved

Possible solutions that need a proof of concept:
- Dependency Injection
- Singleton Manager (app level only, can never be more than 1, dont use scoped elements mixin to scope this)
- First singleton instance wins
- ... feel free to comment more ideas, I'll edit this comment to add them

LarsDenBakker:
One other option is to separately publish singletons, and use package managers to installs single versions.

Also good to look at the requirements, you may not need each singleton to be an actual singleton. Maybe for some of them the behavior can be slightly adjusted to allow multiple versions, or behavior can be extracted.

jorenbroekema:
I added some requirements for each of the packages that use singleton (I hope I didn't miss any).

jorenbroekema:
Here's another scenario on the extension layer which to me seems problematic no matter the reason why something is a singleton:

Feature A on v1 consumes utils which are on v2.
[...]

PR #703, opened 2020-05-06
feat: use singleton manager to support nested npm installations

daKmoR:
closes https://github.com/ing-bank/lion/issues/690

Outline how to use a singleton manager to make sure we stay a singleton even if multiple nested versions are installed by npm/yarn.

### DIFF
--- a/packages/ajax/src/ajax.js
+++ b/packages/ajax/src/ajax.js
@@ -1,9 +1,10 @@
+import { singletonManager } from 'singleton-manager';
 import { AjaxClass } from './AjaxClass.js';
 
 /**
  * @typedef {ajax} ajax the global instance for handling all ajax requests
  */
-export let ajax = AjaxClass.getInstance(); // eslint-disable-line import/no-mutable-exports
+export let ajax = singletonManager.get('@lion/ajax::ajax::0.3.x') || AjaxClass.getInstance(); // eslint-disable-line import/no-mutable-exports
 
 /**
  * setAjax allows the Application Developer to override the globally used instance of {@link:ajax}.
--- a/packages/singleton-manager/README.md
+++ b/packages/singleton-manager/README.md
@@ -1,0 +1,302 @@
+# Singleton Manager
+
+A singleton manager provides a way to make sure a singleton instance loaded from multiple file locations stays a singleton.
+Primarily useful if two major version of a package with a singleton is used.
+
+## How to use
+
+### Installation
+
+```sh
+npm i --save singleton-manager
+```
+
+⚠️ You need to make SURE that only ONE version of `singleton-manager` is installed. For how see [Non Goals](#non-goals).
+
+### Example Singleton Users
+
+Use the same singleton for both versions (as we don't use any of the breaking features)
+
+```js
+// managed-my-singleton.js
+import { singletonManager } from 'singleton-manager';
+import { mySingleton } from 'my-singleton'; // is available as 1.x and 2.x via node resolution
+
+singletonManager.set('my-singleton::index.js::1.x', mySingleton);
+singletonManager.set('my-singleton::index.js::2.x', mySingleton);
+```
+
+OR create a special compatible version of the singleton
+
+```js
+// managed-my-singleton.js
+import { singletonManager } from 'singleton-manager';
+import { MySingleton } from 'my-singleton'; // is available as 1.x and 2.x via node resolution
+
+class CompatibleSingleton extends MySingleton {
+  // add forward or backward compatibility code
+}
+const compatibleSingleton = new CompatibleSingleton();
+
+singletonManager.set('my-singleton::index.js::1.x', compatibleSingleton);
+singletonManager.set('my-singleton::index.js::2.x', compatibleSingleton);
+```
+
+AND in you App then you need to load the above code BEFORE loading the singleton or any feature using it.
+
+```js
+import './managed-my-singleton.js';
+
+import { mySingleton } from 'my-singleton'; // will no always be what is "defined" in managed-my-singleton.js
+```
+
+### Warning
+
+Overriding version is an App level concern hence components or "features" are not allowed to use it.
+If you try to call multiple times for the same key then it will throw an error.
+
+```js
+// on app level
+singletonManager.set('my-singleton/index.js::1.x', compatibleSingleton);
+
+// somewhere in a dependency
+singletonManager.set('my-singleton/index.js::1.x', compatibleSingleton);
+
+// will throw an error that it is already defined
+```
+
+### Example Singleton Maintainers
+
+If you are a maintainer of a singleton be sure to check if a singleton manager version is set.
+If that is the case return it instead of your default instance.
+
+It could look something like this:
+
+```js
+// my-singleton.js
+import { singletonManager } from 'singleton-manager';
+import { MySingleton } from './src/MySingleton.js';
+
+export const overlays =
+  singletonManager.get('my-singleton/my-singleton.js::1.x') || new MySingleton();
+```
+
+## Convention Singleton Key
+
+The key for a singleton needs to be "unique" for the package.
+Hence the following convention helps maintaining this.
+
+As a key use the `<package>::<unique-variable>::<semver-range>`.
+
+Examples Do:
+
+- `overlays::overlays::1.x` - instance created in index.js
+- `@scope/overlays::overlays::1.x` - with scope
+- `overlays::overlays::1.x` - version 1.x.x (> 1.0.0 you do 1.x, 2.x)
+- `overlays::overlays::2.x` - version 2.x.x (> 1.0.0 you do 1.x, 2.x)
+- `overlays::overlays::0.10.x` - version 0.10.x (< 1.0.0 you do 0.1.x, 0.2.x)
+
+Examples Don't:
+
+- `overlays` - too generic
+- `overlays::overlays` - you should include a version
+- `overlays::1.x` - you should include a package name & unique var
+- `./index.js::1.x` - it should start with a package name
+
+---
+
+## Singleton Manager Rationale
+
+We have an app with 2 pages.
+
+- page-a uses overlays 1.x
+- page-b uses overlays 2.x (gets installed nested)
+
+```txt
+my-app (node_modules)
+├── overlays (1.x)
+├── page-a
+│   └── page-a.js
+└── page-b
+    ├── node_modules
+    │   └── overlays (2.x)
+    └── page-b.js
+```
+
+The tough part in this case is the OverlaysManager within the overlays package as it needs to be a singleton.
+
+It starts of simplified like this
+
+```js
+export class OverlaysManager {
+  name = 'OverlayManager 1.x';
+  blockBody = false;
+  constructor() {
+    this._setupBlocker();
+  }
+  _setupBlocker() {
+    /* ... */
+  }
+  block() {
+    this.blockBody = true; // ...
+  }
+
+  unBlock() {
+    this.blockBody = false; // ...
+  }
+}
+```
+
+## Example A (fail)
+
+See it "fail" e.g. 2 separate OverlaysManager are at work and are "fighting" over the way to block the body.
+
+```bash
+npm run start:fail
+```
+
+Steps to reproduce:
+
+1. Page A click on block
+2. Page B => "Blocked: false" (even when hitting the refresh button)
+
+See [the code](./demo/fail/demo-app.js).
+
+---
+
+## Example B (singleton manager)
+
+The breaking change in `OverlayManager` was renaming of 2 function (which has been deprecated before).
+
+- `block()` => `blockingBody()`
+- `unBlock()` => `unBlockingBody()`
+
+knowing that we can create a Manager that is compatible with both via
+
+```js
+import { OverlaysManager } from 'overlays';
+
+class CompatibleOverlaysManager extends OverlaysManager {
+  blockingBody() {
+    this.block();
+  }
+  unBlockingBody() {
+    this.unBlock();
+  }
+}
+```
+
+all that is left is a to "override" the default instance of the "users"
+
+```js
+import { singletonManager } from 'singleton-manager';
+
+const compatibleOverlaysManager = new CompatibleOverlaysManager();
+singletonManager.set('overlays::overlays::1.x', compatibleOverlaysManager);
+singletonManager.set('overlays::overlays::2.x', compatibleOverlaysManager);
+```
+
+See it in action
+
+```bash
+npm run start:singleton
+```
+
+and [the code](./demo/singleton/demo-app.js).
+
+---
+
+## Example C (singleton and complex patching on app level)
+
+The breaking change in `OverlayManager` was converting a property to a function and a rename of a function.
+
+- `blockBody` => `_blockBody`
+- `block()` => `blockBody()`
+- `unBlock()` => `unBlockBody()`
+
+e.g. what is impossible to make compatible with a single instance is to have `blockBody` act as a property for 1.x and as a function `blockBody()` for 2.x.
+
+So how do we solve it then?
+
+We will make 2 separate instances of the `OverlayManager`.
+
+```js
+compatibleManager1 = new CompatibleManager1(); // 1.x
+compatibleManager2 = new CompatibleManager2(); // 2.x
+console.log(typeof compatibleManager1.blockBody); // Boolean
+console.log(typeof compatibleManager2.blockBody); // Function
+
+// and override
+singletonManager.set('overlays::overlays::1.x', compatibleManager1);
+singletonManager.set('overlays::overlays::2.x', compatibleManager2);
+```
+
+and they are "compatible" to each other because they sync the important data to each other.
+e.g. even though there are 2 instances there is only `one` dom element inserted which both can write to.
+When syncing data only the initiator will update the dom.
+This makes sure even though functions and data is separate it will be always consistent.
+
+See it in action
+
+```bash
+npm run start:singleton-complex
+```
+
+and [the code](./demo/singleton-complex/demo-app.js).
+
+---
+
+## How does it work?
+
+As a user you can override what the import of `overlays/instance.js` provides.
+You do this via a singletonManager and a "magic" string.
+
+- Reason be that you can target ranges of versions
+
+```js
+singletonManager.set('overlays::overlays::1.x', compatibleManager1);
+singletonManager.set('overlays::overlays::2.x', compatibleManager2);
+```
+
+### Potential Improvements
+
+Potentially we could have "range", "exacts version" and symbol for unique filename.
+So you can override with increasing specificity.
+If you have a use case for that please open an issue.
+
+## Non Goals
+
+Making sure that there are only 2 major versions of a specific packages.
+npm is not meant to handle it - and it never will
+
+```txt
+my-app
+├─┬ feat-a@x
+│ └── foo@2.x
+├─┬ feat-a@x
+│ └── foo@2.x
+└── foo@1.x
+```
+
+dedupe works by moving dependencies up the tree
+
+```txt
+// this app
+my-app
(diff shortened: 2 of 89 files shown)


packages/singleton-manager/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "dependency-injection",
```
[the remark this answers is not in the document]
tmsns: The comment was done before the implementation, but the review only send afterwards. This is fixed now. :-)

packages/singleton-manager/demo/singleton-complex/overlayCompatibility.js
```diff
@@ -0,0 +1,64 @@
+import { OverlaysManager } from 'overlays';
```
tmsns, line 1: As this compatibility file is located in the root of the app, is it the goal/guideline that the app (always) provides it for every singleton? Or only for singletons that have clashes?

I imagine that a v2 of a feature could also provide the compatibility layer?

packages/singleton-manager/demo/singleton-complex/node_modules/overlays/instance.js
```diff
@@ -0,0 +1,4 @@
+import { OverlaysManager } from './index.js';
+import { singletonManager } from '../../../../index.js';
+
+export const overlays = singletonManager.get('overlays/index.js::1.x') || new OverlaysManager();
```
tmsns, line 4: Would it be useful for features to add a default value in the singletonManager for the singletons they provide?

ie.
```javascript
singletonManager.set('overlays::index.js::1.x', new OverlaysManager());
export const overlays = singletonManager.get('overlays::index.js::1.x');
```

jorenbroekema:
My general impression is still very positive, same as when you presented it in the call. I've only reviewed the implementation commit so far, haven't gotten yet to the one where you apply it to lion.

My concern is a bit about "you need to make sure only 1 singleton manager is imported". Maybe a dumb question but what if a feature decides they need it, and then that feature is used by an app that also needs it? Like perhaps an application that is deployed by itself, but that can simultaneously be used as a widget? I'm a little worried that we're "delaying" the problem one step.

Also do you have a link / documentation on what that sideEffects field in package.json does ? Never seen it before.

tlouisse:
> My concern is a bit about "you need to make sure only 1 singleton manager is imported". Maybe a dumb question but what if a feature decides they need it, and then that feature is used by an app that also needs it?

@jorenbroekema I think we discussed this would be handled by releasing just one version of the package (that has an the api similar to a Map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).
There might still be caveats though (different bundles on the same page, yarn sometimes installing multiple versions of the same package even when versions exactly match).
Since the Map api is wrapped, this could be mitigated by writing to the window. 
But if we do that, we should probably do that from the first (preferable only) release of singletonManeger.

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+Use the same singleton for both versions (as we don't use any of the breaking features)
+
+```js
+// managed-my-singleton.js
+import { singletonManager } from 'singleton-manager';
+import { mySingleton } from 'my-singleton'; // is available as 1.x and 2.x via node resolution
+
+singletonManager.set('my-singleton/index.js::1.x', mySingleton);
+singletonManager.set('my-singleton/index.js::2.x', mySingleton);
```
jorenbroekema, line 26:
```suggestion
singletonManager.set('my-singleton::index.js::1.x', mySingleton);
singletonManager.set('my-singleton::index.js::2.x', mySingleton);
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+```js
+import './managed-my-singleton.js';
+
+import { mySingleton } from 'my-singleton'; // will no always be what is "defined" in managed-my-singleton.js
+```
+
+### Warning
+
+Overriding version is an App level concern hence components of "features" are not allowed to use it.
```
jorenbroekema, line 55:
```suggestion
Overriding version is an App level concern hence components or "features" are not allowed to use it.
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+// on app level
+singletonManager.set('my-singleton/index.js::1.x', compatibleSingleton);
+
+// somewhere in a dependency
+singletonManager.set('my-singleton/index.js::1.x', compatibleSingleton);
+
+// will throw an error that it is already defined
+```
+
+### Example Singleton Maintainers
```
jorenbroekema, line 68:
```suggestion
### Example Singleton Manager
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+import { singletonManager } from 'singleton-manager';
+import { MySingleton } from 'my-singleton'; // is available as 1.x and 2.x via node resolution
+
+class CompatibleSingleton extends MySingleton {
+  // add forward or backward compatibility code
+}
+const compatibleSingleton = new CompatibleSingleton();
+
+singletonManager.set('my-singleton/index.js::1.x', compatibleSingleton);
+singletonManager.set('my-singleton/index.js::2.x', compatibleSingleton);
```
jorenbroekema, line 42:
```suggestion
singletonManager.set('my-singleton::index.js::1.x', compatibleSingleton);
singletonManager.set('my-singleton::index.js::2.x', compatibleSingleton);
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+Examples Don't:
+
+- `overlays` - too generic
+- `overlays::overlays` - you should include a version
+- `overlays::1.x` - you should include a package name & unique var
+- `./index.js::1.x` - it should start with a package name
+
+---
+
+## Singleton Manager Rational
```
jorenbroekema, line 103:
```suggestion
## Singleton Manager Rationale
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+- `block()` => `blockingBody()`
+- `unBlock()` => `unBlockingBody()`
+
+knowing that we can create a Manager that is compatible with both via
+
+```js
+import { OverlaysManager } from 'overlays';
+
+class CompatibleManager extends OverlaysManager {
```
jorenbroekema, line 174:
```suggestion
class CompatibleOverlaysManager extends OverlaysManager {
```

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+all that is left is a to "override" the default instance of the "users"
+
+```js
+import { singletonManager } from 'singleton-manager';
+
+const manager = new CompatibleManager();
+singletonManager.set('overlays::overlays::1.x', compatibleManager);
+singletonManager.set('overlays::overlays::2.x', compatibleManager);
```
jorenbroekema, line 191:
```suggestion
const compatibleOverlaysManager = new CompatibleOverlaysManager();
singletonManager.set('overlays::overlays::1.x', compatibleOverlaysManager);
singletonManager.set('overlays::overlays::2.x', compatibleOverlaysManager);
```

jorenbroekema: Makes it a bit easier to read the docs for me, because manager is a little too generic

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+- `blockBody` => `_blockBody`
+- `block()` => `blockBody()`
+- `unBlock()` => `unBlockBody()`
+
+e.g. what is impossible to make compatible with a single instance is to have `blockBody` act as a property for 1.x and as a function `blockBody()` for 2.x.
+
+So how do we solve it then?
+
+We will make 2 separate instances of the `OverlayManager`.
```
jorenbroekema, line 216: This is really hard to understand. Why are the typeof the same manager -> blockbody different here? Why does making 2 instances help the problem? I think we should see the code inside these docs

packages/singleton-manager/README.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Potential Improvements
+
+Potentially we could have "range", "exacts version" and symbol for unique filename.
+So you can override with increasing specificity.
+If you have a use case for that please open an issue.
+
+## Non Goals
+
+Making sure that there are only 2 mayor versions of a specific packages.
```
jorenbroekema, line 264:
```suggestion
Making sure that there are only 2 major versions of a specific packages.
```

packages/singleton-manager/demo/singleton-complex/node_modules/overlays/index.js
```diff
@@ -0,0 +1,31 @@
...
+    this.blocker.style.backgroundColor = '#ff6161';
+  }
+
+  unBlock() {
+    this.blockBody = false;
+    this.blocker.style.backgroundColor = 'transparent';
+  }
+}
+
+export const overlaysId = Symbol('overlays');
```
jorenbroekema, line 31: ? what is this for

packages/singleton-manager/demo/singleton-complex/node_modules/page-a/page-a.js
```diff
@@ -0,0 +1,51 @@
+import { LitElement, html, css } from 'lit-element';
+import { overlays } from 'overlays/instance.js';
+
+export class PageA extends LitElement {
+  // getInstance(sym, fallback) {
```
jorenbroekema, line 5: related to the symbol I think, do we need to remove or is this useful code?

packages/singleton-manager/demo/singleton-complex/overlayCompatibility.js
jorenbroekema: Only for singletons that have clashes I guess. I do think it's our responsibility as maintainers of lion provide compatibility layers that our users can import, so we do some of the work for them already. 

So we would export based on 3 hypothetically breaking majors:
- OverlaysManager__OneTwo
- OverlaysManager__TwoOne
- OverlaysManager__ThreeOne
- OverlaysManager__OneThree

packages/singleton-manager/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "singleton-manager",
+  "version": "0.0.0",
```
jorenbroekema, line 3: What do we do about this? Do we wanna release it straight as 1.0.0 because we can never break it?

packages/singleton-manager/README.md
daKmoR: it's an example for people who create and maintain a singleton

packages/singleton-manager/README.md
daKmoR: 👍

packages/singleton-manager/package.json
daKmoR: yes... will do via a commit message with "BREAKING CHANGE" 👍

jorenbroekema:
@tlouisse  hmm yeah I forgot that npm will of course dedupe if you have a feature that also uses singleton manager. So importing multiple singleton manager should never happen.

To get back to my example though, if you had a feature (a widget) on v2 of overlays with a dependency on something that has a v1 of overlays, the feature will probably use the singleton manager to resolve the blocking problem. What if you use this widget in an application where you meet the same issue. Then the singleton manager will throw because you're trying to set already existing keys. Which I guess is fine and intended, but then the application needs to know about which features have already managed singleton conflicts. And also about which features have nested features that manage singleton conflicts.

My point is, we throw an error, but it may be tough for the application developer (the "top level developer") to understand where it's coming from. Perhaps having a super descriptive error is good enough.. But I'm also a bit worried that an application could break all of a sudden because a feature decides to use the singleton manager in a non breaking update. Maybe a softer warning is more friendly? Because nothing breaks if you try to set an existing key, everything should work just fine, so a hard error throw is maybe too much, since a dependency can mess things up for you without you even changing any code in your app.

packages/singleton-manager/demo/singleton-complex/overlayCompatibility.js
daKmoR: We can start with providing these compatibilities once we have 1.x

packages/singleton-manager/test/singleton-demo.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import '../demo/singleton/overlayCompatibility.js';
+
+import { overlays } from '../demo/singleton/node_modules/overlays/instance.js';
+import { overlays as overlays2 } from '../demo/singleton/node_modules/page-b/node_modules/overlays/instance.js';
+
+describe('singleton-demo', () => {
+  it('uses the compatibility overrides', async () => {
+    // Note: we can not test how it would work without applying the compatibility layer
+    //       as it is a global side effect.
```
daKmoR, line 11:
```suggestion
    //       as it is a global side effect and there is only one karma instance running.
```

packages/singleton-manager/README.md
jorenbroekema: Wouldn't it still be more clear then to write:
```md
### Example Singleton Manager

If you are a maintainer of a singleton.... etc.
```
?

packages/singleton-manager/demo/singleton-complex/demo-app.js
```diff
@@ -0,0 +1,93 @@
+import { LitElement, css, html } from 'lit-element';
```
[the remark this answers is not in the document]
daKmoR: added package.json files 👍

packages/singleton-manager/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "description": "Manage singletons",
```
jorenbroekema, line 4:
```suggestion
  "description": "Manage singletons across multiple major versions so they converge to a single instance",
```

packages/singleton-manager/demo/singleton-complex/node_modules/overlays/instance.js
daKmoR: no as these are app level overrides... e.g. it can only be set "once" (on app level) any additional settings to the same "key" will throw an error

packages/singleton-manager/demo/singleton-complex/node_modules/overlays/index.js
daKmoR: indeed not needed anymore -> removed

packages/singleton-manager/demo/singleton-complex/node_modules/page-a/page-a.js
daKmoR: removed

packages/singleton-manager/README.md
daKmoR: added more info 👍

daKmoR:
This will release the following versions:

e.g. all a feature release (breaking change in semver) + `singleton-manager: 1.0.0`

Full list:

```txt
 - @lion/ajax: 0.2.10 => 0.3.0
 - @lion/button: 0.5.12 => 0.6.0
 - @lion/calendar: 0.7.2 => 0.8.0
 - @lion/checkbox-group: 0.8.8 => 0.9.0
 - @lion/choice-input: 0.8.2 => 0.9.0
 - @lion/core: 0.5.2 => 0.6.0
 - @lion/dialog: 0.4.17 => 0.5.0
 - @lion/field: 0.12.2 => 0.13.0
 - @lion/fieldset: 0.10.2 => 0.11.0
 - @lion/form-system: 0.7.11 => 0.8.0
 - @lion/form: 0.4.23 => 0.5.0
 - @lion/helpers: 0.3.6 => 0.4.0
 - @lion/icon: 0.4.4 => 0.5.0
 - @lion/input-amount: 0.5.23 => 0.6.0
 - @lion/input-date: 0.5.23 => 0.6.0
 - @lion/input-datepicker: 0.11.6 => 0.12.0
 - @lion/input-email: 0.6.7 => 0.7.0
 - @lion/input-iban: 0.7.2 => 0.8.0
 - @lion/input-range: 0.2.23 => 0.3.0
 - @lion/input: 0.5.22 => 0.6.0
 - @lion/localize: 0.9.2 => 0.10.0
 - @lion/overlays: 0.13.2 => 0.14.0
 - @lion/radio-group: 0.8.8 => 0.9.0
 - @lion/select-rich: 0.15.3 => 0.16.0
 - @lion/select: 0.5.22 => 0.6.0
 - singleton-manager: 0.0.0 => 1.0.0
 - @lion/steps: 0.3.9 => 0.4.0
 - @lion/switch: 0.8.2 => 0.9.0
 - @lion/tabs: 0.2.9 => 0.3.0
 - @lion/textarea: 0.5.23 => 0.6.0
 - @lion/tooltip: 0.8.12 => 0.9.0
 - @lion/validate: 0.9.2 => 0.10.0
```

packages/ajax/stories/index.stories.mdx
```diff
@@ -4,7 +4,7 @@ import {
   html,
 } from '@open-wc/demoing-storybook';
 
-import { ajax } from '../src/ajax.js';
+import { ajax } from '../ajax.js';
```
jorenbroekema, line 7: This seems like the wrong path, because ajax.js is indeed inside ../src/

daKmoR: good catch 👍 fixed 💪

tlouisse:
> @tlouisse hmm yeah I forgot that npm will of course dedupe if you have a feature that also uses singleton manager. So importing multiple singleton manager should never happen.
> 
> To get back to my example though, if you had a feature (a widget) on v2 of overlays with a dependency on something that has a v1 of overlays, the feature will probably use the singleton manager to resolve the blocking problem. What if you use this widget in an application where you meet the same issue. Then the singleton manager will throw because you're trying to set already existing keys. Which I guess is fine and intended, but then the application needs to know about which features have already managed singleton conflicts. And also about which features have nested features that manage singleton conflicts.
> 
> My point is, we throw an error, but it may be tough for the application developer (the "top level developer") to understand where it's coming from. Perhaps having a super descriptive error is good enough.. But I'm also a bit worried that an application could break all of a sudden because a feature decides to use the singleton manager in a non breaking update. Maybe a softer warning is more friendly? Because nothing breaks if you try to set an existing key, everything should work just fine, so a hard error throw is maybe too much, since a dependency can mess things up for you without you even changing any code in your app.

Yeah, good points.
I think we make the assumption that only the app ("top level developer") is allowed to define the singletons and all features should just import the default version of that singleton they need.

But indeed, what if a feature depends on a feature? I think they should still import the versions of overlay manager, but only for testing and demoing purposes (in their index.html and tests (code that they don't publish in npm)) they should use the singletonManager.

@daKmoR is this what you had in mind and do you think it will be 100% bullet proof?

packages/ajax/src/ajax.js
```diff
@@ -1,9 +1,10 @@
+import { singletonManager } from 'singleton-manager';
 import { AjaxClass } from './AjaxClass.js';
 
 /**
  * @typedef {ajax} ajax the global instance for handling all ajax requests
  */
-export let ajax = AjaxClass.getInstance(); // eslint-disable-line import/no-mutable-exports
+export let ajax = singletonManager.get('@lion/ajax::ajax::0.2.x') || AjaxClass.getInstance(); // eslint-disable-line import/no-mutable-exports
```
daKmoR, line 7: hmmm when we release this will be 0.3.x 😱

daKmoR:
> But indeed, what if a feature depends on a feature? I think they should still import the versions of overlay manager they need, but only for testing and demoing purposes (in their index.html and tests (code that they don't publish in npm)) they should use the singletonManager.

indeed 👍

> @daKmoR is this what you had in mind and do you think it will be 100% bullet proof?

as long as we throw an error and "force" application to ensure non of their features override in a "production" code - yes 👍

packages/ajax/src/ajax.js
daKmoR: updated version strings

PR #704, opened 2020-05-07
fix(calendar): reset centralDate, focusedDate and disableDates

gerjanvangeest:
closes https://github.com/ing-bank/lion/issues/1338

### DIFF
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -325,6 +325,19 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
   }
 
   /**
+   * This exposes an interface for datepickers that want to
+   * reinitialize when calendar is opened
+   */
+  initCentralDate() {
+    if (this.selectedDate) {
+      this.focusSelectedDate();
+    } else {
+      this.centralDate = /** @type {Date} */ (this.__initialCentralDate);
+      this.focusCentralDate();
+    }
+  }
+
+  /**
    * @private
    */
   __calculateInitialCentralDate() {
@@ -334,6 +347,8 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     } else {
       this.__ensureValidCentralDate();
     }
+    /** @type {Date} */
+    this.__initialCentralDate = this.centralDate;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/calendar/src/LionCalendar.js
```diff
@@ -223,6 +223,11 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     }
   }
 
+  reset() {
+    this.centralDate = this.__initialCentralDate;
+    this._requestUpdate('centralDate');
```
tlouisse, line 228: Is this line needed? I would expect _requestUpdate to be called under the hood by UpdatingMixin from `lit-element`, since `centralDate` is already part of `static get properties`?

tlouisse:
Seems like a good approach 👍 I think the original bug report mentioned not only reset not being synchronized, but more like `datepicker.modelValue` in general.
We might also need a solution like: 'whenever `selectedDate` changes, centralDate is aligned to this'.  That should then also solve the reset issue I think.

packages/calendar/src/LionCalendar.js
```diff
@@ -413,6 +415,12 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     }
   }
 
+  __selectedDateChanged() {
+    if (!this.selectedDate && this.__initialCentralDate) {
+      this.centralDate = this.__initialCentralDate;
+    }
```
tlouisse, line 421: Think this is the right direction. I think there was also some complaint about the selectedDate not syncing when programatically set. We could add that here as well if needed

gerjanvangeest: There is a test to prevent this from happening: https://github.com/ing-bank/lion/blob/master/packages/calendar/test/lion-calendar.test.js#L144 
Can you remember why we have this test?

tlouisse: It makes sense that it doesn;t update while opened. It should however sync next time it's opened: https://jsfiddle.net/4Lbzyc1m/4/.  (open with chromium)

tlouisse: So it's safe to say that the calendar shouldn't do this by default, but the calendar should provide an interface to be used by the datepicker, so that:
```
if (!datepicker.opened) calender.syncCentralDateWithSelectedDate()
```

jorenbroekema:
@gerjanvangeest  @tlouisse  can we finish this PR or close it? Is something still missing?

tlouisse (approved):

PR #713, opened 2020-05-14
fix(tabs): tab keyboard navigation trap

### LINKED ISSUE #712: [tabs] trapped in tabs when focused
## Describe the bug

Tab is a trap!

When navigating with keyboard on tabs, you can't leave them with keyboard

## To Reproduce

1. Go to https://lion-web-components.netlify.app/?path=/story/navigation-tabs--default-story
2. Click on first tab
3. Try to navigate to anything else with _Tab_ key
4. You're trapped

## Expected behavior

Only arrow keys should be trapped

MathieuPuech:
Closes #712

### DIFF
--- a/packages/tabs/src/LionTabs.js
+++ b/packages/tabs/src/LionTabs.js
@@ -16,22 +16,22 @@ const deselectPanel = element => {
   element.removeAttribute('selected');
 };
 
-const setupButton = ({ element, uid, clickHandler, keydownHandler }) => {
+const setupButton = ({ element, uid, clickHandler, keydownHandler, keyupHandler }) => {
   element.setAttribute('id', `button-${uid}`);
   element.setAttribute('role', 'tab');
   element.setAttribute('aria-controls', `panel-${uid}`);
   element.addEventListener('click', clickHandler);
-  element.addEventListener('keyup', keydownHandler);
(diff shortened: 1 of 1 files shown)


packages/tabs/src/LionTabs.js
```diff
@@ -51,6 +51,17 @@ const deselectButton = element => {
   element.setAttribute('tabindex', -1);
 };
 
+const handleButtonKeydown = e => {
+  switch (e.key) {
+    case 'ArrowDown':
+    case 'ArrowRight':
+    case 'ArrowUp':
+    case 'ArrowLeft':
+      e.preventDefault();
```
gerjanvangeest, line 60: I'm missing `home` & `end`.

packages/tabs/src/LionTabs.js
```diff
@@ -168,7 +180,7 @@ export class LionTabs extends LitElement {
     };
   }
 
-  __handleButtonKeydown(e) {
+  __handleButtonKeyup(e) {
     switch (e.key) {
       case 'ArrowDown':
       case 'ArrowRight':
```
gerjanvangeest, line 186: I believe all `e.preventDefault`s can be removed from the keyup handler

PR #719, opened 2020-05-19
Feat pattern validator

takrishna:
1. Added IsPattern validator for strings validators
2. Modified - Default Translation files for the error message
3. Added tests 

#My first attempt to contribute - #caution

### DIFF
--- a/packages/validate/src/validators/StringValidators.js
+++ b/packages/validate/src/validators/StringValidators.js
@@ -89,3 +89,25 @@ export class IsEmail extends Validator {
     return hasError;
   }
 }
+
+const hasPattern = (value, pattern) => pattern.test(value);
+export class Pattern extends Validator {
+  static get validatorName() {
+    return 'Pattern';
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  execute(value, pattern = this.param) {
+    if (!(pattern instanceof RegExp)) {
+      throw new Error(
+        'Psst... Pattern validator expects RegExp object as parameter e.g, new Pattern(/#LionRocks/) or new Pattern(RegExp("#LionRocks")',
+      );
+    }
+    let hasError = false;
+    if (!isString(value) || !hasPattern(value, pattern)) {
+      hasError = true;
+    }
+
+    return hasError;
+  }
+}
--- a/packages/validate/test/StringValidators.test.js
+++ b/packages/validate/test/StringValidators.test.js
@@ -7,6 +7,7 @@ import {
   MaxLength,
   MinMaxLength,
   IsEmail,
+  Pattern,
 } from '../src/validators/StringValidators.js';
 
 describe('String Validation', () => {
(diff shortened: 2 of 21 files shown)


tlouisse:
Looks good! 
Nitpicking: I think the name `isPattern` is not completely correct, because you test the modelValue _against_ a pattern instead of testing whether the modelValue _is_ a pattern.

I think just `Pattern` would be a good name for this validator.

takrishna:
Agreed, "Pattern" is simple and easy - I was originally considering "IsAPatternOf" to be verbose .. phew.. !

packages/validate/src/validators/StringValidators.js
```diff
@@ -89,3 +89,19 @@ export class IsEmail extends Validator {
     return hasError;
   }
 }
+
+const isPattern = (value, pattern) => pattern.test(value);
```
tlouisse, line 93: And one more nitpick: maybe call this `matchesPattern` or `hasPattern` then?

packages/validate/test/StringValidators.test.js
```diff
@@ -111,4 +112,16 @@ describe('String Validation', () => {
     isEnabled = validator.execute('foo@120.120.120.93');
     expect(isEnabled).to.be.true;
   });
+
+  it('provides new Pattern() to allow only valid patterns', () => {
+    let isEnabled;
+    const validator = new Pattern(/#LionRocks/);
+    expect(validator.constructor.name).to.equal('Pattern');
```
tlouisse, line 119: We should test for `validator.constructor.validatorName`.
(We probably do this wrong in the other tests as well (we couldn't use `constructor.name` because it gets lost in code minification/transpilation) )

packages/validate/translations/it.js
```diff
@@ -6,6 +6,7 @@ export default {
     MaxLength: 'Inserire un(a) {fieldName} corretto(a) (fino a {params} caratteri).',
     MinMaxLength:
       'Inserire un(a) {fieldName} corretto(a) (tra {params.min} e {params.max} caratteri).',
+    Pattern: 'Kérjük, adjon meg érvényes {fieldName} értéket.',
```
gerjanvangeest, line 9: I believe you copied here the Hungarian translation, instead of the Italian `IsNumber`

packages/validate/translations/sk.js
```diff
@@ -5,6 +5,7 @@ export default {
     MinLength: 'Uveďte správne {fieldName} (najmenej {params}).',
     MaxLength: 'Uveďte správne {fieldName} (maximálne {params} znakov).',
     MinMaxLength: 'Uveďte správne {fieldName} ({params.min} až {params.max} znakov).',
+    Pattern: 'Введите действительное значение поля {fieldName}.',
```
gerjanvangeest, line 8: I believe this is Russian

tlouisse:
I think if you fix the suggestions from @gerjanvangeest and put everything in one commit, it will be good to merge 💪

packages/validate/translations/it.js
takrishna: trust me I am better than this :D - fixed

takrishna:
> I think if you fix the suggestions from @gerjanvangeest and put everything in one commit, it will be good to merge 💪

Fixed and Squashed into one commit

daKmoR:
mind you adding a throw later is a breaking change 😅

while removing a throw is find... so better add now then later... 🤔

jorenbroekema:
Thanks for contributing :)!

PR #720, opened 2020-05-19
fix: add brief explanation of hasFeedbackFor and showsFeedbackFor

### LINKED ISSUE #714: Documentation on field validation states
## Describe the bug

There currently seems to be a gap in the documentation for checking the validity of a field. From a conversation on the polymer slack it seems that using hasFeedbackFor/showsFeedbackFor are what should be used.

## To Reproduce

1. Go to https://lion-web-components.netlify.app/?path=/docs/forms-validation-overview--page

## Expected behavior
I would expect a section showing how to get these states for a field

An example given in the slack group was `el.hasFeedbackFor.includes('error')` to get whether or not a field is valid. 

##Additional Context
https://polymer.slack.com/archives/CJGFWJN9J/p1589539615208400

j--w:
Fixes #714 

Add short explanation/example of using `hasFeedbackFor` or `showsFeedbackFor` in JS code to determine a fields validity.

### DIFF
--- a/packages/validate/stories/Overview.stories.mdx
+++ b/packages/validate/stories/Overview.stories.mdx
@@ -22,6 +22,24 @@ not always lead to the desired User Experience.
 Together with [interaction states](?path=/docs/forms-system-interaction-states--interaction-states),
 validity states can determine whether a validation message should be shown along the input field.
 
+## Assessing field validation state programmatically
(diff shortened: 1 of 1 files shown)


jorenbroekema:
Awesome thanks :)!

PR #726, opened 2020-05-26
[tooltip] accessibility fixes

tlouisse:
Fixes:
- clean up a11y attrs for all types of overlays on teardown
- enables a11y in lion-tooltip
- expose tooltip prop `invokerRelation` and connect to OverlayCtrl

closes https://github.com/ing-bank/lion/issues/754

### DIFF
--- a/packages/tooltip/src/LionTooltip.js
+++ b/packages/tooltip/src/LionTooltip.js
@@ -12,6 +12,10 @@ export class LionTooltip extends OverlayMixin(LitElement) {
         reflect: true,
         attribute: 'has-arrow',
       },
+      invokerRelation: {
+        type: String,
+        attribute: 'invoker-relation',
+      },
     };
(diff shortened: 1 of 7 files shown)


packages/tooltip/src/LionTooltip.js
```diff
@@ -70,6 +74,17 @@ export class LionTooltip extends OverlayMixin(LitElement) {
...
     super();
+    /**
+     * Whether an arrow should be displayed
+     * @type {boolean}
+     */
+    this.hasArrow = false;
+    /**
+     * Decides whether the tooltip invoker text should be considered a description
+     * (sets aria-describedby) or a label (sets aria-labelledby).
+     * @type {'label'\'description'}
```
tlouisse, line 85: was my own code, but '\' should be '|'

PR #728, opened 2020-05-26
fix(select-rich): prevent scrolling when focused

### LINKED ISSUE #724: [select-rich] missing preventDefault on ArrowUp and ArrowDown
## Describe the bug

When using up and down keys on a select-rich element focused, the page is scrolling.

## To Reproduce

1. Go to https://lion-web-components.netlify.app/?path=/docs/forms-select-rich--default-story
2. Focus on the first select-rich element
3. Change value with down key
4. The value is changed and the page is scrolled

## Expected behavior

Page should not scroll

## Additional context

On windows 10 Chrome Version 81.0.4044.138

MathieuPuech:
Closes #724

### DIFF
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -69,6 +69,30 @@ export class LionSelectInvoker extends LionButton {
     this.type = 'button';
   }
 
+  connectedCallback() {
+    if (super.connectedCallback) {
+      super.connectedCallback();
+    }
+
+    const handleKeydown = event => {
+      switch (event.key) {
+        case 'ArrowDown':
+        case 'ArrowUp':
+          event.preventDefault();
+        /* no default */
+      }
+    };
+    this.handleKeydown = handleKeydown;
+    this.addEventListener('keydown', this.handleKeydown);
+  }
+
+  disconnectedCallback() {
+    if (super.disconnectedCallback) {
+      super.disconnectedCallback();
+    }
+    this.removeEventListener('keydown', this.handleKeydown);
+  }
+
   _contentTemplate() {
     if (this.selectedElement) {
       const labelNodes = Array.from(this.selectedElement.querySelectorAll('*'));


packages/select-rich/src/LionSelectRich.js
```diff
@@ -388,6 +399,7 @@ export class LionSelectRich extends ScopedElementsMixin(
 
     this._listboxNode.addEventListener('active-changed', this.__onChildActiveChanged);
     this._listboxNode.addEventListener('model-value-changed', this.__proxyChildModelValueChanged);
+    this.addEventListener('keydown', handleKeydown);
```
jorenbroekema, line 402: Could we move this to LionSelectRichInvoker? Because that's the actual `activeElement` when you tab into the rich select. Right now this "works" because LionSelectRich happens to be a parent, so the listener is triggered, but I think it would be better to move it to the invoker where the event happens.

packages/select-rich/src/LionSelectRich.js
```diff
@@ -38,6 +38,17 @@ function isInView(container, element, partial = false) {
   return isTotal || isPartial;
 }
 
+function handleKeydown(event) {
+  switch (event.key) {
+    case 'ArrowDown':
+    case 'ArrowUp':
+    case 'Home':
```
jorenbroekema, line 45: I would argue Home and End keys should not be prevent defaulted here. They have no meaningful interaction on the invoker, only if you are inside the listbox. And on the listbox node those keys are already prevent defaulted. Here on the invoker they should act as normal and scroll the user down right?

jorenbroekema:
My final code would probably look like this inside the LionSelectRichInvoker:

```js
connectedCallback() {
  if (super.connectedCallback) {
    super.connectedCallback();
  }

  const handleKeydown = event => {
    switch (event.key) {
      case 'ArrowDown':
      case 'ArrowUp':
        event.preventDefault();
      /* no default */
    }
  };
  this.boundHandleKeydown = handleKeydown.bind(this);
  this.addEventListener('keydown', this.boundHandleKeydown);
}

disconnectedCallback() {
  if (super.disconnectedCallback) {
    super.disconnectedCallback();
  }
  this.removeEventListener('keydown', this.boundHandleKeydown);
}
```

packages/select-rich/src/LionSelectRich.js
MathieuPuech: Yes it's working when moved in the invoker

packages/select-rich/src/LionSelectRich.js
MathieuPuech: Yes it's working when moved in the invoker

jorenbroekema:
Awesome! Thanks for contributing this fix :)

PR #731, opened 2020-05-27
fix(select-rich): set inherits reference width to min

jorenbroekema:
Fixes:
Screenshot 2020-05-27 at 10 36 00

To:
Screenshot 2020-05-27 at 11 33 02

Overriding this inheritsReferenceWidth for noDefaultSelected still works as before, but is now less useful as a default because we now set the width to 'min' by default already (before it was 'full').

### DIFF
--- a/packages/select-rich/test/lion-select-rich.test.js
+++ b/packages/select-rich/test/lion-select-rich.test.js
@@ -400,6 +400,23 @@ describe('lion-select-rich', () => {
       expect(elSingleoption.opened).to.be.false;
     });
 
+    it('sets inheritsReferenceWidth to min by default', async () => {
+      const el = await fixture(html`
+        <lion-select-rich name="favoriteColor" label="Favorite color">
+          <lion-options slot="input">
+            <lion-option .choiceValue=${'red'}>Red</lion-option>
+            <lion-option .choiceValue=${'hotpink'}>Hotpink</lion-option>
+            <lion-option .choiceValue=${'teal'}>Teal</lion-option>
+          </lion-options>
+        </lion-select-rich>
+      `);
+
+      expect(el._overlayCtrl.inheritsReferenceWidth).to.equal('min');
+      el.opened = true;
+      await el.updateComplete;
+      expect(el._overlayCtrl.inheritsReferenceWidth).to.equal('min');
+    });
+
     it('should override the inheritsWidth prop when no default selected feature is used', async () => {
       const el = await fixture(html`
         <lion-select-rich name="favoriteColor" label="Favorite color" has-no-default-selected>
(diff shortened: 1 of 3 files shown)


packages/select-rich/stories/index.stories.mdx
```diff
@@ -312,17 +312,11 @@ It is up to you how to render this Object in the DOM.
       );
```
jorenbroekema, line 312: My Prettier went kinda nuts on this. Might be nice to change: `"format:prettier": "prettier \"**/*.{js,md}\" --write",` to `"format:prettier": "prettier \"**/*.{js,md,mdx}\" --write",`, we already have it on our extension layer since last week and it works pretty well

jorenbroekema: Soon we won't have .mdx though so I guess it's not worth the effort now :P

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -410,10 +427,14 @@ describe('lion-select-rich', () => {
           </lion-options>
         </lion-select-rich>
       `);
+      // The default is min, so we override that behavior here
+      el._overlayCtrl.inheritsReferenceWidth = 'full';
```
tlouisse, line 431: We should probably always use `updateConfig` as a best practice. 
I think those props that are directly available on the controller are mainly there for backwards compatibiliy and if we want to keep them, they should be readonly.

Setting props on overlayCtrl directly is not guaranteed to properly reload init/teardown phases etc.

packages/select-rich/stories/index.stories.mdx
```diff
@@ -312,17 +312,11 @@ It is up to you how to render this Object in the DOM.
...
     }
     return html`
-      <lion-select-rich
-        label="Credit Card"
-        name="color"
-        @select-model-value-changed=${showOutput}
-      >
+      <lion-select-rich label="Credit Card" name="color" @select-model-value-changed=${showOutput}>
```
tlouisse, line 315: Do we have our Prettier settings aligned well across all our contributors? I also tend to get those diffs a lot lately.

packages/select-rich/test/lion-select-rich.test.js
jorenbroekema: Yeah but it's a test so it doesn't matter right? All I wanna test here is if the hasNoDefault case overrides the inherits width but I can't properly do this if the initial inherits width is the same width as the override

packages/select-rich/stories/index.stories.mdx
jorenbroekema: Yes, but not for .mdx because that file ext is not added to the `yarn format:prettier` script, I wanted to add it but we will soon not have any .mdx files so I figured let's leave it be for now.

tlouisse: Ah, so I guess it will be fixed once we move to mdjs and makes this an incidental one 👍

packages/select-rich/test/lion-select-rich.test.js
tlouisse: Well, of course we should use the right public apis preferrably when possible, since tests also serve as examples.
So far we don't do it (change props directly on the controller) in our demos/tests at least, but I agree this should be better documented: we should explicitly say that changing a config like this doesn't trigger the right init and teardown actions.

Don't want to nitpick too much, so if you want to do this later, that's fine 👍

PR #736, opened 2020-05-28
fix: throw when passing unconnected contentNode

jorenbroekema:
This was already not working and would give you problems, now we explicitly mention why it fails by throwing an error

### DIFF
--- a/packages/overlays/test/OverlayController.test.js
+++ b/packages/overlays/test/OverlayController.test.js
@@ -1,20 +1,20 @@
 /* eslint-disable no-new */
+import '@lion/core/test-helpers/keyboardEventShimIE.js';
 import {
-  expect,
-  html,
-  fixture,
   aTimeout,
   defineCE,
-  unsafeStatic,
+  expect,
+  fixture,
+  html,
   nextFrame,
+  unsafeStatic,
 } from '@open-wc/testing';
 import { fixtureSync } from '@open-wc/testing-helpers';
-import '@lion/core/test-helpers/keyboardEventShimIE.js';
 import sinon from 'sinon';
-import { keyCodes } from '../src/utils/key-codes.js';
-import { simulateTab } from '../src/utils/simulate-tab.js';
 import { OverlayController } from '../src/OverlayController.js';
 import { overlays } from '../src/overlays.js';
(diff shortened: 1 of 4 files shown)


packages/overlays/test/OverlayController.test.js
```diff
@@ -153,6 +153,30 @@ describe('OverlayController', () => {
...
+      it('throws when passing a content node that was created "offline"', async () => {
+        const contentNode = document.createElement('div');
+        const createOverlayController = () => {
+          new OverlayController({
+            ...withLocalTestConfig(),
+            contentNode,
+          });
+        };
+        expect(createOverlayController).to.throw(
+          'The contentNode you passed is not connected anywhere in the DOM',
```
daKmoR, line 166:
```suggestion
          'The OverlayController could not find a render target as the provided contentNode "div" is not connected to the DOM. Make sure to use "[parentNode].appendChild(contentNode)" before passing it on.',
```

daKmoR: imho when you have multiple error messages in an app it's rather important to have a clear path on finding where this error belongs to

jorenbroekema: Hmm bit confusing maybe to the average user, maybe:
```
The OverlayController could not find a render target, since the provided contentNode is not connected to the DOM. Make sure that it is connected, e.g. by doing "document.body.appendChild(contentNode)", before passing it on.
```
What do you think?

tlouisse: If we do it, we should do it consistently throughout our code base, right?
In order to keep messages succint we could do:

```
[OverlayController] The contentNode you passed is (etc...)
```
?

jorenbroekema: Probably a good idea indeed, let's start this trend here and maybe a contributor can do it everywhere else. Trend being: errors thrown should be prepended with `[ClassName] ` whenever possible (if not thrown inside a class, we can use file name or function name or something else that's descriptive)

jorenbroekema: Changed it

daKmoR:
will merged to dev first 👍

PR #737, opened 2020-05-28
fix(button): aria-disabled for button and switch

### LINKED ISSUE #722: [button] Disabled button can be focused & activated when using a screen reader
## Describe the bug

Disabled buttons can be focused & activated by navigating through the form controls with a screen reader.

## To Reproduce

- Go to the lion demo page on https://lion-web-components.netlify.app/?path=/docs/buttons-button--disabled
- Open your screen reader (e.g. VoiceOver on Mac)
- Navigate to the disabled button (without using the tab key)
- The screen reader will read out "<label>, button" instead of the expected "<label>, dimmed, button"
- Activate the button
- The screen reader will read out "press <label>, button" instead of the expected "dimmed"

## Expected behavior

Same behavior as with a disabled `<button>` element. Users should not be able to activate it.

## Screenshots

Expected behavior on focus:
image

Actual behavior on focus:
image

Expected behavior on activate:
image

Actual behavior on activate:
image

## Proposed solution

Make the `lion-button` an actual `button` element instead of a `div`.

Or add the attribute `aria-disabled="true"` when `disabled` is true.
I have tried the latter in the browser's dev tools and it gave the expected behavior. ([more info on aria-disabled](https://a11y-101.com/development/aria-disabled))

## Additional context

Tested with macOS built-in VoiceOver application on

- Chrome
- Firefox
- Safari

Hzunax:
Fixes #722

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -184,6 +184,9 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
         native.type = this.type;
       }
     }
+    if (changedProperties.has('disabled')) {
+      this.setAttribute('aria-disabled', `${this.disabled}`); // create mixin if we need it in more places
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


jorenbroekema:
Thanks! :)

PR #738, opened 2020-05-28
fix(field): remove validation toggled disable

### LINKED ISSUE #705: [validation] validation is not removed when disabled state toggles
## Describe the bug

The scenario is 
   1) Required validator on the input field
   2) Disabled state toggles on property change
   3) Model value is empty by default.

In the above scenario, when the input becomes enabled and if it is not filled, it shows proper validation message.
But when the input is disabled again, the message disappears but the input is still in invalid state.
This **forces** user to enable the input and provide value, even when they want to disable it.

## To Reproduce
```js
function toggleDisabled() {
  const fieldset = document.querySelector('#fieldset');
  fieldset.disabled = !fieldset.disabled;
}

function submitForm(){
  const formValue=document.querySelector('#form');
  if (!formValue.hasFeedbackFor.includes('error')) {
    console.log("No error");
  }
}
```
render:
```html
<lion-form id="form">
  <form>
  <lion-fieldset name="nameGroup" label="Name" id="fieldset" disabled>
      <lion-input
        name="FirstName"
        label="First Name"
        .modelValue=${'Foo'}
        disabled
      ></lion-input>
      <lion-input name="LastName" label="Last Name" .validators="${[new Required()]}"></lion-input>
  </lion-fieldset>
  <button type="button" @click=${submitForm}>Submit</button>
  </form>
</lion-form>
<button @click=${toggleDisabled}>
  Toggle disabled
</button>
```

Initially form will be in disable state. And steps to reproduce
1) click on submit button, then it displays "No error" in console.
2) click on toggle disabled, last name is not provide, so it gives validation message
3) Again click on toggle disabled, to move the form to disable state.
4) And on click of submit button again, "No error" is not displayed in console.

## Expected behavior

When the input switches back to disabled state, then validation on it should be removed and the console should also display "No error" as it passes the validation.

Hzunax:
Closes #705 

When a field is enabled and then toggled to disabled while being invalid, it still had the `validationStates` and `hasFeedbackFor` properties set from when it was enabled.

Calling `this.__finishValidation` synchronously after `this.__clearValidationResults`, ensures that those properties are reset.

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -265,7 +265,7 @@ export const ValidateMixin = dedupeMixin(
       async validate({ clearCurrentResult } = {}) {
         if (this.disabled) {
           this.__clearValidationResults();
-          this.__validationResult = [];
+          this.__finishValidation({ source: 'sync', hasAsync: true });
           this._updateFeedbackComponent();
           return;
         }
--- a/packages/form-core/test/lion-field.test.js
+++ b/packages/form-core/test/lion-field.test.js
@@ -380,7 +380,36 @@ describe('<lion-field>', () => {
       expect(el.validationStates.error).to.have.a.property('HasX');
 
       expect(disabledEl.hasFeedbackFor).to.deep.equal([]);
-      expect(disabledEl.validationStates.error).to.equal(undefined);
+      expect(disabledEl.validationStates.error).to.deep.equal({});
+    });
+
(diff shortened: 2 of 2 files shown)


jorenbroekema:
Hey, we recently merged something quite huge, so a rebase on master is required. Fix looks good though, thanks for taking it on!

Hzunax:
Rebased on master. Thanks for the heads-up!

jorenbroekema:
Thanks a lot 😁

PR #743, opened 2020-06-02
feat(localize): allow overriding of translations

igomezal:
This pull request allows the possibility to override the translations defined by a component so it can be fully reused.


jorenbroekema:
Hey, I will close this PR, we won't implement this directly right now but we are going to do a refactor of the localize system where we will take this use case into account. https://github.com/ing-bank/lion/discussions/787 it can be tracked for now in our v1 release tracker. Overriding translations directly in some form or another for a certain namespace is definitely a requirement that I think we should add :) so thanks for the PR, we can take it as food for thought on how to do that

PR #746, opened 2020-06-02
fix(overlays): body resize on global overlay open

### LINKED ISSUE #745: [overlays] When a dialog is opened, the body moved
## Describe the bug

There is a movement in body when opening a dialog because scrollbars are not here when the dialog is opened (overflow of body is hidden).

## To Reproduce

1. Go to https://lion-web-components.netlify.app/?path=/docs/overlays-dialog--main
2. Open a dialog

## Expected behavior

Body size should not change

## Additional context

The bug is not visible on OSX because scrollbars does not take width on OSX.

MathieuPuech:
Closes #745


daKmoR:
awesome thxxxxxxx 🤗

PR #748, opened 2020-06-03
Chore/demo updates

jorenbroekema:


packages/ajax/README.md
```diff
@@ -13,10 +13,6 @@ export default {
 };
 ```
 
-## Live Demo/Documentation
```
gerjanvangeest, line 16: I thought we would leave these lines in all the READMEs, but remove them in a `./stories/all.overrides.md` file for storybook.

jorenbroekema: We don't have them in the overrides files, we only have them here in lion. https://lion-web-components.netlify.app/?path=/docs/forms-input--main this will show you the awkwardness of having the link in lion, because it links to itself everywhere

gerjanvangeest: But having them in the [./packages/**/README.md](https://github.com/ing-bank/lion/tree/master/packages/input/README.md) still makes sense to me.

So why not adding it to `./stories/all.overrides.md`?

jorenbroekema: because then they show up in your storybook, and you get the awkwardness. Mind that README.md and Stories are now the exact same thing.

PR #752, opened 2020-06-04
chore: fix more links for storybook

jorenbroekema:


packages/form-integrations/test/form-integrations.test.js
```diff
@@ -1,7 +1,7 @@
 import { expect, fixture, html } from '@open-wc/testing';
 import './helpers/umbrella-form.js';
 
-// Test umbrella form
+// Test umbrella form.
```
tlouisse, line 4: Is this a linter or a newly proposed standard ? :) (the dot)

jorenbroekema: Making sure this package gets released (.md gets ignored by lerna but we actually want it to be released so I can upgrade to latest lion in our extension and extend the .md)

PR #766, opened 2020-06-11
chore: use web test runner

LarsDenBakker:
Testing out web test runner on lion, making a PR to test the CI.

There are test failures, they're doing some pretty fancy DOM stuff and probably relied on some environment in Karma that's not there here. Would be good to get some extra 👀 on that if anyone has some time to spare :)


jorenbroekema:
I'll have time to look into this tomorrow afternoon I think :) should be able to help at least with those nested overlays tests and button tests

package.json
```diff
@@ -42,6 +42,7 @@
     "@open-wc/testing-karma-bs": "^1.3.30",
```
daKmoR, line 42: should be removed right?
- `@open-wc/testing-karma`
- `@open-wc/testing-karma-bs`

daKmoR: also karma.config files

jorenbroekema: also what do we do with test:bs and test:prune-snapshots and test:update-snapshots

daKmoR: we don't use browserstack anyways... so remove

do we have failing snapshots tests? are we using shapshots? if no then remove

tlouisse:
Looks promising 💪 

It would be nice if we also fix all 'collateral' errors (so not failing tests, but uncaught errors). 
My guess is that we can do this by guarding/rejecting 'outstanding Promises' on disconnectedCallback...

jorenbroekema:
Would it be acceptable to do it in a separate PR? and get this merged. I'm getting a little tired of karma for multiple reasons 🙈

tlouisse:
> Would it be acceptable to do it in a separate PR? and get this merged. I'm getting a little tired of karma for multiple reasons 🙈

Would be fine with me. Only generates a lot of 'noise' in your console and ci now.
So that's why we need to fix it soon I think.
Will make a task for that...

tlouisse:
One more question @jorenbroekema / @LarsDenBakker : will it be possible still to use Browserstack later?

LarsDenBakker:
@tlouisse yep: https://github.com/modernweb-dev/web/tree/master/packages/test-runner-browserstack

PR #767, opened 2020-06-12
test: add visual regression testing with puppeteer

MathieuPuech:


package.json
```diff
@@ -19,20 +19,29 @@
...
     "husky": "^1.0.0",
     "lerna": "3.4.3",
     "lint-staged": "^8.0.0",
+    "looks-same": "^7.2.3",
     "markdownlint-cli": "^0.17.0",
+    "minimist": "^1.2.5",
+    "mkdirp-promise": "^5.0.1",
     "mocha": "^7.1.1",
     "npm-run-all": "^4.1.5",
+    "pixelmatch": "^5.2.0",
```
MathieuPuech, line 42: should be remove, not used

MathieuPuech:
inspiration from https://github.com/jakwuh/ava-puppeteer

test/screenshots/index.js
```diff
@@ -0,0 +1,265 @@
...
+ * @param {string} root
+ * @param {string} id
+ * @param {string} selector
+ * @return {{file: string, folder: string, path: string}}
+ */
+async function buildPath({ root, id, selector }) {
+  log(`Building path for ${id} - ${selector}`);
+
+  const [title, file] = id.split('--');
+  const [category, component] = title.split('-');
```
MathieuPuech, line 36:
```suggestion
  const [, category, component, file] = /(\w*)-(.*)--(.*)$/g.exec(id);
```

jorenbroekema:
Heads up, we are done with migrating to web test runner from karma now, I'm not sure if this PR is affected much by it, but let's see if we can rebase and get this merged, because the setup looks good to me :) and for things like buttons and rich select and probably a few other things like tabs, calendar and overlays, it's quite useful to do screenshot testing in order to flag visual regressions.

MathieuPuech:
I will try using the server and mocha. Using the puppeteer launcher is not yet possible on wtr now because the pages are not exposed. For screenshots test, it needs to have access to pages to make screenshots with puppeteer.

jorenbroekema:
Hmm I see, could you run it concurrently with es-dev-server which serves storybook-static folder? Or maybe I misunderstand the problem. I think for screenshot testing there may not be that much extra value to run them on WTR, I think running them with mocha and by having a before hook that starts es-dev-server instance is actually good enough :). So maybe WTR doesn't affect this PR that much after all.

MathieuPuech:
I made a new updated PR with a branch on the repo here: https://github.com/ing-bank/lion/pull/849

PR #771, opened 2020-06-18
feat(providence): add providence package

jorenbroekema:


packages/providence/README.md
```diff
@@ -0,0 +1,215 @@
...
+npm i --save-dev providence
+```
+
+### Add a providence script to package.json
+
+```js
+...
+"scripts": {
+    ...
+    "providence": "providence analyze match-imports -r 'node_modules/lion-based-ui'",
```
jorenbroekema, line 22: The reference project would usually be node_modules/@lion/package right? lion-based-ui is the target usually right where you run providence in its root?

I wonder if in general we should have a command/flag that supports multi-packages, `-r 'node_modules/@lion' --mono-repo` or something, I dunno. Maybe not in the scope of this story though.

packages/providence/providence.conf.js
```diff
@@ -0,0 +1,98 @@
+// This file is read by dashboard and cli and needs to be present under process.ced()
```
jorenbroekema, line 1: process.cwd()?

packages/providence/README.md
```diff
@@ -0,0 +1,215 @@
+# Providence
+
+Providence is the 'All Seeing Eye' that generates usage statistics by analyzing code.
+It measures the effectivity and popularity of your software.
+In just a few clicks you can measure the impact for (breaking) changes; making
+your release process more stable and predictable.
+
```
jorenbroekema, line 7: I think we should add a small section about who the main user is, and an example analysis of what they can do.

packages/providence/README.md
jorenbroekema: Edit: Maybe I'm confused about what the perspective is in these docs.

packages/providence/src/cli/generate-extend-docs-data.js
```diff
@@ -0,0 +1,43 @@
...
+const { QueryService } = require('../program/services/QueryService.js');
+const { LogService } = require('../program/services/LogService.js');
+
+async function launchProvidenceWithExtendDocs(referencePaths, prefixObj, outputFolder) {
+  const t0 = performance.now();
+
+  const results = await providence(
+    QueryService.getQueryConfigFromAnalyzer('match-paths', { prefix: prefixObj }),
+    {
+      gatherFilesConfig: {
```
jorenbroekema, line 16: Good defaults, but this is currently not overridable, do we have plans for providing extensions / exclude rules through CLI or through a providence config object that you can pass? Should it be a blocker for now?

packages/providence/src/cli/generate-extend-docs-data.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const fs = require('fs');
+const pathLib = require('path');
```
jorenbroekema, line 3: Bit confusing, would call it path to be conventional

packages/providence/src/cli/generate-extend-docs-data.js
```diff
@@ -0,0 +1,43 @@
...
+        excludeFolders: ['coverage', 'test'],
+      },
+      queryMethod: 'ast',
+      report: false,
+      targetProjectPaths: [pathLib.resolve(process.cwd())],
+      referenceProjectPaths: referencePaths,
+    },
+  );
+
+  const outputFilePath = pathLib.join(outputFolder, 'providence-extend-docs-data.json');
```
jorenbroekema, line 27: Currently this file already exists with an empty array in it, is that intended?

packages/providence/src/cli/generate-extend-docs-data.js
tlouisse: The reason I do this consistently throughout the whole package is that within Babel ast traversal, the `path` variable is used for the ast path

packages/providence/src/cli/generate-extend-docs-data.js
tlouisse: Nope :) Will remove it

packages/providence/providence.conf.js
tlouisse: I saw it as well. Will be fixed in next push

packages/providence/README.md
tlouisse: Good idea

packages/providence/README.md
tlouisse: No, you're right :) will fix it

packages/providence/src/cli/generate-extend-docs-data.js
tlouisse: Yes, you'r right. Definitely about excludeFolders. Will pick that up

husky.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   hooks: {
-    'pre-commit': 'lint-staged && node ./scripts/yarn-lock-scan.js',
+    'pre-commit': 'lint-staged',
```
jorenbroekema, line 3: The reason we added yarn lock scan here is because we wanted to always scan it no matter what. But I think it's a bit too much indeed. We should rebase to latest master though, where lint-staged is handled in lintstagedrc file (has to do with package.json formatter I added last week)

tlouisse: Yeah, I thought it was a bit too much, but I think it's better to be safe than sorry, since you could still circumvent  alltogether with `-n`

Can drop this commit again...

PR #773, opened 2020-06-18
feat(button): add contentTemplate

gerjanvangeest:


gerjanvangeest:
It prevents me from doing `display: flex` on `.btn`. So found another way to style the accordion content.

When we reopen this MR, we need to change line 165 too, to make button-id work.

PR #779, opened 2020-06-19
feat(component): add collapsible

### LINKED ISSUE #777: Add collapsible component
The basic version of the collapsible component would be a nice addition to the Lion. I am currently working on it and will soon create the MR.

hardikpthv:
- Introduces the `lion-collapsible`

Closes #777


erikkroes:
A quick peek makes me wonder why you're using aria-hidden. The content should either show or hide, not specifically for assistive technology.

packages/collapsible/src/LionCollapsible.js
```diff
@@ -0,0 +1,127 @@
...
+
+        :host ::slotted([slot='content']) {
+          transition: var(--animation);
+          max-height: 0;
+          opacity: 0;
+          overflow: hidden;
+        }
+
+        :host([expanded]) ::slotted([slot='content']) {
+          max-height: 100vh;
```
LarsDenBakker, line 46: this would make the animation speed depend on screen size. You could take a look at https://github.com/PolymerElements/iron-collapse/blob/master/iron-collapse.js which dynamically calculates the height of the content. And it uses display: none so you don't need extra tricks for a11y like @erikkroes mentioned :)

packages/collapsible/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@lion/collapsible",
+  "version": "0.1.0",
```
gerjanvangeest, line 3: Keep this on "0.0.0" before you release anything

packages/collapsible/README.md
```diff
@@ -0,0 +1,85 @@
+# Collapsible
+
+`lion-collapsible` is a combination of a button (the trigger), a chunk of 'extra content' and an animation that is used to disclose the extra content.
```
gerjanvangeest, line 3: `trigger` -> `invoker`

packages/collapsible/src/LionCollapsible.js
```diff
@@ -0,0 +1,127 @@
...
+    if (invokerNode) {
+      const uid = uuid();
+      this._setupExpandCollapseListeners();
+      this._setupTransitionListener();
+      this.__setCommonAriaAttributes(this.expanded);
+      invokerNode.setAttribute('id', `collapsible-invoker-${uid}`);
+      invokerNode.setAttribute('aria-controls', `collapsible-content-${uid}`);
+      this._contentNode.setAttribute('aria-labelledby', `collapsible-invoker-${uid}`);
+      this._contentNode.setAttribute('id', `collapsible-content-${uid}`);
+      this.setAttribute('aria-labelledby', `collapsible-invoker-${uid}`);
```
gerjanvangeest, line 98: I don't see any reason why the `lion-collapsible` needs to have these aria labels.

packages/collapsible/test/lion-collapsible.test.js
```diff
@@ -0,0 +1,86 @@
...
+      invoker.dispatchEvent(new Event('click'));
+      expect(collapsible.expanded).to.equal(true);
+    });
+  });
+  */
+
+  describe('Accessibility', () => {
+    it('is a11y AXE accessible', async () => {
+      const collapsible = await fixture(defaultCollapsible);
+      expect(collapsible).to.be.accessible();
```
gerjanvangeest, line 44: Can you also add a test for the expanded version?

packages/collapsible/test/lion-collapsible.test.js
```diff
@@ -0,0 +1,86 @@
...
+    });
+
+    it('does not make contents focusable', async () => {
```
gerjanvangeest, line 47: I think this test can be deleted.

hardikpthv:
Thanks, @erikkroes, @LarsDenBakker and @gerjanvangeest for the review. :) I will take a look and update it.

packages/collapsible/src/LionCollapsible.js
```diff
@@ -76,6 +72,26 @@ export class LionCollapsible extends LitElement {
     }
   }
 
+  toggle() {
+    this.__toggle();
+  }
+
+  expand() {
```
gerjanvangeest, line 79: Could we rename this to `show`, to stay inline with our other components? (e.g. overlays)

packages/collapsible/src/LionCollapsible.js
```diff
@@ -76,6 +72,26 @@ export class LionCollapsible extends LitElement {
...
+    if (!this.expanded) {
+      this.expanded = true;
+      this._updateContentSize(this.expanded);
+      this.__dispatchEvent(this.expanded);
+    }
+  }
+
+  collapse() {
```
gerjanvangeest, line 87: Could we rename this to `hide`, to stay inline with our other components? (e.g. overlays)

packages/collapsible/src/LionCollapsible.js
hardikpthv: Sure 👍

packages/collapsible/src/LionCollapsible.js
hardikpthv: 👍

packages/collapsible/test/lion-collapsible.test.js
hardikpthv: Added

packages/collapsible/src/LionCollapsible.js
hardikpthv: Thanks for sharing it. I made the changes similarly. :)

tlouisse:
Good work 👍 

I have a few suggestions/questions:

- would it be possible to give the default collapsible no animation at all and make it possible for an extension layer to add whatever animation (we could show in a demo how a subclasser would do this and apply the current animation)? 
We could think of returning a Promise, allowing the subclasser to choose whether he uses CSS, js, or a lib lik GSAP. Something similar (not activated yet, but in 'pseudo code') is here: https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayController.js#L520

- what should happen when you click the invoker while animating? would it be possible to 'animate back' instead of ignoring the click?

- I think it's nice in general that the api is aligned to a large extent to the overlay system already 👍  Maybe we could also think of:
  - event best practices (no bubble, no compose)
  - event names. When possible align event names with property names. For the overlay system we have `opened` and `opened-changed`

hardikpthv:
Thanks for the review, @tlouisse.

-  _Would it be possible to give the default collapsible no animation at all and make it possible for an extension layer to add whatever animation (we could show in a demo how a subclasser would do this and apply the current animation)?_ **That sounds like a good idea. I will have to think about it further. Or maybe can we make it as a separate release?**
-  _What should happen when you click the invoker while animating?_ **Right now, nothing.**
- _Would it be possible to 'animate back' instead of ignoring the click?_ **Yes**
- _About events:_
    -   I will remove the bubble and compose.
    -   Maybe I can rename the property to `opened` or event name to `collapsible-expanded`/`collapsed` :thinking:

packages/collapsible/README.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+#### Custom Invoker Template
+
+A custom template can be specified to the `invoker` slot. It can be any button or custom component which mimics the button behavior for better accessibility support. In the below example, `lion-button` is used as a collapsible invoker.
+
+```js preview-story
+export const customInvokerTemplate = () => html`
+  <lion-collapsible>
+    <lion-button slot="invoker">More about cars</lion-button>
```
jorenbroekema, line 85: I don't think this is imported right? Maybe change it to <div> or <small> as an example or something, prevents having to depend on lion-button.

packages/collapsible/src/LionCollapsible.js
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Calculate total content height after collapsible opens
+   * @param {Object} contentNode content node
+   * @private
+   */
+  async __calculateHeight(contentNode) {
+    contentNode.style.setProperty('max-height', '');
+    await new Promise(resolve => requestAnimationFrame(() => resolve()));
+    return this._contentHeight; // Expected height i.e. actual size once collpased after animation
```
jorenbroekema, line 172:
```suggestion
    return this._contentHeight; // Expected height i.e. actual size once collapsed after animation
```

packages/collapsible/src/LionCollapsible.js
```diff
@@ -0,0 +1,207 @@
...
+  async _hideAnimation() {}
+
+  /**
+   * Calculate total content height after collapsible opens
+   * @param {Object} contentNode content node
+   * @private
+   */
+  async __calculateHeight(contentNode) {
+    contentNode.style.setProperty('max-height', '');
+    await new Promise(resolve => requestAnimationFrame(() => resolve()));
```
jorenbroekema, line 171: Doesn't this mean that we get a content flash in order to calculate?

LarsDenBakker: You can use iron-collapse's trick to trigger a style recalculation synchronously: https://github.com/PolymerElements/iron-collapse/blob/master/iron-collapse.js#L193 we've used this for another collapsible component and it works there.

jorenbroekema: Ah yeah true! I know it as `void this.offsetWidth;` which I used to create the action logger:

https://github.com/ing-bank/lion/blob/master/packages/helpers/sb-action-logger/src/SbActionLogger.js#L227

Has a short explanation why it works

packages/collapsible/README.md
hardikpthv: I can add custom styling to the native button, just to demonstrate that invoker can be customizable. Same as `customAnimation`.

packages/collapsible/src/LionCollapsible.js
LarsDenBakker: Yep that works as well.

packages/collapsible/README.md
jorenbroekema: Sounds good as well :)

packages/collapsible/src/LionCollapsible.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+
+  /**
+   * Set common aria attributes for invoker and content nodes.
+   * @private
+   */
+  __setCommonAriaAttributes() {
+    this._invokerNode.setAttribute('aria-expanded', this.opened);
+    this._contentNode.setAttribute('aria-hidden', !this.opened);
```
jorenbroekema, line 173: I mentioned it in private but just that we don't forget: probably don't need aria-hidden anymore ;)

hardikpthv: Yes, just removed it. 😅

PR #780, opened 2020-06-19
cache correct locale and format hungarian & bulgarian dates

### LINKED ISSUE #540: [validate] Date formats of some countries not supported
## Describe the bug

Date formats of some countries not supported:
  
image

> The help-text isn't translated in above example.

The validation messages in languages like `Hungarian` or `Bulgarian` suggest a specific date format, but the validators break if that format is typed. Typing in another format validates correctly, and then formats the input in the correct format.

## To Reproduce

1. go to the [input-datepicker validation](https://lion-web-components.netlify.com/?path=/story/forms-input-datepicker--validation) story
2. Change the language on the `html` tag.
3. Select a date via the datepicker
4. Change the date manually and the validators will fail.

raichuAK:
Earlier localize.locale evaluation was static for class; now pass localize.locale for each change  (  example directly changing html lang attribute )

Modify addleadingzero method for parsing bulgarian & hungarian languages.

BREAKING CHANGE: N

closes #540


packages/localize/src/date/parseDate.js
```diff
@@ -2,9 +2,9 @@ import { localize } from '../localize.js';
 import { getDateFormatBasedOnLocale } from './getDateFormatBasedOnLocale.js';
 import { addLeadingZero } from './addLeadingZero.js';
 
-const memoize = (fn, parm) => {
+const memoize = fn => {
   const cache = {};
-  return () => {
+  return parm => {
```
tlouisse, line 7: we could also make this function more generic by doing `const n = [...arguments]` in the function above

packages/localize/test/date/parseDate.test.js
```diff
@@ -32,8 +32,6 @@ describe('parseDate()', () => {
     expect(equalsDate(parseDate('12.12.1976'), new Date('1976/12/12'))).to.equal(true);
     expect(equalsDate(parseDate('13.12.1976'), new Date('1976/12/13'))).to.equal(true);
     expect(equalsDate(parseDate('14.12.1976'), new Date('1976/12/14'))).to.equal(true);
-    expect(equalsDate(parseDate('14.12-1976'), new Date('1976/12/14'))).to.equal(true);
```
gerjanvangeest, line 35: Could we add:

```js
expect(equalsDate(parseDate('14. 12. 1976.'), new Date('1976/12/14'))).to.equal(true);
expect(equalsDate(parseDate('14.12.1976 г.'), new Date('1976/12/14'))).to.equal(true);
```

gerjanvangeest:
Something i saw that was also wrong: https://github.com/ing-bank/lion/blob/master/packages/validate-messages/translations/bg.js#L13 (& line 31)

```text
(ДД-ММ-ГГГГ) -> (ДД.ММ.ГГГГ  г.)
```

Could you add this to the MR?

packages/localize/src/date/parseDate.js
raichuAK: Do you think passing multiple parameters to function will improve it?

packages/localize/test/date/parseDate.test.js
raichuAK: Added cases for above mentioned scenarios

packages/localize/src/date/parseDate.js
tlouisse: Currently it's only used for this purpose. If it needs to be a generic memoize function (which the name suggests, and it makes sense to promote it to at some later point) then yes.
For now it's fine ;)

tlouisse:
Looks good. Thanks a lot 👍 

Could you:
- sign the CLA? :)
- squash all commits into one? (`fix(localize): format hungarian & bulgarian dates` for instance)

raichuAK:
Due to commit issue; creating seperate PR https://github.com/ing-bank/lion/pull/798.
Please refer new PR.

PR #782, opened 2020-06-22
fix(select-rich): instantiation after first update

MathieuPuech:
See discussion about that here:

https://polymer.slack.com/archives/CJGFWJN9J/p1591901856272600

Issue was revealed when select-rich is rendered with react

### DIFF
--- a/packages/form-core/src/registration/FormRegistrarPortalMixin.js
+++ b/packages/form-core/src/registration/FormRegistrarPortalMixin.js
@@ -32,7 +32,6 @@ export const FormRegistrarPortalMixin = dedupeMixin(
         if (super.connectedCallback) {
           super.connectedCallback();
         }
-        this.__checkRegistrationTarget();
 
         formRegistrarManager.add(this);
         if (this.__hasBeenRendered) {
@@ -66,6 +65,7 @@ export const FormRegistrarPortalMixin = dedupeMixin(
       }
 
       firstUpdated(changedProperties) {
+        this.__checkRegistrationTarget();
         super.firstUpdated(changedProperties);
         this.__resolveRegistrationReady();
         this.__readyForRegistration = true;
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -203,15 +203,7 @@ export class LionSelectRich extends ScopedElementsMixin(
     this.__hasInitialSelectedFormElement = false;
     this.hasNoDefaultSelected = false;
     this._repropagationRole = 'choice-group'; // configures FormControlMixin
-    this.__setupEventListeners();
     this.__initInteractionStates();
-  }
-
-  connectedCallback() {
-    this._listboxNode.registrationTarget = this;
-    if (super.connectedCallback) {
-      super.connectedCallback();
-    }
   }
 
   disconnectedCallback() {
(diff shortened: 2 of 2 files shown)


tlouisse:
Thanks a lot for all the effort put in this.
I locally rebased against latest master and get 9 failing tests (latest master is based on WTR instead of karma, but I'm not sure if that's the issue?)

Since I cannot reproduce the issue (I would need an integration environment with React), it's hard now to fix this myself.
It could be an idea to have some 'framework integration sandboxes' for this? @jorenbroekema @daKmoR @gerjanvangeest what do you think?
My feeling is that refactoring the form registration system (also mentioned here: https://github.com/ing-bank/lion/issues/788#issuecomment-654028687) might also solve this bug.

MathieuPuech:
> Thanks a lot for all the effort put in this.
> I locally rebased against latest master and get 9 failing tests (latest master is based on WTR instead of karma, but I'm not sure if that's the issue?)
> 
> Since I cannot reproduce the issue (I would need an integration environment with React), it's hard now to fix this myself.
> It could be an idea to have some 'framework integration sandboxes' for this? @jorenbroekema @daKmoR @gerjanvangeest what do you think?
> My feeling is that refactoring the form registration system (also mentioned here: [#788 (comment)](https://github.com/ing-bank/lion/issues/788#issuecomment-654028687)) might also solve this bug.

I can create a repository with basic react integration if you want. I've not tested all components in react yet.

jorenbroekema:
Seems like registration of form elements is broken with this change. Locally it doesn't work for me either. Definitely not WTR's fault, on master all these tests pass without problems. So definitely need to look into this more deeply.

tlouisse:
> I can create a repository with basic react integration if you want. I've not tested all components in react yet.

Cool. Thanks a lot for the offer, it's something we might be interested in in the future (but not yet because of maintenance reasons).
We just discussed internally and testing `document.createElement` would also work for this particular case.

MathieuPuech:
> > I can create a repository with basic react integration if you want. I've not tested all components in react yet.
> 
> Cool. Thanks a lot for the offer, it's something we might be interested in in the future (but not yet because of maintenance reasons).
> We just discussed internally and testing `document.createElement` would also work for this particular case.
> 
> I added a failing test for this here (there's also a potential fix, but the inital a11y (aria-activedescendant) is still broken) :
> https://github.com/ing-bank/lion/blob/chore/failingTestSelectRichReact/packages/select-rich/test/lion-select-rich.test.js#L862

I think a simple rebase is mixing all things, I will do a rebase manually to fix this.

For the react repo, I will wait for the extending documentation to do it

MathieuPuech:
it seems that we just don't wait enough in tests because form are added with an event 'form-element-register'.

It's working when the event is sent in connectedCallback because it's sooner but the test could fail if the event don't have the time to be received. I think the bug is in tests or maybe we should have something to await.

MathieuPuech:
> it seems that we just don't wait enough in tests because form are added with an event 'form-element-register'.
> 
> It's working when the event is sent in connectedCallback because it's sooner but the test could fail if the event don't have the time to be received. I think the bug is in tests or maybe we should have something to await.

no, I was wrong, even with a wait of 1s it's the same. I continue searching what's the issue

MathieuPuech:
It should be ok like that (it's working in my react component wrapper too)

tlouisse:
> It should be ok like that (it's working in my react component wrapper too)

Yes, it also works with my test. Thanks 👍 Will add that test later and merge it.

PR #784, opened 2020-06-23
fix(button): prevent scrolling when repeating space key

### LINKED ISSUE #740: [button] page is scrolling when space key is pressed
## Describe the bug

I don't know if it is a desired behavior but the button does not prevent the scrolling in a page when you are focused on it and you keep the space key pressed

## To Reproduce

1. Go to Storybook button documentation page https://lion-web-components.netlify.app/?path=/docs/buttons-button--main
2. Click on the first button on page to focus it (or focus it with tab key)
3. Press space bar and keep it pressed
4. The page is scrolled

## Expected behavior

Page does not scroll

## Additional context

It seems that native button prevent the scroll when pressing space bar event when you keep it pressed.

MathieuPuech:
Closes #740


gerjanvangeest:
Thanks for the fix :)

PR #786, opened 2020-06-23
tests: timezone and locale

### LINKED ISSUE #776: 2 Tests are failing depending the environment
## Describe the bug

2 tests are failing on certain environments.

```
FAILED TESTS:
  Form Integrations
    × ".serializedValue" returns all non disabled fields based on form structure
      HeadlessChrome 83.0.4103 (Windows 10.0.0)
    AssertionError: expected { Object (full_name, date, ...) } to deeply equal { Object (bio, checkers[], ...) }
      + expected - actual
         "checkers[]": [
           []
         ]
         "comments": ""
      -  "date": "2000-12-11"
      -  "datepicker": "2020-12-11"
      +  "date": "2000-12-12"
      +  "datepicker": "2020-12-12"
         "dinosaurs": ""
         "email": ""
         "favoriteColor": "hotpink"
         "full_name": {
    at Context.<anonymous> (packages/form-integrations/test/form-integrations.test.js:9:39)
  parseAmount()
    × parseAmount with locale set and length is more than four
      HeadlessChrome 83.0.4103 (Windows 10.0.0)
    AssertionError: expected 6 to equal 6000
      + expected - actual
      -6
      +6000
    at Context.<anonymous> (packages/input-amount/test/parsers.test.js:158:10)
    at Test.Runnable.run (node_modules/karma-mocha-snapshot/lib/adapter.js:14:17)
```

## To Reproduce

It's hard to know how to reproduce it, it seems that my local environment is making the tests failing. It could be something about timezone and locale settings.

## Additional context

My current locale is canadian french and my timezone is Eastern Daylight Time (GMT-4).

MathieuPuech:
Closes #776

### DIFF
--- a/packages/form-integrations/test/helpers/umbrella-form.js
+++ b/packages/form-integrations/test/helpers/umbrella-form.js
@@ -44,13 +44,13 @@ export class UmbrellaForm extends LitElement {
           <lion-input-date
             name="date"
             label="Date of application"
-            .modelValue="${new Date('2000-12-12')}"
+            .modelValue="${new Date('2000/12/12')}"
             .validators="${[new Required()]}"
           ></lion-input-date>
           <lion-input-datepicker
             name="datepicker"
             label="Date to be picked"
-            .modelValue="${new Date('2020-12-12')}"
+            .modelValue="${new Date('2020/12/12')}"
             .validators="${[new Required()]}"
           ></lion-input-datepicker>
           <lion-textarea
--- a/packages/input-amount/test/parsers.test.js
+++ b/packages/input-amount/test/parsers.test.js
@@ -153,7 +153,7 @@ describe('parseAmount()', () => {
   it('parseAmount with locale set and length is more than four', () => {
     expect(
       parseAmount('6,000', {
-        locale: 'gb-GB',
+        locale: 'en-GB',
       }),
     ).to.equal(6000);
     expect(


packages/form-integrations/test/helpers/umbrella-form.js
```diff
@@ -44,13 +44,13 @@ export class UmbrellaForm extends LitElement {
           <lion-input-date
             name="date"
             label="Date of application"
-            .modelValue="${new Date('2000-12-12')}"
+            .modelValue="${new Date('2000-12-12T00:00:00')}"
```
gerjanvangeest, line 47: Does this only occur in the tests, or do you also need to provide the time when you use it?

MathieuPuech: I suppose giving the time is the only way to give the timezone. If you don't give the timezone, 

If I do a `new Date('2000-12-12')`, I have `Mon Dec 11 2000 19:00:00 GMT-0500 (Eastern Standard Time)`. The date selected is december 11.

MathieuPuech: there is a lot of references to `new Date` in documentation, errors could happen for all people in a timezone with negative GMT time. We should maybe change the documentation about that?

MathieuPuech: No, it's good in this code, so maybe it's something to be fixed with input-date?

Capture

gerjanvangeest: Yes, changing it in input-date sounds like the right approach.

Let me test your suggestion locally, to see if it works.

MathieuPuech: I pushed a fix on the branch

MathieuPuech: we should maybe have tests in a different timezone to avoid these issues. Chrome can be started with a different timezone:

https://gist.github.com/dustinsmith1024/dab3560abaa62cbe0754117d44c71026

MathieuPuech: you just need to pass TZ environment variable to chrome. With karma, I think we need to start another script because karma use 1 browser for all tests.

packages/input-date/src/LionInputDate.js
```diff
@@ -50,7 +50,7 @@ export class LionInputDate extends LocalizeMixin(LionInput) {
     // modelValue is localized, so we take the timezone offset in milliseconds and subtract it
     // before converting it to ISO string.
     const offset = modelValue.getTimezoneOffset() * 60000;
-    return new Date(modelValue - offset).toISOString().slice(0, 10);
+    return new Date(modelValue.getTime() + offset).toISOString().slice(0, 10);
```
gerjanvangeest, line 53: I believe for us. it should be `-`, since this change made one test fail.

I'll dive in it more deeply tomorrow.

gerjanvangeest: The problem is in the way dates get handled:

Screenshot 2020-06-25 at 11 35 52

So for now all we need to do is change the value in the test to `new Date('2000/12/12')` for `input-date` & `input-datepicker`.

With your fix the `serializedValue` is correct (for negative timezones), but the `formattedValue` still displays wrong.

MathieuPuech: it seems that `new Date('2000/12/12')` is equivalent to `new Date('2000-12-12T00:00:00')`

I can change the tests to 2000/12/12 and remove the "fix"

MathieuPuech: I pushed new commit with this and removed the fix

PR #789, opened 2020-06-29
feat(providence-analytics): filepath matching based on anymatch

tlouisse:


packages/providence-analytics/src/program/services/InputDataService.js
```diff
@@ -218,8 +253,12 @@ class InputDataService {
     this.__targetProjectPaths = ensureArray(v);
   }
 
-  static getDefaultGatherFilesConfig() {
-    return defaultGatherFilesConfig;
+  static get defaultGatherFilesConfig() {
+    return {
+      extensions: ['.js'],
+      filter: ['!node_modules/**', '!bower_components/**', '!**/*.conf.js', '!**/*.config.js'],
```
jorenbroekema, line 259: Are these needed? Most devs have these in gitignore and they already get ignored. For config it makes sense

tlouisse: True, although I think it doesn't hurt to keep them for now to be sure (don't know if there could be 'exotic' notations in gitignore that could be missed by anymatch)

PR #791, opened 2020-06-30
feat(choice-input): add rendering of help-text

MathieuPuech:
help-text attribute is available on choice inputs (checkbox and radio), but it's not rendered.

Here is a proposal for rendering them.

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceInputMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceInputMixin.js
@@ -92,6 +92,7 @@ export const ChoiceInputMixin = superclass =>
         css`
           :host {
             display: flex;
+            flex-wrap: wrap;
           }
 
           :host([hidden]) {
@@ -100,6 +101,10 @@ export const ChoiceInputMixin = superclass =>
 
           .choice-field__graphic-container {
             display: none;
+          }
+          .choice-field__help-text {
+            display: block;
+            flex-basis: 100%;
           }
         `,
       ];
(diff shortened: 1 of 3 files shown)


packages/form-core/src/choice-group/ChoiceInputMixin.js
```diff
@@ -101,6 +102,10 @@ export const ChoiceInputMixin = superclass =>
           .choice-field__graphic-container {
             display: none;
           }
+          .choice-field__help-text {
+            display: block;
+            flex-basis: 100%;
```
jorenbroekema, line 107: This makes the help-text full width basically right? Is there a specific reason for it?

jorenbroekema: Ah, is it to make sure it wraps to the next line :P?

MathieuPuech: Yes with flex layout, I didn’t find another way to do it without changing HTML markup. With CSS grid, it’s easier But without ie11. 

Do you think it’s an issue?

jorenbroekema: Seems good to me, it's "minimal styling" to make it look acceptable which is always fine for us :), as long as it's easy to override as a subclasser, which it is. So I think I'll just merge it 👍

jorenbroekema: @tlouisse  you good with this? I think on our ing extension layer we'll need to delete the storybook section since we don't allow using help-text on individual checkbox/radio per our UX specs, but I think that should be it. Can you take care of that once we merge 😇 ?

tlouisse: Seems good. I will merge it 👍

Issue #781, opened 2020-06-22
Uncaught (in promise) in nested lion dialog


remcovaes:
## Describe the bug

When using nested lion dialogs, it shows an error in the browser console.

## To Reproduce

Nest 2 lion dialogs.

Example:
https://github.com/remcovaes/lion-dialog-uncaught-in-promise

Only tested in chrome deskptop 83.0.4103.106 (Official Build) (32-bit)

## Current behavior

Error in console:
```
OverlayMixin.js:152 Uncaught (in promise) undefined
connectedCallback @ OverlayMixin.js:152
__initConnectionTarget @ OverlayController.js:288
_init @ OverlayController.js:259
updateConfig @ OverlayController.js:233
OverlayController @ OverlayController.js:146
_defineOverlay @ OverlayMixin.js:58
_setupOverlayCtrl @ OverlayMixin.js:205
(anonymous) @ OverlayMixin.js:144
Promise.then (async)
connectedCallback @ OverlayMixin.js:142
__insert @ parts.js:207
__commitNode @ parts.js:214
__commitTemplateResult @ parts.js:249
commit @ parts.js:189
render @ render.js:44
render$1 @ shady-render.js:251
update @ lit-element.js:160
performUpdate @ updating-element.js:551
_enqueueUpdate @ updating-element.js:510
async function (async)
_enqueueUpdate @ updating-element.js:504
_requestUpdate @ updating-element.js:476
initialize @ updating-element.js:328
initialize @ lit-element.js:84
UpdatingElement @ updating-element.js:97
LitElement @ lit-element.js:31
(anonymous) @ index.js:5
(anonymous) @ index.js:5
```

## Expected behavior

No error

gerjanvangeest:
Thanks for the isolated example, it helped me to see that the bug is introduced in v0.14.0 of overlays, thus in v0.5.0 of the dialog.

PR #794, opened 2020-07-04
fix(overlays): only teardown when overlayCtrl defined

tlouisse:
closes https://github.com/ing-bank/lion/issues/781

### DIFF
--- a/packages/overlays/src/OverlayMixin.js
+++ b/packages/overlays/src/OverlayMixin.js
@@ -158,6 +158,10 @@ export const OverlayMixin = dedupeMixin(
           super.disconnectedCallback();
         }
 
+        if (!this._overlayCtrl) {
+          return;
+        }
+
         this._overlayDisconnectComplete = new Promise((resolve, reject) => {
           this.__resolveOverlayDisconnectComplete = resolve;
           this.__rejectOverlayDisconnectComplete = reject;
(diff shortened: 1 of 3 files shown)


packages/overlays/src/OverlayMixin.js
```diff
@@ -158,6 +158,10 @@ export const OverlayMixin = dedupeMixin(
           super.disconnectedCallback();
         }
 
+        if (!this._overlayCtrl) {
+          return;
+        }
+
         this._overlayDisconnectComplete = new Promise((resolve, reject) => {
```
jorenbroekema, line 165: Do we rely on this promise somewhere, where we need it even though there is no overlay controller? E.g. for nested overlay nodes, I believe the disconnectedCallback of the nested node gets called before overlay controller is set, and then we do need to make sure that we do some teardown stuff for the content node etc. right?

tlouisse: As long as there's no overlay controller (yet), there will be nothing to teardown, right?
I think the overlay setup starts at updateComplete, so the question would be whether `firstUpdated` still fires after `disconnectedCallback`.

jorenbroekema: Actually there is something to teardown https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayMixin.js#L218

The open/close handlers need to be torn down as well as the content node from the global-overlays container. Which I think is why I made the code as it was before this change, to only guard the `this._overlayCtrl.teardown()` with `this._overlayCtrl`  but not guarding `this._teardownOverlayCtrl()` with `this._overlayCtrl`

Does that make sense?

jorenbroekema: Hmm true :\ not sure then... might need to test it with a nested overlay demo to make sure maybe, you can take the nested dialog test case from lion-dialog to check it with, where it creates the nested node first, and then adds it to the parent node template. Because that's the use case where this went weirdly in the past

tlouisse: Yes, I thin you mean this story: https://lion-web-components.netlify.app/?path=/story/overlays-system-overview--nested-overlays (this is where the error originally occurred)

jorenbroekema: Actually it's this use case https://github.com/ing-bank/lion/blob/master/packages/overlays/test-suites/OverlayMixin.suite.js#L232. which is slightly different from the demo you linked and also the previous "it" in that test file, because the nested overlay node is constructed before, and then passed as a node into the parent template. This matters, because the connected/disconnected/parent-setupOverlay lifecycles happen in different orders for these different use cases. For the normal use case, it is parent setup overlay --> child connected. When creating the child node first, it's child connected --> parent setup overlay --> child disconnected --> child reconnected.

tlouisse: This works (because tests run). It also shows that it reconstructs (it hits the line after check for overlayCtrl in disconnected)

PR #796, opened 2020-07-06
Chore/blog journey

onbekendnow:
- Fully focus on a good developer API, without caring too much on how it will look.

- Being able to create a component without have the designs ready.

- Keeping a clean code base by having the backwards compatibilty code on a different layer. Therefore allowing ourselves to innovate and break often (also known as


tlouisse:
This code/documentation seems to be outdated as well and is written by one of our colleagues.
We assume this merge request was not created intentionally and we will therefore close it. 
Please let us know if you think it should be reopened.

PR #799, opened 2020-07-06
Chore/extending documentation

gerjanvangeest:
First draft for "extending documentation" docs

### DIFF
--- a/docs/extending-documentation.md
+++ b/docs/extending-documentation.md
@@ -1,0 +1,287 @@
+# Extending Lion Documentation
+
+```js script
+export default {
+  title: 'Guidelines/Extending documentation',
+};
+```
+
+If you extend [Lion](https://lion-web-components.netlify.app/) components, you don't only want to reuse the components, but you probably want to reuse the documentation (Storybook demos) as well. Wouldn't it be nice to just take it all from lion, but replace it with your own design system extension, so you don't have the extra maintenance of essentially copying the docs from `Lion` for your own design system implementation?
+
+In this blog we will explain how `Lion` supports this use case, and allows you to extend not just the components, but also the documentation.
+
+> Important to note is that you must have a naming convention in your extension that is similar to `Lion*` for class names and `lion-*` for tag names, e.g. `LeaTabs` and `lea-tabs`. So essentially, have a consistent prefix for all your extension classes and tagnames.
+
+## Steps involved
+
+In order to extend documentation, there are a few things that we need to enable you to do:
+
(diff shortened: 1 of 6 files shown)


docs/extending-documentation.md
```diff
@@ -0,0 +1,289 @@
...
+
+You can remove content:
+
+````md
+```
+::removeFrom('heading:has([value=Distribute New Elements])')
+```
+````
+
+Or you can add an extra paragraph below the content:
```
jorenbroekema, line 255: We should explain the selectors shortly, being CSS selectors essentially, and link to `unist-util-select` or link to the remark-extend documentation README.md

docs/extending-documentation.md
```diff
@@ -0,0 +1,289 @@
...
+module.exports = {
+  [...],
+  setupMdjsPlugins: (plugins, filePath) => {
+    if (!isLion(filePath)) {
+      return plugins;
+    }
+    const newPlugins = [...plugins];
+    const markdownIndex = newPlugins.findIndex(plugin => plugin.name === 'markdown');
+    const overridePaths = [`${process.cwd()}/.storybook/all.override.md`];
+    overridePaths.push(getLocalOverridePath(filePath));
```
jorenbroekema, line 194:
```suggestion
    const overridePaths = [`${process.cwd()}/.storybook/all.override.md`, getLocalOverridePath(filePath)];
```
probably nicer, now that we don't have a conditional installation override file.

docs/extending-documentation.md
```diff
@@ -0,0 +1,289 @@
...
+      i += 1;
+    }
+    return newPlugins;
+  },
+  [...],
+};
+```
+
+In the code above we have 2 places in where you can do overrides: `./.storybook/all.override.md` for generic overrides and via `getLocalOverridePath` for each component, when needed, the `rel` needs to be the same in `lion` and your own project to be able to override the right file.
+
```
jorenbroekema, line 215: The code is still a little magical without more explanation.

Might wanna explain a bit more what setupMdjsPlugins means, and what we actually do here. 

Essentially, we figure out where the "markdown" Mdjs plugin is, and insert our own remark-extend plugins after it, one for each of the override paths that we define (in our case, only 2, local and global override). And the reason we do a .reverse() is because we reverse the order of override.md plugins, so that local overrides are done before the global ones (all.override.md).

So maybe it's nicer to do:
```js
const overridePaths = [`${process.cwd()}/.storybook/all.override.md`, getLocalOverridePath(filePath)];
```
and 
```js
for (const overridePath of overridePaths) { }
```
to make it a little bit more simpler code for our reader.

PR #800, opened 2020-07-06
chore: fix flaky test, prevent unhandled promise undefined feedbackNode

jorenbroekema:


packages/fieldset/test/lion-fieldset.test.js
```diff
@@ -622,8 +622,7 @@ describe('<lion-fieldset>', () => {
       const spyB = sinon.spy(childB, 'initInteractionState');
       expect(fieldset.prefilled).to.be.false;
       expect(fieldset.dirty).to.be.false;
-      await fieldset.registrationReady;
```
jorenbroekema, line 625: field initInteractionState only happens when fieldset parent registration has completed, from FormGroupMixin:
```js
async __initInteractionStates() {
        if (!this.registrationHasCompleted) {
          await this.registrationComplete;
        }
        this.formElements.forEach(el => {
          if (typeof el.initInteractionState === 'function') {
            el.initInteractionState();
          }
        });
      }
```

yarn.lock
```diff
@@ -2513,20 +2513,6 @@
     mocha "^6.2.2"
     sinon-chai "^3.3.0"
 
-"@polymer/iron-test-helpers@^3.0.1":
```
jorenbroekema, line 2516: 🎉  dunno why this was still in our lock file but it is gone now at last :D

PR #805, opened 2020-07-08
chore: update to test runner v0.6

LarsDenBakker:
This updates to the latest version of web test runner which among other changes is a lot faster.

I notice this test is flaky:

```
packages/overlays/test/OverlayMixin.test.js:

 ❌ OverlayMixin integrations > OverlayMixin > syncs opened to overlayController
      at: packages/overlays/test-suites/OverlayMixin.suite.js
      error: expected true to be false
      + expected - actual
      
      -true
      +false
```

Is that already the case on master?

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@open-wc/testing": "^2.5.0",
     "@open-wc/testing-helpers": "^1.0.0",
     "@storybook/addon-a11y": "~5.0.0",
-    "@web/test-runner": "^0.2.9",
+    "@web/test-runner": "^0.6.7",
     "@webcomponents/webcomponentsjs": "^2.2.5",
     "babel-eslint": "^8.2.6",
     "babel-polyfill": "^6.26.0",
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
I changed the config from MJS to CJS because the CI uses an older version of node v12.

jorenbroekema:
Hmmm haven't seen that test be flaky before, but if it is, we should make it non-flaky :D I did fix a bunch of flaky ones so I'll just add this to the list I guess

tlouisse:
> This updates to the latest version of web test runner which among other changes is a lot faster.
> 
> I notice this test is flaky:
> 
> ```
> packages/overlays/test/OverlayMixin.test.js:
> 
>  ❌ OverlayMixin integrations > OverlayMixin > syncs opened to overlayController
>       at: packages/overlays/test-suites/OverlayMixin.suite.js
>       error: expected true to be false
>       + expected - actual
>       
>       -true
>       +false
> ```
> 
> Is that already the case on master?

There have been a few flaky ones, not sure if this was amongst them. I will merge it though and see if we can get rid of the flakiness later.

PR #807, opened 2020-07-09
chore(select-rich): test React-like instantiation

tlouisse:


jorenbroekema:
Nice :) it fails though atm in the CI.

You can ignore the accessible test that fails, just restart for now and get lucky.

In the types PR I've already set concurrency to 5 (instead of default Infinity), hopefully that will mitigate the problem, assuming it is caused by the RAM being the bottleneck and therefore the accessible test being too slow (> 2000ms).

tlouisse:
> Nice :) it fails though atm in the CI.
> 
> You can ignore the accessible test that fails, just restart for now and get lucky.
> 
> In the types PR I've already set concurrency to 5 (instead of default Infinity), hopefully that will mitigate the problem, assuming it is caused by the RAM being the bottleneck and therefore the accessible test being too slow (> 2000ms).

It keeps failing. Maybe the upgrade of WTR helps as well here. Who knows...

Issue #806, opened 2020-07-08
[dialog] select rich in the content of a dialog


MathieuPuech:
## Describe the bug

When a select-rich is in the content of a dialog, there is an error in console.

## To Reproduce

I created a branch to show the bug:

https://github.com/MathieuPuech/lion/tree/fix/select-rich-in-a-dialog

1. Go to storybook page of dialog (main)
2. See error

## Expected behavior

No error in console ;)

## Logs

```
Uncaught TypeError: Cannot read property 'removeEventListener' of undefined
    at HTMLElement.__teardownOverlay (LionSelectRich.js:699)
    at HTMLElement.disconnectedCallback (LionSelectRich.js:222)
    at OverlayController.__initConnectionTarget (OverlayController.js:289)
    at OverlayController._init (OverlayController.js:274)
    at OverlayController.updateConfig (OverlayController.js:234)
    at new OverlayController (OverlayController.js:147)
    at HTMLElement._defineOverlay (OverlayMixin.js:58)
    at HTMLElement._setupOverlayCtrl (OverlayMixin.js:207)
    at OverlayMixin.js:144
```

PR #810, opened 2020-07-09
fix(select-rich): align teardown phase with OverlayMixin

tlouisse:
closes https://github.com/ing-bank/lion/issues/806

### DIFF
--- a/packages/select-rich/test/lion-select-rich-dialog-integration.test.js
+++ b/packages/select-rich/test/lion-select-rich-dialog-integration.test.js
@@ -1,0 +1,53 @@
+import { OverlayMixin } from '@lion/overlays';
+import { LitElement } from 'lit-element';
+import { defineCE, fixture, html, expect, unsafeStatic } from '@open-wc/testing';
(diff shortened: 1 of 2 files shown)


packages/select-rich/test/lion-select-rich-dialog-integration.test.js
```diff
@@ -0,0 +1,53 @@
...
+          <div slot="content">content of the overlay</div>
+        </div>
+      `;
+    }
+  },
+);
+const tag = unsafeStatic(tagString);
+
+describe('Select Rich Integration tests', () => {
+  it('works inside a dialog', async () => {
```
jorenbroekema, line 24: Should we do some more smoke testing, e.g. open the dialog and assert it, open the rich select and assert it, select an option and assert the modelValue? Now we just check that no error is thrown (good start though).

tlouisse: Good be a nice idea for the future indeed.

Issue #803, opened 2020-07-07
[dialog] click on a label in a dialog with hidesOnOutsideClick on


MathieuPuech:
## Describe the bug

When you have a label for an input in a dialog, clicking on the label close the dialog

## To Reproduce

Change the main story of dialog with this (added label, input node and config property on dialog:
```js story
export const main = () => html`
  <style>
    ${demoStyle}
  </style>
  <lion-dialog .config=${{ hidesOnOutsideClick: true }}>
    <button slot="invoker">Click me to open dialog</button>
    <div slot="content" class="dialog">
      <label for="test">test</label>
      <input id="test">
      Hello! You can close this dialog here:
      <button
        class="close-button"
        @click=${e => e.target.dispatchEvent(new Event('close-overlay', { bubbles: true }))}
      >
        ⨯
      </button>
    </div>
  </lion-dialog>
`;
```
2. click on the label

## Expected behavior

Not closing the dialog

tlouisse:
Interesting one. The problem is that a click on a label that points to an input, causes the platform to focus the input. Apparently the platform does this via a click on this input.
Two consecutive clicks lead to a mismatch in our code (it makes the 'body handler' misalign with the 'content handler'). Fixed by: https://github.com/ing-bank/lion/pull/815/commits/ece7c0f33e4a4be888bcbc20d90a2b65e261a9c9

PR #815, opened 2020-07-13
Fix/overlay bugs

tlouisse:
- closes https://github.com/ing-bank/lion/issues/817
- closes https://github.com/ing-bank/lion/issues/816
- closes https://github.com/ing-bank/lion/issues/803

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -67,8 +67,7 @@ const supportsCSSTypedObject = window.CSS && CSS.number;
  * Note that a contentWrapperNode should be provided for [l2], [l3] and [l4]
  * In case of a global overlay ([g1]), it's enough to provide just the contentNode.
  * In case of a local overlay or a responsive overlay switching from placementMode, one should
- * always configure as if it was a local overlay.
- *
+ * always configure as if it were a local overlay.
  */
 
 export class OverlayController {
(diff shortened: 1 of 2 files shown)


packages/overlays/src/OverlayController.js
```diff
@@ -358,11 +357,14 @@ export class OverlayController {
   __setupTeardownAccessibility({ phase }) {
     if (phase === 'init') {
       this.__storeOriginalAttrs(this.contentNode, ['role', 'id']);
```
jorenbroekema, line 359: Should we guard this.contentNode as well?

packages/overlays/test/OverlayController.test.js
```diff
@@ -430,6 +430,8 @@ describe('OverlayController', () => {
         // Don't hide on inside (content) click
         ctrl.contentNode.click();
         await aTimeout();
```
jorenbroekema, line 432: These two timeouts are necessary for the double setTimeout in the src? So if we removed these timeouts here in the test, and revert the fix, this test would pass, there we keep the timeouts in this test so it's a proper regression test? :) Just for my understanding

packages/overlays/src/OverlayController.js
tlouisse: It already is: https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayController.js#L249

packages/overlays/test/OverlayController.test.js
tlouisse: They are not neccessary (anymore), removed them 👍

PR #820, opened 2020-07-14
Fix/registration portal

jorenbroekema:


jorenbroekema:
@MathieuPuech  I saw some changes were done in the instantiation logic, but I think the timings are a little off with regards to registration targets for portal elements (e.g. LionOptions inside LionSelectRich). I think it's important that the implementer sets the registration target before the parent firstUpdated, and then make sure that the portal mixin checks for this registration target after the implementer has set it. So I think this PR fixes that timing problem.

What do you think about it? It seems like the "React" simple integration/smoke test still passes, but maybe you want to make sure first on your end?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -206,6 +206,13 @@ export class LionSelectRich extends ScopedElementsMixin(
     this.__initInteractionStates();
   }
 
+  connectedCallback() {
+    this._listboxNode.registrationTarget = this;
```
jorenbroekema, line 210: putting this before super.firstUpdated is too late, because there, the firstUpdated of the listboxNode is already called. If we put it before super.connectedCallback, we're on time, because the firstUpdated of the listboxNode hasn't been called yet, because no render has happened yet.

tlouisse:
Would it be possible to add a failing test as well for the previous situation?

MathieuPuech:
Here is what I get with this implementation on a reactified component

```
LionSelectRich.js:210 Uncaught TypeError: Cannot set property 'registrationTarget' of undefined
    at HTMLElement.connectedCallback (LionSelectRich.js:210)
    at appendChildToContainer (react-dom.development.js:7572)
    at insertOrAppendPlacementNodeIntoContainer (react-dom.development.js:20350)
    at insertOrAppendPlacementNodeIntoContainer (react-dom.development.js:20356)
    at insertOrAppendPlacementNodeIntoContainer (react-dom.development.js:20356)
    at insertOrAppendPlacementNodeIntoContainer (react-dom.development.js:20356)
    at commitPlacement (react-dom.development.js:20334)
    at commitMutationEffects (react-dom.development.js:22748)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
connectedCallback @ LionSelectRich.js:210
appendChildToContainer @ react-dom.development.js:7572
insertOrAppendPlacementNodeIntoContainer @ react-dom.development.js:20350
insertOrAppendPlacementNodeIntoContainer @ react-dom.development.js:20356
insertOrAppendPlacementNodeIntoContainer @ react-dom.development.js:20356
insertOrAppendPlacementNodeIntoContainer @ react-dom.development.js:20356
commitPlacement @ react-dom.development.js:20334
commitMutationEffects @ react-dom.development.js:22748
callCallback @ react-dom.development.js:188
invokeGuardedCallbackDev @ react-dom.development.js:237
invokeGuardedCallback @ react-dom.development.js:292
commitRootImpl @ react-dom.development.js:22509
unstable_runWithPriority @ scheduler.development.js:653
runWithPriority$1 @ react-dom.development.js:11039
commitRoot @ react-dom.development.js:22381
finishSyncRender @ react-dom.development.js:21807
performSyncWorkOnRoot @ react-dom.development.js:21793
scheduleUpdateOnFiber @ react-dom.development.js:21188
updateContainer @ react-dom.development.js:24373
(anonymous) @ react-dom.development.js:24758
unbatchedUpdates @ react-dom.development.js:21903
legacyRenderSubtreeIntoContainer @ react-dom.development.js:24757
render @ react-dom.development.js:24840
(anonymous) @ render.js:74
render @ render.js:73
_callee$ @ render.js:165
tryCatch @ runtime.js:45
invoke @ runtime.js:274
prototype.<computed> @ runtime.js:97
asyncGeneratorStep @ render.js:48
_next @ render.js:50
(anonymous) @ render.js:50
(anonymous) @ render.js:50
_renderMain @ render.js:174
renderMain @ render.js:138
_callee$ @ start.js:383
tryCatch @ runtime.js:45
invoke @ runtime.js:274
prototype.<computed> @ runtime.js:97
asyncGeneratorStep @ start.js:97
_next @ start.js:99
(anonymous) @ start.js:99
(anonymous) @ start.js:99
renderMain @ start.js:402
renderUI @ start.js:429
emit @ index.js:180
(anonymous) @ story_store.js:357
setTimeout (async)
setSelection @ story_store.js:351
(anonymous) @ start.js:499
emit @ index.js:180
_renderMain @ config_api.js:106
render @ config_api.js:54
ConfigApi.configure @ config_api.js:81
configure @ start.js:672
configure @ index.js:35
(anonymous) @ generated-entry.js:5
./.storybook/generated-entry.js @ generated-entry.js:5
__webpack_require__ @ bootstrap:853
fn @ bootstrap:150
0 @ index.stories.js:42
__webpack_require__ @ bootstrap:853
checkDeferredModules @ bootstrap:45
webpackJsonpCallback @ bootstrap:32
(anonymous) @ main.0aecb20f88ecd383b498.bundle.js:1
Show 17 more frames
```

```
FormRegistrarPortalMixin.js:78 Uncaught (in promise) Error: A FormRegistrarPortal element requires a .registrationTarget
    at HTMLElement.__checkRegistrationTarget (FormRegistrarPortalMixin.js:78)
    at HTMLElement.firstUpdated (FormRegistrarPortalMixin.js:69)
    at HTMLElement.performUpdate (updating-element.js:568)
    at HTMLElement._enqueueUpdate (updating-element.js:510)
```

jorenbroekema:
Yeah makes sense, happens when listbox node is not yet connected inside the rich select....

But that's pretty tricky to solve, because then we need to delay setting the registration target until the listboxnode is available. And then we also need to delay checking the registration target in the portal mixin until the registration target it set...

We could create such promises and resolve them etc. but will require some more work. And we should then also decide on a timeout probably for __checkRegistrationTarget so it doesn't delay/wait indefinitely...

Not sure what to do.. open for suggestions

MathieuPuech:
example of use of lion in react is available here:

https://codesandbox.io/s/eager-aryabhata-4mvyh?file=/src/App.js

tlouisse:
Hmm, this test misses a very vital thing:
https://github.com/ing-bank/lion/blob/master/packages/select-rich/test/lion-select-rich.test.js#L870
The whole fixture is not connected to the DOM. Maybe you can connect it to the dom (and cleanup afterwards) to create a failing test for Mathieu's case?

jorenbroekema:
Okay so concluding, I want to close this PR, as it makes things worse not better. You cannot, before the connectedCallback of your parent, assume that something like a listboxNode is already available and rendered, because it usually isn't. @MathieuPuech 's case shows that as well.

The reason we get problems in our extension layer is because the order of what gets firstUpdated first (the parent or the child) is reversed. In general, what gets rendered in which order is not something you can reliably expect between different components, so your logic should not be based on it.

I was able to create a workaround 🎉  on our extension layer to fix things, I basically just do:
```js
connectedCallback() {
  if (super.connectedCallback) {
    super.connectedCallback();
    if (!this.registrationTarget) {
      this.registrationTarget = this.parentElement;
    }
  }
}
```

Which temporarily sets the registration target to the parent element if it wasn't set by the rich select yet (because its firstUpdated didn't get called yet). This seems to work okay, because later the child element gets moved across the DOM because we're portalling to global overlay content. So then the registrationTarget is set again (but this time by the parent), and form registration is done from the start again, with the right target this time.

I don't understand the process for a full 100% and it's probably hard to understand from my description, but @daKmoR  is making a POC to simplify our registration logic and make it synchronous again. The whole reason we have this async events-based registration system is due to some IE11 problems we had, that we might now have a better solution for. So I'm not gonna dive into it much further right now :P

PR #827, opened 2020-07-22
chore: fix extend docs fenced codeblocks in docs

jorenbroekema:
This was the only way for me to make it look good in storybook

### DIFF
--- a/docs/extending-documentation.md
+++ b/docs/extending-documentation.md
@@ -214,23 +214,34 @@ In the code above we have 2 places in where you can do overrides: `./.storybook/
 
 We insert a remarkExtend plugin instance for both the local overrides and the global overrides, and pass the necessary arguments. Then we return the new array of plugins for Mdjs.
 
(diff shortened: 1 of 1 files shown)


docs/extending-documentation.md
```diff
@@ -214,23 +214,34 @@ In the code above we have 2 places in where you can do overrides: `./.storybook/
...
 
-````md
-```
+Use triple backtick \` to create a fenced codeblock ` ``` `, and put inside it:
+
+```md
 ::addMdAfter(':root')
 ```
 
-```js script
```
daKmoR, line 224: did you try using 4 back ticks? or 3 `~` tildes as a guard around the guard?

jorenbroekema: I tried 4 back ticks, looks very ugly because the storybook md parser doesn't handle it well

jorenbroekema: I will try tildes now

jorenbroekema: Storybook's md parser will take ~~~ as opening and then the ``` as closing :( so also not possible

PR #829, opened 2020-07-22
fix(select-rich): add serializedValue setter due to getter override

jorenbroekema:


packages/select-rich/src/LionSelectRich.js
```diff
@@ -159,6 +159,12 @@ export class LionSelectRich extends ScopedElementsMixin(
     return this.modelValue;
   }
 
+  // Duplicating from FormGroupMixin, because you cannot independently inherit/override getter + setter.
+  // If you override one, gotta override the other, they go in pairs.
+  set serializedValue(value) {
+    this._setCheckedElements(value, (el, val) => el.serializedValue.value === val);
```
daKmoR, line 165: I think this should work... e.g. we have it but we still call the parent one

```suggestion
    super.serializedValue = value;
```

jorenbroekema: Good one, works!

PR #830, opened 2020-07-22
WIP: Feat/type localize

narzac:


packages/localize/src/LocalizeManager.js
```diff
@@ -138,18 +163,41 @@ export class LocalizeManager extends LionSingleton {
...
+    return Promise.all(
+      namespaces.map(
+        /** @param {Object.<string, Function> | string} namespace */
+        namespace => this.loadNamespace(namespace, { locale }),
+      ),
+    );
   }
 
+  /**
+   * @param {Object.<string, Function> | string} namespaceObj
```
daKmoR, line 190: this type get's used a lot... probably best to define it and then use it

either via a JsDoc TypeDef or within the `types.ts`... not sure what the best type would be here 😅

but it would be nice to just be able to write
```
* @param {NamespaceObj} namespaceObj
```

narzac: Good point

PR #835, opened 2020-07-25
fix: input-date safari issue

### LINKED ISSUE #167: [input-date] Safari incorrectly reformats dates
Use Safari and enter a date like 12/12/1212. Tab out of the field. The information in the field becomes 11/12/1212, instead of staying 12/12/1212. 

I could only reproduce this with dates in the past. The edge of this case seems to be in 1939/1940. 12/12/1939 turns into 11/12/1939 when tabbing out. 12/12/1940 does not change.

onkar75:
Fixes #167

### DIFF
--- a/packages/localize/src/date/isoUTCConverter.js
+++ b/packages/localize/src/date/isoUTCConverter.js
@@ -1,0 +1,26 @@
+/**
+ * To parse a string into the UTC Date format
+ *
+ * @param string
+ * @returns {Date}
+ */
+const parseISOString = s => {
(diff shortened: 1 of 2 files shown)


daKmoR:
hey there 👋

are you going to keep working on this? if so a rebase to latest master would be good 🤗
and there are quite some failing tests 😅

is there any way we can help?

gerjanvangeest:
@onkar75 thanks for the initial attempt.
We fixed the issue in another MR: https://github.com/ing-bank/lion/pull/1021

PR #839, opened 2020-07-27
chore: migrate to changesets and github actions

jorenbroekema:


.lintstagedrc.js
```diff
@@ -3,12 +3,6 @@ module.exports = {
   '*': ['eclint fix', 'git add'],
   '*.js': ['eslint --fix', 'prettier --write', 'git add'],
   '*.md': ['prettier --write', 'markdownlint', 'git add'],
-  'yarn.lock': ['node ./scripts/yarn-lock-scan.js'],
```
daKmoR, line 6: still valuable?

.lintstagedrc.js
```diff
@@ -3,12 +3,6 @@ module.exports = {
...
-  '*package.json': absolutePaths => {
-    const sortPackages = [];
-    absolutePaths.forEach(p => {
-      sortPackages.push(`node ./scripts/sort-package-json.js ${p}`);
-    });
-    return [...sortPackages, `node ./scripts/lint-versions.js`];
```
daKmoR, line 12: linting still makes sense?

.lintstagedrc.js
jorenbroekema: Oops I removed the wrong one :P

.lintstagedrc.js
jorenbroekema: I don't think so right? Because `changeset version` already ensures that interdependent packages are kept updated to latest (or if incompatible, it patches them).

.changeset/config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.1.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "linked": [],
```
daKmoR, line 5: for lion we probably need linked... I'm for example not 100% sure if I need to release overlays if there is an update to localize 🤔

jorenbroekema: What does lerna do currently? I think it does release all right if you release localize..

.lintstagedrc.js
jorenbroekema: still makes sense for new packages getting out of sync and root dev deps

.github/workflows/verify.yml
```diff
@@ -0,0 +1,30 @@
...
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Test
+        run: yarn test
```
daKmoR, line 27: won't this run browser tests in both version of node? 🤔

.changeset/config.json
jorenbroekema: https://github.com/atlassian/changesets/issues/415 created an issue here to get some clarifications and see if it makes sense for us to use linked packages.

.github/workflows/verify.yml
jorenbroekema: resolved this one :)

.changeset/config.json
Andarist: Without the usage of `linked` you get a reasonable flow. Given your example - for any release type of `localize` there would always be an accompanying patch bump of overlays. Unless you mark overlays for a minor or major release, of course.

jorenbroekema: now that we added the changeset action it should be pretty clear also what will happen in terms of bumping in the PR that it auto-generates :) so we can see if things make sense before we actually publish. Thanks for taking a look @Andarist

Andarist: No problem. That's a pretty good trait of Changesets - that you can review the release before you actually make it. If you encounter any problems with the tool - just ping me and I will try to help.

PR #845, opened 2020-07-28
feat(providence-analytics): allow target dependencies via cli

tlouisse:
adds:
- `--include-target-deps`: For all search targets, will include all its dependencies
    (node_modules and bower_components). When --target-deps-filter is applied, a subset
    will be applied that matches the filter condition
- `--target-deps-filter`: Regex condition to be applied to dependencies of search targets.

- extensive cli tests

### DIFF
--- a/packages/providence-analytics/src/cli/cli.js
+++ b/packages/providence-analytics/src/cli/cli.js
@@ -8,22 +8,18 @@ const providenceModule = require('../program/providence.js');
 const { LogService } = require('../program/services/LogService.js');
 const { QueryService } = require('../program/services/QueryService.js');
 const { InputDataService } = require('../program/services/InputDataService.js');
-const { promptAnalyzerMenu, promptAnalyzerConfigMenu } = require('./prompt-analyzer-menu.js');
-const {
-  extensionsFromCs,
-  setQueryMethod,
-  targetDefault,
-  appendProjectDependencyPaths,
-  installDeps,
-  pathsArrayFromCollectionName,
-  pathsArrayFromCs,
-} = require('./cli-helpers.js');
+const promptModule = require('./prompt-analyzer-menu.js');
+const cliHelpers = require('./cli-helpers.js');
 const extendDocsModule = require('./generate-extend-docs-data.js');
+
+const { extensionsFromCs, setQueryMethod, targetDefault, installDeps } = cliHelpers;
+
 const { version } = require('../../package.json');
 
-async function cli({ cwd, addProjectDependencyPaths } = {}) {
+async function cli({ cwd } = {}) {
   let resolveCli;
   let rejectCli;
+
   const cliPromise = new Promise((resolve, reject) => {
     resolveCli = resolve;
     rejectCli = reject;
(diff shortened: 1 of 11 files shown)


packages/providence-analytics/src/cli/cli.js
```diff
@@ -157,39 +161,49 @@ async function cli({ cwd, addProjectDependencyPaths } = {}) {
...
-    .option('--write-log-file', `Writes all logs to 'providence.log' file`);
+    .option('--write-log-file', `Writes all logs to 'providence.log' file`)
+    .option(
+      '--include-target-deps',
+      `For all search targets, will include all its dependencies
+    (node_modules and bower_components). When --target-deps-filter is applied, a subset
+    will be applied that matches the filter condition`,
+    )
+    .option(
+      '--target-deps-filter [target-deps-filter]',
```
jorenbroekema, line 204: Do we need both `include-target-deps` and `target-deps-filter` ?

Proposal:
`--target-dependencies` --> enum `true | false | regex | glob` ?

tlouisse: Yeah, was also thinking about that, but don't think commander really supports that. Or do you mean the string 'true' like `--target-dependencies 'true'`?

jorenbroekema: Yeah exactly, a string enum

jorenbroekema: Could also make it more clear by saying `all | none | <regex> | <glob>` (none = default) so people don't get confused thinking it's a boolean flag.

tlouisse: Yeah, the problem is also a bit that a package could theoretically be called 'true', 'false' etc.

But I like the idea of globs, makes it more consistent with the rest (difference though is that we are dealing with package names instead of paths here). It would also be possible to discriminate between globs and regexes by requiring the latter to be enclosed by '/'. 

So allow `--target-dependencies /^my-prefix.*/` or `--target-dependencies glob*` ...?

jorenbroekema: I think if you have globs you probably don't need regex right? Since globs pretty much covers "regex for file paths". And if you have a package called `all` or `none` all you gotta do is make the path to it non-absolute `all/` or `./all`, should be okay I think :P

tlouisse: The match is actually not based on path names, but on folder names.
We use `read-package-tree`: https://github.com/ing-bank/lion/blob/master/packages/providence-analytics/src/program/utils/read-package-tree-with-bower-support.js#L153 (an inline fork, because we need bower support).
This could possibly be rewritten later, but for now I would propose the regex notation

tlouisse: FIxed 💪

jorenbroekema: Nice, what is the plan for migrating to globs? Any timeline or open issue for it? I think it's good for now but for many people this will be rather unexpected to use a regex instead of glob for file pattern.

PR #853, opened 2020-08-03
feat(form-core): add types for form-core

jorenbroekema:


packages/form-core/src/registration/FormControlsCollection.js
```diff
@@ -91,8 +93,9 @@
 export class FormControlsCollection extends Array {
   /**
    * @desc Gives back the named keys and filters out array indexes
+   * @return {string[]}
    */
-  keys() {
+  _keys() {
```
jorenbroekema, line 98: Technically breaking change for `formElements`, but we assume no one uses formElemenets.keys() ? Not entirely sure if this is ok, but if we extend Array, TS will not let us override private `keys` method :(

daKmoR: imho that's good 👍

packages/form-core/types/FormatMixinTypes.d.ts
```diff
@@ -0,0 +1,52 @@
...
+  formatOn: string;
+  formatOptions: FormatOptions;
+  value: string;
+  __preventRecursiveTrigger: boolean;
+  __isHandlingUserInput: boolean;
+
+  parser(v: string, opts: FormatOptions): unknown;
+  formatter(v: unknown, opts: FormatOptions): string;
+  serializer(v: unknown): string;
+  deserializer(v: unknown): unknown;
```
tlouisse, line 30: this should get `(v:string)`

PR #857, opened 2020-08-04
feat: add loading indicator

ruphin:
This adds a loading indicator package. There are some things I'm not super satisfied with, so please review carefully and check my notes.


packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -0,0 +1,110 @@
+import { svg, css, LitElement } from '@lion/core';
+import { localize, LocalizeMixin } from '@lion/localize';
+
+export class LionProgressIndicator extends LocalizeMixin(LitElement) {
+  static get styles() {
```
ruphin, line 5: By adding this, these styles will be added to any sub-implementation that uses this to extend the base styles: 

```
static get styles() {
   return [...super.styles, css`...`]
}

So if you extend this component specifically you need to NOT include the styles from super. Is that a pattern that you use elsewhere?

File: packages/progress-indicator/src/LionProgressIndicator.js [hunk 2]
```diff
@@ -0,0 +1,110 @@
...
+          display: inline-block;
+        }
+
+        svg.lion-progress {
+          animation: progress-indicator-rotate 2s linear infinite;
+          height: 40px;
+          width: 40px;
+        }
+
+        @keyframes progress-indicator-rotate {
```
ruphin, line 18: These keyframes are registered "globally" so there is risk of naming collision with an extension of this component.

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -0,0 +1,110 @@
...
+
+export class LionProgressIndicator extends LocalizeMixin(LitElement) {
+  static get styles() {
+    return [
+      css`
+        :host {
+          display: inline-block;
+        }
+
+        svg.lion-progress {
```
ruphin, line 12: I have to add a class specifier to the svg tag in order to avoid naming collision with extensions of this component to be safe. I don't particularly like that

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -0,0 +1,110 @@
...
+            default:
+              return import('../translations/en.js');
+          }
+        },
+      },
+    ];
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  template() {
```
ruphin, line 88: This and the styles are more or less superfluous as they will always be overridden. I don't feel like they add much value, unless someone intends to use Lion directly without extending the components, which I think is not a use case, right?

packages/progress-indicator/translations/en.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  loading: 'Loading',
```
ruphin, line 2: The accessibility readers will see this component as "Loading", which is probably correct in nearly all cases, but there may be places where a progress indicator is used that does not indicate "Loading" but something else. Do we want to make it possible to override this?

ruphin:
TODO: Add a changeset to this pull request

packages/progress-indicator/package-lock.json
```diff
@@ -0,0 +1,46 @@
+{
```
gerjanvangeest, line 1: We don't save `package.lock` files

packages/progress-indicator/src/LionProgressIndicator.js
gerjanvangeest: We don't use `lion` in the classname, so if you fully override the styles you don't need to keep calling it `lion-`. Just use `.progress-indicator`.

packages/progress-indicator/src/LionProgressIndicator.js
gerjanvangeest: In all cases so far we have some minimal styling we would like to reuse. In e.g. the button we have put commends like `/* if you extend, please overwrite */` inside the css on parts that can be overwritten.

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: maybe call this `_graphicTemplate` ? (to comply with our naming conventions)

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -0,0 +1,110 @@
...
+    `;
+  }
+
+  render() {
+    return this.template();
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute('role', 'status');
```
tlouisse, line 102: so we decided not to go for https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role?

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: I think the button is not the right component to take as an example.

Ideally, it should always be:

```
static get styles() {
   return [...super.styles, css`...`]
}
```

I think because of all reasons mentioned (keyframe name collision, lion prefixed classes, redundant css) it 
would be way nicer to give this component empty styles by default (maybe only keep inline-block) and create an extension demo like here: https://github.com/ing-bank/lion/blob/master/packages/collapsible/README.md


The main section of the demo can then just explain what it does: a11y and i18n.
Unless @daKmoR sees harm in this approach.

daKmoR: it's fine for now... we will need to improve the lion demos at some point... but that's for later

packages/progress-indicator/src/LionProgressIndicator.js
ruphin: I will change this

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -0,0 +1,110 @@
...
+            case 'ru-RU':
+            case 'ru':
+              return import('../translations/ru.js');
+            case 'sk-SK':
+            case 'sk':
+              return import('./translations/sk.js');
+            case 'uk-UA':
+            case 'uk':
+              return import('../translations/uk.js');
+            default:
```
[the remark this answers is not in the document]
ruphin: I will add these

packages/progress-indicator/package-lock.json
ruphin: I will remove this

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: It probably will be needed if we want to make this the basis for a linear progress indicator as well and if the progress needs to have a concrete value

ruphin: I think the progressbar role only applies to determinate progress indicators. This implementation is the basis for indeterminate progress indicators, and we could provide a basis for a determinate indicator as an extension of this. We could also decide later to extend this and provide explicit support for determinate indicators.

Do you think this is an ok way forward?

packages/progress-indicator/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@lion/progress-indicator",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
ruphin: Fixed

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: After our discussion last week, I was under the impression that we would build a component that lays the foundation for all appearances as listed here: https://material-ui.com/components/progress/
In that case, it should probably be a role=progressbar.

In case of only supporting indetermate indicators, this might be sufficient.
Maybe we should do a follow up story then with a good a11y assessment and cross browser tests....

packages/progress-indicator/package-lock.json
MathieuPuech: We could add package-lock.json files to the .gitignore file:
`/packages/*/package-lock.json`

packages/progress-indicator/src/LionProgressIndicator.js
ruphin: We could add a `determinate` attribute or something similar, and have the internal behaviour set the correct aria attributes based on the chosen mode. Always setting role=progressbar will cause problems for non-determinate indicators, so I think having the current behaviour as a default will be fine. We can add additional support for determinate indicators later in a non-breaking change.

packages/progress-indicator/src/LionProgressIndicator.js
ruphin: Removed the styling altogether

packages/progress-indicator/src/LionProgressIndicator.js
ruphin: I implemented the approach described by @tlouisse

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: I think it's possible to have indeterminate progressbars as well: 
- https://material-ui.com/components/progress/#circular

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role ->
> If the progress value  is indeterminate, the developer should omit the aria-valuenow attribute.

packages/progress-indicator/translations/en.js
tlouisse: Makes sense. We could consider calling a method `_setLabel()` inside `onLocaleUpdated`, so Sub classers can override this

packages/progress-indicator/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Example
+
+```html
+<lion-progress-indicator></lion-progress-indicator>
+```
+
+## Extended indicator with a custom visual
+
+`LionProgressIndicator` is designed to be extended to add visuals. Implement the `_graphicTempplate` method to set the rendered content, and apply styles normally.
```
daKmoR, line 44:
```suggestion
`LionProgressIndicator` is designed to be extended to add visuals. Implement the `_graphicTemplate` method to set the rendered content, and apply styles normally.
```

packages/progress-indicator/src/LionProgressIndicator.js
tlouisse: For now we can probably leave at is and later extend a11y, without breaking the public api.

ruphin:
I added a changeset to release version 0.1.0 of this. I think all comments have been addressed.

The tests are failing because test coverage is not high enough, although I don't think it's caused by this package. The tests for this are passing.

Feel free to merge if it passes review.

Issue #859, opened 2020-08-05
[checkbox-group and radio-group]: formattedValue is undefined


palash2601:
## Describe the bug

The `formattedValue` for `lion-checkbox-group` and `lion-radio-group` is undefined.

## To Reproduce

This is the most important part. Without a simple way of reproducing the issue, we can not verify your bug.

Here are a few ways you can help us to reproduce the issue:

Can be reproduced in Lion Demo storybook

1. Go to https://lion-web-components.netlify.app/?path=/docs/forms-checkbox-group--main
2. CHeck for `formattedValue` of checkbox-group in the given example

## Expected behavior

The `formattedValue` for `checkbox-group` and `radio-group` should return formatted value of 'checked' child elements(checkbox or radio element) in array

in demo example, if all elements are checked then: ['Archimedes', 'Francis Bacon', 'Marie Curie']

in demo example, if first 2 elements are checked then: ['Archimedes', 'Francis Bacon']

PR #862, opened 2020-08-06
fix(form-core): form-group should have a valid formattedValue

daKmoR:
closes https://github.com/ing-bank/lion/issues/859

still working on it


package.json
```diff
@@ -9,6 +9,7 @@
     "build:docs": "wca analyze \"packages/tabs/**/*.js\"",
     "build:types": "tsc -p tsconfig.build.types.json",
     "bundlesize": "rollup -c bundlesize/rollup.config.js && bundlesize",
+    "debug": "web-test-runner \"packages/form-core/test/**/*.test.js\" --watch",
```
daKmoR, line 12: added this so you can always change it to which package you are currently working on...
later I wanna make it a real command so you can do 
```
yarn debug -g form-core
yarn debug -g overlays
...
```

requires some changes in web-test-runner so postponing this until then

tlouisse: so now only form-core supported on purpose?

packages/form-core/test/choice-group/ChoiceGroupMixin.test.js
```diff
@@ -121,6 +136,19 @@ describe('ChoiceGroupMixin', () => {
     expect(el.formElements[2].checked).to.be.true;
   });
 
+  it('can set initial formattedValue on creation', async () => {
```
tlouisse, line 139: should we also check with `multipleChoice`?

daKmoR: added 👍

package.json
daKmoR: right now I edit `form-core` to what I am working on in the `package.json`

daKmoR: which is of course a workaround 😅

PR #863, opened 2020-08-07
fix: [overlay] fix unhandled rejection

Evittta:


Evittta:
I'll create another one

Issue #826, opened 2020-07-22
[overlays] Test "uses contentWrapperNode as provided for local positioning" has Unhandled rejection


daKmoR:
## Describe the bug

`Unhandled rejection: Error: Failed to execute 'appendChild' on 'Node': The new child element contains the parent.`

## Expected behavior

No Unhandled rejection

daKmoR:
fixed by https://github.com/ing-bank/lion/pull/864

PR #864, opened 2020-08-07
Fix/overlay unhandled rejection

Evittta:
Fix of issue - https://github.com/ing-bank/lion/issues/826

OverlayController.js has unhandled rejection -
Unhandled rejection: Error: Failed to execute 'appendChild' on 'Node': The new child element contains the parent.


Evittta:
Fix in #865

Issue #851, opened 2020-08-02
[providence] Extend docs config is empty for found dependencies


jorenbroekema:
## Describe the bug

When I create a new project and try to use some lion dependency, providence does not output the query results properly, or does not find the matches.

providence-extend-docs-data.json
```js
["[no-dependency]","[no-dependency]","[no-dependency]"]
```

When using `@lion/input`. The no-dependency ones make sense because `@lion/input` installs `localize`, `form-core` and `core`, upon which there is no direct dependency.

## To Reproduce

1, Create a new project/folder
2. `git init && npm init -y`
3. `yarn add lit-element @lion/input && yarn add --dev providence-analytics
4. Create a .gitignore
```
/dist/
/node_modules/
```
5. Create a custom input --> `/src/CustomInput.js`
```js
import { LionInput } from '@lion/input';
  
export class CustomInput extends LionInput {}
```
6. Create an element definition file --> `/custom-input.js`
```js
import { CustomInput } from './src/custom-input/CustomInput.js';

customElements.define('custom-input', CustomInput);
```
7. Add these two files to `package.json` --> `files` property (optional, does not make a difference for the result)
8. `yarn providence extend-docs -r 'node_modules/@lion/*'  --prefix-from lion --prefix-to custom`

## Expected behavior

A `providence-extend-docs-data.json` file with some meaningful output but I get:

providence-extend-docs-data.json
```js
["[no-dependency]","[no-dependency]","[no-dependency]"]
```

instead.

jorenbroekema:
Note: it works properly in `0.1.0` but is then broken from `0.2.0` onwards

tlouisse:
Somehow it draws the conclusion that there is no dependency (for instance` @lion/input`) present: https://github.com/ing-bank/lion/blob/master/packages/providence-analytics/src/program/analyzers/helpers/Analyzer.js#L118
This should be based on all data package jsons ref and target (name, dependencies, devDependencies)

jorenbroekema:
Yes that conclusion is correct. But apart from the three times no-dependency, it should also push a sibling with the data for the one that does have a direct dependency.
- @lion/input, dependency --> missing data :(
- @lion/localize, no dependency --> correct
- @lion/form-core, no dependency --> correct
- @lion/core, no dependency --> correct

tlouisse:
Ah. Seems related to anymatch/ 'auto whitelist' introduction as well (v0.2.0).
[...]

Issue #850, opened 2020-08-02
[providence] Does not search distribution files


jorenbroekema:
## Describe the bug

Providence does not analyze my target project properly. I publish my library to NPM with the following config, where I build the `dist` on prepublish:
```json
{
  "files": [
    "dist"
  ]
}
```

This means that the `dist` folder ends up in the `keepFilter` here: https://github.com/ing-bank/lion/blob/master/packages/providence-analytics/src/program/services/InputDataService.js#L297

However, since it's an auto-generated folder, it also ends up in the `removeFilter` because I put `/dist/` in my `.gitignore` file.

Therefore, neither my `src` nor my `dist` gets searched, and I end up with an empty `providence-extend-docs-data.json`.

## To Reproduce

1. Create a project with `/dist/` in .gitignore, but with `dist` inside the files property in `package.json`
2. Install `@lion/dialog` for example, so we have a reference project
3. Install providence `yarn add --dev providence-analytics`
4. `yarn providence extend-docs -r 'node_modules/@lion/*' --prefix-from lion --prefix-to pre`

## Expected behavior

A non-empty providence-extend-docs-data.json, because I have some components that extend LionDialog in both my 'src' and 'dist'.

## Screenshots

```
No files found for path '<localPath>'
```
Where localPath is your root folder of your project.

## Potential Solution

It definitely needs to analyze either your source code or your published code. For the purpose of extending lion docs, I would prefer if it analyzed and rewrote paths to my `src` folder, because that makes my docs/demos easier to debug if something goes wrong there. Furthermore, I don't want my docs/demos to be reliant on my build-process that creates the `/dist` folder that I publish to NPM.

So, question is then, what should the solution be? Should the `keepFilter` be empty if the user hasn't supplied one themselves? That way it searches (`anymatch`es) everything that's not in the `removeFilter`, which means that it will search `src` unless otherwise specified by the user. In my opinion, relying on the `package.json` `files` entry, is a bad idea, since so many projects publish a built folder (e.g. `dist`) that does not get committed to git.

And if the keepFilter is not prepopulated with the NPM files field, how should the user specify one?

Some ideas (which involves an empty / unused keepFilter if not used):

- Use the already available whitelist CLI flags
- providence external config only solution, `keepFilter`
- A new CLI flag `--keep-filter <glob>` but this is a bit harder if you want to allow an array of globs or more complex things which `anymatch` accepts. Alternatively, could make it a boolean flag: `--scan-published-files-only`?

[...]

PR #870, opened 2020-08-10
feat(providence-analytics): add --allowlist-mode option

tlouisse:
closes https://github.com/ing-bank/lion/issues/850
closes https://github.com/ing-bank/lion/issues/851

### DIFF
--- a/.changeset/clever-trains-remember.md
+++ b/.changeset/clever-trains-remember.md
@@ -2,6 +2,14 @@
 'providence-analytics': minor
 ---
 
-Allowlist mode: autodetects whether analyzed repository is a "git" or "npm" (published artifact) repository.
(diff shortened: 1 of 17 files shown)


.changeset/clever-trains-remember.md
```diff
@@ -0,0 +1,15 @@
...
+
+#### Features
+
+- Allowlist mode: autodetects whether analyzed repository is a "git" or "npm" (published artifact) repository.
+  Via the cli `--allowlist-mode 'npm|git|all'` and `--allowlist-mode-reference 'npm|git|all'` can be
+  configured to override the autodetected mode.
+
+#### Bugfixes
+
+- clean output extend-docs
```
jorenbroekema, line 15: This is to remove the [no-dependency] right? Could be clearer maybe

PR #878, opened 2020-08-12
fix(providence-analytics): allowlist takes precedence over -mode

tlouisse:


packages/providence-analytics/src/program/services/InputDataService.js
```diff
@@ -338,32 +341,47 @@ class InputDataService {
...
+      const { negated, pattern } = isNegatedGlob(allowEntry);
       if (negated) {
         removeFilter.push(pattern);
       } else {
-        keepFilter.push(filterEntry);
+        keepFilter.push(allowEntry);
       }
     });
-
+    // console.log('keepFilter', keepFilter);
```
gerjanvangeest, line 352: console?

tlouisse: yeah, I fixed already locally :)

PR #881, opened 2020-08-12
fix(form-core): form should be able to access formattedValue of children

daKmoR:


packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -157,7 +157,11 @@ export const ChoiceGroupMixin = dedupeMixin(
       _getFromAllFormElements(property, filterCondition = () => true) {
         // For modelValue and serializedValue, an exception should be made,
         // The reset can be requested from children
-        if (property === 'modelValue' || property === 'serializedValue') {
```
tlouisse, line 160: nitpick: comment should also mention formattedValue

packages/form-integrations/test/helpers/umbrella-form.js
```diff
@@ -64,11 +64,11 @@ export class UmbrellaForm extends LitElement {
           <lion-input-email name="email" label="Email"></lion-input-email>
           <lion-checkbox-group
             label="What do you like?"
-            name="checkers[]"
```
jorenbroekema, line 67: Do we no longer strongly encourage the `[]` here for backend form integration?

daKmoR: a group should never had it 😅  
all it's children should have gotten it...

and when using it we run into a different bug 🙈
but that is for another merge request

packages/form-core/src/choice-group/ChoiceGroupMixin.js
daKmoR: updated

packages/form-integrations/test/helpers/umbrella-form.js
tlouisse: It currently breaks if you do.
This could be fixed by changin line 88 of FormRegistrarMixin: `if (name.substr(-2) === '[]' && this._repropagationRole !== 'choice-group') {`

PR #882, opened 2020-08-13
feat(form-core): type validate

jorenbroekema:


packages/form-core/src/validate/Validator.js
```diff
@@ -75,29 +89,32 @@ export class Validator {
...
    */
-  static async getMessage(/* data */) {
+  // eslint-disable-next-line no-unused-vars
+  static async getMessage(data) {
     return `Please configure an error message for "${this.name}" by overriding "static async getMessage()"`;
   }
 
   /**
-   * @param {FormControl} formControl
+   * @param {HTMLElement} formControl
```
tlouisse, line 112: Would it be possible to make this of type FormControlMixin?

packages/form-core/src/validate/Validator.js
```diff
@@ -111,6 +128,38 @@ export class Validator {
    * - Or, when a webworker was started, its process could be aborted and then restarted.
    */
   abortExecution() {} // eslint-disable-line
+
+  __fakeExtendsEventTarget() {
```
tlouisse, line 132: Would it be the right moment to abstract this throughout the whole codebase?

Have the feeling that tech debt increases as this gets more verbose

packages/form-core/src/validate/resultValidators/DefaultSuccess.js
```diff
@@ -1,14 +1,29 @@
...
+   * @param  {...any} args
+   */
   constructor(...args) {
     super(...args);
     this.type = 'success';
   }
 
+  /**
+   *
+   * @param {Object} context
```
tlouisse, line 18: what is actually the difference between `Object` and `object`? 
Do we use the habit of lowercasing platform primitives (like we do with string and boolean?)

packages/form-core/src/validate/validators/DateValidators.js
```diff
@@ -14,6 +17,9 @@ export class IsDate extends Validator {
     return 'IsDate';
   }
 
+  /**
+   * @param {?} value
```
tlouisse, line 21: type Date

packages/form-core/test/validate/ResultValidator.test.js
```diff
@@ -3,18 +3,29 @@ import { ResultValidator } from '../../src/validate/ResultValidator.js';
...
+       */
+      executeOnResults({ regularValidationResult, prevValidationResult }) {
+        const hasSuccess = regularValidationResult.length && !prevValidationResult.length;
+        return !!hasSuccess;
       }
     }
     expect(
       new MyResultValidator().executeOnResults({
-        regularValidateResult: [new Required(), new MinLength(3)],
+        regularValidationResult: [new Required(), new MinLength(3)],
```
tlouisse, line 28: Agree with the name change, but potentially breaking

packages/form-core/test/validate/Validator.test.js
```diff
@@ -103,15 +116,19 @@ describe('Validator', () => {
   it('fires "param-changed" event on param change', async () => {
     const vali = new Validator('foo');
     const cb = sinon.spy(() => {});
-    vali.addEventListener('param-changed', cb);
+    if (vali.addEventListener) {
```
tlouisse, line 119: Is this needed because TS doesn't know about existance of `addEventListener` on prototype?

In that case we could better add (gets rid of tech debt and these workarounds):
```
EventTargetMixin extends superclass {
   addEventListener(...){...}
}
```

packages/form-core/test/validate/Validator.test.js
```diff
@@ -128,22 +145,14 @@ describe('Validator', () => {
...
     class MyValidator extends Validator {
+      /**
+       * @param {string} modelValue
+       * @param {string} param
+       */
       execute(modelValue, param) {
         const hasError = modelValue === 'forbidden' && param === 'values';
         return hasError;
       }
-
```
tlouisse, line 135: I know these lines are not needed for the test to pass.
But since we cannot document every small feature, tests can serve here as a backup wrt how things are meant to be used.
Could yoou maybe add this info to https://github.com/ing-bank/lion/blob/master/packages/form-core/src/validate/Validator.js#L97 instead?

packages/form-core/src/validate/LionValidationFeedback.js
```diff
@@ -1,37 +1,51 @@
...
+    super.updated(changedProperties);
     if (this.feedbackData && this.feedbackData[0]) {
       this.setAttribute('type', this.feedbackData[0].type);
       this.currentType = this.feedbackData[0].type;
       window.clearTimeout(this.removeMessage);
       if (this.currentType === 'success') {
         this.removeMessage = window.setTimeout(() => {
           this.removeAttribute('type');
-          this.feedbackData = '';
+          /** @type {messageMap[]} */
```
tlouisse, line 47: Where is this type defined? Shouldn't custom types always be capitalized?

packages/form-core/src/validate/Validator.js
```diff
@@ -51,16 +64,17 @@ export class Validator {
...
-   * @param {string} data.fieldName
-   * @param {*} data.params
-   * @param {string} data.type
-   * @returns {string|Node|Promise<stringOrNode>|() => stringOrNode)}
+   * @param {object} [data]
+   * @param {*} [data.modelValue]
+   * @param {string} [data.fieldName]
+   * @param {*} [data.params]
+   * @param {string|undefined} [data.type]
+   * @returns {Promise<string|Node>}
```
tlouisse, line 72: It can still return a string/Node withour promise?

packages/form-core/src/validate/Validator.js
```diff
@@ -75,29 +89,32 @@ export class Validator {
...
-   * @param {string} data.type
-   * @returns {string|Node|Promise<stringOrNode>|() => stringOrNode)}
+   * @param {object} [data]
+   * @param {*} [data.modelValue]
+   * @param {string} [data.fieldName]
+   * @param {*} [data.params]
+   * @param {string} [data.type]
+   * @param {Object.<string,?>} [data.config]
+   * @param {string} [data.name]
+   * @returns {Promise<string|Node>}
```
tlouisse, line 104: It can still return a string/Node withour promise?

packages/form-core/src/validate/LionValidationFeedback.js
jorenbroekema: Yeah as a convention we should probably capitalize. Defined at the top of the file, but perhaps better to put it in the types folder in one of the type definition files and import it always from this one place. I'm thinking ValidateMixinTypes.d.ts but I didn't create this file yet.

packages/form-core/src/validate/Validator.js
jorenbroekema: Since it's an async method, it always returns a promise no matter what. What is resolved is either a string or a Node. Or am I misunderstanding?

packages/form-core/src/validate/Validator.js
jorenbroekema: Would it make sense to eventually put {LionField} here? Since that's the most logic common denominator I think? Otherwise FormControlMixin would make sense yes.

packages/form-core/src/validate/Validator.js
jorenbroekema: We want to get rid of this eventually right? We only do it in two places, here and in overlay controller I believe (or mixin..). For me personally, I don't like to abstract until we get 3 instances and now we have two. Also, abstracting this is really really tough for typing, which is one of the reasons why I added it here inside the class instead of as an external util. The reason why it's tough to type when you add these methods through an external method, is because you need to somehow tell this class that those methods (addEventListener etc.) have now become available because we called fakeExtendsEventTarget in the constructor. To my knowledge there was no elegant way to do that unless we make the fakeExtendsEventTarget a mixin so the methods can be inherited.

packages/form-core/test/validate/ResultValidator.test.js
jorenbroekema: Actually I didn't change the name, I just aligned it with the source code. I got a type error that showed that regularValidateResult prop does not exist, so yay for types showing us we have a bug in our test :). too bad the test still passed before this change though

packages/form-core/test/validate/Validator.test.js
jorenbroekema: Ah yeah I suggested the same in a previous reply to your comments :P I think a mixin is the most elegant solution, and although I try not to make hasty abstractions, here it makes sense for ease of typing.. but perhaps we can do this in a separate MR for keeping the scope on form-core. 

We will probably do it in lion core? Then we release everything, so also good idea to sync that MR with this MR, since this one will release all form related packages.

packages/form-core/test/validate/Validator.test.js
jorenbroekema: Sure, will do!

packages/form-core/src/validate/Validator.js
tlouisse: You can determine whether it's async or not when you override it

packages/form-core/src/validate/Validator.js
tlouisse: No, because fieldsets/formgroups don't inherit from LionField :)

packages/form-core/src/validate/Validator.js
tlouisse: I think we use it in 4 places throughout our code base (overlay, validate, localize, form registration).
Let's fix it in a later pr 👍

packages/form-core/src/validate/resultValidators/DefaultSuccess.js
jorenbroekema: I think me and @narzac  agreed we should uppercase non-primitives as a convention, although I'm sure you can find some lowercased `function` type annotations that I made a mistake in. To JSDocs it makes 0 difference so it's just a convention we can choose.

narzac:
Looks good

narzac (approved):

gerjanvangeest (approved):

PR #886, opened 2020-08-16
fix(select-rich): event listeners can't be set more than once

fpc97:
fix for issue #884


daKmoR:
thank you for looking into this... but it seems the root cause lies a little deeper 😭

We do have a synchronous setup for events but an asynchronous teardown of events in the OverlayMixin 🤔 
That seems to cause this side effect...

I'm gonna look into it 😅

daKmoR:
in the end the underlying problem was a typing issue introduced by trying to execute two async tasks in a specific order...

it's now fixed via https://github.com/ing-bank/lion/pull/911

again thank you for looking into 🤗

PR #887, opened 2020-08-17
fix(validate-mixin): remove event listener on disconnect instead of add

mblankenstein:


daKmoR:
nice catch 💪

thank you 🙇‍♂️

PR #889, opened 2020-08-19
feat(calendar): add year navigation

gerjanvangeest:


packages/calendar/src/LionCalendar.js
```diff
@@ -165,12 +165,12 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     return this.__focusedDate;
   }
 
-  goToNextMonth() {
-    this.__modifyDate(1, { dateType: 'centralDate', type: 'Month', mode: 'both' });
+  goToNextDate(type) {
```
daKmoR, line 168: I think having 2 separate functions for this would be a nicer user api? 🤔

e.g. for users it would be
- `el.goToNextMonth()`
- `el.goToNextYear()`

brings nicer autocomplete and you can't get it "wrong" if you would for example pass a type `'asdf'`

tlouisse: On top of that, we should not change public method names :)

.changeset/shiny-drinks-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/calendar': patch
```
tlouisse, line 2: Should be a minor I think

gerjanvangeest: Yes, i was doubting. Minor makes sense.

packages/calendar/src/LionCalendar.js
```diff
@@ -302,40 +326,96 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
...
+        }
+      }
+    }
+    return { disabled, month };
+  }
+
+  __renderPreviousButton(type, previousMonth, previousYear) {
+    const { disabled, month } = this.__getPreviousDisabled(type, previousMonth, previousYear);
+    const previousButtonTitle = this.__getNavigationLabel('previous', type, month, previousYear);
+    const clickDateDelegation = function () {
```
tlouisse, line 370: probably better to make it `function clickDateDelegation() {` (for linting)

packages/calendar/src/utils/dayTemplate.js
```diff
@@ -14,14 +14,50 @@ const defaultMonthLabels = [
...
+
+  let firstDay = false;
+  let endOfFirstWeek = false;
+  let startOfFirstFullWeek = false;
+  let endOfLastFullWeek = false;
+  let startOfLastWeek = false;
+  let lastDay = false;
+
+  firstDay = dayNumber === 1;
+  endOfFirstWeek = day.weekOrder === 6 && firstWeekDays.includes(dayNumber);
```
tlouisse, line 34: Maybe just define them as consts here directly (without giving boolean values to what ends up to be numbers)

daKmoR (approved):

PR #890, opened 2020-08-19
test(calendar): add visual regression tests

MathieuPuech:


jorenbroekema (approved):

jorenbroekema:
Sorry it took a while, we were busy typing all of lion and deciding what to do with these globals, decided in the end to just do `@ts-nocheck` here 👍 

Thanks for adding!

PR #894, opened 2020-08-20
fix(core): fix types for core mixins to inherit Lit/HTMLElement super

jorenbroekema:


package.json
```diff
@@ -51,6 +51,7 @@
     "@web/test-runner": "^0.7.3",
     "@web/test-runner-browserstack": "^0.1.1",
     "@web/test-runner-playwright": "^0.5.1",
+    "@web/test-runner-puppeteer": "^0.6.1",
```
jorenbroekema, line 54: Allows WSL/WSL2 to at least run browser tests with `--puppeteer` flag :P

packages/core/src/DisabledMixin.js
```diff
@@ -19,23 +23,23 @@ const DisabledMixinImplementation = superclass =>
 
     constructor() {
       super();
-      this.__requestedToBeDisabled = false;
+      this._requestedToBeDisabled = false;
```
jorenbroekema, line 26: Made protected, because the disabledmixinwithtabindex uses it.

packages/core/src/DelegateMixin.js
```diff
@@ -40,9 +44,7 @@ const DelegateMixinImplementation = superclass =>
     }
 
     connectedCallback() {
-      if (super.connectedCallback) {
```
jorenbroekema, line 43: guard not necessary if you pass LitElement as a minimum superclass, which always has this method

packages/core/src/SlotMixin.js
```diff
@@ -23,7 +26,9 @@ const SlotMixinImplementation = superclass =>
     }
 
     connectedCallback() {
+      // @ts-ignore checking this in case we pass LitElement, found no good way to type this...
```
tlouisse, line 29: Is there a reason this mixin is not dependent on LitEl?

tlouisse (approved):

packages/core/src/SlotMixin.js
jorenbroekema: Because it could be that you pass HTMLElement instead. So it's either HTMLElement or LitElement, but you can't do something like `@type {LitElement | HTMLElement}` properly :( but feel free to experiment and see if you can find a better way

Issue #899, opened 2020-08-21
[form-core] validations are not filtered according the feedback condition


manolakis:
## Describe the bug

Having different validation types, and specific feedback conditions to show those types, validations seems to not be filtered by those conditions, which causes wrong validation feedback messages to be shown.

## To Reproduce

With the following class

```js
class TestClass extends ValidateMixin(LitElement) {
  static get validationTypes() {
    return ['error', 'info'];
  }

  static get properties() {
    return {
      modelValue: { type: String },
    };
  }

  _showFeedbackConditionFor(type) {
    switch (type) {
      case 'error':
        return ['A', 'B'].includes(this.modelValue);
      default:
        return ['B', 'C'].includes(this.modelValue);
    }
  }
},
```

and using some custom validators that always returns `true`

```html
<test-class
  .validators=${[
    new AlwaysInvalid({}, { type: 'error' }),
    new AlwaysInvalid({}, { type: 'info' }),
  ]}
></test-class>
```

then, depending on the `modelValue` the following feedback message are shown.

| modelValue | feedback message | Comment |
|------|--------------------| ----------|
| 'A' | 'error'| |
| 'B' | 'error' | Both are true, but 'error' has precedence over 'info' |
| 'C' | **'error'** | **It should be 'info', but 'error' has not been filtered and has more precedence than 'info'** |
| 'D' | | |

## Expected behaviour

| modelValue | feedback message | Comment |
|------|--------------------| ----------|
| 'A' | 'error' ||
| 'B' | 'error' | Both are true, but 'error' has precedence over 'info' |
| 'C' | 'info' ||
| 'D' | | |

PR #900, opened 2020-08-21
fix(form-core): filter feedback messages according feedback conditions

manolakis:
This commit filters the validators by the showing feedback conditions in order to fix https://github.com/ing-bank/lion/issues/899

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -636,7 +636,9 @@ export const ValidateMixinImplementation = superclass =>
         .constructor);
       const types = ctor.validationTypes;
       // Sort all validators based on the type provided.
-      const res = validationResult.sort((a, b) => types.indexOf(a.type) - types.indexOf(b.type));
+      const res = validationResult
+        .filter(v => this._showFeedbackConditionFor(v.type))
+        .sort((a, b) => types.indexOf(a.type) - types.indexOf(b.type));
       return res.slice(0, this._visibleMessagesAmount);
     }
   };
--- a/.changeset/cold-panthers-tie.md
+++ b/.changeset/cold-panthers-tie.md
@@ -1,0 +1,5 @@
+---
+'@lion/form-core': patch
+---
+
+filter feedback messages according feedback conditions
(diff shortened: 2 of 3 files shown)


packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -556,7 +556,9 @@ export const ValidateMixin = dedupeMixin(
           this.__feedbackQueue.add(async () => {
             /** @type {Validator[]} */
             this.__prioritizedResult = this._prioritizeAndFilterFeedback({
-              validationResult: this.__validationResult,
+              validationResult: this.__validationResult.filter(({ type }) =>
```
tlouisse, line 559: Ah, I thought you meant changing `this.__validationResult` when we spoke before. Altering the param is indeed better (then we keep the 'validation truth regardless of interaction states'), but we would have to see if this is the parameter that we expect in  `_prioritizeAndFilterFeedback`.

I will plan a story on our backlog to write Subclasser (or maybe actually Application Developer, then we need to 'promote' this api to be public) documentation for this and better explore all edge cases.
After that, we will decide whether we solve it like this or differently.

Rest of the merge request (tests) is really great, thanks 👍

manolakis: ok, perfect 👍

tlouisse: I think the best solution would be to do: 
```js
_prioritizeAndFilterFeedback({ validationResult }) {
    return super._prioritizeAndFilterFeedback({
      validationResult: validationResult.filter(v => this._showFeedbackConditionFor(v.type)),
    });
  }
}
```

Almost the same, but it still gives you the freedom to override `_prioritizeAndFilterFeedback` and have access to the full `validationResult` :)

jorenbroekema:
@manolakis  hey! Are you still working on this? Was Thijs' comment addressed already?

manolakis:
I think there was a misunderstanding because I though you were going to address it. I'll update the PR. Is that ok for you?

jorenbroekema:
Aahh yeahh my bad I didn't see this part:

> I will plan a story on our backlog to write Subclasser (or maybe actually Application Developer, then we need to 'promote' this api to be public) documentation for this and better explore all edge cases.

Probably should align with @tlouisse  then 😅

manolakis:
I'm updating the PR with @tlouisse comment addressed. Feel free to close it.

manolakis:
@jorenbroekema @tlouisse PR updated and rebased with the master. A lot of changes related to types! 😅

jorenbroekema (approved):

jorenbroekema:
Thanks @manolakis  :)

PR #908, opened 2020-08-27
feat: update to latest lit-element version

gerjanvangeest:


tlouisse (approved):

tlouisse (changes requested):
providence-analytics can be a patch release I think

tlouisse (approved):

PR #911, opened 2020-08-31
feat(overlays): make OverlayController constructor tasks sync

daKmoR:
closes https://github.com/ing-bank/lion/issues/884

So that was way more work than expected 😅

What I did:

- Make the OverlayController constructor phase synchronous. 
- Trigger a setup of the OverlayController on every connectedCallback
- Execute a new OverlayController after (shadowDom) rendering is of the element is done.
- Teardown the OverlayController on every disconnectedCallback
- This means moving a dialog triggers teardown in the old location and setup in the new location
- Restore the original light dom if needed in the teardown phase of the OverlayController


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -207,7 +207,7 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
   }
 
   /**
-   * Defining this overlay as a templates lets OverlayInteraceMixin
+   * Defining this overlay as a templates lets OverlayInteractionMixin
```
tlouisse, line 210: OverlayMixin  :)

daKmoR: changed

packages/tooltip/src/LionTooltip.js
```diff
@@ -90,12 +90,6 @@ export class LionTooltip extends OverlayMixin(LitElement) {
     this.__setupRepositionCompletePromise();
   }
 
-  connectedCallback() {
```
tlouisse, line 93: i will miss this ;)

tlouisse (approved):

Issue #730, opened 2020-05-27
[input] can't change name dynamically


tescuderoSopra:
## Describe the bug

Changing a name dynamically keeps the original name.

## To Reproduce

```javascript
render() {
  // add option (Boolean) to properties to re-render correctly
  const name = this.option ? 'two' : 'one';
  const value = `${name}value`;
  return html`
    <ing-form @submit=${ev => {ev.preventDefault(); console.log(this.shadowRoot.getElementById('group').serializedValue);}}>
      <form action="/">
        <ing-fieldset name="group" id="group">
          <ing-input .name=${name} .modelValue=${value} id="input"></ing-input>
          <ing-button @click=${() => {this.option = !this.option;}}>toggle and serialize</ing-button>
        </ing-fieldset>
      </form>
    </ing-form>
  `;
}
// {one: "onevalue"}
// {one: "twovalue"}
// {one: "onevalue"}
// {one: "twovalue"}
```
## Expected behavior

Changing input names dynamically should work.

## Additional context

We don't have a fix, we just saw in the code what was the most likely cause. We first noticed this in ing-web 1. The code has changed somewhat so in ing-web 2 but the issue is basically the same: in FormRegistrarMixin (lion) this.formElements is only modified in add/removeFormElement. If an element's name changes this.formElements isn't updated and the form serializes incorrectly.

jorenbroekema:
As mentioned in our extension layer, we should probably listen for changes to name attribute and unregister + register the field if it has changed, so that it's updated with the correct name.

PRs / failing tests are welcome :)

nikhilgupta2001:
can you provide screen shot of the issue please

sposh:
Not really a visual bug. We are currently looking (finally) at implementing a fix to send you a MR. Cheers

sposh:
Trying to submit MR (https://github.com/ing-bank/lion/pull/934) but getting a "license/cla Pending — Contributor License Agreement is not signed yet.". Try to sign but it doesn't update.

PR #934, opened 2020-09-14
Feature/change name

tescuderoSopra:
Issue: https://github.com/ing-bank/lion/issues/730


jorenbroekema:
Hey, I tried to rebase your commits to latest master but it wasn't really working out, and I also figured that the implementation should be different --> only in the ChoiceGroupMixin, since this delegation of name attribute is "special" only for choice groups where the fieldset acts as a field, and its fields are "sub-fields" and should have the same name.

The implementation ended up a bit simpler because of this, and I added some regression tests: https://github.com/ing-bank/lion/pull/1022

jorenbroekema:
Misread / misunderstood the issue a bit, I created a test now for formElements/serializedValue properly updating on `name` changes and used most of your code (although slightly changed) :) thanks for the PR! I think the idea with the event is good!

Turns out I fixed an unrelated bug 🙈

PR #935, opened 2020-09-14
Types: input & input-amount

jorenbroekema:


packages/form-core/index.d.ts
```diff
@@ -0,0 +1,52 @@
+export { FocusMixin } from './src/FocusMixin.js';
+export { FormatMixin } from './src/FormatMixin.js';
```
daKmoR, line 2: I thought that an index.d.ts is only needed if you reexport things from node_modules?

jorenbroekema: Nope. Always needed. I triple checked 2 weeks ago

daKmoR: was not needed after all... removed

packages/localize/types/LocalizeMixinTypes.d.ts
```diff
@@ -5,6 +5,29 @@ export interface FormatNumberPart {
   value: string;
 }
 
+// Take the DateTimeFormat and add the missing resolved options as well as optionals
+export declare interface FormatDateOptions extends Intl.DateTimeFormatOptions {
```
daKmoR, line 9: very nice 👍

daKmoR (approved):

PR #922, opened 2020-09-07
Feat/combobox

tlouisse:
TODO:
- [ ] a11y tests in all browsers and screenreaders
- [x] fix all tests
- [x] move generic listbox tests to ListboxMixin suite
- [x] run ListboxMixin suite inside combobox package
- [x] example of how to add selection-display in demos
- [ ] functional testing + fixing open ends

PR #937, opened 2020-09-16
feat(combobox): new package 'combobox'

daKmoR:
supersedes https://github.com/ing-bank/lion/pull/922 as there where too many merge conflicts


packages/combobox/docs/lion-combobox-selection-display.js
```diff
@@ -0,0 +1,192 @@
...
+        deletedEls.forEach(delEl => {
+          this.__prevSelectedEls.splice(this.__prevSelectedEls.indexOf(delEl), 1);
+        });
+        this.selectedElements = this.__prevSelectedEls;
+      }
+    }
+    this.__prevSelectedEls = this.selectedElements;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
```
daKmoR, line 130:
```suggestion
```

packages/combobox/test/lion-combobox.test.js
```diff
@@ -543,6 +543,128 @@ describe('lion-combobox', () => {
...
+         */
+        const el = /** @type {LionCombobox} */ (await fixture(html`
+          <lion-combobox name="foo" autocomplete="none">
+            <lion-option .choiceValue="${'Artichoke'}">Artichoke</lion-option>
+            <lion-option .choiceValue="${'Chard'}">Chard</lion-option>
+            <lion-option .choiceValue="${'Chicory'}">Chicory</lion-option>
+            <lion-option .choiceValue="${'Victoria Plum'}">Victoria Plum</lion-option>
+          </lion-combobox>
+        `));
+        // https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/grid-combo.html
```
tlouisse, line 576: We only implement the listbox right now. Maybe in the future we can support a grid as popup, but it would need to be a different component (based on a 'GridMixin').

I do see the behavior here indeed : https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html
In example 2, the matching option gets aria-selected="true", but aria-activedescendant is still empty. A bit weird imho, since a new `ArrowDown` key activates the sibling after selected, indicating the prev activedescendant was equal to aria-selected=true.
Also, the puts a class 'focused' on the element. Since active === focused, I'm not sure if th behavior from w3c example is exactly as intended...

tlouisse:
See: https://github.com/ing-bank/lion/pull/972

PR #940, opened 2020-09-22
Proposal: Use global cache for storing objects for singletons

rafabernad:
Currently data stored in singletons is scoped to it's module, so it cannot be accessed by coexisting compatible versions of the singleton being consumed.

By using globally registered symbols we can store the singleton data safely in a global scope, so compatible instances can make use of the same objects set safely.

This way, the "singleton" would be the objects stored in the global symbol, and compatible instances could make use of it, not having to worry about dependency trees in large projects, removing the need of artifacts like the singleton manager. 

This PR show an implementation of this pattern in LocalizeManager to open the discussion.

### DIFF
--- a/packages/icon/package.json
+++ b/packages/icon/package.json
@@ -33,7 +33,6 @@
   ],
   "dependencies": {
     "@lion/core": "0.12.0",
-    "singleton-manager": "1.1.2"
   },
   "keywords": [
     "icon",
--- a/packages/icon/package.json
+++ b/packages/icon/package.json
@@ -32,7 +32,7 @@
     "./docs/icon-resolvers.js"
   ],
   "dependencies": {
-    "@lion/core": "0.12.0",
+    "@lion/core": "0.12.0"
   },
   "keywords": [
     "icon",
--- a/packages/icon/src/IconManager.js
+++ b/packages/icon/src/IconManager.js
@@ -1,8 +1,9 @@
-export class IconManager {
-  constructor() {
-    this.__iconResolvers = new Map();
-  }
+const sym = Symbol.for('@lion/icon@0');
+if (!document[sym]) {
+  document[sym] = new Map();
+}
 
+export class IconManager {
   /**
    * Adds an icon resolver for the given namespace. An icon resolver is a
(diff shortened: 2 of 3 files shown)


daKmoR:
interesting approach 🤗

we will discuss it in the team to get more opinions on it 🤗

LarsDenBakker:
I think it's a good approach, the singleton manager does pretty much the same thing and it indeed avoids the complexity of managing a dependency.

manolakis:
I'm not sure about this approach. As [`Symbol('foo') === Symbol('foo')` is `false`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) having different instances of the same module in the `node_modules` would mean that we would have different symbols for the same definition, so we would not have singletons.

Maybe I'm missing something? 🤔

rafabernad:
> ~I'm not sure about this approach. As [`Symbol('foo') === Symbol('foo')` is `false`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) having different instances/versions of the same module would mean that we would have different symbols for the same definition, so we would not have singletons.~
> 
> ~Maybe I'm missing something? 🤔~
> 
> Yes, I was missing the `Symbol.for` 🙈
> 
> Don't you think that this could be done inside the `SingletonManager` instead of doing it in several places?

I think the main benefit of this proposal is simplifying the dependency tree, so singletons can "manage themselves" without the need of using facades like the singleton manager, so the ideal expected outcome of applying this pattern would be removing the singleton manager entirely, as the globally stored cache would guarantee the same behavior as the singleton manager. 

That said, yes, you could apply this pattern in the singleton manager, but what's the point of doing it? The Singleton Manager is supposed to be stable (or it wouldn't work as it's supposed to work due to versioning conflicts).

manolakis:
The point is to hide the complexity to the developer with a simple interface and use a single way of use singletons.

developers should use `Symbol.for` instead of `new Symbol()` and where are we going to store it? into de document? into the window? globalThis? What if we decide to change it later? how many places should change it?

IMHO having a little dependency hiding this complexity worth it, although the code is simple.

PR #948, opened 2020-09-27
feat(icon): allow custom iconManager

jcperea:
## What I did

Add a _iconManager function at LionIcon to allow components extending LionIcon to use their own IconManager, or at least, set a project specific key for the singletonManager

### DIFF
--- a/.changeset/forty-buttons-occur.md
+++ b/.changeset/forty-buttons-occur.md
@@ -1,0 +1,5 @@
+---
+'@lion/icon': minor
+---
+
+Allow setting a custom icon manager, which is important if you as a subclasser are managing your singletons.
--- a/packages/icon/src/LionIcon.js
+++ b/packages/icon/src/LionIcon.js
@@ -161,6 +161,11 @@ export class LionIcon extends LitElement {
     }
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  get _iconManager() {
+    return icons;
+  }
+
   /**
    * @param {string} prevIconId
    */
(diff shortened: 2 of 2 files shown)


daKmoR:
hmm `icons` should be singleton? 🤔
so allowing to define your own without actually "replacing" icons itself could lead to hard to debug issues 😅

can you elaborate why you would need this? 🤗

jcperea:
yes :) sorry, let me explain myself

Two scenarios:

* First. Imagine a library (i.e. library-web) with its own LibraryIcon object, extending LionIcon. If you have an scenario where you need to implement your own IconManager, you can't extend from LionIcon because this is directly coupled to https://github.com/ing-bank/lion/blob/master/packages/icon/src/icons.js.

* Second. It is ok to use the IconManager provided by lion, but you have multiple instances/versions of Lion running on your app. To avoid the registration of icons in different instances of IconManager, you can fix the actual IconManager via the singletonManager, but then this library-web should use the `@lion/icon::icons::0.5.x` key to update/customize the manager, and probably it'll make sense for this library to use its own key, for instance, `@library-web::icons::1.0.x`.

With this MR, you can allow this LibraryIcon to define its own `icons`, and do something like this:

```js
// file: './icons.js'
import { singletonManager } from 'singleton-manager';
import { icons, setIcons } from '@lion/icon';

if (singletonManager.has('@library-web::icons::1.0.x')) {
  setIcons(singletonManager.get('@library-web::icons::1.0.x'));
}

export { icons, setIcons };
```
so the LibraryIcon can do:

```js
// file: './LibraryIcon.js'
import { LionIcon } from '@lion/icon';
import { icons } from './icons.js';

export class LibraryIcon extends LionIcon {

  /** @override */
  _iconManager() {
    return icons;
  }
}
```

gerjanvangeest:
We are probably going to remove the `singletonManager` and go for this approach: https://github.com/ing-bank/lion/pull/940/files

What is your need to extend `lion-icon`, maybe we want this as well?

jorenbroekema (approved):

PR #956, opened 2020-09-28
feat(collapsible): add types for collapsible

jorenbroekema:
## What I did

1. add types for collapsible

### DIFF
--- a/packages/collapsible/demo/CustomCollapsible.js
+++ b/packages/collapsible/demo/CustomCollapsible.js
@@ -6,10 +6,9 @@ const EVENT = {
 };
 /**
  * `CustomCollapsible` is a class for custom collapsible element (`<custom-collapsible>` web component).
- *
  * @customElement custom-collapsible
- * @extends LionCollapsible
  */
+// @ts-expect-error false positive for incompatible static get properties. Lit-element merges super properties already for you.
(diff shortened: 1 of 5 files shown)


packages/collapsible/demo/CustomCollapsible.js
```diff
@@ -66,20 +64,22 @@ export class CustomCollapsible extends LionCollapsible {
...
    * @param {Object} options - element node and its options
+   * @param {HTMLElement} options.contentNode
    * @override
    */
   async _hideAnimation({ contentNode }) {
     if (this._contentHeight === '0px') {
       return;
     }
-    ['opacity', 'padding', 'max-height'].map(prop => contentNode.style.setProperty(prop, 0));
+    ['opacity', 'padding', 'max-height'].map(prop => contentNode.style.setProperty(prop, `${0}`));
```
daKmoR, line 74:
```suggestion
    ['opacity', 'padding', 'max-height'].map(prop => contentNode.style.setProperty(prop, '0'));
```

should be good enough right 😅

jorenbroekema: Hehehe yeah, my brain was on autopilot for these type issues and just put `${}` without thinking much about it haha

PR #959, opened 2020-09-28
chore(pagination): extract nav list logic into protected method

hardikpthv:
## What I did

1. Extract the nav list template logic into a protected method so that the derived component can easily reuse it.

### DIFF
--- a/.changeset/slow-scissors-fail.md
+++ b/.changeset/slow-scissors-fail.md
@@ -1,0 +1,5 @@
+---
+'@lion/pagination': patch
+---
+
+fix(pagination): extract nav list template to reuse in derived component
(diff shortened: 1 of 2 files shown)


daKmoR:
if you want this to be released then you will need to add a "changeset".

e.g. do `yarn changeset` then select pagination and to a fix release + add your message for the changelog.
this will create a md file you should commit (probably best to do `git commit --amend` to add it to this single commit)

PS: I think a commit message with `fix(pagination)` would be better

hardikpthv:
Thanks, @daKmoR for the tip. I followed the steps you suggested. Do let me know if any changes are required.

daKmoR:
oh can you move this to a "local" branch? e.g. not in your fork 😅

hardikpthv:
There you go https://github.com/ing-bank/lion/pull/961

PR #961, opened 2020-09-28
fix(pagination): extract nav list template to reuse in derived component

hardikpthv:
## What I did

1. Extract the nav list template logic into a protected method so that the derived component can easily reuse it.


daKmoR (approved):

daKmoR:
thaaankks 🤗

PR #963, opened 2020-09-28
feat: add types for input components

jorenbroekema:
## What I did

1. added types for input components: 
- date
- email
- iban 
- range
- stepper


packages/input-date/src/LionInputDate.js
```diff
@@ -24,15 +28,19 @@ export class LionInputDate extends LocalizeMixin(LionInput) {
 
   constructor() {
     super();
-    this.parser = (value, options) => (value === '' ? undefined : parseDate(value, options));
+    /**
+     * @param {string} value
+     */
+    this.parser = value => (value === '' ? undefined : parseDate(value));
```
daKmoR, line 34: we no longer parse the options?

packages/input-range/src/LionInputRange.js
```diff
@@ -3,18 +3,31 @@ import { css, html, unsafeCSS } from '@lion/core';
...
+// @ts-expect-error https://github.com/microsoft/TypeScript/issues/40110 + false positive for incompatible static get properties. Lit-element merges super properties already for you.
 export class LionInputRange extends LocalizeMixin(LionInput) {
   static get properties() {
     return {
-      min: Number,
-      max: Number,
-      unit: String,
+      min: {
+        type: Number,
+        reflect: true,
```
daKmoR, line 21: why reflect?

packages/input-stepper/src/LionInputStepper.js
```diff
@@ -22,30 +22,53 @@ export class LionInputStepper extends LionInput {
...
     return {
-      min: Number,
-      max: Number,
-      step: Number,
-      modelValue: Number,
-      __disableIncrementor: Boolean,
-      __disableDecrementor: Boolean,
+      min: {
+        type: Number,
+        reflect: true,
```
daKmoR, line 27: why reflect?

packages/input-range/src/LionInputRange.js
jorenbroekema: If a user changes the `min` attribute, it should reflect to the property and vice versa.

packages/input-date/src/LionInputDate.js
jorenbroekema: parseDate does not accept any 🙈  it uses localize.locale. There's no other parse options like there is with numbers.

daKmoR: hmm ok, but `parseDate` should accept options like a different locale... 😅

but yeah that is for another pull request

daKmoR (approved):

PR #710, opened 2020-05-14
fix(select-rich): enable reset

gerjanvangeest:

PR #964, opened 2020-09-29
fix(listbox): add reset function

gerjanvangeest:
## What I did

1. Add reset function to listbox
  - improved version of: https://github.com/ing-bank/lion/pull/710

### DIFF
--- a/.changeset/slimy-windows-sing.md
+++ b/.changeset/slimy-windows-sing.md
@@ -1,0 +1,5 @@
+---
+'@lion/listbox': minor
+---
+
+Add reset function to listbox and all extentions
--- a/packages/form-integrations/docs/15-features-overview.md
+++ b/packages/form-integrations/docs/15-features-overview.md
@@ -9,6 +9,7 @@ For usage and installation please see the appropriate packages.
 import { html } from 'lit-html';
 import '@lion/checkbox-group/lion-checkbox-group.js';
 import '@lion/checkbox-group/lion-checkbox.js';
+import '@lion/combobox/lion-combobox.js';
 import '@lion/fieldset/lion-fieldset.js';
 import '@lion/form/lion-form.js';
 import '@lion/input-amount/lion-input-amount.js';
@@ -19,11 +20,12 @@ import '@lion/input-iban/lion-input-iban.js';
 import '@lion/input-range/lion-input-range.js';
 import '@lion/input-stepper/lion-input-stepper.js';
 import '@lion/input/lion-input.js';
+import '@lion/listbox/lion-listbox.js';
+import '@lion/listbox/lion-option.js';
+import '@lion/listbox/lion-options.js';
 import '@lion/radio-group/lion-radio-group.js';
 import '@lion/radio-group/lion-radio.js';
 import '@lion/select/lion-select.js';
-import '@lion/listbox/lion-option.js';
-import '@lion/listbox/lion-options.js';
 import '@lion/select-rich/lion-select-rich.js';
 import '@lion/textarea/lion-textarea.js';
 import { MinLength, Required } from '@lion/form-core';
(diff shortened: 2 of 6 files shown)


packages/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -383,6 +383,52 @@ export function runListboxMixinSuite(customConfig = {}) {
...
+        `);
+        el.setCheckedIndex(1);
+        expect(el.modelValue).to.equal(20);
+        el.reset();
+        expect(el.modelValue).to.equal('');
+      });
+
+      it('should reset selected value to initial values when multiple-choice', async () => {
+        const el = await fixture(html`
+          <${tag} multiple-choice has-no-default-selected>
```
gerjanvangeest, line 418: Not fully happy to have to add `has-no-default-selected`, but that is not part of this pull request

packages/listbox/src/ListboxMixin.js
```diff
@@ -252,6 +252,7 @@ const ListboxMixinImplementation = superclass =>
       super.firstUpdated(changedProperties);
 
       this.__moveOptionsToListboxNode();
+      this._initialCheckedIndex = this.checkedIndex;
```
tlouisse, line 255: Why not save modelValue instead (we do it in other places as wel, since this is SSOT and all other values are derived from it)?

.changeset/slimy-windows-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/listbox': patch
```
tlouisse, line 2: maybe a minor even

packages/listbox/src/ListboxMixin.js
gerjanvangeest: Since we loop through all child form elements to reset the old checked state, which goes the easiest with `index`

gerjanvangeest: But using the `modelValue` makes all other code way easier. So i'll update it

packages/listbox/src/ListboxMixin.js
```diff
@@ -252,6 +252,8 @@ const ListboxMixinImplementation = superclass =>
       super.firstUpdated(changedProperties);
 
       this.__moveOptionsToListboxNode();
+      /** @type {any[]} */
+      this._initialModelValue = this.modelValue;
```
daKmoR, line 256: I am wondering if we should better use
```js
this.registrationComplete.then(() => {
  this._initialModelValue = this.modelValue;
});
```

to be more "safe" if some child maybe takes slightly longer 😅

packages/listbox/src/ListboxMixin.js
```diff
@@ -277,6 +281,20 @@ const ListboxMixinImplementation = superclass =>
...
 
+    resetInteractionState() {
+      super.resetInteractionState();
+      this.submitted = false;
+    }
+
+    reset() {
+      this.formElements.forEach(formElement => {
+        // eslint-disable-next-line no-param-reassign
+        formElement.checked = formElement.choiceValue === this._initialModelValue;
```
daKmoR, line 292: would that not put a `checked` property an every form element? even normal inputs? 🤔

gerjanvangeest: Normal inputs can't be a child of the listbox (ChoiceGroupMixin will throw an error).
Checkbox and radio can be placed inside the listbox, but this breaks the whole listbox, but i don't think we have to guard this, and especially not within the reset function.

packages/form-integrations/docs/15-features-overview.md
```diff
@@ -94,12 +96,23 @@ export const main = () => {
           <lion-radio .choiceValue=${'brontosaurus'} label="brontosaurus"></lion-radio>
           <lion-radio .choiceValue=${'diplodocus'} label="diplodocus"></lion-radio>
         </lion-radio-group>
+        <lion-listbox name="favoriteFruit" label="Favorite fruit">
```
tlouisse, line 99: 👍

packages/listbox/src/ListboxMixin.js
```diff
@@ -381,6 +385,21 @@ const ListboxMixinImplementation = superclass =>
       /* eslint-enable no-param-reassign */
     }
 
+    resetInteractionState() {
+      super.resetInteractionState();
+      this.submitted = false;
+    }
+
+    reset() {
+      this.formElements.forEach(formElement => {
```
tlouisse, line 394: It should be possible to do: `this.modelValue = this._initialModelValue` (since it's SSOT and will automatically sync checked state under the hood)

gerjanvangeest: Yes indeed, although it required a small fix under that hood :)

tlouisse (approved):

PR #966, opened 2020-09-29
fix(overlays): add transition hooks to overlays

daKmoR:
## What I did

1. Add `async transitionShow` and `async transitionHide` to OverlayController to enable basic support for transitions/animations

### DIFF
--- a/.changeset/little-jeans-yell.md
+++ b/.changeset/little-jeans-yell.md
@@ -1,0 +1,5 @@
+---
+'@lion/overlays': patch
+---
+
+Adds `async transitionShow` and `async transitionHide` to OverlayController to enable basic support for transitions/animations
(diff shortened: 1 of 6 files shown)


.changeset/little-jeans-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/overlays': patch
```
tlouisse, line 2: Patch instead of minor
 because those methods already existed?

daKmoR: because semver < 1.x.x and minor would therefore be a breaking change... 

but it doesn't really matter as I think there is already a minor change for overlays lined up for the next release... jup will be 0.20.0 https://github.com/ing-bank/lion/pull/936

so can do minor or patch here... both is fine with me

gerjanvangeest:
Can we also have a demo how to make use of the transition hooks?

gerjanvangeest (approved):

daKmoR:
I think we should add demos in an "animations system" story... if that doesn't happen too soon then we can add it specifically for this case here

Issue #960, opened 2020-09-28
[tabs] Tabs nested inside a tab have indexing issues


weriji:
## Expected behavior

When you Nest tabs inside a tab, I expect them to behave the same way tabs do. Currently I am experiencing two issues with that.

First, when changing tabs inside a tab, it closes the main tab. Example, I have 3 main tabs(Work, Info, Contact), Inside the Contact tab I have 3 nested tabs (Corporate, Dev, Manager). When changing from Dev to Manager this close the Contact tab which then needs to be re-opened.

Second, when you go to the Manager tab and then back to Dev(after re-opening the Contact tab from first issue above) both tabs are opened simultaneously(Manager and Dev). The expected behaviour would be to close Manager tab and open the Dev one. Instead they're both opened and selected.

## Actual Behavior

Reproduction Link : https://webcomponents.dev/preview/DCcnvo5rqSeaxk5Bwuz8

## Additional context

@lion/lion-tabs and the issue exist in the version that is used for the reproduction.

daKmoR:
@weriji thank you for the issue and especially for the reproduction 🤗

it's fixed on master and we will release it today or tomorrow 👍

PR #973, opened 2020-09-30
fix(tabs): allow nested tabs

gerjanvangeest:
## What I did

fixed: https://github.com/ing-bank/lion/issues/960

By only looking for tabs and panels as direct children.


daKmoR (approved):

daKmoR:
uh very nice 💪

PR #975, opened 2020-10-01
feat(combobox): improvements api, demos and autocompletion ux

tlouisse:
- renamed `filterOptionCondition ` (similarity to `match-mode`, since this is basically an override)
- demos for `matchCondition`
- inline autocompletion edge cases solved (that would be inconsistent ux otherwise)
- demos took a long time render: introduced a lazyRender directive that only adds (expensive) lionOptions after first meaningful paint has happened
- made clearer from the code that selectionDisplay component is for demo purposes only at this moment

### DIFF
--- a/.changeset/heavy-ghosts-sell.md
+++ b/.changeset/heavy-ghosts-sell.md
@@ -1,0 +1,11 @@
+---
+'@lion/combobox': patch
+---
+
+Combobox api, demo and ux improvements
+
+- renamed `filterOptionCondition ` (similarity to `match-mode`, since this is basically an override)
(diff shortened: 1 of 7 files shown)


.changeset/new-tools-march.md
```diff
@@ -0,0 +1,25 @@
+---
```
jorenbroekema, line 1: I think this file is as a result from not having rebased on latest master? :P

tlouisse: I think as well. Was just removing it 👍

jorenbroekema (approved):

PR #978, opened 2020-10-01
fix: assure proper typings _inputNode across lion fields

jorenbroekema:
## What I did

1. Refactor of some fields to ensure that \_inputNode has the right type. It starts as HTMLElement for LionField, and all HTMLInputElement, HTMLSelectElement and HTMLTextAreaElement logic, are moved to the right places.


packages/input/src/LionInput.js
```diff
@@ -49,6 +50,42 @@ export class LionInput extends LionField {
     };
   }
 
+  get _inputNode() {
+    return /** @type {HTMLInputElement} */ (super._inputNode); // casts type
+  }
+
+  /** @type {number} */
+  get selectionStart() {
```
tlouisse, line 58: api should also be applied to LionTextarea

packages/input-amount/src/LionInputAmount.js
```diff
@@ -53,14 +53,11 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
...
   static get styles() {
-    return [
-      super.styles,
-      css`
-        .input-group__container > .input-group__input ::slotted(.form-control) {
-          text-align: right;
-        }
-      `,
-    ];
+    return css`
```
tlouisse, line 56: Forgot why the parent does not have any styles. Everything inheriting from FormControlMixin should add `super.styles` right?

packages/input-stepper/src/LionInputStepper.js
```diff
@@ -10,14 +10,11 @@ import { IsNumber, MinNumber, MaxNumber } from '@lion/form-core';
...
   static get styles() {
-    return [
-      super.styles,
-      css`
-        .input-group__container > .input-group__input ::slotted(.form-control) {
-          text-align: center;
-        }
-      `,
-    ];
+    return css`
```
tlouisse, line 13: Forgot why the parent does not have any styles. Everything inheriting from FormControlMixin should add `super.styles` right?

packages/form-core/test-suites/form-group/FormGroupMixin-input.suite.js
```diff
@@ -0,0 +1,197 @@
...
+
+  describe('FormGroupMixin with LionInput', () => {
+    it('serializes undefined values as "" (nb radios/checkboxes are always serialized)', async () => {
+      const fieldset = /**  @type {FormGroup} */ (await fixture(html`
+        <${tag}>
+          <${childTag} name="custom[]"></${childTag}>
+          <${childTag} name="custom[]"></${childTag}>
+        </${tag}>
+      `));
+      console.log(fieldset.formElements['custom[]'][1]);
```
tlouisse, line 49: 🤡

tlouisse: Seems to pass linting staged on precommit recenlty ☹️

packages/input-stepper/src/LionInputStepper.js
tlouisse: We still don't have it. It misses styles right now. I would rather go for:

```
  static get styles() {
    return [
      // @ts-expect-error
      super.styles,
      css`
        .input-group__container > .input-group__input ::slotted(.form-control) {
          text-align: center;
        }
      `,
    ];
  }
```

tlouisse (approved):

PR #986, opened 2020-10-04
chore(combobox): advanced extension demos github + whatsapp

tlouisse:
## What I did

Added advanced extension demos github + whatsapp:

It will tell a Subclasser:
- how to extend (advanced use cases)
- how to animate options (whatsapp example)
- how to rearrange templates (github example)

combobox_demos


packages/combobox/src/LionCombobox.js
```diff
@@ -493,17 +496,19 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
-        return;
-      }
-
-      // [4]. Cleanup previous visibility and a11y states
+      // [3]. Cleanup previous visibility and a11y states
       option.setAttribute('aria-hidden', 'true');
       option.removeAttribute('aria-posinset');
       option.removeAttribute('aria-setsize');
 
+      // [4]. If ._inputNode is empty, no filtering will be applied
```
palash2601, line 504: Hi @tlouisse ,

I tried this behaviour by adding below test in code from master branch, i.e without this change and the test passes.

```
it('keeps showing list box when text in input node is cleared after typing something', async () => {
      const el = /** @type {LionCombobox} */ (await fixture(html`
        <lion-combobox name="foo" autocomplete="list">
          <lion-option .choiceValue="${'Artichoke'}">Artichoke</lion-option>
          <lion-option .choiceValue="${'Chard'}">Chard</lion-option>
          <lion-option .choiceValue="${'Chicory'}">Chicory</lion-option>
          <lion-option .choiceValue="${'Victoria Plum'}">Victoria Plum</lion-option>
        </lion-combobox>
      `));

        // mimicUserTyping(el, 'c');
        // await el.updateComplete;
        // expect(el.opened).to.equal(true);
        mimicUserTyping(el, '');
        await el.updateComplete;
        expect(el.opened).to.equal(true);
    });
```

I also tested this behaviour in storybook and Only with `autocomplete` as `none` it works fine (overlay remains open) but for other demos list is not displayed when inputnode text is cleared (overlay closes)

tlouisse: Thanks! 
Yeah, it would have state opened on keyup, but all options would have `aria-hidden=true`
Maybe we should finetune it a bit more later (if the empty box takes up space and you don't want to see it. On the other hand you maybe want to show a message there like 'loading' or smth else, so it needs some more thinking)
For now I will make the tests work by checking for aria-hidden 👍

tlouisse: I will rename this behavior to `showAllOptionsOnEmpty` (by default (wai-aria/datalist) it hides)

tlouisse: will be handled in separate pr...

daKmoR (approved):

PR #987, opened 2020-10-05
chore(combobox): demo for server side autocompletion

tlouisse:
## What I did

Demo for handling server side autocomplete

TODO:
- [ ] after first server call, autocomplete is not called yet in _onListboxContentChanged
- [ ] tests

### DIFF
--- a/packages/combobox/docs/demo-server-side.js
+++ b/packages/combobox/docs/demo-server-side.js
@@ -1,0 +1,48 @@
+import { LitElement, html } from '@lion/core';
+
+function fetchMyData() {
+  return new Promise(resolve => {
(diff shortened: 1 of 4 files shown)


packages/combobox/docs/demo-server-side.js
```diff
@@ -0,0 +1,48 @@
+import { LitElement, html } from '@lion/core';
+
+function fetchMyData() {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve('abcdefghijk'.split(''));
```
tlouisse, line 6: make this more useful array of data....

PR #990, opened 2020-10-05
chore(combobox): server-side demo fix

palash2601:
## What I did

1.


packages/listbox/src/ListboxMixin.js
```diff
@@ -732,6 +732,7 @@ const ListboxMixinImplementation = superclass =>
           slot.assignedNodes().forEach(node => {
             this._listboxNode.appendChild(node);
           });
+          this._onListboxContentChanged();
```
tlouisse, line 735: So it did work and I just forgot to add it here :)

tlouisse (approved):

PR #992, opened 2020-10-05
chore(listbox): types

tlouisse:
Typed lion-listbox + select-rich package


jorenbroekema (changes requested):
Nice. Can we change the ts-ignores to ts-expect-error though? That way we are more explicit about the fact that we expect a TS error, and it's not a fluke.

Not a blocker but in the future it may be easier to typecast fixtures like this:
```js
import { expect, fixture as _fixture, html } from '@open-wc/testing';
/**
 * @typedef {import('../src/LionListbox').LionListbox} LionListbox
 * @typedef {import('lit-html').TemplateResult} TemplateResult
 */

const fixture = /** @type {(arg: TemplateResult) => Promise<LionListbox>} */ (_fixture);
```
That way all your fixtures are typecasted automatically so it saves a lot of manual labour :)

tlouisse:
That would save some find and replaces indeed 👍 

The choice for ts-ignore vs ts-expect-error is always a conscious one:
- expect-error: there is a real error and when that is solved you want to get notified
- ignore: you deliberately op-out of the rule. In case of a protected option that you want to access in a test for instance.

I agree we should use expect-error as much as possible in the first category

jorenbroekema (approved):

packages/listbox/types/ListboxMixinTypes.d.ts
```diff
@@ -41,10 +41,17 @@ export declare class ListboxHost {
...
   public setCheckedIndex(index: number): void;
 
+  /** Reset interaction states and modelValue */
+  public reset(): void;
+
   protected _scrollTargetNode: LionOptions;
 
   protected _listboxNode: LionOptions;
 
+  protected _listboxReceivesNoFocus: Boolean;
```
jorenbroekema, line 51: not blocking but we tend to use lowercase for primitives. For typescript I think there's actually a significant difference, for JSDocs it doesn't matter

tlouisse: makes sense. Will adapt in the other mr 👍

PR #1008, opened 2020-10-07
fix: fix backdrop teardown/init bugs and align APIs

jorenbroekema:
## What I did

1. Fix a bug where teardown did not properly teardown the right backdropNode for animation = true
2. Fix a bug where content node was not torn down properly when updateConfig is called every time we show (when for example the config is updated on before-show based on viewport)
3. Simplify the code for backdrop feature
4. Added demos for backdrop
5. Align API for backdrop (can pass `hasBackdrop: true` which spawns a default backdropNode, can pass `backdropNode` or declaratively `<div slot="backdrop"></div>` to pass your own, and can pass `hasBackdropAnimation: true` which turns on animation, assumes you have CSS animations applied to your backdropNode)
6. Small fix for overlayContentNode in OverlayMixin, to support imperatively passed contentNode

The two bugs were quite difficult to create an automated test for, visual regression would really help for these two 😅

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -251,7 +251,7 @@ export class OverlayController extends EventTargetShim {
    * @type {boolean}
    */
   get hasBackdrop() {
-    return /** @type {boolean} */ (this.config?.hasBackdrop);
+    return /** @type {boolean} */ (!!this.backdropNode || this.config?.hasBackdrop);
   }
 
   /**
@@ -618,7 +618,7 @@ export class OverlayController extends EventTargetShim {
         }
       }
     } else if (phase === 'teardown') {
-      this.__restorOriginalAttrs();
+      this.__restoreOriginalAttrs();
     }
   }
 
@@ -634,7 +634,7 @@ export class OverlayController extends EventTargetShim {
     this.__originalAttrs.set(node, attrMap);
   }
 
-  __restorOriginalAttrs() {
+  __restoreOriginalAttrs() {
     for (const [node, attrMap] of this.__originalAttrs) {
       Object.entries(attrMap).forEach(([attrName, value]) => {
         if (value !== null) {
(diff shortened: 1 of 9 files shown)


packages/overlays/src/OverlayController.js
```diff
@@ -891,97 +894,90 @@ export class OverlayController extends EventTargetShim {
...
-          this.backdropNode,
-          this.contentWrapperNode,
-        );
+      case 'init': {
+        if (!this.backdropNode) {
+          this.__backdropNode = document.createElement('div');
+          /** @type {HTMLElement} */
+          (this.backdropNode).classList.add(`${this.placementMode}-overlays__backdrop`);
+        }
+        this.backdropNode.slot = 'backdrop';
```
tlouisse, line 909: Why should OverlayController know OverlayMixin.
It's cleaner if the connection is one way (OverlayMixins knows controller, not the other way around)

packages/overlays/src/OverlayController.js
```diff
@@ -102,6 +102,7 @@ export class OverlayController extends EventTargetShim {
       contentWrapperNode: config.contentWrapperNode,
       invokerNode: config.invokerNode,
       backdropNode: config.backdropNode,
+      hasBackdropAnimation: false,
```
tlouisse, line 105: Could we do animation in transitionHide/transitonShow, just like with the contentNode? I think it's a Deisgn System implementation detail.
And then we are consistent: all animation happen in the Subclasser layer. We only provide proper timing alignments by exposing promise hooks

packages/overlays/src/OverlayController.js
```diff
@@ -891,97 +894,90 @@ export class OverlayController extends EventTargetShim {
...
+        insertionAnchor.insertBefore(this.backdropNode, insertionBefore);
         break;
+      }
       case 'show':
-        this.backdropNode.classList.add('global-overlays__backdrop--visible');
-        if (animation === true) {
-          this.backdropNode.classList.add('global-overlays__backdrop--fade-in');
+        this.backdropNode.classList.add(`${this.placementMode}-overlays__backdrop--visible`);
+        if (this.config?.hasBackdropAnimation === true) {
+          this.backdropNode.classList.add(`${this.placementMode}-overlays__backdrop--animation-in`);
```
tlouisse, line 923: maybe we can do this as a default implimentation ofn transitionShow? So the user can override it?
Then this would be the only place where we have a default animation (kind of fine, since native <dialog> has it as well I think)
And all other apis would be the same:
Animations can be done via Promise functions (for maximum flexibility (you can use greensock whatever) and are not toggled via boolean flags)

packages/overlays/src/OverlayController.js
```diff
@@ -418,7 +421,7 @@ export class OverlayController extends EventTargetShim {
    */
   updateConfig(cfgToAdd) {
     // Teardown all previous configs
-    this._handleFeatures({ phase: 'teardown' });
+    this.teardown();
```
tlouisse, line 424: 👍

packages/overlays/src/OverlayController.js
jorenbroekema: I would actually like to remove this. It's leftover from the first commit, but I don't think this is needed anymore except when the user sets hasBackdrop: true and does not pass their own backdrop node (declaratively or imperatively), right?

jorenbroekema:
```js
if (!this.backdropNode) {
  this.__backdropNode = document.createElement('div');
  /** @type {HTMLElement} */
  (this.backdropNode).slot = 'backdrop';
  /** @type {HTMLElement} */
  (this.backdropNode).classList.add(`${this.placementMode}-overlays__backdrop`);
}
``` 
sound good?

tlouisse (approved):

PR #1017, opened 2020-10-09
fix(localization): was not formatting minus values

demeteren:
The old code was taking only the index 0 of formatted parts. This cause a bug in minus amount values. This foreach function changes the specific 'currency' value.

### DIFF
--- a/packages/localize/src/number/forceTryCurrencyCode.js
+++ b/packages/localize/src/number/forceTryCurrencyCode.js
@@ -4,13 +4,17 @@
  * @param {import('../../types/LocalizeMixinTypes').FormatNumberOptions} [options]
  * @returns {FormatNumberPart[]}
  */
+
 export function forceTryCurrencyCode(formattedParts, { currency, currencyDisplay } = {}) {
   const result = formattedParts;
-  // Change the currency code from TRY to TL, for Turkey
   if (currency === 'TRY' && currencyDisplay === 'code') {
-    if (result[0].value === 'TRY') {
-      result[0].value = 'TL';
-    }
+    result.map(part => {
+      const newPart = part;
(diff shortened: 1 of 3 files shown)


packages/localize/src/number/forceTryCurrencyCode.js
```diff
@@ -4,13 +4,16 @@
...
   const result = formattedParts;
-  // Change the currency code from TRY to TL, for Turkey
+  // Change the currency code from TRY to TL
   if (currency === 'TRY' && currencyDisplay === 'code') {
-    if (result[0].value === 'TRY') {
-      result[0].value = 'TL';
-    }
+    result.forEach(part => {
+      part.type === 'currency' ? (part.value = 'TL') : '';
+    });
```
daKmoR, line 16: this code seems a little hard to read (to me) 😅

could you add a test for it?

and afterward, potentially change it to something like this?

```suggestion
    return formattedParts.map(part => {
      const newPart = part;
      if (part.type === 'currency') {
         newPart.value = 'TL';
      }
      return newPart;
    });
```

this would also mean that both eslint disabled would not be needed 💪

demeteren: Thank you :) But with map, objects have immutability therefor I used forEach function to modify array parts.. And also I didn't want to use object assign and create a new array of parts because it would slow things down.  Maybe if I changed arrow function to default function and used "if" statement, would it help?

demeteren: Thanks for your review. I made it easier to read and added tests for both positive and negative numbers.

jorenbroekema (approved):

jorenbroekema:
Thanks for the PR @demeteren  👍

PR #1019, opened 2020-10-11
refactor(button): several improvements

jorgecasar:
## What I did

- remove click-area --> move styles to host::before
- reduce css so that extending styles makes sense. Merge .btn with host.
- reduce the template and remove the if else construction inside the template.
- hide focus styles if they're not needed, for example, when an element receives focus via the mouse.
- improve `__clickDelegationHandler`. Use current slotted native button instead of create new one.
- fix vertical alignment when 2 buttons are inline and one has icon. Example included.
image
- add min-with to ensure target size for mobile, not only height
image
- add test to check event.target in all situations (inside and outside form)

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -8,16 +8,6 @@ import {
 } from '@lion/core';
 import '@lion/core/src/differentKeyEventNamesShimIE.js';
 
-// TODO: several improvements:
-// [1] remove click-area
-// [2] remove the native _button slot. We can detect and submit parent form without the slot.
-// [3] reduce css so that extending styles makes sense. Merge .btn with host
-// [4] reduce the template and remove the if else construction inside the template (an extra
-// div by default to support IE is fine) => <div id="${this._buttonId}"><slot></slot></div>
-// should be all needed
-// [5] do we need the before and after templates? Could be added by subclasser
-// [6] extract all functionality (except for form submission) into LionButtonMixin
-
 const isKeyboardClickEvent = (/** @type {KeyboardEvent} */ e) => e.key === ' ' || e.key === 'Enter';
 const isSpaceKeyboardClickEvent = (/** @type {KeyboardEvent} */ e) => e.key === ' ';
 
@@ -41,15 +31,10 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
 
   render() {
     return html`
-      <div class="btn">
-        <div class="click-area"></div>
-        ${this._beforeTemplate()}
-        ${browserDetection.isIE11
-          ? html`<div id="${this._buttonId}"><slot></slot></div>`
-          : html`<slot></slot>`}
-        ${this._afterTemplate()}
-        <slot name="_button"></slot>
-      </div>
+      ${this._beforeTemplate()}
+      <div class="button-content" id="${this._buttonId}"><slot></slot></div>
+      ${this._afterTemplate()}
+      <slot name="_button"></slot>
     `;
   }
 
@@ -67,24 +52,39 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
     return [
       css`
         :host {
+          position: relative;
           display: inline-block;
-          min-height: 40px; /* src = https://www.smashingmagazine.com/2012/02/finger-friendly-design-ideal-mobile-touchscreen-target-sizes/ */
-          outline: 0;
-          background-color: transparent;
           box-sizing: border-box;
+          vertical-align: middle;
+          line-height: 24px;
+          background: #eee; /* minimal styling to make it recognizable as btn */
+          padding: 8px; /* padding to fix with min-height */
+          outline: none; /* focus style handled below */
+          cursor: default; /* /* we should always see the default arrow, never a caret */
         }
 
-        .btn {
-          min-height: 24px;
+        :host::before {
+          content: '';
+
+          /* center vertically and horizontally */
+          position: absolute;
+          top: 50%;
+          left: 50%;
+          transform: translate(-50%, -50%);
+
+          /* touch area (comes into play when button height goes below this one) */
+          /* src = https://www.smashingmagazine.com/2012/02/finger-friendly-design-ideal-mobile-touchscreen-target-sizes/ */
+          min-height: 40px;
+          min-width: 40px;
+        }
+
+        .button-content {
           display: flex;
           align-items: center;
-          position: relative;
-          background: #eee; /* minimal styling to make it recognizable as btn */
-          padding: 8px; /* vertical padding to fix with host min-height */
-          outline: none; /* focus style handled below, else it follows boundaries of click-area */
+          justify-content: center;
         }
 
-        :host .btn ::slotted(button) {
+        :host ::slotted(button) {
           position: absolute;
           top: 0;
           left: 0;
@@ -98,41 +98,36 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
           border: 0; /* reset default agent styles */
         }
 
-        .click-area {
-          position: absolute;
-          top: 0;
-          right: 0;
-          bottom: 0;
-          left: 0;
-          margin: 0;
-          padding: 0;
-        }
-
-        :host(:focus:not([disabled])) .btn {
+        /* Show focus styles on keyboard focus. */
+        :host(:focus:not([disabled])),
+        :host(:focus-visible) {
           /* if you extend, please overwrite */
           outline: 2px solid #bde4ff;
         }
 
-        :host(:hover) .btn {
+        /* Hide focus styles if they're not needed, for example,
+        when an element receives focus via the mouse. */
+        :host(:focus:not(:focus-visible)) {
+          outline: 0;
+        }
+
+        :host(:hover) {
           /* if you extend, please overwrite */
           background: #f4f6f7;
         }
 
-        :host(:active) .btn, /* keep native :active to render quickly where possible */
-        :host([active]) .btn /* use custom [active] to fix IE11 */ {
+        :host(:active), /* keep native :active to render quickly where possible */
+        :host([active]) /* use custom [active] to fix IE11 */ {
           /* if you extend, please overwrite */
           background: gray;
-        }
-
-        :host([disabled]) {
-          pointer-events: none;
         }
 
         :host([hidden]) {
           display: none;
         }
 
-        :host([disabled]) .btn {
+        :host([disabled]) {
+          pointer-events: none;
           /* if you extend, please overwrite */
           background: lightgray;
           color: #adadad;
@@ -213,14 +208,9 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    * @param {Event} e
    */
   __clickDelegationHandler(e) {
-    if ((this.type === 'submit' || this.type === 'reset') && e.target === this) {
-      if (this._form) {
-        const nativeButton = document.createElement('button');
-        nativeButton.type = this.type;
-        this._form.appendChild(nativeButton);
-        nativeButton.click();
-        this._form.removeChild(nativeButton);
-      }
+    if ((this.type === 'submit' || this.type === 'reset') && e.target === this && this._form) {
+      e.stopImmediatePropagation();
+      this._nativeButtonNode.click();
     }
   }
 
@@ -240,6 +230,7 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
     this.removeEventListener('mousedown', this.__mousedownHandler);
     this.removeEventListener('keydown', this.__keydownHandler);
     this.removeEventListener('keyup', this.__keyupHandler);
+    this.removeEventListener('click', this.__clickDelegationHandler);
   }
 
   __mousedownHandler() {
--- a/packages/button/test/lion-button.test.js
+++ b/packages/button/test/lion-button.test.js
@@ -1,5 +1,5 @@
 import { browserDetection } from '@lion/core';
-import { aTimeout, expect, fixture, html, oneEvent } from '@open-wc/testing';
+import { aTimeout, expect, fixture, html, oneEvent, unsafeStatic } from '@open-wc/testing';
 import sinon from 'sinon';
 import '@lion/core/src/differentKeyEventNamesShimIE.js';
 import '../lion-button.js';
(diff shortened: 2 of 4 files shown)


packages/button/src/LionButton.js
```diff
@@ -213,14 +203,9 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    * @param {Event} e
    */
   __clickDelegationHandler(e) {
-    if ((this.type === 'submit' || this.type === 'reset') && e.target === this) {
```
jorenbroekema, line 216: Now I am really questioning why we had this logic here in the first place. I doubt it was without reason. Could you check if implicit form submission works in IE11 with this? We currently don't run the tests automatically in IE11 so we have to kinda check it manually.

packages/button/test/lion-button.test.js
```diff
@@ -206,7 +206,7 @@ describe('lion-button', () => {
       const wrapperId = el.getAttribute('aria-labelledby');
       expect(/** @type {ShadowRoot} */ (el.shadowRoot).querySelector(`#${wrapperId}`)).to.exist;
       expect(/** @type {ShadowRoot} */ (el.shadowRoot).querySelector(`#${wrapperId}`)).dom.to.equal(
-        `<div id="${wrapperId}"><slot></slot></div>`,
+        `<div class="slot" id="${wrapperId}"><slot></slot></div>`,
```
jorenbroekema, line 209: Maybe we should rename the class it to `button-content` or `btn-content` to be a bit more descriptive.

packages/button/src/LionButton.js
jorgecasar: It was because of a double dispatch. But with stopImmediatePropagation we avoit to dispatch host one and only dispatch the native one. If you are listening the event in the lion-button it will be the currentTarget and the target will be the native button.

jorgecasar: It works!

[image](https://user-images.githubusercontent.com/948953/95856645-a0beab00-0d5a-11eb-80c8-799bfda706da.png)

jorgecasar:
There are only a bug left.

image

IE11 doesn't have any issue. But with inline-grid it breaks.

packages/button/src/LionButton.js
```diff
@@ -98,41 +94,36 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-          bottom: 0;
-          left: 0;
-          margin: 0;
-          padding: 0;
-        }
-
-        :host(:focus:not([disabled])) .btn {
+        /* Show focus styles on keyboard focus. */
+        :host(:focus:not([disabled])),
+        :host(:focus-visible) {
```
jorenbroekema, line 99: This is a progressive enhancement for browsers that support it, but does not break anything for browsers that don't support this selector, correct?

jorgecasar: Correct. It works as before with `:host(:focus:not([disabled]))` and it will hide the focus for `:host(:focus:not(:focus-visible))` if it is supported.

Chrome 86 support it. Previous versions works fine.

packages/button/src/LionButton.js
```diff
@@ -213,14 +204,9 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
-      if (this._form) {
-        const nativeButton = document.createElement('button');
-        nativeButton.type = this.type;
-        this._form.appendChild(nativeButton);
-        nativeButton.click();
-        this._form.removeChild(nativeButton);
-      }
+    if (e.target === this) {
+      e.stopImmediatePropagation();
+      this._nativeButtonNode.click();
```
tlouisse, line 209: The reason why it's not fired from here was the following scenario:

```html
<div @click="${delegatedClickHandler}">
  <lion-button></lion-button>
</div>
```

```js
function delegatedClickHandler(ev) {
  console.log(ev.target); // lion-button, button (2 clicks). 
}

lionButton.click(); // will trigger two clicks

```

In delegatedClickHandler we only want `lion-button` to be found.

Therefore the code we had before should be kept.

TODO: add a unit test for this use case :)

packages/button/src/LionButton.js
```diff
@@ -67,24 +52,35 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
+          vertical-align: middle;
+          line-height: 24px;
+          background: #eee; /* minimal styling to make it recognizable as btn */
+          padding: 8px; /* padding to fix with min-height */
+          outline: none; /* focus style handled below */
         }
 
-        .btn {
-          min-height: 24px;
+        :host::after {
```
tlouisse, line 66: what does this do?

packages/button/src/LionButton.js
```diff
@@ -67,24 +52,35 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
...
+          content: '';
+          top: 0;
+          right: 0;
+          bottom: 0;
+          left: 0;
+          margin: 0;
+          padding: 0;
+        }
+
+        .button-content {
```
tlouisse, line 77: Not sure if it would be possible to omit this layer and apply flex-inline on the parent?

packages/button/test/lion-button.test.js
```diff
@@ -451,10 +451,10 @@ describe('lion-button', () => {
         });
       });
 
-      it('has host in the target property', async () => {
+      it('has native button in the target property', async () => {
         const el = /** @type {LionButton} */ (await fixture('<lion-button>foo</lion-button>'));
         const event = await prepareClickEvent(el);
-        expect(event.target).to.equal(el);
+        expect(event.target).to.equal(el._nativeButtonNode);
```
tlouisse, line 457: Ah, this is the test I was talking about. It should not be changed :)

packages/button/src/LionButton.js
jorgecasar: The problem of the current way is that you are triggering 2 clicks from one user interaction. It's not delegation, it's duplication. One from `lion-button` and other from the fake button created on the fly. If you listen the click in the form or above you will get 2 events instead of one as expected. Look at what's the current behavior:

image

I guess is more correct only dispatch one event from the children button because lion-button is on the event path as father. Look at the proposed behavior:

image

If you listen the event in the lion-button you can get it with currentTarget, because the real target who trigger the event is a native button in both cases.

packages/button/src/LionButton.js
jorgecasar: Changed the the "click area".

About the `cursor: pointer`, it's use for links, the native button has default cursor. I'm not sure if we want to change the cursor for buttons 🤔

packages/button/src/LionButton.js
jorgecasar: The slotted svg cause some problems with vertical align. I couldn't get a better approach. I'm open to suggestions.

Currently if you put a button inline with other with icon you can check the misalignment

packages/button/src/LionButton.js
jorgecasar: I added this test, because the current test 'is fired once' doesn't cover this case:

`
    it('is fired one inside a form', async () => {
      const clickSpy = /** @type {EventListener} */ (sinon.spy());
      const el = /** @type {LionButton} */ (await fixture(
        html`<form @click="${clickSpy}">
          <lion-button>foo</lion-button>
        </form>`,
      ));

      el.click();

      // trying to wait for other possible redispatched events
      await aTimeout(0);
      await aTimeout(0);

      expect(clickSpy).to.have.been.calledOnce
    });
`

tlouisse: That is indeed true, but:
- the currentTarget would not work for listeners using event delegation (so in case you don't attach handler to lion-button)
- it creates drawbacks for more commonly used scenarios (see below)
- we would advise to put click listeners (if needed on a form) on ing-form

I think we had your proposed approach in the past, but we decided to go for allowing event delegation for parent elements of lion-button that are in close proximity. Not just for **submit** actions, but mainly for regular clicks on lion-buttons with **type 'button'**.

So we want to support scenarios like these:

```
<consumed-component @click="${onClick}">
    #shadow-root
    <lion-button id="accept">accept</lion-button>
    <lion-button id="cancel">cancel</lion-button>
</consumed-component>
```

As consumer of consumed-component, I'm allowed to know what buttons exist and what ids they have. But I cannot directly add click handlers to those buttons, I need to do it on the host:

```js
function onClick(ev) {
  if (ev.target.id === 'accept') {
    proceed();
  }
}
```
A scenario like this is more likely to happen then that of listening on the form.

But you still have a valid point that on submit we generate 'noise' for the form element (and all its parents) that listen for clicks. We should actually reduce that nosie to a minimum and make sure that propagation is stopped on the form.

Then you got the alternative, not firing a click, but doing `form.submit()`. This creates other drawbacks, @jorenbroekema might know the details about this.

tlouisse: ### Long story short :)

In case of event delegation you are usually looking for something (like an id). Many events pass, but you filter the things you need (the rest is 'noise').

By using the logic we had, we only introduce a bit (very minimal) extra noise and only on reset/submit and on a native form.

This event would not be catched by the selection criterium of the one creating the event (delegated) listener, because it would not match his/her filter criteria.

packages/button/src/LionButton.js
tlouisse: Yep, we cannot omit this layer because of IE and this needs to be the flex-parent for the content

packages/button/src/LionButton.js
tlouisse: So checked it in the browser and we only need:

:host::before {
  /* occupy full width of host */
  left: 0;
  right: 0;
}

.button-content {
    position: relative; /* we need to win in stacking order from other positioned element (::before), so we can still selecr text */
    cursor: default; /* we should always see the default arrow, never a caret */
}

packages/button/src/LionButton.js
jorgecasar: What about this approaach?
```
      it('has host in the target property', async () => {
        const el = /** @type {LionButton} */ (await fixture('<lion-button type="button">foo</lion-button>'));
        const event = await prepareClickEvent(el);
        expect(event.target).to.equal(el);
      });

      it('has native button in the target property on submit or reset', async () => {
        const el = /** @type {LionButton} */ (await fixture('<lion-button>foo</lion-button>'));
        const event = await prepareClickEvent(el);
        expect(event.target).to.equal(el._nativeButtonNode);
      });
```

In your example the button type should be `button` instead of the default `submit`, then I suggest:

```
<consumed-component @click="${onClick}">
    #shadow-root
    <lion-button type="button" id="accept">accept</lion-button>
    <lion-button type="button" id="cancel">cancel</lion-button>
</consumed-component>
```

In this case it will work as spected

packages/button/src/LionButton.js
jorgecasar: I added an example, taking care about min-height and min-width. I guess `translate` approach works better because create an area button centric, instead of top left corner

packages/button/src/LionButton.js
tlouisse: Unfortunately the 'problem' with a native button (which we mimic) is that it has type submit by default.
As long as your button is not inside a form, it acts as a regular button (type button).

I think we should mimic that behavior, since most people would not use type=button in the above scenario (and it would be dom that the one applying onClick does not own).

But I agree that it would be nicest to not have the click on the form at all and that your test (the one below) would pass indeed.

```js
it('is fired one inside a form', async () => {
      const clickSpy = /** @type {EventListener} */ (sinon.spy());
      const el = /** @type {HTMLFormElement} */ (await fixture(
        html`<form @click="${clickSpy}">
          <lion-button>foo</lion-button>
        </form>`,
      ));

      // @ts-ignore
      el.querySelector('lion-button').click();

      // trying to wait for other possible redispatched events
      await aTimeout(0);
      await aTimeout(0);

      expect(clickSpy).to.have.been.calledOnce;
    });
```

So maybe we could try to see if we could solve without that extra native button click?
- since we have `this._form`we can try to call `submit()` instead (but I think @jorenbroekema tried before and there were some pitfalls. If I'm correct we listen to the submit event of the native form in lion-form and that wouldn't work anymore)
- we also tried to intercept the event from the native button and adjust the target to lion-button. Although this also led to issues in the end.

packages/button/src/LionButton.js
tlouisse: Nice 💪  Makes sense

packages/button/src/LionButton.js
jorgecasar: About the test, currently in master is not passing this test, because lion-button create a native button on the fly and throw the event, then the form and above receive 2 events (fake button and lion-button).

Natural behavior of a `<button type="submit">` inside a form dispatch the button click on submit (by clicking in the button and by pressing enter in a input field).

I tried to simulate `this._form.submit()`, but it send the fom, you can't preventDefault in the form listener.

Change the target it's not allowed, it's a readonly property.

I added `&& this._form` in the condition to check if the button is inside a form to trigger the native click. Otherwise trigger the `lion-button` click.

tlouisse: > I added `&& this._form` in the condition to check if the button is inside a form to trigger the native click. Otherwise trigger the `lion-button` click.

Sounds like a very reasonable tradeoff 👍  In that case it needs to act as a 'real submit button' only if part of a form.
Maybe we should then add some tests for this to make the difference (with and without form) clear?

jorgecasar: The tests are already added, check if they are enough

tlouisse: Ok thanks, could you make sure the previous use case is non breaking in case of a regular button (type button)?

```js
it('has host in the target property on `submit` or `reset` inside a form', async () => {
        const el = /** @type {LionButton} */ (await fixture('<lion-button>foo</lion-button>'));
        const clickSpy = /** @type {EventListener} */ (sinon.spy(e => e.preventDefault()));
        await fixture(html`<div @click="${clickSpy}">${el}</div>`);

        const event = await prepareClickEvent(el);

        expect(event.target).to.equal(el);
});
```

I think we can then wrap it up :) Thanks so far 💪

jorgecasar: I automated all use cases:

```
      const useCases = [
        { container: 'div', type: 'submit', targetHost: true },
        { container: 'div', type: 'reset', targetHost: true },
        { container: 'div', type: 'button', targetHost: true },
        { container: 'form', type: 'submit', targetHost: false },
        { container: 'form', type: 'reset', targetHost: false },
        { container: 'form', type: 'button', targetHost: true },
      ];
```

Feedback is welcome 👍

tlouisse: awesome, thanks 💪

tlouisse (approved):

daKmoR:
maybe add a changeset to actually release it?

PR #1022, opened 2020-10-12
fix(form-core): update child choice fields name attr to parent

### LINKED ISSUE #730: [input] can't change name dynamically

jorenbroekema:
## What I did

1. fixes issue where name delegation in ChoiceGroupMixin doesn't happen on updated callback when parent name changes
2. fixes #730

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceInputMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceInputMixin.js
@@ -110,6 +110,17 @@ const ChoiceInputMixinImplementation = superclass =>
       if (changedProperties.has('modelValue')) {
         this.__syncCheckedToInputElement();
       }
+
+      if (
+        changedProperties.has('name') &&
+        // @ts-expect-error not all choice inputs have a parent form group, since this mixin does not have a strict contract with the registration system
+        this.__parentFormGroup &&
+        // @ts-expect-error
+        this.__parentFormGroup.name !== this.name
+      ) {
+        // @ts-expect-error not all choice inputs have a name prop, because this mixin does not have a strict contract with form control mixin
+        this.name = changedProperties.get('name');
+      }
     }
 
     constructor() {
--- a/packages/listbox/test-suites/ListboxMixin.suite.js
+++ b/packages/listbox/test-suites/ListboxMixin.suite.js
@@ -96,24 +96,6 @@ export function runListboxMixinSuite(customConfig = {}) {
         );
       });
 
-      it('throws if a child element with a different name than the group tries to register', async () => {
-        const el = await fixture(html`
-        <${tag} name="gender">
-          <${optionTag} .choiceValue=${'female'} checked></${optionTag}>
-          <${optionTag} .choiceValue=${'other'}></${optionTag}>
-        </${tag}>
-      `);
-        const invalidChild = await fixture(html`
-        <${optionTag} name="foo" .choiceValue=${'male'}></${optionTag}>
-      `);
-
-        expect(() => {
-          el.addFormElement(invalidChild);
-        }).to.throw(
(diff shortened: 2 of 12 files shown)


packages/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -96,24 +96,6 @@ export function runListboxMixinSuite(customConfig = {}) {
         );
       });
 
-      it('throws if a child element with a different name than the group tries to register', async () => {
```
jorenbroekema, line 99: Duplicate test, implementation is not in ListboxMixin but rather in ChoiceGroupMixin

daKmoR: you now longer throw at all?

tlouisse:
Are you sure that it fixes #730? 

I think in a regular fieldset we do allow inputs to have different names and we want our formElements to get notified when a child input (not a choice inut) changes its name

jorenbroekema:
> Are you sure that it fixes #730?
> 
> I think in a regular fieldset we do allow inputs to have different names and we want our formElements to get notified when a child input (not a choice inut) changes its name

I think I fixed a different bug lol, since I copied the demo and replaced it with radio-group + radio. If I check the demo with fieldset + input or form + input, the bug is not there, so it seems like it's already fixed.

packages/listbox/test-suites/ListboxMixin.suite.js
jorenbroekema: Not in this instance no, it just syncs it to the parent's name without error'ing. If I kept the error, the situation would be this:

- parent name updates --> child name !== parent name --> error. So we have to add an edge case to not error in this case
- child formElement gets added --> child name !== parent name --> still error.

I figured it was better to align behaviours and just sync in all cases. I don't like the error, it's a bit unhelpful, and in all cases the solution to fix the error is to sync the name with the parent first. So why not do this by default for our users and prevent them from having to do manual work.

packages/form-core/src/choice-group/ChoiceInputMixin.js
```diff
@@ -110,6 +110,17 @@ const ChoiceInputMixinImplementation = superclass =>
       if (changedProperties.has('modelValue')) {
         this.__syncCheckedToInputElement();
       }
+
```
jorenbroekema, line 113: This enforces that you can't change radio `name` if it's different from the radio-group `name`. If you wanna change the name, it must be done through the parent.

gerjanvangeest (approved):

Issue #994, opened 2020-10-06
[combobox]: allow formatting of textbox content


palash2601:
--- edit by @tlouisse ---
> after having spoken to @palash2601, the goal is to allow formatting of textbox content. One way to achieve this is to allow plotting an input inside (like lion-input) that already supports formatting. Another way is to add FormatMixin to LionCombobox

--- end edit ---

## Expected behavior

Combobox should be rendered without any error when used with other lion components

## Actual Behavior

When lion field is used in `input` slot of combobox element, below error is thrown
```
ChoiceGroupMixin.js:210 Uncaught TypeError: Cannot read property 'checked' of undefined
    at HTMLElement._throwWhenInvalidChildModelValue (ChoiceGroupMixin.js:210)
    at HTMLElement.addFormElement (ChoiceGroupMixin.js:173)
    at HTMLElement.addFormElement (ListboxMixin.js:366)
    at HTMLElement._onRequestToAddFormElement (FormRegistrarMixin.js:163)
    at HTMLElement.connectedCallback (FormRegisteringMixin.js:31)
    at HTMLElement.connectedCallback (FormControlMixin.js:216)
    at HTMLElement.connectedCallback (ValidateMixin.js:163)
    at HTMLElement.connectedCallback (FormatMixin.js:381)
    at HTMLElement.connectedCallback (FocusMixin.js:26)
    at HTMLElement.connectedCallback (InteractionStateMixin.js:106)
```

This error is not thrown when native input is used.

Using `lion-input` triggers the form-element registration cycle. Since Combobox element extends ListBoxMixin which extends `ChoiceGroupMixin` this error is thrown. 

As per my understanding `ChoiceGroupMixin` [expects](https://github.com/ing-bank/lion/blob/master/packages/form-core/src/choice-group/ChoiceGroupMixin.js#L211) all the child elements to have modelValue as `Boolean` value and having `lion-input` as child element here results in this error

https://webcomponents.dev/edit/1Xuf6Y72LFuP8kuovLOj

## Additional context

Which component and version of it are you using.

@lion/combobox: 0.1.1

tlouisse:
It was not designed with this use case in mind. Only fieldset and forms allow nesting form controls atm (and choice groups allow choice inputs, which are the options in this particular case).
For a11y you need the input element. Maybe there's other ways to add advanced behavior like formatters. 
Could you make a feature request clearly describing the goals of this?

tlouisse:
For this to work you would need:

[...]

PR #1024, opened 2020-10-12
fix(form-core): add missing interaction states for feedback conditions

jorenbroekema:
## What I did

1. Add missing interactions states 'filled' and 'focused' to be considered in validation.
2. Fix our long-broken demo that actually very neatly shows how you can vary the interaction state conditions that apply to whether feedback gets shown or not.

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -192,7 +192,17 @@ export const ValidateMixinImplementation = superclass =>
         this.validate({ clearCurrentResult: true });
       }
 
-      if (['touched', 'dirty', 'prefilled', 'submitted', 'hasFeedbackFor'].includes(name)) {
+      if (
+        [
+          'touched',
+          'dirty',
+          'prefilled',
+          'focused',
+          'submitted',
+          'hasFeedbackFor',
+          'filled',
+        ].includes(name)
+      ) {
         this._updateShouldShowFeedbackFor();
       }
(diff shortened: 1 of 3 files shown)


packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -192,7 +192,17 @@ export const ValidateMixinImplementation = superclass =>
         this.validate({ clearCurrentResult: true });
       }
 
-      if (['touched', 'dirty', 'prefilled', 'submitted', 'hasFeedbackFor'].includes(name)) {
+      if (
```
tlouisse, line 195: Nice. Didn't know demos were broken actually :)

If I would be nitpicking, I would say the ValidateMixin knows nothing about the InteractionStateMixin.
If all elements would extend for either FIeld or Fieldset, we could place it there later (only Listbox (and thus SelectRich and Combobox do not extend from Field and we might want to if we want to support [formatting](https://github.com/ing-bank/lion/issues/994)).

tlouisse (approved):

packages/form-core/src/validate/ValidateMixin.js
jorenbroekema: Yeahh for subclassers it would indeed be nicer if ValidateMixin lived more agnostically of interaction states, as someone could use the validation system but have their own interaction states or something to align with their design system / methodology. Definitely a good idea for refactoring that later on.

Issue #1018, opened 2020-10-09
[validate][combobox] Validators do not work on Combobox


davidoff31:
## Expected behavior

I should be able to apply at least a Required validator to a combobox

## Actual Behavior

When I set the .validators property on the combobox I get an error 

implementation
![image](https://user-images.githubusercontent.com/63789961/95575088-638eac00-0a2e-11eb-9499-1115755ce947.png)

error 

![image](https://user-images.githubusercontent.com/63789961/95575122-743f2200-0a2e-11eb-941d-da52d3a6535b.png)

## Additional context

As seen on Combobox v0.1.2

jorenbroekema:
Yep, reproduced it, looking into a fix 👍  thanks for notifying

davidoff31:
@jorenbroekema The problem is still present with version 0.1.3 :disappointed:

jorenbroekema:
Hmm I spoke to Palash as well, apparently I solved a different bug (validation wasn't working with combobox, but I didn't get an error thrown).

The error you are seeing is quite common, it usually happens when you have multiple installs of @lion/form-core, so I would try to do:

```sh
npm ls @lion/form-core
```

And verify that there is only 1 install and the rest is deduped.

davidoff31:
I did have conflicting versions of form-core indeed.
Unfortunately the problem is there even when I resolve them:
![image](https://user-images.githubusercontent.com/63789961/96605403-750d6900-12f6-11eb-9def-de77b40a89fb.png)

Implement this:
![image](https://user-images.githubusercontent.com/63789961/96605450-83f41b80-12f6-11eb-9f5b-e05a0c3de50a.png)

and I get these errors:
![image](https://user-images.githubusercontent.com/63789961/96605546-9ec69000-12f6-11eb-9803-aa60724ab7ab.png)

jorenbroekema:
That's weird, npm should be able to dedupe the lion/form-core from the combobox up the tree to the lion/form-core in your root node modules, since they are the same version. The errors you get are expected if you have multiple installations, so that part is not really a bug on our end. 

Did you try:

```sh
npm dedupe
```

Or nuke your node_modules completely and reinstall?

davidoff31:
That made the trick, thanks

Issue #993, opened 2020-10-06
[combobox]: _inputNode value gets overrided with modelValue


palash2601:
On Implementing combobox with server-side interaction, i.e options list is received from backend as implemented in `feat/comboboxDemos` branch, below issue happens

## Expected behavior

When clearing the _inputNode value, the characters should get cleared

## Actual Behavior

When clearing the inputNode value, _inputNode gets filled with previously selected modelValue.

Steps:
1. Enter text like `App` in _inputNode and select `Apple` from list
2. start clearing the text in the _inputNode and wait for update to be complete

The _inputNode gets filled with previously selected modelValue which is `Apple`. Expected behaviour would be that the inputNode value gets cleared and updated list of options is visible

According to me it is happening as the component is getting re-rendered when new options list based on the new _inputNode node is rendered. This makes the call to [this](https://github.com/ing-bank/lion/blob/master/packages/combobox/src/LionCombobox.js#L277) code which sets the _inputNode with modelValue

## Additional context

Which component and version of it are you using.
`@lion/combobox: "0.1.1"`

tlouisse:
Interesting. So this only happens for the demo with server side interaction and not in general, right?

palash2601:
Yes....I think this is happening because new search text in `_inputNode` results in new list i.e new `lion-option` components are rendered in listbox. This triggers `form-element-register` for new child components(options) of combobox which results in call to [this](https://github.com/ing-bank/lion/blob/master/packages/listbox/src/ListboxMixin.js#L673) code which sets the `modelValue` to `_inputNode` value

palash2601:
I have created this PR with the change. Not 100% sure about the fix. What do you guys think?

PR #1026, opened 2020-10-12
fix(combobox): dispatch model-value-changed properly on unselect

jorenbroekema:
## What I did

1. Allow flexibility for extending the repropagation prevention conditions, which is needed for combobox, so that a model-value-changed event is propagated when no option matches after an input change. This allows validation to work properly e.g. for Required. fixes https://github.com/ing-bank/lion/issues/1018

It's a tricky one, because of two reasons: 

- First, the redispatching of modelValue update to Lit was not done correctly. The oldValue was passed as the already updated value since it goes through a custom getter, so Lit didn't see a change, and no updated callback would happen for modelValue. This part is fixed by storing and passing the oldModelValue, which makes validation work.
- Second, dirty is not set to true when pressing backspace (see demo). The reason is that the model-value-changed event that gets dispatched never arrives at the InteractionStateMixin, because the FormControlMixin does event.stopImmediatePropagation. Usually we dispatch a new event with the updated formPath, but this does not happen because the target is not checked. We assume that there will be another event from a target that IS checked, but this is not the case for combobox, where you can unselect by selecting "nothing". For a listbox / radio group, you can only unselect by selecting another option.

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -648,7 +648,6 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
         this.opened = false;
         this.__shouldAutocompleteNextUpdate = true;
         this._setTextboxValue('');
-        // this.checkedIndex = -1;
         break;
       case 'Enter':
         if (!this.formElements[this.activeIndex]) {
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -420,13 +420,30 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
     }
   }
 
-  /* eslint-disable no-param-reassign, class-methods-use-this */
+  /**
+   * We need to extend the repropagation prevention conditions here.
+   * Usually form groups with single choice will not repropagate model-value-changed of an option upwards
+   * if this option itself is not the checked one. We want to prevent duplicates. However, for combobox
+   * it is reasonable that an option can become unchecked without another one becoming checked, because
+   * users can enter any text they want, whether it matches an option or not.
+   *
+   * Therefore, extend the condition to fail by checking if there is any elements checked. If so, then we
+   * should indeed not repropagate as normally. If there is no elements checked, this will be the only
+   * model-value-changed event that gets received, and we should repropagate it.
+   *
+   * @param {EventTarget & import('../types/choice-group/ChoiceInputMixinTypes').ChoiceInputHost} target
+   */
+  _repropagationConditionFails(target) {
+    return super._repropagationConditionFails(target) && this.formElements?.some(el => el.checked);
+  }
 
+  /* eslint-disable no-param-reassign */
   /**
    * @overridable
    * @param {LionOption & {__originalInnerHTML?:string}} option
    * @param {string} matchingString
    */
+  // eslint-disable-next-line class-methods-use-this
   _onFilterMatch(option, matchingString) {
     const { innerHTML } = option;
     option.__originalInnerHTML = innerHTML;
@@ -443,7 +460,7 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    * @param {string} [curValue]
    * @param {string} [prevValue]
    */
-  // eslint-disable-next-line no-unused-vars
+  // eslint-disable-next-line no-unused-vars, class-methods-use-this
   _onFilterUnmatch(option, curValue, prevValue) {
     if (option.__originalInnerHTML) {
       option.innerHTML = option.__originalInnerHTML;
@@ -451,12 +468,14 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
     // Alternatively, an extension can add an animation here
     option.style.display = 'none';
   }
+  /* eslint-enable no-param-reassign */
 
   /**
    * Computes whether a user intends to autofill (inline autocomplete textbox)
    * @overridable
    * @param {{ prevValue:string, curValue:string }} config
    */
+  // eslint-disable-next-line class-methods-use-this
   _computeUserIntendsAutoFill({ prevValue, curValue }) {
     const userIsAddingChars = prevValue.length < curValue.length;
     const userStartsNewWord =
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -433,8 +433,8 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    *
    * @param {EventTarget & import('../types/choice-group/ChoiceInputMixinTypes').ChoiceInputHost} target
    */
-  _repropagationConditionFails(target) {
-    return super._repropagationConditionFails(target) && this.formElements?.some(el => el.checked);
+  _repropagationCondition(target) {
+    return super._repropagationCondition(target) || this.formElements.every(el => !el.checked);
   }
 
   /* eslint-disable no-param-reassign */
@@ -482,6 +482,7 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
       prevValue.length &&
       curValue.length &&
       prevValue[0].toLowerCase() !== curValue[0].toLowerCase();
+
     return userIsAddingChars || userStartsNewWord;
   }
 
--- a/packages/form-core/src/FormControlMixin.js
+++ b/packages/form-core/src/FormControlMixin.js
@@ -794,13 +794,14 @@ const FormControlMixinImplementation = superclass =>
         return;
       }
 
-      // B2. Are we a single choice choice-group? If so, halt when unchecked
+      // B2. Are we a single choice choice-group? If so, halt when target unchecked
+      // and something else is checked, meaning we will get
+      // another model-value-changed dispatch for the checked target
       //
       // We only send the checked changed up (not the unchecked). In this way a choice group
       // (radio-group, checkbox-group, select/listbox) acts as an 'endpoint' (a single Field)
       // just like the native <select>
-      // @ts-expect-error multipleChoice is not directly available but only as side effect
-      if (this._repropagationRole === 'choice-group' && !this.multipleChoice && !target.checked) {
+      if (this._repropagationConditionFails(target)) {
         return;
       }
 
@@ -819,6 +820,20 @@ const FormControlMixinImplementation = superclass =>
       // Since for a11y everything needs to be in lightdom, we don't add 'composed:true'
       this.dispatchEvent(
         new CustomEvent('model-value-changed', { bubbles: true, detail: { formPath } }),
+      );
+    }
+
+    /**
+     * TODO: Extend this in choice group so that target is always a choice input and multipleChoice exists.
+     * This will fix the types and reduce the need for ignores/expect-errors
+     * @param {EventTarget & import('../types/choice-group/ChoiceInputMixinTypes').ChoiceInputHost} target
+     */
+    _repropagationConditionFails(target) {
+      return (
+        this._repropagationRole === 'choice-group' &&
+        // @ts-expect-error multipleChoice is not directly available but only as side effect
+        !this.multipleChoice &&
+        !target.checked
       );
     }
 
--- a/packages/form-core/src/FormControlMixin.js
+++ b/packages/form-core/src/FormControlMixin.js
@@ -801,7 +801,7 @@ const FormControlMixinImplementation = superclass =>
       // We only send the checked changed up (not the unchecked). In this way a choice group
       // (radio-group, checkbox-group, select/listbox) acts as an 'endpoint' (a single Field)
       // just like the native <select>
-      if (this._repropagationConditionFails(target)) {
+      if (!this._repropagationCondition(target)) {
         return;
       }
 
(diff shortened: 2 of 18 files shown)


packages/listbox/src/ListboxMixin.js
```diff
@@ -670,7 +670,9 @@ const ListboxMixinImplementation = superclass =>
         ev.stopPropagation();
       }
       this.__onChildCheckedChanged(ev);
-      this.requestUpdate('modelValue', this.modelValue);
+
+      // don't send this.modelValue as oldValue, since it will take modelValue getter which takes it from child elements, which is already the updated value
```
tlouisse, line 674: We need to pass along the old value (it's a best practice and we should be able to detect)

@palash2601  did this as a merge request recently. It prevented unintended behavior inside the combobox textbox syncing.
@palash2601 could you elaborate on this an would it be possible to add a unit test for this?

Also I think if we want to do it correctly, I was thinking, we should do:

```js
setCheckedIndex(...) {
const oldModelValue = this.multipleChoice ? [...this.modelValue] : this.modelValue; 
... // body of method
this.requestUpdate('modelValue', oldModelValue);
}
```

And then we should add a `hasChanged` method to the static prop getter as well that is able to do a deep equal for arrays.

Also, we need to make sure all 'checked changed logic' is handled via this method (everything is now, except when checked events are fired by LionOptions).

palash2601: Hey @tlouisse , @jorenbroekema ,

This change earlier was done for [https://github.com/ing-bank/lion/issues/993](issue). More detailed RCA is in comments of this issue

jorenbroekema:
```js
const oldModelValue = this.multipleChoice ? [...this.modelValue] : this.modelValue; 
```

Won't work. It goes through the modelValue getter which fetches from the lion-option's --> checked. This is already "updated" so you don't actually get the old model value, you get the new one. And then if you pass that to `this.requestUpdate` the update never happens because the diff does not pass.

jorenbroekema:
```js
  __proxyChildModelValueChanged(ev) {
      // We need to redispatch the model-value-changed event on 'this', so it will
      // align with FormControl.__repropagateChildrenValues method. Also, this makes
      // it act like a portal, in case the listbox is put in a modal overlay on body level.
      if (ev.stopPropagation) {
        ev.stopPropagation();
      }
      this.__onChildCheckedChanged(ev);

      // don't send this.modelValue as oldValue, since it will take modelValue getter which takes it from child elements, which is already the updated value
      this.requestUpdate('modelValue', this.__oldModelValue);
      this.dispatchEvent(
        new CustomEvent('model-value-changed', { detail: { element: ev.target } }),
      );
      this.__oldModelValue = this.modelValue;
    }
```

This seems to work, now it passes the diff check and I actually get the updated callback for modelValue when I do backspace on a preselected option.

tlouisse: > ```js
> const oldModelValue = this.multipleChoice ? [...this.modelValue] : this.modelValue; 
> ```
> 
> Won't work. It goes through the modelValue getter (ChoiceGroupMixin) which fetches from the lion-option's --> checked. This is already "updated" so you don't actually get the old model value, you get the new one. And then if you pass that to `this.requestUpdate` the update never happens because the diff does not pass.

If I'm right you only get the new one if 'checked' was set on an individual option, right? (because it's already updated then by the child itself)
In all other cases, it's the body of `setCheckedIndex` that does the change. 
Ideally, all checked states would be managed by the parent via this method and a child would only request a checked state change.

This would be some work to implement, so maintaining a `this.__oldModelValue` would also do the job.

packages/combobox/src/LionCombobox.js
```diff
@@ -420,13 +420,30 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
+  /**
+   * We need to extend the repropagation prevention conditions here.
+   * Usually form groups with single choice will not repropagate model-value-changed of an option upwards
+   * if this option itself is not the checked one. We want to prevent duplicates. However, for combobox
+   * it is reasonable that an option can become unchecked without another one becoming checked, because
+   * users can enter any text they want, whether it matches an option or not.
+   *
+   * Therefore, extend the condition to fail by checking if there is any elements checked. If so, then we
+   * should indeed not repropagate as normally. If there is no elements checked, this will be the only
+   * model-value-changed event that gets received, and we should repropagate it.
```
tlouisse, line 432: Nice debugging 👍 
Should we allow for unselecting a single choice even on choicegroup level?

packages/form-core/src/FormControlMixin.js
```diff
@@ -812,6 +813,20 @@ const FormControlMixinImplementation = superclass =>
       );
     }
 
+    /**
+     * TODO: Extend this in choice group so that target is always a choice input and multipleChoice exists.
```
tlouisse, line 817: I considered this in the beginning and decided to keep all logic in one place for readability.

If we only have one condition method to override in choiceGroup (like this) then it would be acceptable.

Could we maybe give this method a 'positive name' (like `_repropagationCondition`)?

packages/listbox/src/ListboxMixin.js
```diff
@@ -670,10 +673,13 @@ const ListboxMixinImplementation = superclass =>
...
+      this.requestUpdate('modelValue', this.__oldModelValue);
       this.dispatchEvent(
         new CustomEvent('model-value-changed', { detail: { element: ev.target } }),
       );
+      this.__oldModelValue = this.modelValue;
```
tlouisse, line 682: Does this properly trigger modelValue change on multipleChoice?

Maybe we can guard this by adding this to ListboxMixinSuite?

```js
      it('requests update for modelValue when checkedIndex changes', async () => {
        <${tag} name="gender" .modelValue=${'other'} >
          <${optionTag} .choiceValue=${'male'}></${optionTag}>
          <${optionTag} .choiceValue=${'female'}></${optionTag}>
          <${optionTag} .choiceValue=${'other'}></${optionTag}>
        </${tag}>
      `);
        expect(el.checkedIndex).to.equal(2);
        const spy = sinon.spy(el, 'requestUpdate');
        el.setCheckedIndex(1);
        expect(spy).to.have.been.calledWith('modelValue', 'other');
      });

      it('requests update for modelValue after click', async () => {
        <${tag} name="gender" .modelValue=${'other'} >
          <${optionTag} .choiceValue=${'male'}></${optionTag}>
          <${optionTag} .choiceValue=${'female'}></${optionTag}>
          <${optionTag} .choiceValue=${'other'}></${optionTag}>
        </${tag}>
      `);
        expect(el.checkedIndex).to.equal(2);
        const spy = sinon.spy(el, 'requestUpdate');
        el.formElements[0].click();
        expect(spy).to.have.been.calledWith('modelValue', 'other');
      });
```

And then also for `multipleChoice` :)

We might need this as well:

```js
        modelValue: {
          ...super.modelValue,
          type: Object,
          hasChanged: (/** @type {string|string[]} */ nV, /** @type {string|string[]} */ oV) => {
            return Array.isArray(nV) ? JSON.stringify(nV) !== JSON.stringify(oV) : nV !== oV;
          },
        },
```

packages/combobox/src/LionCombobox.js
jorenbroekema: Perhaps.. Maybe the radio group is special in that it doesn't allow for unselecting, and this check is radio group specific. Then again, this behaviour works for listbox select (like rich select). So.. I'm not sure, this will be clearer if we have more use cases for single choice form fields I think. Maybe leave it as it is (default behaviour that you can't unselect) for now?

packages/listbox/src/ListboxMixin.js
jorenbroekema: Added a commit, small fix, after that those tests pass for both single and multiple choice 👍

packages/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -127,6 +128,62 @@ export function runListboxMixinSuite(customConfig = {}) {
...
+            <${optionTag} .choiceValue=${'male'}></${optionTag}>
+            <${optionTag} .choiceValue=${'female'}></${optionTag}>
+            <${optionTag} .choiceValue=${'other'}></${optionTag}>
+          </${tag}>
+        `);
+        expect(el.checkedIndex).to.eql([2]);
+        const spy = sinon.spy(el, 'requestUpdate');
+        el.setCheckedIndex(1);
+        expect(spy).to.have.been.calledWith('modelValue', sinon.match.array.deepEquals(['other']));
+      });
```
tlouisse, line 171: Nice. To make it really complete, I think we should also check whether it will be recognized in the `updated` callback via the `hasChanged` function. (if modelValue is an array reference, then we would not get notified if we have the default dirty check).

packages/listbox/src/ListboxMixin.js
tlouisse: Nice. I added on more remark in the test

packages/listbox/test-suites/ListboxMixin.suite.js
jorenbroekema: Hmm so just spy the updated callback you think? I guess it's kinda needed to really test that the update is done, and not cancelled because the dirty check fails.

packages/combobox/src/LionCombobox.js
tlouisse: That would be fine by me. The platform doesn't allow unselecting via UI, although I think when you build groups with role=radio this is a possibility (the polymer radio-group supported this at least)

packages/listbox/test-suites/ListboxMixin.suite.js
tlouisse: Indeed, spy on updated and inspect changedProperties

jorenbroekema: Done, was a bit annoying to assert it because my sinon matcher wouldn't match with the multiple choice updated model value because it's an array and doesn't "true equals" the expected match, but I worked around it 👍

jorenbroekema: Turns out we don't actually need a hasChanged btw, I checked with and without and it doesn't make a difference 🤷

tlouisse: That's weird. Do we change the reference to the array somewhere?

If so, we would also have a change detected on similar arrays

packages/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -781,7 +781,7 @@ export function runListboxMixinSuite(customConfig = {}) {
           listbox.dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter' }));
           el.activeIndex = 1;
           listbox.dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter' }));
-          expect(options[0].checked).to.equal(true);
+          // expect(options[0].checked).to.equal(true);
```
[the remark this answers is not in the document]
jorenbroekema: no, I saw it too, already uncommented

packages/listbox/src/ListboxMixin.js
```diff
@@ -338,23 +338,29 @@ const ListboxMixinImplementation = superclass =>
...
-     * @param {'set'|'unset'|'toggle'} multiMode
+     * If an array is passed for multiple-choice, it will check the indexes in array, and uncheck the rest
+     * If a number is passed, the item with the passed index is checked without unchecking others
+     * For single choice, __onChildCheckedChanged we ensure that we uncheck siblings
+     * @param {number|number[]} index
      */
-    setCheckedIndex(index, multiMode = 'toggle') {
-      if (this.formElements[index]) {
-        if (!this.multipleChoice) {
+    setCheckedIndex(index) {
```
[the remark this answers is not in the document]
jorenbroekema: I didn't understand the use cases and there were no tests, so I ended up removing it. For single choice you can do everything by 'set' because untargeted children get unchecked automatically. Toggle? just set to -1 to toggle it off, set to index to toggle it on. Unset? Just set to -1. 

I'm okay with adding it back in for multiple choice, as it would be useful there but maybe we should have tests so the behaviour is clearly specced, I couldn't really infer it from the implementation.

packages/listbox/test-suites/ListboxMixin.suite.js
jorenbroekema: I guess we do, couldn't really figure that one out..

jorenbroekema: Maybe you can create an issue + failing test later if you find that doing setCheckedIndex = [1, 3] when 1 and 3 are already selected results in a model-value-changed :D?

packages/listbox/src/ListboxMixin.js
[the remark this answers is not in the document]
tlouisse: Indeed, the 2nd param only makes sense for multipleChoice. We used it before in _listboxOnClick, so it was indirectly tested before. A unit test on top would be nice indeed.
We can also add it back once we have a use case for it, so we can get this going 👍

packages/listbox/test-suites/ListboxMixin.suite.js
tlouisse: Ok, I will 👍

tlouisse (approved):

PR #1029, opened 2020-10-15
fix(tooltip): keeps tooltip closed when invoker is disabled

gerjanvangeest:
## What I did

1. Add a check if the invoker is disabled, if so don't open the tooltip

### DIFF
--- a/packages/tooltip/README.md
+++ b/packages/tooltip/README.md
@@ -47,18 +47,21 @@ const tooltipDemoStyles = css`
 
 ```js preview-story
 export const main = () => html`
-    <style>${tooltipDemoStyles}</style>
-    <lion-tooltip>
-      <button slot="invoker" class="demo-tooltip-invoker">Hover me</button>
-      <div slot="content" class="demo-tooltip-content">This is a tooltip<div>
-    </lion-tooltip>
-  `;
+  <style>
+    ${tooltipDemoStyles}
+  </style>
+  <lion-tooltip>
+    <button slot="invoker" class="demo-tooltip-invoker">Hover me</button>
+    <div slot="content" class="demo-tooltip-content">This is a tooltip</div>
+  </lion-tooltip>
+`;
 ```
 
 ## Features
 
 - Show content when hovering the invoker
 - Show content when the invoker is focused
+- Does not show content when invoker is disabled
 - Uses Popper.js under the hood, to have the content pop up relative to the invoker
 - Use `.config` to override the overlay configuration
 - Config has `popperConfig` property that has a one to one relation with Popper.js configuration API.
--- a/packages/tooltip/src/LionTooltip.js
+++ b/packages/tooltip/src/LionTooltip.js
@@ -66,6 +66,16 @@ export class LionTooltip extends ArrowMixin(OverlayMixin(LitElement)) {
     });
   }
 
+  _hasDisabledInvoker() {
+    if (this._overlayCtrl && this._overlayCtrl.invoker) {
+      return (
+        /** @type {HTMLElement & { disabled: boolean }} */ (this._overlayCtrl.invoker).disabled ||
+        this._overlayCtrl.invoker.getAttribute('aria-disabled') === 'true'
+      );
+    }
+    return false;
+  }
+
   _setupOpenCloseListeners() {
     super._setupOpenCloseListeners();
     this.__resetActive = this.__resetActive.bind(this);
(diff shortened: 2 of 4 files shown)


packages/tooltip/README.md
```diff
@@ -47,18 +47,21 @@ const tooltipDemoStyles = css`
...
+    <div slot="content" class="demo-tooltip-content">This is a tooltip</div>
+  </lion-tooltip>
+`;
 ```
 
 ## Features
 
 - Show content when hovering the invoker
 - Show content when the invoker is focused
+- Does not show content when invoker is disabled
```
jorenbroekema, line 64: Can we create a demo for that?

packages/tooltip/src/LionTooltip.js
```diff
@@ -66,6 +66,14 @@ export class LionTooltip extends ArrowMixin(OverlayMixin(LitElement)) {
     });
   }
 
+  _hasDisabledInvoker() {
+    return (
+      this._overlayCtrl &&
+      this._overlayCtrl.invoker &&
+      /** @type {HTMLElement & { disabled: boolean }} */ this._overlayCtrl.invoker.disabled
```
jorenbroekema, line 73:
```suggestion
      /** @type {HTMLElement & { disabled: boolean }} */ (this._overlayCtrl.invoker).disabled
```

packages/tooltip/README.md
gerjanvangeest: I felt it bit overkill to add a demo that doesn't "work", but if you feel it necessary, i could add it easily.

packages/tooltip/src/LionTooltip.js
```diff
@@ -66,6 +66,14 @@ export class LionTooltip extends ArrowMixin(OverlayMixin(LitElement)) {
     });
   }
 
+  _hasDisabledInvoker() {
+    return (
+      this._overlayCtrl &&
+      this._overlayCtrl.invoker &&
+      /** @type {HTMLElement & { disabled: boolean }} */ (this._overlayCtrl.invoker).disabled
```
tlouisse, line 73: could we also check for aria-disabled === 'true'?

gerjanvangeest: As in `&&` or `||`?

packages/tooltip/README.md
jorenbroekema: Hmm true, but uninspiring to show a disabled button with a non-working tooltip 👍

packages/tooltip/src/LionTooltip.js
jorenbroekema: Good point, I would assume ||

tlouisse:
Was actually wondering whether this behavior should be on OverlayController in general.
(we can do that later as well and start with just the tooltip btw)

packages/tooltip/src/LionTooltip.js
```diff
@@ -66,6 +66,15 @@ export class LionTooltip extends ArrowMixin(OverlayMixin(LitElement)) {
...
   }
 
+  _hasDisabledInvoker() {
+    return (
+      this._overlayCtrl &&
+      this._overlayCtrl.invoker &&
+      /** @type {HTMLElement & { disabled: boolean }} */ ((this._overlayCtrl.invoker).disabled ||
+        this._overlayCtrl.invoker.getAttribute('aria-disabled') === 'true')
+    );
+  }
```
jorenbroekema, line 76:
```suggestion
  _hasDisabledInvoker() {
    return (
      this._overlayCtrl &&
      this._overlayCtrl.invoker &&
      (/** @type {HTMLElement & { disabled: boolean }} */ (this._overlayCtrl.invoker).disabled ||
        this._overlayCtrl.invoker.getAttribute('aria-disabled') === 'true')
    );
  }
```

should fix the typecast

gerjanvangeest:
> Was actually wondering whether this behavior should be on OverlayController in general.
> (we can do that later as well and start with just the tooltip btw)

This is only needed for overlays that open on `hover`, which so far is only the tooltip. But yes, could be placed on the OverlayController.

tlouisse:
> > Was actually wondering whether this behavior should be on OverlayController in general.
> > (we can do that later as well and start with just the tooltip btw)
> 
> This is only needed for overlays that open on `hover`, which so far is only the tooltip. But yes, could be placed on the OverlayController.

Where is this specced? Doesn't make much sense to me tbh. Disabled invoker is disabled for a reason, right?

gerjanvangeest:
> Where is this specced? Doesn't make much sense to me tbh. Disabled invoker is disabled for a reason, right?

If a normal button is disabled, it wont work on `click` by default.

packages/tooltip/src/LionTooltip.js
gerjanvangeest: The linter says otherwise..

jorenbroekema (approved):

Issue #855, opened 2020-08-03
[accordion] resizing of textarea not called when in accordion


MathieuPuech:
## Describe the bug

When using a lion-textarea in a lion-accordion, the lion-textarea is not resized when the accordion show the content

## To Reproduce

Reproduced on the demo of accordion:
https://github.com/ing-bank/lion/tree/bug/textareaInAccordion

1. Go to accordion storybook to the With textarea story
2. Click on 'Lorem' button
3. See error

## Expected behavior

Resizing of textarea called when the visibility changed

## Screenshots

Not resized:
![image](https://user-images.githubusercontent.com/9367152/89220218-2ece2680-d59f-11ea-9fb9-b7e4c461dfc4.png)

After manually calling resizeTextarea() of lion-textarea:
![image](https://user-images.githubusercontent.com/9367152/89220361-70f76800-d59f-11ea-84f4-8f56c730e96f.png)

jorenbroekema:
Hmm this is a tough one. An easy fix is to call `resizeTextarea()` on `lion-textarea` when the accordion is toggled.

However, you will run into this problem constantly, as this would also show in a dialog that is hidden by default. Opening the dialog with the textarea inside would also not make it do `resizeTextarea()`. MutationObserver would be the wrong choice in my opinion, as you don't know the level of nesting of textarea in relation to the container that toggles the visibility indirectly.

So the only feasible way in my opinion would be IntersectionObserver (I haven't tried it but I assume that would fix it). This has decent support, except IE11 of course. So for that we need to document clearly that a polyfill has to be added. If we check that IntersectionObserver exists before instantiating, I think we can do this and it won't be a breaking change. This means if people run into this bug on IE11, they need to just add the polyfill in their app and it's fixed.

Sound good?

MathieuPuech:
I think a fix working with a polyfill for old browsers is good enough. It’s a very specific use case.

PR #1030, opened 2020-10-15
fix(textarea): add intersection observer for textarea recalculate

jorenbroekema:
## What I did

1. Added Intersection Observer to ensure textarea gets recalculated on visibility changes, e.g. when it is inside a dialog, tabs, or accordion when they are hidden initially

fixes https://github.com/ing-bank/lion/issues/855

### DIFF
--- a/packages/textarea/src/LionTextarea.js
+++ b/packages/textarea/src/LionTextarea.js
@@ -73,6 +73,8 @@ export class LionTextarea extends NativeTextFieldMixin(LionFieldWithTextArea) {
     // eslint-disable-next-line wc/guard-super-call
     super.connectedCallback();
     this.__initializeAutoresize();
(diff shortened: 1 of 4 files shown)


packages/textarea/src/LionTextarea.js
```diff
@@ -73,6 +73,11 @@ export class LionTextarea extends NativeTextFieldMixin(LionFieldWithTextArea) {
     // eslint-disable-next-line wc/guard-super-call
     super.connectedCallback();
     this.__initializeAutoresize();
+    this.__intersectionObserver = new IntersectionObserver(() => {
+      console.log('i');
```
gerjanvangeest, line 77: console...

jorenbroekema: oh oops, fixed

gerjanvangeest (approved):

PR #1034, opened 2020-10-16
feat(listbox): add click on enter for options with href

Bubbit:
## What I did

1. Current implementation with LinkMixin allows to add links to the options of a combox, however link is only triggered on a 'click' and not on 'enter' as the listboxmixin blocks this. 
In order to get the same behaviour an extra check is implemented to click on 'enter' of an option with a 'href' attribute.
2. Added example doc.
3. added Rel option to LinkMixin in order to allow passing in 'noopener noreferrer' when required.

Test to follow (let me know if you have a suggestion on how to test this properly)

### DIFF
--- a/.changeset/eight-lions-provide.md
+++ b/.changeset/eight-lions-provide.md
@@ -1,0 +1,5 @@
+---
+'@lion/listbox': minor
+---
+
+Add click on enter for options with href, to ensure that anchors are navigated towards, for example when applying LinkMixin to LionOption as part of a listbox.
--- a/packages/combobox/docs/LinkMixin.js
+++ b/packages/combobox/docs/LinkMixin.js
@@ -51,6 +51,9 @@ const LinkMixinImplementation = superclass =>
       if (changedProperties.has('target')) {
         this._nativeAnchor.target = this.target;
       }
+      if (changedProperties.has('rel')) {
+        this._nativeAnchor.rel = this.rel;
+      }
     }
 
     __navigate() {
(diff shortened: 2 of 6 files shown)


tlouisse:
Nice. I think the demo would be a good fundament for a google search box extension 👍

About the test:
- it should not be dependent on LinkMixin (since it doesn't exist officially, only in the docs)
- It should make clear that we allow for listbox options being anchors (like google search)
- It should make clear that because we use activedescendent (and not a roving tabindex) an option will not be the activeElement and therefore the [enter] listener will not be received on the individual option.


pseudo code for a test:

```js
it('allows anchors to be clicked when a [href] attribute is present', async () => {
      <lion-combobox autocomplete="none" opened>
         <lion-option href="https://google.com">Google</lion-option>
         <lion-option href="https://duckduckgo.com">DuckDuck Go</lion-option>
      </lion-combobox>
    `);

  // Make second option the activedescendant. This means it has visual focus and 'screen reader focus', 
  // but _inputNode has 'browser focus' (meaning it is the activeElement)
  el.activeIndex = 1;

  // Allow options that behave like anchors (think of Google Search) to trigger the anchor behavior
  const activeOption = el.formElements[1];
  const clickSpy = sinon.spy(activeOption, 'click')

  // Now the user selects via keyboard interaction
  el._inputNode.dispatchEvent(new KeyboardEvent('Enter'));
 
  expect(clickSpy).to.have.been.calledOnce; 
});
```  

Ideally we would also test that no click will be triggered without an href

Bubbit:
Added both the test cases 👍

jorenbroekema (approved):

jorenbroekema:
Thanks @Bubbit  🤗

PR #1041, opened 2020-10-21
[draft] chore: add index.d.ts for all packages

tlouisse:
## What I did

- add index.d.ts for all packages, so they can be imported on extension layer
- make LocalizeMixin depend on LitElement, so `styles` etc. will not error
- make `slots` a getter in types, so sublcasses will not complain when redefined


daKmoR:
> add index.d.ts for all packages

but why?

tlouisse:
> but why?
The implementing party needs them to recognize exported types inside node_modules

daKmoR:
these are already auto generated by tsc itself... no need to manually add or maintain them

so for example
https://unpkg.com/browse/@lion/combobox@0.1.3/index.d.ts

tlouisse:
> these are already auto generated by tsc itself... no need to manually add or maintain them

That's a way better solution indeed 👍  My local node_modules were a bit behind, so they weren't exported yet.
I will close this one, since the other 'fixes' turned out not to solve the problem in the end (took a while for type checks to 'kick in' after a vscode restart)

PR #1049, opened 2020-10-23
chore: type enhancements SlotMixin, LocalizeMixin, SelectRichInvoker

tlouisse:
## What I did

type enhancements SlotMixin, LocalizeMixin, SelectRichInvoker

### DIFF
--- a/packages/localize/types/LocalizeMixinTypes.d.ts
+++ b/packages/localize/types/LocalizeMixinTypes.d.ts
@@ -1,4 +1,5 @@
 import { Constructor } from '@open-wc/dedupe-mixin';
+import { LitElement } from '@lion/core';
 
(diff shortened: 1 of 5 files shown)


packages/localize/types/LocalizeMixinTypes.d.ts
```diff
@@ -1,4 +1,5 @@
 import { Constructor } from '@open-wc/dedupe-mixin';
+import { LitElement } from 'lit-element';
```
jorenbroekema, line 2: import from `@lion/core` so we control the version, in case lit-element releases an update that we need to hold off on because of incompatibilities.

jorenbroekema (changes requested):

jorenbroekema (approved):

PR #1053, opened 2020-10-26
fix(localize): do not parse non-existing dates

gerjanvangeest:
`parseDate('31.02.2020') returned 'Mon Mar 02 2020 00:00:00 GMT+0100....'`

But not anymore, now it returns `undefined`.

### DIFF
--- a/packages/localize/src/date/parseDate.js
+++ b/packages/localize/src/date/parseDate.js
@@ -56,8 +56,17 @@ export function parseDate(dateString) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
-  if (year > 0 && month > 0 && day > 0) {
-    return new Date(Date.UTC(year, month - 1, day));
+  const parsedDate = new Date(Date.UTC(year, month - 1, day));
+
+  // Check if parsedDate is not `Invalid Date` or that the date has changed (e.g. the not existing 31.02.2020)
+  if (
+    year > 0 &&
+    month > 0 &&
+    day > 0 &&
+    parsedDate.getDate() === day &&
+    parsedDate.getMonth() === month - 1
+  ) {
+    return parsedDate;
   }
   return undefined;
 }
--- a/.changeset/swift-windows-train.md
+++ b/.changeset/swift-windows-train.md
@@ -1,0 +1,5 @@
+---
+'@lion/localize': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/localize/src/date/parseDate.js
```diff
@@ -53,10 +53,11 @@ export function parseDate(dateString) {
     default:
       parsedString = '0000/00/00';
   }
+
   const parsedDate = new Date(parsedString);
-  // Check if parsedDate is not `Invalid Date`
+  // Check if parsedDate is not `Invalid Date` or that the date has changed (e.g. the not existing 31.02.2020)
   // eslint-disable-next-line no-restricted-globals
-  if (!isNaN(parsedDate.getTime())) {
+  if (!isNaN(parsedDate.getTime()) && parsedDate.getDate() === Number(parsedString.slice(8, 10))) {
```
tlouisse, line 60: `Number(parsedString.slice(8, 10)` seems to be based on DateFormatBasedOnLocale `year-month-day`. On a locale with `month-day-year`, it would fail because the day would be somewhere else inside parsedString?

gerjanvangeest: In the function above the code (`parsedString `) is always formatted in the same order `year-month-day`.

tlouisse: That would either mean [getDateFormatBasedOnLocale](https://github.com/ing-bank/lion/blob/52bf194d3961db24b4a896d0c1d2f500fb1e0385/packages/localize/src/date/getDateFormatBasedOnLocale.js#L8) always gives back `year-month-day` or the switch statement evaluates values it could never get :)

I think we should run `return undefined when no valid date provided` for a different locale as well:

```js
    expect(parseDate('31.02.2020')).to.equal(undefined);
    expect(parseDate('02.31.2020')).to.equal(undefined); // probably fails
```

gerjanvangeest: I added those tests (and some extra) all succeeded

tlouisse: That's true, I debugged it.
The reason is that `new Date('2020/31/02')` gives back null and `new Date('2020/02/31')` does not apparently :)

Ok, I approve this, but an explanation that `parsedDate.getDate() === Number(parsedString.slice(8, 10))` can only get hit with `year-month-day` due to the platform handling different constructor formats differently, wouldn't hurt imho.

And those extra tests to confirm this (with locales with different orders) would then help as well.

tlouisse (approved):
Approved it, but an extra expect and an extra comment wouldn't hurt

PR #1054, opened 2020-10-26
fix(input-stepper): make use of prefix and suffix slots

hardikpthv:
## What I did

1. Use prefix and suffix existing slots for the increment and decrement buttons
2. Add methods that can be overridden for increment and decrement icons

### DIFF
--- a/packages/input-stepper/src/LionInputStepper.js
+++ b/packages/input-stepper/src/LionInputStepper.js
@@ -251,15 +251,6 @@ export class LionInputStepper extends LionInput {
   }
 
   /**
-   * Override before template to none
-   * @override
-   */
-  // eslint-disable-next-line class-methods-use-this
-  _inputGroupBeforeTemplate() {
-    return html``;
-  }
-
-  /**
    * Get the decrementor button sign template
    * @returns {String|import('lit-element').TemplateResult}
    */
--- a/packages/input-stepper/src/LionInputStepper.js
+++ b/packages/input-stepper/src/LionInputStepper.js
@@ -34,8 +34,6 @@ export class LionInputStepper extends LionInput {
         type: Number,
         reflect: true,
       },
-      __disableIncrementor: { attribute: false },
-      __disableDecrementor: { attribute: false },
     };
   }
 
@@ -50,8 +48,6 @@ export class LionInputStepper extends LionInput {
     super();
     /** @param {string} modelValue */
     this.parser = modelValue => parseFloat(modelValue);
-    this.__disableIncrementor = false;
-    this.__disableDecrementor = false;
     this.min = Infinity;
     this.max = Infinity;
     this.step = 1;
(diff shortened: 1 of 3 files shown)


packages/input-stepper/src/LionInputStepper.js
```diff
@@ -103,6 +99,15 @@ export class LionInputStepper extends LionInput {
     }
   }
 
+  // @ts-ignore
+  get slots() {
+    return {
+      ...super.slots,
+      prefix: () => this.__getButtonNode(),
+      suffix: () => this.__getButtonNode(true),
```
gerjanvangeest, line 107: This feels strange, better split it up in 2 functions.

packages/input-stepper/src/LionInputStepper.js
```diff
@@ -195,61 +235,67 @@ export class LionInputStepper extends LionInput {
   }
 
   /**
-   * Override after template to none
+   * Override before template to none
    * @override
    */
   // eslint-disable-next-line class-methods-use-this
-  _inputGroupAfterTemplate() {
+  _inputGroupBeforeTemplate() {
```
gerjanvangeest, line 242: I didn't see review first version of the stepper, but I was wondering why do we override these?

packages/input-stepper/src/LionInputStepper.js
```diff
@@ -195,61 +235,67 @@ export class LionInputStepper extends LionInput {
...
+   * @returns {import('lit-element').TemplateResult}
+   */
+  _decrementorTemplate() {
     return html`
       <button
-        ?disabled=${this.disabled || this.readOnly || this.__disableDecrementor}
+        ?disabled=${this.disabled || this.readOnly}
         @click=${this.__decrement}
         tabindex="-1"
         name="decrement"
```
gerjanvangeest, line 274: What is the need of a `name` attribute on the button?

gerjanvangeest (commented):
Hi @hardikpthv, looking good.

Just some small remarks. And you should also run `yarn changeset` to make sure you release a new version.

packages/input-stepper/src/LionInputStepper.js
hardikpthv: I think we need it otherwise `before` slot will show the text/content above the text box which is not expected.

packages/input-stepper/src/LionInputStepper.js
hardikpthv: I think it's not needed anymore, I will remove it.

packages/input-stepper/src/LionInputStepper.js
gerjanvangeest: 1. It only appears if you add anything to the slot
2. You can style the way you want it (most likely inline) since this is a white label component. 
3. We allow it for all other input fields, so i don't see a reason to disallow it for the stepper.

hardikpthv: Okay, let's allow it. 😅

gerjanvangeest (approved):

PR #1055, opened 2020-10-27
chore: upgrade to npm 7 (workspaces), get rid of yarn

jorenbroekema:
## What I did

1. Replace all yarn usages / recommendations to recommend npm 7.
2. Use npm + npm cache in CI, install npm 7 globally in CI so it works.
3. Verified tests + storybook still work fine everywhere. Looks like the hoisting works properly, since we always run stuff from the root (web-dev-server etc.).

question: do we need to release anything? I think this is a pure dev-workflow related PR, even though it has some changes in packages --> package.jsons, I don't think those changes have to be released. It's only important for local.

TODO:

https://github.com/ing-bank/lion/pull/1055/files#diff-d249536d464658bab74496b4a0355dc1bf3dbdbb5a94ea055d042d71a6b70c43R97

NPM doesn't have a --scope flag like yarn does, so that's a bit annoying for updating your `@lion` dependencies. I found `update-by-scope` package which is simple enough, but doesn't allow you to add flags to the install command like `--save-exact`. I think it's probably easier to do a PR to `update-by-scope` than to add a feature to NPM 🙈  so I'll reach out to the maintainer of `update-by-scope` whether he's down for a PR to add support.


.github/workflows/release.yml
```diff
@@ -18,33 +18,36 @@ jobs:
...
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: ${{ steps.npm-cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
-            ${{ runner.os }}-yarn-
+            ${{ runner.os }}-npm-
 
       - name: Install Dependencies
-        run: yarn --frozen-lockfile
+        run: npm install
```
daKmoR, line 43:
```suggestion
        run: npm ci
```
faster and will throw if there is no lock file

.github/workflows/verify.yml
```diff
@@ -14,26 +14,29 @@ jobs:
...
+          path: ${{ steps.npm-cache-dir.outputs.dir }}
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
-            ${{ runner.os }}-yarn-
+            ${{ runner.os }}-npm-
 
-      - name: Install dependencies
-        run: yarn --frozen-lockfile
+      - name: Install Dependencies
+        run: npm install
```
daKmoR, line 33:
```suggestion
        run: npm ci
```

.github/workflows/verify.yml
```diff
@@ -46,25 +49,28 @@ jobs:
...
```
daKmoR, line 68:
```suggestion
        run: npm ci
```

.github/workflows/verify.yml
```diff
@@ -80,20 +86,23 @@ jobs:
...
```
daKmoR, line 105:
```suggestion
        run: npm ci
```

daKmoR (approved):

PR #1061, opened 2020-10-28
fix(radio-group): reset bug

gerjanvangeest:
## What I did

1. Look into a reset bug

 > If a checked radio, which comes after the initial checked radio, is reset, all radios gets unchecked.

2. Add test suites to radio & checkbox groups


packages/radio-group/test/lion-radio-group.test.js
```diff
@@ -11,6 +11,43 @@ import '../lion-radio.js';
...
+        <lion-radio label="Other" .choiceValue=${'other'}></lion-radio>
+      </lion-radio-group>
+    `);
+    el.formElements[0].checked = true;
+    expect(el.modelValue).to.deep.equal('male');
+
+    el.resetGroup();
+    expect(el.modelValue).to.deep.equal('female');
+
+    el.formElements[2].checked = true;
```
gerjanvangeest, line 42: this section is still failing, and why i started to add all the other tests.

packages/checkbox-group/test/lion-radio.test.js
```diff
@@ -0,0 +1,26 @@
+import { expect, fixture, html } from '@open-wc/testing';
+import '../lion-checkbox.js';
+
```
jorenbroekema, line 3: Filename should be `lion-checkbox.test.js`

packages/radio-group/src/LionRadioGroup.js
```diff
@@ -10,4 +10,26 @@ export class LionRadioGroup extends ChoiceGroupMixin(FormGroupMixin(LitElement))
     super.connectedCallback();
     this.setAttribute('role', 'radiogroup');
   }
+
+  /**
+   * @override FormatMixin, during a reset if the current checked value is behind
+   * the initial checked value, they both got unchecked
+   */
```
jorenbroekema, line 17: This is an override of the FormGroupMixin I believe. I am wondering why it is necessary.

`child.reset()` should call ---> LionField.reset(), because LionRadio is a LionField extension. Here it does:

```js
reset() {
  this.modelValue = this._initialModelValue;
  this.resetInteractionState();
}
```
So my question would be, why is the initialModelValue "wrong" for radios? And maybe that's where we should implement a fix

packages/radio-group/test/lion-radio-group.test.js
```diff
@@ -11,6 +10,41 @@ import '../lion-radio.js';
 const fixture = /** @type {(arg: TemplateResult) => Promise<LionRadioGroup>} */ (_fixture);
 
 describe('<lion-radio-group>', () => {
+  it('restores to empty modelValue if changes were made', async () => {
```
jorenbroekema, line 13: Should these tests be moved to the ChoiceGroupMixin suite? I think these tests apply for all single choice groups right?

packages/radio-group/test/lion-radio.test.js
```diff
@@ -1,11 +1,39 @@
...
   it('should have type = radio', async () => {
-    const el = await fixture(`
-      <lion-radio name="radio" value="male"></lion-radio>
+    const el = await fixture(html`
+      <lion-radio name="radio" .choiceValue="${'male'}"></lion-radio>
     `);
     expect(el.getAttribute('type')).to.equal('radio');
   });
+
+  it('can be reset when unchecked by defualt', async () => {
```
jorenbroekema, line 16: Should these tests be moved to the ChoiceInputMixin suite? I think these tests apply for all single choice input fields right? The only specific thing is type="radio" so that should be the only test in this file I think

packages/radio-group/test/lion-radio-group.test.js
jorenbroekema: Is this resolved or still failing?

jorenbroekema (changes requested):

packages/radio-group/test/lion-radio-group.test.js
gerjanvangeest: This is fixed via: https://github.com/ing-bank/lion/pull/1061#discussion_r529391587

packages/radio-group/src/LionRadioGroup.js
gerjanvangeest: Yes, I quoted the wrong mixin..

The `_initialModelValue` is correct, the problem is that at some point you get 2 checked items (one restored from initial value and one still checked). Radio group can only have 1 selected child, which causes that the initial value gets deselected again once the selected child gets a reset.

packages/radio-group/test/lion-radio-group.test.js
gerjanvangeest: Only the reset code is not part of the `ChoiceGroupMixin`, but the `FormGroupMixin`, should be applied there. Only the `FormGroupMixin-suite` as is, can't be used in other components. It needs a big refactor to make it suitable.

Will add a TODO

packages/radio-group/src/LionRadioGroup.js
jorenbroekema: Ah yeah fair enough that's a bit tricky

jorenbroekema (approved):

PR #1063, opened 2020-10-29
feat: Adds Input-iban country blacklisting

andrei1902:
## What I did

1. Added IsBlacklistedCountryIBAN class on validators (error messages are the same as IsCountryIBAN validator)
2. Added unit test for new validator

### DIFF
--- a/packages/input-iban/README.md
+++ b/packages/input-iban/README.md
@@ -90,3 +90,23 @@ export const countryRestrictions = () => html`
   <small>Demo instructions: you can use NL20 INGB 0001 2345 67</small>
 `;
 ```
+
+### Blacklisted Country
+
+By default, we validate the input to ensure the IBAN is valid.
+To get the default feedback message for this default validator, use `loadDefaultFeedbackMessages` from `@lion/form-core`.
+
+In the example below, we show how to use an additional validator that restricts the `input-iban` to IBANs from certain countries.
+
+```js preview-story
+export const blacklistedCountry = () => html`
+  <lion-input-iban
+    .modelValue=${'RO89RZBR6997372848645577'}
+    .validators=${[new IsBlacklistedCountryIBAN('RO')]}
+    name="iban"
+    label="IBAN"
+  ></lion-input-iban>
+  <br />
+  <small>Demo instructions: you can use RO 89 RZBR 6997 3728 4864 5577</small>
+`;
+```
--- a/packages/input-iban/README.md
+++ b/packages/input-iban/README.md
@@ -96,13 +96,13 @@ export const countryRestrictions = () => html`
 By default, we validate the input to ensure the IBAN is valid.
 To get the default feedback message for this default validator, use `loadDefaultFeedbackMessages` from `@lion/form-core`.
 
-In the example below, we show how to use an additional validator that restricts the `input-iban` to IBANs from certain countries.
+In the example below, we show how to use an additional validator that blocks IBANs from certain countries.
 
 ```js preview-story
 export const blacklistedCountry = () => html`
   <lion-input-iban
     .modelValue=${'RO89RZBR6997372848645577'}
-    .validators=${([new IsBlacklistedCountryIBAN('RO')], [new IsBlacklistedCountryIBAN('ES')])}
+    .validators=${([new IsNotCountryIBAN('RO')], [new IsNotCountryIBAN('ES')])}
     name="iban"
     label="IBAN"
   ></lion-input-iban>
--- a/packages/input-iban/README.md
+++ b/packages/input-iban/README.md
@@ -102,7 +102,7 @@ In the example below, we show how to use an additional validator that restricts 
 export const blacklistedCountry = () => html`
   <lion-input-iban
     .modelValue=${'RO89RZBR6997372848645577'}
-    .validators=${[new IsBlacklistedCountryIBAN('RO')]}
+    .validators=${([new IsBlacklistedCountryIBAN('RO')], [new IsBlacklistedCountryIBAN('ES')])}
     name="iban"
     label="IBAN"
   ></lion-input-iban>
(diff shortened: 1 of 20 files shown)


gerjanvangeest:
Hi @andrei1902,

Thank you for the new feature, but i have some questions.

1. A blacklisted validator would mean in my opinion that IBANs from that country should not be used, but the validation feedback message says I should use (in this case) a Romanian IBAN.
2. Why not use an array to define multiple countries at once (could make the default Validation message more complex)?

andrei1902:
> Hi @andrei1902,
> 
> Thank you for the new feature, but i have some questions.
> 
> 1. A blacklisted validator would mean in my opinion that IBANs from that country should not be used, but the validation feedback message says I should use (in this case) a Romanian IBAN.
> 2. Why not use an array to define multiple countries at once (could make the default Validation message more complex)?

1. I added translations for this particular validator to keep it in step with the IsCountryIBAN validator.
2. Correct! The usage of an array of countries would complicate the error message. I updated the README.md to show multiple validator usage.

packages/input-iban/README.md
```diff
@@ -90,3 +90,23 @@ export const countryRestrictions = () => html`
   <small>Demo instructions: you can use NL20 INGB 0001 2345 67</small>
 `;
 ```
+
+### Blacklisted Country
+
+By default, we validate the input to ensure the IBAN is valid.
+To get the default feedback message for this default validator, use `loadDefaultFeedbackMessages` from `@lion/form-core`.
+
+In the example below, we show how to use an additional validator that restricts the `input-iban` to IBANs from certain countries.
```
jorenbroekema, line 99:
```suggestion
In the example below, we show how to use an additional validator that blocks IBANs from certain countries.
```

jorenbroekema (changes requested):
I would suggest changing the name from IsBlacklistedCountryIBAN to `IsNotCountryIBAN` to make it consistent with `IsCountryIBAN` and to make clear that it is the opposite behaviour. Other than that, I think it looks good 👍 .

We can look some other time into making both of them accept an array perhaps, but as you mentioned it will complicate error messages, so might be a tricky improvement.

andrei1902:
I pushed the changes you suggested.

packages/input-iban/README.md
```diff
@@ -90,3 +90,23 @@ export const countryRestrictions = () => html`
...
+```js preview-story
+export const blacklistedCountry = () => html`
+  <lion-input-iban
+    .modelValue=${'RO89RZBR6997372848645577'}
+    .validators=${([new IsNotCountryIBAN('RO')], [new IsNotCountryIBAN('ES')])}
+    name="iban"
+    label="IBAN"
+  ></lion-input-iban>
+  <br />
+  <small>Demo instructions: you can use RO 89 RZBR 6997 3728 4864 5577</small>
```
jorenbroekema, line 110: Maybe a little confusing :P since RO is blacklisted, but you want to show off this validation by putting an invalid country.

```suggestion
  <small>Demo instructions: Try <code>RO 89 RZBR 6997 3728 4864 5577</code> and watch it fail</small>
```

packages/input-iban/src/validators.js
```diff
@@ -153,3 +153,39 @@ export class IsCountryIBAN extends IsIBAN {
...
+   * @param {string} [data.fieldName]
+   * @param {*} [data.params]
+   * @param {string} [data.type]
+   * @param {Object.<string,?>} [data.config]
+   * @param {string} [data.name]
+   * @returns {Promise<string|Node>}
+   */
+  static async getMessage(data) {
+    await loadTranslations();
+    return localize.msg('lion-validate+iban:error.IsBlaklistedCountryIBAN', data);
```
jorenbroekema, line 189: I would also change the name of the validation messages, makes it easier to find if it's the same name as the validator. 
```suggestion
    return localize.msg('lion-validate+iban:error.IsNotCountryIBAN', data);
```

packages/input-iban/translations/bg.js
```diff
@@ -16,5 +16,20 @@ export default {
       'RO {Румънски}\n' +
       'other {{params}}\n' +
       '} {fieldName}.',
+    IsBlaklistedCountryIBAN:
```
jorenbroekema, line 19: Same comment as above

packages/input-iban/test/validators.test.js
```diff
@@ -19,4 +19,14 @@ describe('IBAN validation', () => {
...
   });
+
+  it('provides IsNotCountryIBAN to prevent IBANs from specific countries', () => {
+    const nlValidator = new IsNotCountryIBAN('NL');
+    const deValidator = new IsNotCountryIBAN('DE');
+    expect(nlValidator.execute('NL17INGB0002822608')).to.be.true;
+    expect(deValidator.execute('DE89370400440532013000')).to.be.true;
+    expect(nlValidator.execute('DE89370400440532013000')).to.be.false;
+    expect(deValidator.execute('NL17INGB0002822608')).to.be.false;
+    expect(nlValidator.execute('foo')).to.be.false;
```
jorenbroekema, line 30: I think test is failing on this line. Which is expected. The IsNotCountryIBAN does not fail on "foo", but the IsIBAN validator that is applied by default on input-iban, does fail. I think this line can be removed, since it's not a test for the IsNotCountryIBAN, but more of a input-iban/IsIBAN specific test.

jorenbroekema:
Hey @andrei1902  I added some final comments, just small things. After that, you just need to sign the CLA and then I'm fine to merge it 👍  thanks for your work so far on contributing this validator!

jorenbroekema:
Closing in favor of https://github.com/ing-bank/lion/pull/1120 , made some general improvements to these validators there and added array support for this one, and changed the translation key as suggested to IsNotCountryIBAN.

PR #1072, opened 2020-11-02
chore: add ArrowMixin improvements, add Datepicker local/global switch

daKmoR:
## What I did

- ArrowMixin needs to extend styles and not overwrite them
- ArrowMixin add an `_arrowNodeTemplate` which can be used to only override the arrow content
- InputDatepicker switch between bottomsheet style (on mobile) popover (on desktop/table with screensize bigger then 600px)


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -280,6 +290,18 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
     `;
   }
 
+  _setupOverlayCtrl() {
+    super._setupOverlayCtrl();
+
+    this._overlayCtrl.addEventListener('before-show', () => {
+      if (window.innerWidth >= 600) {
+        this._overlayCtrl.updateConfig(this._defineOverlayConfig());
```
ruphin, line 298: I feel like the configuration of the overlay is too spread out across different parts. I would add the `window.innerWidth` switch to `_defineOverlayConfig()`:

```js
_defineOverlayConfig() {
  if (window.innerWidth >= 600) {
    return {
      ...withModalDialogConfig(),
      hidesOnOutsideClick: true,
      ...super._defineOverlayConfig(),
      popperConfig: {
        ...super._defineOverlayConfig().popperConfig,
        placement: 'bottom',
      },
    };
  } else {
    return withBottomSheetConfig();
  }
}
```

and

```js
this._overlayCtrl.addEventListener('before-show', () => {
  this._overlayCtrl.updateConfig(this._defineOverlayConfig());
}
```

daKmoR: changed

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -280,17 +290,36 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
     `;
   }
 
+  _setupOverlayCtrl() {
+    super._setupOverlayCtrl();
+
+    this.__datepickerBeforeShow = () => {
+      this._overlayCtrl.updateConfig(this._defineOverlayConfig());
```
tlouisse, line 297: Is it needed to always redefine the config?

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -280,17 +290,36 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
...
    * @desc overrides default configuration options for this component
    * @returns {Object}
    */
   // eslint-disable-next-line class-methods-use-this
   _defineOverlayConfig() {
-    return {
-      ...withModalDialogConfig(),
-      hidesOnOutsideClick: true,
-    };
+    if (window.innerWidth >= 600) {
```
tlouisse, line 309: Should we define a hardcoded breakpoint value on the lion layer?

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -280,17 +290,36 @@ export class LionInputDatepicker extends ScopedElementsMixin(OverlayMixin(LionIn
...
+      this.hasArrow = true;
+      return {
+        ...withModalDialogConfig(),
+        hidesOnOutsideClick: true,
+        ...super._defineOverlayConfig(),
```
tlouisse, line 314: This adds placementMode 'local'? https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayMixin.js#L94

Maybe we can make it more explicit by adding `placementMode: 'local'` instead?

daKmoR: this actually adds a lot more from the `ArrowMixin`... for placementMode alone yes copying would be better

packages/input-datepicker/src/LionInputDatepicker.js
daKmoR: I'm not sure if checking a specific overlayCtrl config option would be "better"...

imho setting it every time is good enough for now

packages/input-datepicker/src/LionInputDatepicker.js
daKmoR: probable good enough for now... if someone needs it configurable then a feature request with some discussion will need to happen... (as we currently have no central breakpoints defined anywhere in lion)

and that would definitely be a separate pull request

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: Ah, indeed 👍

packages/input-datepicker/src/LionInputDatepicker.js
tlouisse: Yeah, we do it once per switch on our extension layer.
Let's make a small improvement issue 👍

tlouisse (approved):

PR #1078, opened 2020-11-04
fix(form-core): allow Validators from different form-core versions

tlouisse:
closes https://github.com/ing-bank/lion/issues/1079


tlouisse:
After an internal discussion, we decided it's better to enforce deduping on the developer side, so there should always be on version of form-core (better for perf, less error prone). Will close this

PR #1083, opened 2020-11-05
Fix/validation feedback type

jorenbroekema:
## What I did

1. Fix validation feedback _messageTemplate types
2. Use @lion/core for lit type imports
3. revert back to yarn because npm 7 is unstable (deduping not working properly for monorepo 😢 )


gerjanvangeest (approved):

gerjanvangeest (approved):

package.json
```diff
@@ -77,7 +77,7 @@
     "rimraf": "^2.6.3",
     "rollup": "^2.0.0",
     "sinon": "^7.2.2",
-    "typescript": "^3.8.3",
+    "typescript": "3.9.7",
```
gerjanvangeest, line 80: We now fix the version?

jorenbroekema: Yeah I think that would be a smart thing, since any Typescript minor or even patch can be breaking, due to the nature of Typescript where every change is essentially always breaking

ruphin (approved):
lgtm

PR #1090, opened 2020-11-11
Localize post processors

marcoswebdev:
## What I did

From the proposal from @tlouisse I have added the functionality to allow **formatDate** and **formatNumber** post-processing.

### A neat fix...

Apparently, we want to be able to provide `formatDate` settings that only apply to the ing layer.
In this case, we should allow extra options like a 'post-process function' to `formatDate` in the Lion layer.

We then would have to do:

#### Lion

_formatDate_ (https://github.com/ing-bank/lion/blob/master/packages/localize/src/date/formatDate.js)

```javascript

export function formatDate(date, options) {
   ...
   if (typeof options.postProcessor === 'function') {
     return options.postProcessor(formattedDate);
   }
   return formattedDate;
}
```
_localizeManager_ (https://github.com/ing-bank/lion/blob/master/packages/localize/src/LocalizeManager.js)

```javascript

constructor() {
  ...
  this.__datePostprocessors = [];
}

addDatePostprocessors(processors) {
  this.__datePostprocessors = [...this.__datePostprocessors, processors]
}
```
https://github.com/ing-bank/lion/blob/master/packages/validate-messages/translations/cs.js (for all locales)

```javascript
IsDate: ' Zadejte datum {formattedDateParam}.'
```
The `formattedDateParam` above would be provided by the validator, that calls `formatDate(date, {...options, postProcessors: locale.datePostprocessors})`.
This code change would change all occurrences of IntlMessageFormat date functionality (`{params, date, YYYYMMDD}`) as well, so we can move to IntlMessageFormatLite.

### DIFF
--- a/packages/localize/src/date/formatDate.js
+++ b/packages/localize/src/date/formatDate.js
@@ -1,5 +1,8 @@
+import { localize } from '../localize.js';
 import { getLocale } from './getLocale.js';
 import { normalizeIntlDate } from './normalizeIntlDate.js';
+
+/** @typedef {import('../../types/LocalizeMixinTypes').DatePostProcessor} DatePostProcessor */
 
 /**
  * Formats date based on locale and options
@@ -33,5 +36,22 @@ export function formatDate(date, options) {
   } catch (e) {
     formattedDate = '';
   }
+
+  if (localize.formatDateOptions.postProcessors.size > 0) {
+    Array.from(localize.formatDateOptions.postProcessors).forEach(([locale, fn]) => {
+      if (locale === computedLocale) {
+        formattedDate = fn(formattedDate);
+      }
+    });
+  }
+
+  if (formatOptions.postProcessors && formatOptions.postProcessors.size > 0) {
+    Array.from(formatOptions.postProcessors).forEach(([locale, fn]) => {
+      if (locale === computedLocale) {
+        formattedDate = fn(formattedDate);
+      }
+    });
+  }
+
   return normalizeIntlDate(formattedDate, computedLocale, formatOptions);
 }
--- a/packages/localize/test/date/formatDate.test.js
+++ b/packages/localize/test/date/formatDate.test.js
@@ -232,4 +232,104 @@ describe('formatDate', () => {
     // @ts-ignore tests what happens if you use a wrong type
     expect(formatDate(date)).to.equal('');
   });
+
+  describe('Date post processors', () => {
+    /**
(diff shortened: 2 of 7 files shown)


packages/localize/test/date/formatDate.test.js
```diff
@@ -232,4 +232,43 @@ describe('formatDate', () => {
...
+    expect(formatDate(testDate, options)).to.equal('Monday, 21 May 2012');
+    localize.locale = 'nl-NL';
+    expect(formatDate(testDate, options)).to.equal('maandag 21 mei 2012');
+    localize.locale = 'fr-FR';
+    expect(formatDate(testDate, options)).to.equal('lundi 21 mai 2012');
+    localize.locale = 'de-DE';
+    expect(formatDate(testDate, options)).to.equal('MONTAG, 21. Mai 2012');
+    localize.locale = 'en-US';
+    expect(formatDate(testDate, options)).to.equal('MONDAY, May 21, 2012');
+  });
```
tlouisse, line 273: Will there be a test that handles `__datePostProcessors ` as well? (when options.postProcessor is not provided)?

Is it allowed to add multiple post processors? In what order will they be executed and what happens when you provide both a global and a local one?


I know it's still a draft, so maybe you were already planning to implement this :)

marcoswebdev: I added more tests for the global and local processors 😄 

The strategy followed is that we process first global processors in the `options` property and after that, we launch the localize processors. What do you think?

marcoswebdev:
`formatNumber` post processors added 👍 

I also aligned the logic between `formatNumber` and `formatDate` post-processing 😄

packages/localize/test/date/formatDate.test.js
tlouisse: Thanks. Looks great to me 👍

tlouisse (approved):

packages/localize/src/date/formatDate.js
```diff
@@ -33,5 +36,22 @@ export function formatDate(date, options) {
   } catch (e) {
     formattedDate = '';
   }
+
+  if (localize.formatDateOptions.postProcessors.size > 0) {
+    Array.from(localize.formatDateOptions.postProcessors).forEach(([locale, fn]) => {
```
tlouisse, line 41: small perf improvement: I could just have done `localize.formatDateOptions.postProcessors.get(computedLocale)`

PR #1102, opened 2020-11-19
switch non projected contentNodes

tlouisse:
wip...


tlouisse:
Closing this, this should be solved together with everything mentioned here: https://github.com/ing-bank/lion/discussions/1118

PR #1106, opened 2020-11-30
feat: upgrade to popper 2

jorenbroekema:
Right now, the tooltip demo that I changed ('has-arrow' attribute added), looks good, for at least placement top and placement right. So basically POC is done.

TODO:

- [x] Check whether modifiers that have the same name get properly deduped by popper. If so, we can just stack them and don't have to do any merging/deduping ourselves. If not, we may have to create a mergeModifiers helper or something.. now that modifiers is an array (not uniquely-keyed items)
- [x] I managed to get quite far with fixing types, but there's still some errors. This will be some work to get it all properly..
- [x] Update Mixin type definition files where needed (ArrowMixin, OverlayMixin)
- [x] Update lion packages that use/extend popperConfig
- [x] Make all tests succeed again
- [x] Update demos (tooltip, overlays etc.)
- [x] Migration guidelines for our users (link to popper migration guidelines probably is enough, since we mirror their API)
- [x] Create popper config compatibility layer in our extension library and test it
- [x] Create changeset release


packages/helpers/convertPopperConfig/src/convertPopperConfig.js
```diff
@@ -0,0 +1,64 @@
...
+            .split(',')
+            .map((/** @type {string} */ c) => Number(c.replace('px', '')));
+        }
+        return result;
+      });
+      p2Cfg[option] = modifiers;
+    } else if (option === 'positionFixed') {
+      p2Cfg.strategy = p1Cfg.positionFixed === true ? 'fixed' : 'absolute';
+    } else if (option === 'scheduleUpdate') {
+      // @ts-expect-error
```
jorenbroekema, line 55: Maybe add a small explanation. I assume this is because of the return value`p1Cfg[option]()` not being typed because it uses array accessor to get the method which cannot be typed?

jorenbroekema: Ah I see now, it is because in Popper 1 you used to be able to hook into every update by hooking into scheduleUpdate, but in Popper 2 you can only hook into onFirstUpdate. The recommendation if you need to hook into every update instead of only the first time, is to make your own custom modifier https://popper.js.org/docs/v2/modifiers/#custom-modifiers and use the `fn` callback.

So I think we should remove this part, since adding an update property doesn't anything, to migrate this from popper 1 you will need to make your own modifier, and we cannot "auto convert" that for our users.

tlouisse: Could very well be. I will not have a deeper look till next Monday, but feel free to adjust were needed :)

package.json
```diff
@@ -73,6 +73,7 @@
     "mocha": "^7.1.1",
     "npm-run-all": "^4.1.5",
     "playwright": "^1.2.1",
+    "playwright-chromium": "^1.8.0-next.1608058261773",
```
daKmoR, line 76: why is that needed? 🤔

packages/overlays/src/OverlayController.js
```diff
@@ -7,20 +7,50 @@ import { containFocus } from './utils/contain-focus.js';
...
-  // @ts-ignore
-  return /** @type {Promise<PopperModule>} */ (import('popper.js/dist/esm/popper.min.js'));
+  // @ts-ignore import complains about untyped module, but we typecast it ourselves
+  return /** @type {Promise<PopperModule>} */ (import('@popperjs/core/dist/esm/popper.js'));
 }
 
+// /**
+//  * @returns {Promise<PopperModule>}
+//  */
+// async function preloadPopper() {
```
daKmoR, line 28: can we delete this comment block? 🤔

package.json
jorenbroekema: Hmm we should remove that again, it's only needed if you're on mac big sur

packages/overlays/src/OverlayController.js
jorenbroekema: @tlouisse  what was the purpose of this thing? we need it?

packages/input-amount/test/lion-input-amount.test.js
```diff
@@ -156,7 +156,9 @@ describe('<lion-input-amount>', () => {
       expect(el._currencyDisplayNode?.getAttribute('aria-label')).to.equal('US dollars');
       el.currency = 'PHP';
       await el.updateComplete;
-      expect(el._currencyDisplayNode?.getAttribute('aria-label')).to.equal('Philippine pisos');
+      // TODO: Chrome Intl now thinks this should be pesos instead of pisos. They're probably right.
+      // We could add this to our normalize layer so other browsers also do it correctly?
+      // expect(el._currencyDisplayNode?.getAttribute('aria-label')).to.equal('Philippine pisos');
```
tlouisse, line 161: Indeed definitely make an issue to normalize output for all browsers.
I think Chrome is wrong: https://britsinthephilippines.top/philippines-piso-or-philippines-peso/

packages/localize/test/date/getMonthNames.test.js
```diff
@@ -22,9 +22,10 @@ describe('getMonthNames', () => {
     );
   });
   it('supports "short" style', () => {
-    expect(getMonthNames({ locale: 'en-GB', style: 'short' })).to.deep.equal(
-      s`Jan Feb Mar Apr May Jun Jul Aug Sep Oct Nov Dec`,
-    );
+    // TODO: Chrome thinks it should be Sept, not Sep. Firefox/Webkit disagree. We could normalize it in lion.
```
tlouisse, line 25: Same here. 4 chars doesn't make sense :$ Seems like a buggy Chrome version...

tlouisse: Created an issue: https://github.com/ing-bank/lion/issues/1166

packages/overlays/src/OverlayController.js
tlouisse: No, I pushed a wip state. I tried to lazy load popper-lite including just the modifiers we need. However, the arrow modifier didn't seem to work properly with the code below.

So we can remove it

tlouisse (approved):

PR #1110, opened 2020-12-01
fix: point to parent constructor in static get styles (points to undefined atm)

tlouisse:
## What I did

1.

### DIFF
--- a/packages/overlays/src/ArrowMixin.js
+++ b/packages/overlays/src/ArrowMixin.js
@@ -24,7 +24,8 @@ export const ArrowMixinImplementation = superclass =>
     }
 
     static get styles() {
-      const superCtor = /** @type {typeof import('@lion/core').LitElement} */ (super.constructor);
+      const superCtor = /** @type {typeof import('@lion/core').LitElement} */ (super.prototype
+        .constructor);
       return [
         superCtor.styles ? superCtor.styles : [],
         css`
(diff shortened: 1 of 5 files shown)


gerjanvangeest (approved):

packages/overlays/src/ArrowMixin.js
```diff
@@ -24,7 +24,7 @@ export const ArrowMixinImplementation = superclass =>
     }
 
     static get styles() {
-      const superCtor = /** @type {typeof import('@lion/core').LitElement} */ (super.constructor);
+      const superCtor = /** @type {typeof import('@lion/core').LitElement} */ (this.constructor);
```
jorenbroekema, line 27:
```suggestion
      const ctor = /** @type {typeof import('@lion/core').LitElement} */ (this.constructor);
```

tlouisse: Hmm, this is not the right fix at all. Thanks for noticing, it should have been just `super`, because you are already in 'constructor context'. Still waking up atm :)

jorenbroekema:
Can you explain the difference between `super.constructor.foo` and `super.prototype.constructor.foo`?

tlouisse:
> Can you explain the difference between `super.constructor.foo` and `super.prototype.constructor.foo`?

If you're in a static method, `super` already points to the constructor of the instance

PR #1114, opened 2020-12-02
fix(listbox): using selection-follows-focus with horizontal orientation

### LINKED ISSUE #1105: [listbox] selection-follows-focus check disabled options
## Expected behavior

When using listbox with `selection-follows-focus`, disabled options are checked

## Actual Behavior

https://webcomponents.dev/edit/hCMK42alriu66I3DtNsA

## Additional context

I suppose disabled options should only have the hover state in this case.

Aria for disabled controls https://www.w3.org/TR/wai-aria-practices/#kbd_disabled_controls

MathieuPuech:
## What I did

1. Fix keyboard navigation when `selection-follows-focus` and `orientation="horizontal"` are set on a `<lion-listbox>`
1. Fix keyboard navigation with `selection-follows-focus` and disabled options
1. On click of an option, it become active

Closes #1105

### DIFF
--- a/.changeset/popular-needles-hug.md
+++ b/.changeset/popular-needles-hug.md
@@ -1,0 +1,7 @@
+---
+'@lion/listbox': patch
+---
+
+- Fix keyboard navigation when `selection-follows-focus` and `orientation="horizontal"` are set on a `<lion-listbox>`
+- Fix keyboard navigation with `selection-follows-focus` and disabled options
+- On click of an option, it become active
--- a/packages/listbox/README.md
+++ b/packages/listbox/README.md
@@ -131,7 +131,7 @@ See [wai aria spec](https://www.w3.org/TR/wai-aria-practices/#kbd_selection_foll
 export const selectionFollowsFocus = () => html`
   <lion-listbox name="combo" label="Selection follows focus" selection-follows-focus>
     <lion-option .choiceValue=${'Apple'}>Apple</lion-option>
-    <lion-option .choiceValue=${'Artichoke'}>Artichoke</lion-option>
+    <lion-option .choiceValue=${'Artichoke'} disabled>Artichoke</lion-option>
     <lion-option .choiceValue=${'Asparagus'}>Asparagus</lion-option>
     <lion-option .choiceValue=${'Banana'}>Banana</lion-option>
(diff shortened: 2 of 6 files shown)


jorenbroekema (approved):

jorenbroekema:
Very nice! Thanks!

tlouisse:
Nice job 👍 

@MathieuPuech @jorenbroekema I know it's already merged, but: 
- did we align arrow up/down behavior with orientation (so up/down should **not** go to the prev/next for horizontal orientation and left and right should **not** go to the prev/next for vertical orientation) ?
- if a disabled option is focused via arrow keys (and selectionFollowsFocus) is true, should the previously selected stay selected or should everything be unselected? 
  - What does the platform do (a `<select>` on Windows)?
  - Does it work correctly when multipleChoice is true?

jorenbroekema:
First question: yes

Second question: 
- It's a good point you bring up, in my opinion it should unselect everything but that's not really backed by any existing UI element or recommendation. More an intuitive thing.
  - Not sure, haven't tried
  - How would selectionFollowsFocus work for multiple choice? As you navigate down the list it selects everything on its path :P? My hypothesis is that once you cross the disabled option, it indeed unselects everything so that's definitely not what you want there. May want to put a guard for multiple choice for that.

MathieuPuech:
yes, it's not clear in the code but you can see on lines before that orientation is considered.

For disabled, We could keep the previous selected. While developing, it was like that and I thought it was better not but there is arguments for both behaviors.

The native select don't focus on disabled option but aria rules say [disabled should be focusable](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls) but I didn't see anything for it when the selection follows the focus.

The multiple is working with all of this, the selection-follows-focus is not possible with multiple (it was already like that)

tlouisse:
- Yes indeed, it just sets selectedIndex to the (unchanged) activeIndex, so it already worked  👍 
- Wrt unselecting or keeping the prev checked, I was just wondering whether there were specs or a native example, but I agree the current one is the best choice.
- Good point about multiple, I forgot for a moment selectionFollowsFocus didn't make sense there in the first place 

Alright, everything seems fine then :)

PR #1115, opened 2020-12-02
Fix/overlays focus

jorenbroekema:
## What I did

1. Fix an issue with events no longer working inside the contentNode of local overlays due to body inert
2. Fix an issue with deepContains not properly handling multiple nested shadow roots, and fix a recursion problem where earlier loops would find a target, but later loops would not, and then containsTarget bool would end up as "false" even though we did find a target

### DIFF
--- a/packages/overlays/src/utils/deep-contains.js
+++ b/packages/overlays/src/utils/deep-contains.js
@@ -10,15 +10,13 @@ export function deepContains(el, targetEl) {
     return true;
   }
 
-  /** @param {HTMLElement} elem */
+  /** @param {HTMLElement|ShadowRoot} elem */
   function checkChildren(elem) {
     for (let i = 0; i < elem.children.length; i += 1) {
       const child = /** @type {HTMLElement}  */ (elem.children[i]);
-      if (child.shadowRoot) {
-        containsTarget = deepContains(child.shadowRoot, targetEl);
-        if (containsTarget) {
-          break;
-        }
+      if (child.shadowRoot && deepContains(child.shadowRoot, targetEl)) {
+        containsTarget = true;
+        break;
       }
       if (child.children.length > 0) {
         checkChildren(child);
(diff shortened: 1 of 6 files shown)


packages/overlays/src/utils/deep-contains.js
```diff
@@ -34,7 +32,7 @@ export function deepContains(el, targetEl) {
         return true;
       }
     }
-    checkChildren(el);
   }
+  checkChildren(el);
```
tlouisse, line 36: Difference is very subtle it seems. So basically we stopped traversal for `Element`s that are not `HTMLElement `?

Maybe we can combine the ifs on line 30 and 31?

packages/overlays/test/utils-tests/deep-contains.test.js
```diff
@@ -63,4 +63,42 @@ describe('deepContains()', () => {
...
+    expect(deepContains(element, element.firstElementChild.shadowRoot)).to.be.true;
+    expect(deepContains(element, element.firstElementChild.shadowRoot.children[1])).to.be.true;
+    expect(deepContains(element, element.firstElementChild.shadowRoot.children[1].shadowRoot)).to.be
+      .true;
+    expect(
+      deepContains(
+        element,
+        element.firstElementChild.shadowRoot.children[1].shadowRoot.lastElementChild,
+      ),
+    ).to.be.true;
```
tlouisse, line 102: Good test. Maybe also add a `to.be.false` to make sure we didn't implement `deepContains` as `() => true` ? :)

tlouisse (approved):

packages/overlays/src/utils/deep-contains.js
jorenbroekema: Yes it stopped traversal for the ones that were ShadowRoot, and of course those should be traversed as well :D. If's can now be combined indeed, changed it 👍

packages/overlays/test/utils-tests/deep-contains.test.js
jorenbroekema: I added two tests to assert that deepContains returns false. One for siblings, one for when two nodes are just not related in any way (e.g. a sibling and its siblings child node, I guess a cousin node xD?)

packages/overlays/src/utils/deep-contains.js
tlouisse: Yeah, I just figured it out indeed :) Ok, nice

packages/overlays/test/utils-tests/deep-contains.test.js
tlouisse: Haha, a cousin node I would say indeed. Those tests should be enough, so ready to merge I would say.

PR #1120, opened 2020-12-03
feat(input-iban): add IsNotCountryIBAN validator

jorenbroekema:


packages/input-iban/translations/hu.js
```diff
@@ -16,5 +16,20 @@ export default {
...
+      'DE {Német}\n' +
+      'ES {Spanyol}\n' +
+      'FR {Francia}\n' +
+      'HU {Magyar}\n' +
+      'IT {Olasz}\n' +
+      'NL {Holland}\n' +
+      'PL {Lengyel}\n' +
+      'RO {Román}\n' +
+      'other {{params}}\n' +
+      '} {fieldName} nem engedélyezett',
```
gerjanvangeest, line 33: Its a sentence, so should end with a `.`. (also for all following languages)

packages/input-iban/translations/sk.js
```diff
@@ -16,5 +16,20 @@ export default {
       'RO {rumunský}\n' +
       'other {{params}}\n' +
       '} kód {fieldName}.',
+    IsNotCountryIBAN:
+      '{params, select,\n' +
+      'AT {rakúsky}\n' +
```
gerjanvangeest, line 21: Start of sentence, so with a capital?

jorenbroekema: Hmm yeah in this case we should indeed capitalise, good catch.

jorenbroekema: Pushed as separate commit, I think I did it correctly even with cyrillic haha

packages/input-iban/translations/hu.js
jorenbroekema: Done

packages/input-iban/README.md
```diff
@@ -90,3 +90,25 @@ export const countryRestrictions = () => html`
...
+By default, we validate the input to ensure the IBAN is valid.
+To get the default feedback message for this default validator, use `loadDefaultFeedbackMessages` from `@lion/form-core`.
+
+In the example below, we show how to use an additional validator that blocks IBANs from certain countries.
+
+```js preview-story
+export const blacklistedCountry = () => html`
+  <lion-input-iban
+    .modelValue=${'RO89RZBR6997372848645577'}
+    .validators=${([new IsNotCountryIBAN('RO')], [new IsNotCountryIBAN('ES')])}
```
tlouisse, line 105: Isn't this a bit against how validators work atm?

The system was designed to have one instance of a certain validator per FormControl.
per validator type we can check whether it's either valid or invalid for this Validator.
The 2nd instance in the example above will probably cancel out the first one, leading to:
`validationStates.error.IsNotCountryIban => false`

(=> action point, check for this during setup and when new Validators are added)

Maybe we can do smth like `new isNonBlackListedIban(['RO', 'ES'])` ?

jorenbroekema: I would be down with keeping the current name (IsNotCountryIBAN) and passing in an array, but this significantly complicates the translation message which is why I was okay with having multiple instances instead

tlouisse (changes requested):
- I would change name to `isNonBlackListedIban`
- Adding multiple instances of one Validator to a FormControl should not be allowed

packages/input-iban/README.md
tlouisse: Yeah, I understand. For that to work we would need to allow for the validator to return metadata instead of just a Boolean, maybe.
About the name, it was not completely intuitive to me personallt, but it best aligns with the current `IsCountryIBAN `
indeed.

The validationStates after a better check would actually be marked as error, because it only sets it to true for everything that ends up being invalid (just one instance of `IsNotCountryIBAN`).

Alright, I can approve it then :)

tlouisse (approved):

PR #1121, opened 2020-12-04
fix(overlays): sync config change to controller

### LINKED ISSUE #1075: [Overlays] - Config prop change makes content disappear and desync from opened
## Expected behavior

Either of these would fix my problem:
- If the `config` changes while the component is `opened`, the content says visible
- An update of the OverlayMixin's `config` property with an equal value gets ignored

## Actual Behavior

Updating the config with any value results in the shown overlay content to disappear, even while opened stays true.

### Reproduction

- Open the [Lion Dialog Storybook entry](https://lion-web-components.netlify.app/iframe.html?id=overlays-dialog--main) inline.
- Copy these snippets into the Dev Tools console

```js
const dialog = document.querySelector('lion-dialog');
dialog.opened = true; // The dialog appears, as expected
```

...then add

```js
dialog.config = { ...dialog.config }; // Unfortunately, the content disappears, yet the config is essentially the same
dialog.opened // It's still true, as expected
```

## Additional context

I'm using the component from a Vue app. In the template, I provide a `config` object. (`:config="{ hidesOnEsc: false }"` because I want to handle esc hiding via Vuex) Now an arbitrary update in the content slot causes Vue to re-render the template, apparently also swapping out the `config` prop with an equal, but new reference. The result is that the content disappears, and the dialog won't come back, as the `opened` prop stays `true`.

### Workarounds
While neither is ideal, but:
- when consuming with Vue, don't provide the `config` prop, instead subclass the component with the desired config, assuming it's static
- use a lifecycle hook (`mounted` with Vue) to imperatively set the config, then leave it alone

riovir:
## What I did

Fixes #1075. Since a `config` reference change in the `OverlayMixin` resets the DOM, essentially hiding the opened content...

1. Prevented the teardown and init from happening caused by calling the `_controller.updateConfig()` if the new config is equivalent to the original. Considering the bundle size I've opted for the manual implementation of the `isEqualConfig` over depending on / inlining a deep equal check from `ramda` or `lodash`.
2. Triggered `__syncToOverlayController()` if the config changed intending to re-trigger the necessary DOM updates.

### Other approaches considered

Deeper, on the `OverlayController` level diffing the exact nature of the config change, and doing a minimal DOM change instead of a complete teardown and init. Didn't have the confidence to pull off such an invasive change.

It's also possible to revert the `__syncToOverlayController()` block, only keeping the first change that disregards equivalent changes. While not solving a true config update, it would fix the initial problem: _render functions replacing the config with equivalent values`_. (Reproduced in Vue, but also in LitElement examples.)

### DIFF
--- a/.changeset/dirty-students-smile.md
+++ b/.changeset/dirty-students-smile.md
@@ -1,0 +1,5 @@
+---
+'@lion/overlays': patch
+---
+
+Sync config change to the controller and make sure that an overlay is re-opened/re-hidden after the updateConfig handling, instead of always closed.
--- a/.changeset/purple-beers-applaud.md
+++ b/.changeset/purple-beers-applaud.md
@@ -1,0 +1,5 @@
+---
+'@lion/overlays': patch
+---
+
+Return promises for the OverlayMixin toggle, open and close methods, since the OverlayController hiding/showing is asynchronous.
--- a/packages/input-datepicker/test-helpers/DatepickerInputObject.js
+++ b/packages/input-datepicker/test-helpers/DatepickerInputObject.js
@@ -33,7 +33,7 @@ export class DatepickerInputObject {
    * @param {number} day
    */
   async selectMonthDay(day) {
-    this.overlayController.show();
+    await this.overlayController.show();
     await this.calendarEl.updateComplete;
     this.calendarObj.getDayEl(day).click();
     return true;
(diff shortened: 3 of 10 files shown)


jorenbroekema (approved):
Hi, thanks again for looking into this, very nice contribution!

I'm going to go ahead and add a changeset so that it will release with a changelog for this change. 
I will also do a small edit in the test where you are putting an error throw on the updateConfig method, it's easier to use a sinon stub here and restore it after wards, so that it doesn't affect future tests.

I agree with you that the other approach, which is a more "smart" teardown based on the config changes, is quite a bit harder. You would have to do diffing on the configs to see which props changed, and only teardown those features that changed. Since we don't do this now, the only concern I have is that for global overlays the DOM teardowns may cause state inside contentNode to be lost because of it being moved in the DOM with appendChild, but I could very well be wrong about that so I'll wait for that potential issue to show itself, and handle it when it does :)

riovir:
Thanks for looking into the quirks and patching up my code!

PR #1130, opened 2020-12-09
chore: move tools to tools folder

jorenbroekema:
## What I did

1. Move providence, singleton-manager, babel plugin and remark extend to tools folder (instead of packages)
2. Add tools folder to our workspaces
3. Edit our dev scripts to also check inside tools folder (lint-versions, workspaces-scripts, .storybook/main.js etc.)
4. Verify all tests are being ran for these 4 tools, both node and browser
5. Verify linting/types is still working for these 4 tools
6. Move singleton manager from "Other" to "Tools" in storybook


.gitignore
```diff
@@ -24,6 +24,9 @@ yarn-error.log
 *.d.ts
 !packages/**/*/types/**/*
 !packages/**/index.d.ts
+!tools/**/*/types/**/*
+!tools/**/index.d.ts
+!tools/**/types.d.ts
```
jorenbroekema, line 29: @tlouisse  I saw this file, not sure if it's even used but I didn't wanna throw it away. Can you maybe look into it or put it inside a types folder so that we can remove this line 29 :P because it's only that one file for which I added it.

gerjanvangeest (approved):

.gitignore
tlouisse: Please don't throw it away indeed :) It is used, but Providence is definitele not fully typed.
Feel free to move it around and update references or do you want me to have a look at it?

jorenbroekema: Naah it's okay for now, just as long as we have "some" plan moving forward (type providence, move this file into types folder so we dont need more .<foo>ignore entries)

daKmoR:
imho the singleton manager should remain in packages...

`tools` should be node tools... (so we will not run the browser tests there - so if we move singleton-manger as well then we will miss those tests)...

that way we also don't need to combine them in the `web-test-runner.config.mjs` making the config simpler

daKmoR:
discussed it again and we will name the folder `packages-node` to make it crystal clear

gerjanvangeest (approved):

Issue #756, opened 2020-06-06
Resizing of overlay in select-rich


MathieuPuech:
## Describe the bug

When using keyboard to change the value of a selection in "windows/linux" interaction mode, the value changed making the invoker size changed. The overlay width should be reized too.

## To Reproduce

1. Go to storybook demo https://lion-web-components.netlify.app/?path=/story/forms-select-rich--main
2. expand the select-rich options
3. use keyboard (up and down) to change values

## Expected behavior

When invoker size changed, the width of the overlay should be resized too

## Screenshots

![2020-06-05_20-06-27](https://user-images.githubusercontent.com/9367152/83931148-2dcc7480-a769-11ea-8885-e434e9a25e22.gif)

## Possible Fix

`__syncInvokerElement` of LionSelectRich

```javascript
  __syncInvokerElement() {
    // sync to invoker
    if (this._invokerNode) {
      this._invokerNode.selectedElement = this.formElements[this.checkedIndex];
      await this._invokerNode.updateComplete;
      this._overlayCtrl._handleInheritsReferenceWidth();
    }
  }
```

But this will fix only for select-rich, the issue is maybe same for other components using the local placement overlay

MathieuPuech:
I've seen another small issue on _handleInheritsReferenceWidth, It takes the clientWidth instead of getBoundingClientRect().width. The second width can be a float number (when zooming for example).

You can see the bug when zooming and adding borders on invoker and the first option for example

![chrome_2020-06-05_21-02-40](https://user-images.githubusercontent.com/9367152/83932373-04631700-a770-11ea-87eb-33879fdee74a.png)

But even if you have the good value of width, I don't know how you can apply it to the element.

jorenbroekema:
Hmmm I wonder how much of a bug the first one is. I think if you implement an extension of the rich select, you need to provide some fixed width on your slotted content to get consistent width across items and also inside the invoker. I'm doubting a bit whether we should do something here, or if it should be left up to the user 🤔  and if we do something, how hard/annoying will it be for the user to then override that behaviour?

For the second one with the `clientWidth` vs `getBoundingClientRect().width`, I agree it's a small bug 👍

daKmoR:
when we look at the native select... it's width will always be the width of the longest option... so it's size never changes... (at least on mac)

[...]

PR #1131, opened 2020-12-09
Fix/select rich width

gerjanvangeest:
## What I did

1. fix: https://github.com/ing-bank/lion/issues/756

### DIFF
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -15,6 +15,10 @@ export class LionSelectInvoker extends LionButton {
     return [
       super.styles,
       css`
+        :host {
+          justify-content: space-between;
+        }
+
         #content-wrapper {
           position: relative;
         }
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -112,6 +112,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     this.interactionMode = 'auto';
 
     this.singleOption = false;
+    this._arrowWidth = 28;
 
     this.__onKeyUp = this.__onKeyUp.bind(this);
     this.__invokerOnBlur = this.__invokerOnBlur.bind(this);
@@ -217,6 +218,8 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
       /* eslint-enable no-param-reassign */
       this.__hasInitialSelectedFormElement = true;
     }
+    this._alignInvokerWidth();
+
     this._onFormElementsChanged();
   }
 
@@ -226,6 +229,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
    */
   removeFormElement(child) {
     super.removeFormElement(child);
+    this._alignInvokerWidth();
     this._onFormElementsChanged();
   }
 
@@ -352,6 +356,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
   _setupOverlayCtrl() {
     super._setupOverlayCtrl();
     this._initialInheritsReferenceWidth = this._overlayCtrl.inheritsReferenceWidth;
+    this._alignInvokerWidth();
 
     this._overlayCtrl.addEventListener('before-show', this.__overlayBeforeShow);
     this._overlayCtrl.addEventListener('show', this.__overlayOnShow);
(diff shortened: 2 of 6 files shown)


packages/select-rich/src/LionSelectRich.js
```diff
@@ -369,6 +378,31 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     this._overlayCtrl.removeEventListener('hide', this.__overlayOnHide);
   }
 
+  /**
+   * Align invoker width with content width
+   * Make sure display is not set to "none" while calculating the content width
+   */
+  _alignInvokerWidth() {
+    if (this._overlayCtrl && this._overlayCtrl.content) {
```
gerjanvangeest, line 386: Our internal extension doesn't need this behavior, since it uses a global overlay. Should i restrict this to local overlays?

packages/select-rich/src/LionSelectInvoker.js
```diff
@@ -15,9 +15,17 @@ export class LionSelectInvoker extends LionButton {
...
       css`
+        :host {
+          justify-content: space-between;
+        }
+
         #content-wrapper {
           position: relative;
         }
+
+        :host ::slotted([slot='after']) {
```
jorenbroekema, line 26: For me this doesn't have any effect. Same for align-items on the host. I guess this is because the button (invoker) takes the height of its contents, and the contents happen to be roughly the same height (arrow / font). Maybe it's better if we leave this CSS out, and leave it up to our subclassers to ensure their rich select invoker looks correct?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -47,6 +47,10 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
         reflect: true,
         attribute: 'single-option',
       },
+      arrowWidth: {
+        type: Number,
+        attribute: 'arrow-width',
+      },
```
jorenbroekema, line 53:
```suggestion
      arrowWidth: {
        attribute: false,
      },
```
Since there is no purpose for linking to the attribute as far as I can tell. Unless you think arrow-width should be passed by the app dev declaratively? But it's more of a subclasser thing to take care of

jorenbroekema: Actually why is it a Lit property to begin with? It doesn't seem to respond to imperative changes of the property. Unless the idea is to add an updated hook and call alignInvokerWidth when arrowWidth changes?

packages/select-rich/src/LionSelectRich.js
```diff
@@ -112,6 +116,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     this.interactionMode = 'auto';
 
     this.singleOption = false;
+    this.arrowWidth = 28;
```
jorenbroekema, line 119: The arrow is 15.x px wide, so let's say 16. And then you have to account for a difference of 4px in padding between content and invoker nodes, which amounts to 8px (both sides). I checked, and an arrowWidth of 24px is indeed enough spacing, but 28px definitely looks nicer... 

Seems a little weird, I don't think we should overcomplicate it by calculating padding diffs :P so I think I'm alright with the subclasser just setting a proper arrow width.

Should we make it a protected property though? Since it's not really meant to be used by the app developer, but it should be set by the subclasser?

packages/select-rich/src/LionSelectRich.js
jorenbroekema: Hmmm, but in theory one could create a lion select rich extension using global DOM position but may still want the invoker width to be aligned. Perhaps leave it up to the subclasser to add a guard if they don't want the behaviour.

packages/select-rich/src/LionSelectRich.js
```diff
@@ -369,6 +378,31 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
...
+
+        const contentWidth = this._overlayCtrl.content.getBoundingClientRect().width;
+        this._invokerNode.style.width = `${contentWidth + this.arrowWidth}px`;
+
+        this._overlayCtrl.content.style.display = initContentDisplay;
+        this._overlayCtrl.content.style.minWidth = initContentMinWidth;
+        this._overlayCtrl.content.style.width = initContentWidth;
+      });
+    }
+  }
```
jorenbroekema, line 404:
```suggestion
  async _alignInvokerWidth() {
    if (this._overlayCtrl && this._overlayCtrl.content) {
      await this.updateComplete;
      const initContentDisplay = this._overlayCtrl.content.style.display;
      const initContentMinWidth = this._overlayCtrl.content.style.minWidth;
      const initContentWidth = this._overlayCtrl.content.style.width;

      this._overlayCtrl.content.style.display = '';
      this._overlayCtrl.content.style.minWidth = 'auto';
      this._overlayCtrl.content.style.width = 'auto';

      const contentWidth = this._overlayCtrl.content.getBoundingClientRect().width;
      this._invokerNode.style.width = `${contentWidth + this.arrowWidth}px`;

      this._overlayCtrl.content.style.display = initContentDisplay;
      this._overlayCtrl.content.style.minWidth = initContentMinWidth;
      this._overlayCtrl.content.style.width = initContentWidth;
    }
  }
```

That works for me in the demo, and it's much faster than doing a setTimeout, because promises are queued as microtasks so they happen faster. And it also means that if updateComplete is already resolved, no time is lost.

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -189,6 +189,36 @@ describe('lion-select-rich', () => {
...
+      const arrowWidth = 30;
+      const el = /** @type {LionSelectRich} */ (await fixture(html`
+        <lion-select-rich arrow-width=${arrowWidth}>
+          <lion-option .choiceValue=${10}>Item 1</lion-option>
+          <lion-option .choiceValue=${10}>Item 2 with long label</lion-option>
+        </lion-select-rich>
+      `));
+      el.opened = true;
+      const options = el.formElements;
+      await aTimeout(0);
```
jorenbroekema, line 203: With my suggestion above, this can be changed to `await el.updateComplete;`

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -189,6 +189,36 @@ describe('lion-select-rich', () => {
...
+      expect(el._invokerNode.clientWidth).to.equal(options[1].clientWidth + arrowWidth);
+
+      const newOption = /** @type {LionOption} */ (document.createElement('lion-option'));
+      newOption.choiceValue = 30;
+      newOption.textContent = '30 with longer label';
+
+      el._inputNode.appendChild(newOption);
+      el.requestUpdate();
```
jorenbroekema, line 211: I believe this is not needed, because align invoker width happens as a consequence of add/removeFormElement, which happens if you append a child option that registers to the select rich.

jorenbroekema (changes requested):

packages/select-rich/src/LionSelectRich.js
gerjanvangeest: I tried to get the padding.. but was not working (can't read from :host selector).

I'm fine with no property.

jorenbroekema (approved):

PR #1138, opened 2020-12-11
feat: add shadow part support

jorgecasar:
## What I did

1. Implement [Shadow Parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part)


jorenbroekema:
Nice, thanks for creating the draft PR. I will look into it more closely but after a first scan I have a question:

If we have these parts exposed, can we still "unexpose" them easily in our extension design system?

The reason for my question is because most design systems actually want to lock certain things behind their shadow boundaries, they intentionally do not expose parts. This is to prevent that their app developers take the design system button and make all kinds of variations that the UX department does not approve of. The only way for them to "hack" those variants would be to extend the component class and override the styles, which is a clear anti-pattern for a user of a design system.

So if we do this change, it makes it easy for lion users to customize the components without needing to extend, but I am actually worried that this is something that design systems which extend lion components (our main use case), will be quite unhappy with. Any thoughts on this?

jdelight:
> Nice, thanks for creating the draft PR. I will look into it more closely but after a first scan I have a question:
> 
> If we have these parts exposed, can we still "unexpose" them easily in our extension design systems?
> 
> The reason for my question is because most design systems actually want to lock certain things behind their shadow boundaries, they intentionally do not expose parts. This is to prevent that their app developers take the design system button and make all kinds of variations that the UX department does not approve of. The only way for them to "hack" those variants would be to extend the component class and override the styles, which is a clear anti-pattern for a user of a design system.
> 
> So if we do this change, it makes it easy for lion users to customize the components without needing to extend, but I am actually worried that this is something that design systems which extend lion components (our main use case), will be quite unhappy with. Any thoughts on this?

👋 we're looking at building on combobox (and other components) for our design system and have been wondering about exactly this question. 

Would it be fair to say there are potentially 3 ways to gain more visual control of lion web components?

1. extend the component class and use a template override to change the underlying html (seems to be the approach used [here](https://github.com/ing-bank/lion/blob/master/packages/combobox/docs/gh-combobox/gh-combobox.js#L335)) and similar examples.
2. use CSS shadow parts and style intentionally exposed elements
3. use CSS custom properties (and possibly [registered properties](https://web.dev/at-property/) to set/override specific colour or numerical properties

I wonder if there are different use-cases for each:

For example, 1) is really valuable if you are adopting Lion Web Components (LWC) as the basis of an in-house design system (DS) - but is likely not what you want end users to use. It does come with the downside of extra maintenance - i.e.  accessibility improvements in LWC may need to be re-added to templates.

2) helps to solve the maintenance issue and there are some thoughts on how to deal with [nested components and parts here](https://youtu.be/yVrvOueeb3s?t=407). Basically if a LWC contains parts, a DS author then renders a LWC, they can decide whether or not to re-export the parts from the inner LWC for their users.

So I think by default it functions as you suggest @jorenbroekema and we "unexpose" parts declared in LWC. However, I'm not sure if this can be done with class extension as well as rendering an inner component. Also it appears [exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-exportparts)  which allows you to export inner parts - is not fully supported everywhere yet.

3) seems like the safest the approach a DS author might use for their developers once they have created their custom component (i.e. based on LWC). Some more on this at the [recent Chrome dev summit](https://youtu.be/DBcz_bGcHgk?t=1042)

Hope this help - interested in your thoughts.

jorgecasar:
Maybe we can move the discussion to https://github.com/ing-bank/lion/discussions/1137. I opened the PR to notify I'm working on it to avoid other start the same task.

Issue #1141, opened 2020-12-15
[calendar] previousMonth and nextMonth conditions error 


jorgecasar:
## Expected behavior

Marking correctly previous and next month

## Actual Behavior

```
    day.previousMonth = currentMonth && day.date.getMonth() < currentMonth.getMonth();
    day.currentMonth = currentMonth && day.date.getMonth() === currentMonth.getMonth();
    day.nextMonth = currentMonth && day.date.getMonth() > currentMonth.getMonth();
```

https://webcomponents.dev/edit/kpZmz1CJN580OaXsk56f?pm=1
![image](https://user-images.githubusercontent.com/948953/102084688-cf6a0800-3e15-11eb-97ba-8456dd167b7a.png)

## Additional context

@lion/calendar@0.12.7

PR #1142, opened 2020-12-15
fix(calendar): solve previousMonth and nextMonth conditions error

jorgecasar:
## What I did

1. Solve issue #1141

### DIFF
--- a/.changeset/curly-ladybugs-grab.md
+++ b/.changeset/curly-ladybugs-grab.md
@@ -1,0 +1,5 @@
+---
+'@lion/calendar': patch
+---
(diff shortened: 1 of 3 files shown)


gerjanvangeest:
Hi @jorgecasar,

Thanks a lot for the fix, could you add a test, so we don't make the same mistake next time?

jorgecasar:
Fixed correctly and added test for first and last month of the year

gerjanvangeest (approved):

gerjanvangeest:
Thanks for the fix :)

PR #1147, opened 2020-12-16
feat(ajax): use fetch as a fundament instead of axios

jorenbroekema:
## What I did

1. Rebased Lars' branch on latest
2. Updated some stuff
3. Fix some type issues (but not all)
4. Add a test for AbortController (users can add polyfill themselves if they wanna use it for IE11 as well, should add documentation for it)

## TODOs

- [x] Discuss w team (before anything else)
- [x] Fix types
- [x] Some more tests maybe (check coverage)
- [x] Changeset
- [x] Breaking change message

### Redaxios :(

We wanted to use [redaxios](https://github.com/developit/redaxios) since it's smaller, uses fetch and keeps the axios API, prevents breaking change.

However:
- Does not support interceptors: https://github.com/developit/redaxios/pull/24
- Does not support Cancel with signal/cancel token
- Jason has said in an issue that a wrapper around fetch should be the end result, and I agree. I don't think redaxios purpose is to be axios but smaller, it's to have the basic functionality but use fetch under the hood to save some refactoring. He said that in one of the issues.. Luckily @LarsDenBakker  already created this wrapper :).

### Interceptors

Interceptors are functions that get executed **before the request is sent** or **before the response is resolved**. This is actually precisely what Lars' response/requestTransformers do in this package 🎉  . So it's supported and with a pretty nice API. 

### Extras

Some cool extra features that I don't think we had before (but not sure)

- If you transform a request straight into a response, the rest of the transformers in the queue are skipped (needs a test).
- Syntax sugar for `Response.text()` + `fetch` all into one call instead of doing double await like you normally have to do with fetch
- Allows passing JS Object as `data` to requestJson calls directly, also handles content-type. Parses the Response as JSON as well. Very handy for APIs that expect/return JSON encoded data, don't have to do the encoding/decoding yourself.


rodrigo-garcia-leon:
@jorenbroekema this fetch-based http client implements transformers as a way to modify the requests/responses, while `axios` has support for interceptors.

The main difference is that interceptors are asynchronous and transformers are synchronous, which may break some use cases around `@lion/ajax`.

jorenbroekema:
@rodrigo-garcia-leon good point. We have decided to rename transformers to interceptors. Since request / requestJson are both already async (because fetch is async), we can easily make it a requirement for the interceptors to also be async functions without much effect for the end user. Then it's closer to axios interceptors and makes migration easier as well. Thoughts?

rodrigo-garcia-leon:
Sounds like a good approach, by making the interceptors async I think all possible use cases should be covered

LarsDenBakker:
Making them async makes sense. Note that there is another difference in that axios interceptors can actually avoid doing the actual fetch and return a response directly. Currently the request transformers don't cater for that. 

Probably it makes sense to have a separate property for "replacing" the actual fetch call... because for those you do still want both the request and response transforms to apply I think.

jorenbroekema:
@LarsDenBakker  https://github.com/ing-bank/lion/pull/1147/files#diff-e854d17b9a6c1d34ce659f4bb1e4c650fb31bade36d8a7d582e8ee42272f2f9aR101  I thought this was returning the Response directly and not doing the fetch call?

LarsDenBakker:
oh.. you're right. I forgot we did that. 👍

daKmoR (approved):

PR #1151, opened 2020-12-17
fix(dialog): prevent opened-changed to be sent when it is not changing

MathieuPuech:
## What I did

1. Prevent opened-changed to be sent twice.

If you add an event listener on a `<lion-dialog>`, you have 2 events sent when you open it.


gerjanvangeest (approved):

gerjanvangeest:
Thanks

PR #1154, opened 2020-12-23
feat: add checkbox indeterminate

MatthieuLebigre:
## What I did

Based on https://github.com/ing-bank/lion/discussions/1128, and inspired by https://github.com/ing-bank/lion/pull/1134, I open this PR to add the checkbox with indeterminate state.

What is different from https://github.com/ing-bank/lion/pull/1134 is mostly that I added the nested checkbox in the slot of the `lion-checkbox-indeterminate` instead of being siblings.

This my first PR in lion, so any feedback will be appreciated, I will update the PR based on your comments.

### DIFF
--- a/packages/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -1,6 +1,28 @@
+import { html, css } from '@lion/core';
 import { LionCheckbox } from './LionCheckbox.js';
 
+/**
+ * @typedef {import('./LionCheckboxGroup').LionCheckboxGroup} LionCheckboxGroup
+ */
+
+// @ts-expect-error false positive for incompatible static get properties. Lit-element merges super properties already for you.
 export class LionCheckboxIndeterminate extends LionCheckbox {
+  static get styles() {
+    const superCtor = /** @type {typeof LionCheckbox} */ (super.prototype.constructor);
+    return [
+      superCtor.styles ? superCtor.styles : [],
+      css`
+        :host .choice-field__nested-checkboxes {
+          display: block;
+        }
+
+        ::slotted([slot='checkbox']) {
+          padding-left: 8px;
+        }
+      `,
+    ];
+  }
+
   static get properties() {
     return {
       /**
@@ -14,52 +36,99 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
   }
 
   get _checkboxGroupNode() {
-    return /** @type {import('./LionCheckboxGroup').LionCheckboxGroup} */ (this.parentElement);
+    return /** @type LionCheckboxGroup */ (this._parentFormGroup);
   }
 
   get _subCheckboxes() {
-    return this._checkboxGroupNode.formElements.filter(checkbox => checkbox !== this);
+    let checkboxes = [];
+    if (
+      this._checkboxGroupNode &&
+      this._checkboxGroupNode.formElements &&
+      this._checkboxGroupNode.formElements.length > 0
+    ) {
+      checkboxes = this._checkboxGroupNode.formElements.filter(
+        checkbox => checkbox !== this && this.contains(checkbox),
+      );
+    }
+    return /** @type LionCheckbox[] */ (checkboxes);
   }
 
-  _parentModelValueChanged() {
-    const checkedElements = this._subCheckboxes.filter(checkbox => checkbox.checked);
-    switch (this._subCheckboxes.length - checkedElements.length) {
+  _setOwnCheckedState() {
+    const subCheckboxes = this._subCheckboxes;
+    if (!subCheckboxes.length) {
+      return;
+    }
+
+    this.__settingOwnChecked = true;
+    const checkedElements = subCheckboxes.filter(checkbox => checkbox.checked);
+    switch (subCheckboxes.length - checkedElements.length) {
       // all checked
       case 0:
         this.indeterminate = false;
         this.checked = true;
         break;
       // none checked
-      case this._subCheckboxes.length:
+      case subCheckboxes.length:
         this.indeterminate = false;
         this.checked = false;
         break;
       default:
         this.indeterminate = true;
+        this.checked = false;
     }
+    this.updateComplete.then(() => {
+      this.__settingOwnChecked = false;
+    });
   }
 
-  _ownModelValueChanged(ev) {
-    if (ev.target === this) {
+  /**
+   * @param {Event} ev
+   */
+  __onModelValueChanged(ev) {
+    if (this.disabled) {
+      return;
+    }
+
+    const _ev = /** @type {CustomEvent} */ (ev);
+    if (_ev.detail.formPath[0] === this && !this.__settingOwnChecked) {
       this._subCheckboxes.forEach(checkbox => {
         // eslint-disable-next-line no-param-reassign
-        checkbox.checked = this.checked;
+        checkbox.checked = this._inputNode.checked;
       });
     }
+    this._setOwnCheckedState();
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  _afterTemplate() {
+    return html`
+      <div class="choice-field__nested-checkboxes">
+        <slot name="checkbox"></slot>
+      </div>
+    `;
+  }
+
+  _onRequestToAddFormElement() {
+    this._setOwnCheckedState();
   }
 
   constructor() {
     super();
     this.indeterminate = false;
+    this._onRequestToAddFormElement = this._onRequestToAddFormElement.bind(this);
+    this.__onModelValueChanged = this.__onModelValueChanged.bind(this);
   }
 
   connectedCallback() {
     super.connectedCallback();
-    this._checkboxGroupNode.addEventListener(
-      'model-value-changed',
-      this._parentModelValueChanged.bind(this),
-    );
-    this.addEventListener('model-value-changed', this._ownModelValueChanged);
+    this.addEventListener('model-value-changed', this.__onModelValueChanged);
+    this.addEventListener('form-element-register', this._onRequestToAddFormElement);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('model-value-changed', this.__onModelValueChanged);
+    this.removeEventListener('form-element-register', this._onRequestToAddFormElement);
   }
 
   /** @param {import('lit-element').PropertyValues } changedProperties */
@@ -67,25 +136,6 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
     super.updated(changedProperties);
     if (changedProperties.has('indeterminate')) {
       this._inputNode.indeterminate = this.indeterminate;
-    }
-  }
-
-  /**
-   * @override
-   * clicking on indeterminate status will set the status as checked
-   */
-  __toggleChecked() {
-    if (this.disabled) {
-      return;
-    }
-
-    // always turn off indeterminate
-    // and set checked to true
-    if (this.indeterminate) {
-      this.indeterminate = false;
-      this.checked = true;
-    } else {
-      this.checked = !this.checked;
     }
   }
 }
--- a/packages/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -1,0 +1,91 @@
+import { LionCheckbox } from './LionCheckbox.js';
+
+export class LionCheckboxIndeterminate extends LionCheckbox {
+  static get properties() {
+    return {
+      /**
+       * Indeterminate state of the checkbox
+       */
+      indeterminate: {
+        type: Boolean,
+        reflect: true,
+      },
+    };
+  }
+
+  get _checkboxGroupNode() {
+    return /** @type {import('./LionCheckboxGroup').LionCheckboxGroup} */ (this.parentElement);
+  }
+
+  get _subCheckboxes() {
+    return this._checkboxGroupNode.formElements.filter(checkbox => checkbox !== this);
+  }
+
+  _parentModelValueChanged() {
+    const checkedElements = this._subCheckboxes.filter(checkbox => checkbox.checked);
+    switch (this._subCheckboxes.length - checkedElements.length) {
+      // all checked
+      case 0:
+        this.indeterminate = false;
+        this.checked = true;
+        break;
+      // none checked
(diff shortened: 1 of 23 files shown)


packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+        reflect: true,
+      },
+    };
+  }
+
+  get _checkboxGroupNode() {
+    return /** @type {import('./LionCheckboxGroup').LionCheckboxGroup} */ (this.parentElement);
+  }
+
+  get _subCheckboxes() {
```
gerjanvangeest, line 21: Hi @MatthieuLebigre,

Looking good so far!

In the end we need also support something like:

```html
<lion-checkbox-group name="scientists[]" label="Favorite scientists">
  <lion-checkbox-indeterminate label="Old Greek scientists">
    <lion-checkbox label="Archimedes" .choiceValue=${'Archimedes'}></lion-checkbox>
    <lion-checkbox label="Plato" .choiceValue=${'Plato'}></lion-checkbox>
    <lion-checkbox label="Pythagoras" .choiceValue=${'Pythagoras'}></lion-checkbox>
  </lion-checkbox-indeterminate>
  <lion-checkbox-indeterminate label="17th Century scientists">
    <lion-checkbox label="Isaac Newton" .choiceValue=${'Isaac Newton'}></lion-checkbox>
    <lion-checkbox label="Galileo Galilei" .choiceValue=${'Galileo Galilei'}></lion-checkbox>
  </lion-checkbox-indeterminate>
</lion-checkbox-group>
```

So `_subCheckboxes` should only get its own child `formElements`.

jorenbroekema: yea so perhaps:
```js
return this._checkboxGroupNode.formElements.filter(checkbox => checkbox !== this && this.contains(checkbox));
```

jorenbroekema:
Hey, I think the behaviour looks good and I like the API way better than my POC, good idea to add a small hook afterTemplate and use it in CheckboxIndeterminate with `<slot>`! 

Indeed what @gerjanvangeest said about sibling indeterminate checkboxes is true, we need to ensure that we only check subcheckboxes that are children of the indeterminate one.

So to continue on this PR:
- [ ] Subcheckboxes only consider formElements that are children, see my suggestion in the reply on that comment
- [ ] Tests (let us know if you get stuck! we can help):
  - [ ] Toggling indeterminate toggles/syncs all children
  - [ ] Indeterminate is true only if more than 0 and less than all child checkboxes are checked.
  - [ ] The indeterminate is checked true if all children are checked true
  - [ ] The indeterminate is checked false (and indeterminate also false) if all children are checked false
- [ ] Fix type errors, but we can help you with that if you're not familiar with JSDocs / TSC combination 👍

jorenbroekema:
I couldn't help myself, made some improvements https://github.com/MatthieuLebigre/lion/pull/1 feel free to check them out before merging to this. I just really wanted to see if the nested indeterminates would work and they do 🎉 😱

packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+      this._inputNode.indeterminate = this.indeterminate;
+    }
+  }
+
+  static get styles() {
+    const superCtor = /** @type {typeof LionCheckbox} */ (super.prototype.constructor);
+    return [
+      superCtor.styles ? superCtor.styles : [],
+      css`
+        :host .nestedCheckboxes {
```
tlouisse, line 100: could we make this `.choice-field__nested-checkboxes` or `.choice-field__indeterminate-checkboxes`?

packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+          display: block;
+        }
+
+        :host ::slotted(lion-checkbox) {
```
tlouisse, line 104: If someone would extend LionCheckbox (for instance call it `wolf-checkbox`) this selector would break.
I think `[slot=input]` will be a safe choice

packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+       */
+      indeterminate: {
+        type: Boolean,
+        reflect: true,
+      },
+    };
+  }
+
+  get _checkboxGroupNode() {
+    return /** @type {import('./LionCheckboxGroup').LionCheckboxGroup} */ (this.parentElement);
```
tlouisse, line 18: A safer choice would be `this.__parentFormGroup`. In this way, we can allow users more freedom in terms of html (adding a div, an ul/li etc.)

packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+      checkbox.checked = this._inputNode.checked;
+    });
+  }
+
+  /**
+   * @override
+   * clicking on indeterminate status will set the status as checked
+   */
+  // eslint-disable-next-line class-methods-use-this
+  __toggleChecked() {}
```
tlouisse, line 55: The current behavior is that it sometimes checks all checkboxes and sometimes unchecks all.

Maybe we can try to implement either one of those?
- a toggle between mixed, true and false like here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html
- readonly, like here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

packages/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -0,0 +1,110 @@
...
+
+  // eslint-disable-next-line class-methods-use-this
+  _afterTemplate() {
+    return html`
+    <div class"nestedCheckboxes">
```
tlouisse, line 60: `=` is missing here :)

MatthieuLebigre:
Hello all !! Thanks all for the comments and suggestions ? I'll fix everything when I'll come back from vacation (beginning of January) :)

packages/checkbox-group/src/LionCheckboxIndeterminate.js
jorenbroekema: I think it's using an unnamed slot in the afterTemplate, but I agree it would be better to name it. Not sure what the best name for it is though, but `input` is for sure already taken by the native <input type="checkbox"> so perhaps it should be `after` or `content`

packages/checkbox-group/src/LionCheckboxIndeterminate.js
jorenbroekema: Added a small improvement to my branch to reuse the __toggleChecked method. I believe the behaviour is already correct though. If not, please clarify what behaviour is the current and what you expected to be different.

jorenbroekema:
Huh thanks to Thijs' comment I noticed we dont have to listen to input `change` event manually, we can reuse the `__toggleChecked` method:

```js
  /**
   * @override ChoiceInputMixin
   * first set subCheckboxes if event came from itself
   * then set own checked + indeterminate state
   * @param {Event} ev
   */
  __toggleChecked(ev) {
    if (this.disabled) {
      return;
    }

    if (ev.target === this) {
      this._subCheckboxes.forEach(checkbox => {
        // eslint-disable-next-line no-param-reassign
        checkbox.checked = this._inputNode.checked;
      });
    }
    this._setOwnCheckedState();
  }
```
I updated my PR to your fork @MatthieuLebigre  👍   have a nice holiday and see you in 2021!

packages/checkbox-group/src/LionCheckboxIndeterminate.js
jorenbroekema: Or slot="checkbox", that makes more sense 👍

tlouisse: Yes indeed, I was confused. This is actually something we cannot style, since we don't know the element name and are not sure that it's a direct child of the default slot (would be possible to wrap this in `ul > li` for instance.
Maybe add a margin of 8px  to `.choice-field__nested-checkboxes` and delete this selector?

tlouisse:
> Huh thanks to Thijs' comment I noticed we dont have to listen to input `change` event manually, we can reuse the `__toggleChecked` method:
> 
> ```js
>   /**
>    * @override ChoiceInputMixin
>    * first set subCheckboxes if event came from itself
>    * then set own checked + indeterminate state
>    * @param {Event} ev
>    */
>   __toggleChecked(ev) {
>     if (this.disabled) {
>       return;
>     }
> 
>     if (ev.target === this) {
>       this._subCheckboxes.forEach(checkbox => {
>         // eslint-disable-next-line no-param-reassign
>         checkbox.checked = this._inputNode.checked;
>       });
>     }
>     this._setOwnCheckedState();
>   }
> ```
> 
> I guess __parentFormGroup and __toggleChecked should be made protected with a single underscore 🙈 can fix that too while we're at it
> 
> I updated my PR to your fork @MatthieuLebigre as well as the checklist I wrote above 👍 have a nice holiday and see you in 2021!

Think I created something similar yesterday: https://github.com/ing-bank/lion/blob/feat/checkbox-indeterminate-enhancements/packages/checkbox-group/src/LionCheckboxIndeterminate.js#L85

This handles [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)  and [tristate](https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html) (it captures indeterminate states, also taking into account users 'meddling' with the captured state by (un)selecting all).

Maybe will have a deeper look on Monday. If you want to cover those details later and proceed now, that's fine with me :)

MatthieuLebigre:
@jorenbroekema  I merged your PR, it should be easier to track all changes in this one.

packages/checkbox-group/src/LionCheckboxIndeterminate.js
jorenbroekema: Ah true, even if you put [slot="checkbox"] this will only apply to direct children, and could be a <div> in between or some other html structures. I agree, might be better to apply the styling to the container in the shadow root 👍

jorenbroekema: Hm but hold on, what if you do

```html
<lion-checkbox-indeterminate label="Scientists">
  <ul slot="checkbox">
    <lion-checkbox label="Isaac Newton" .choiceValue=${'Isaac Newton'}></lion-checkbox>
    <lion-checkbox label="Galileo Galilei" .choiceValue=${'Galileo Galilei'}></lion-checkbox>
  </ul>
</lion-checkbox-indeterminate>
```

That works fine actually, also with the ::slotted([slot="checkbox"]) selector. 

What doesn't work is:

```html
<lion-checkbox-indeterminate label="Scientists">
  <ul>
    <lion-checkbox slot="checkbox" label="Isaac Newton" .choiceValue=${'Isaac Newton'}></lion-checkbox>
    <lion-checkbox slot="checkbox" label="Galileo Galilei" .choiceValue=${'Galileo Galilei'}></lion-checkbox>
  </ul>
</lion-checkbox-indeterminate>
```

But this doesn't work from HTML perspective already (apart from the fact that the CSS ::slotted doesn't work anymore as well).

So perhaps the current way is fine? Although slightly "odd" to put a <ul> in a slot that is called "checkbox", so then maybe the name should be a little more generic or we make it a default slot without a name? I am not usually a fan of unnamed slots but here it might make sense, as you can essentially put whatever you want in it, and the only result is that you get some left padding.

erikkroes: A `<ul>` needs `<li>`'s. You probably want to avoid that.

jorenbroekema: True, maybe a bad example, `<div>` could happen though e.g. if someone needs a wrapper around it with a border or something, for their design system.

tlouisse: Or `<ul slot="checkboxes"><li><ing-checkbox>` (something similar is used in MDN and w3c examples).
But I'm not a fan of a named slot in this particular case, unnamed results in best DX imho.
It's also more in line with the pattern we follow for a regular fieldset/checkbox-group/radio-group etc.

packages/checkbox-group/src/LionCheckboxIndeterminate.js
MatthieuLebigre: Don't know if it's linked, but when I programmatically change the checked attribute of a children checkbox, it's not called (which is the expected behavior). So, I need to also register on this event. What's the best way to do it ?

jorenbroekema: That might be a bug then. Does it work when you change the checked attribute of a child checkbox native input type="checkbox" node?

jorenbroekema: Yea ... that's a tricky one. We only respond to `change` event on the native checkbox input node through proxy at the moment. If we want to also catch programmative changes we should listen to `model-value-changed` event which bubbles up, and we can then check `ev.detail.formPath` to see where it originated from, to determine what to do.

I will do a small refactor today to see if I can make that work 👍

jorenbroekema: I pushed a commit for this, it now uses model-value-changed. The logic to determine what to do depending on where the event comes from and to ensure we don't undo our own checked state setting was a bit tricky but it ended up acceptable to me ;)

MatthieuLebigre:
I did a few changes/fix based on the feedback. I also started to add unit tests. I'll continue on that topic, but I believe for the code itself it should be fine now

jorenbroekema:
Tests are looking good @MatthieuLebigre ! Just a few left, I updated the [checklist](https://github.com/ing-bank/lion/pull/1154#issuecomment-751718406). After those last tests are in, I'll do a full review and finish things up and release 👍

MatthieuLebigre:
> Tests are looking good @MatthieuLebigre ! Just a few left, I updated the [checklist](https://github.com/ing-bank/lion/pull/1154#issuecomment-751718406). After those last tests are in, I'll do a full review and finish things up and release 👍

Done @jorenbroekema

jorenbroekema (approved):

jorenbroekema:
Yay, thanks @MatthieuLebigre  for contributing and for collaborating on this with us 🤗

PR #1156, opened 2020-12-24
chore: improve validators instance error message

gerjanvangeest:
## What I did

1. Based on: https://polymer.slack.com/archives/CJGFWJN9J/p1608528031390200Improve, the error message is improved.
2. In my opinion, no real need to increase the version, it would be okay to wait a little for this to be released.


jorenbroekema:
Agreed, no rush to release, it will go next time form-core gets a more important update

jorenbroekema (approved):

PR #1163, opened 2021-01-04
Iban

sidona:
## What I did

1. improve validation for blacklist country - iban

### DIFF
--- a/.changeset/curvy-suns-cross.md
+++ b/.changeset/curvy-suns-cross.md
@@ -1,0 +1,5 @@
+---
+'@lion/input-iban': patch
+---
+
+allow to specify countries in lowercase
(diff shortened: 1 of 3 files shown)


jorenbroekema (changes requested):
Seems like a good idea to me, so lowercase IBANs get properly validated, and you can pass `['en']` as well as `['EN']` as a param, and it just gets normalized to uppercase.

However, some things you need to do still:
- [ ] Sign CLA
- [ ] Remove debugger statements
- [ ] Revert removing the .editorconfig file, we still want that 😅 
- [ ] Add a test to confirm normalisation behaviour of the country param
- [ ] Add a test to confirm normalisation behaviour of the modelValue
- [ ] Add a changeset with a patch, run `yarn changeset` `npx changeset` and follow the menu and create a good changelog message for this change. If needed we can also do this for you :)

tlouisse:
I completed the suggesions from @jorenbroekema 
Thanks for the contribution @sidona!

tlouisse (approved):

PR #1164, opened 2021-01-05
fix: normalize dependencies

jorgecasar:
## What I did

1. remove dependencies that aren't used in src code
2. include README.md imports as devDependencies
3. move test import to root devDependencies


packages/ajax/package.json
```diff
@@ -32,9 +32,11 @@
   "sideEffects": false,
   "dependencies": {
     "@bundled-es-modules/axios": "0.18.1",
-    "@lion/core": "0.13.6",
```
jorenbroekema, line 35: This may be correct, because if you extend only `@lion/ajax` and its demos, you need the installation of `@lion/core` for the demo extension to function properly, and it only gets installed if it's a real dependency.

packages/checkbox-group/package.json
```diff
@@ -39,6 +39,9 @@
     "@lion/form-core": "0.6.14",
     "@lion/input": "0.10.12"
   },
+  "devDependencies": {
+    "@lion/validate-messages": "0.3.12"
```
jorenbroekema, line 43: This may need to be a real dependency to ensure it is actually installed in extension layers that extend demos.

packages/form-core/types/FormControlMixinTypes.d.ts
```diff
@@ -1,5 +1,5 @@
 import { CSSResultArray } from 'lit-element';
-import { LitElement, nothing, TemplateResult } from '@lion/core';
+import { LitElement, nothing, TemplateResult, CSSResultArray } from '@lion/core';
```
jorenbroekema: We would need to re-export this first in `@lion/core`, the problem is that `lit-element` only exports this as a type export. So we would need to re-export this type export (if possible?) in our index.d.ts file in `@lion/core`.

packages/input-date/package.json
```diff
@@ -33,7 +33,6 @@
     "lion-input-date.js"
   ],
   "dependencies": {
-    "@lion/core": "0.13.6",
```
jorenbroekema, line 36: I guess practically speaking this can be removed because you always get `@lion/core` installed if you depend on any `@lion/<package>`, but it would be more technically correct to have it added here because we import directly from `@lion/core`, therefore the best practice is to add it as a real dependency (npm will dedupe)

packages/input-iban/package.json
```diff
@@ -33,7 +33,6 @@
     "lion-input-iban.js"
   ],
   "dependencies": {
-    "@lion/core": "0.13.6",
```
jorenbroekema, line 36: See comment above on ajax package.json

packages/input-stepper/package.json
```diff
@@ -37,6 +37,9 @@
     "@lion/form-core": "0.6.14",
     "@lion/input": "0.10.12"
   },
+  "devDependencies": {
+    "@lion/form": "0.7.12"
```
jorenbroekema, line 41: We can also just remove the form demo from input-stepper package so we don't need any dependency on @lion/form. This is why we have the form-integrations package anyway, to demo our fields inside a form.

jorenbroekema (changes requested):
To summarize, I think this is what we ought to do:

- 0 devDependencies inside any of lion packages and node-packages. We always develop from the project root, and all develop dependencies should therefore be in the root package.json.
- Any demo dependencies are now real dependencies, because we tell people they can extend demos, this makes our .md files with demos "production files".
- We need to minimise the amount of imports in our demos. 
  - We should not import `@lion/form` to demo integration with a form, instead we should use `form-integrations` package to show how lion fields integrate with forms. This keeps our field packages clean.
  - We should not import `@lion/validate-messages` or `@lion/localize` to demo how lion fields integrate with validation or localisation. This should also be shown in form-integrations package demos.

The other changes are good, we definitely should import from `@lion/core` instead of lit directly. If this happens in a demo file, we should add `@lion/core` as a real dependency as a best practice (always add dependencies you import from as a dependency, don't rely on them being installed as a side effect of installing another package).

Before you go ahead making all these changes, I will also give my team the chance to object  ;)

jorgecasar:
OK, I only move all devDeps to deps. And export CSSResultArray

packages/input-stepper/package.json
jorgecasar: `lion-form` is used in the README.md in the using Form story

jorgecasar: Then we should move this story to `form-integration`, right?

packages/input-date/package.json
jorgecasar: I didn't see the import anywhere, that's why I removed it. It will be loaded by other dependencies, but technically it's not a input-date direct dependency

packages/form-core/types/FormControlMixinTypes.d.ts
jorgecasar: Added to '@lion/core'

packages/input-date/package.json
jorenbroekema: Ah than it's indeed not needed here.

packages/input-stepper/package.json
jorenbroekema: Yep! But this PR is plenty large already, I think I'll go through every package today and minimise dependencies if they are caused by form integration or validation or localisation examples. Will likely do this today and we can release this PR and mine simultaneously.

jorenbroekema (approved):

Issue #1169, opened 2021-01-06
Lion with Angular, Unpkg (optional chaining)


MendyBerger:
## Expected behavior

Importing and using lion in Angular should work

## Actual Behavior

When importing `@lion/select-rich` I'm getting the following error:
```
node_modules/@lion/select-rich/src/LionSelectInvoker.d.ts:8:14 - error TS2417: Class static side 'typeof LionSelectInvoker' incorrectly extends base class static side 'typeof LionButton'.
  Types of property 'styles' are incompatible.
    Type '(CSSResult | CSSResult[])[]' is not assignable to type 'CSSResult[]'.
      Type 'CSSResult | CSSResult[]' is not assignable to type 'CSSResult'.
        Type 'CSSResult[]' is missing the following properties from type 'CSSResult': cssText, styleSheet
```

## Additional context

Which component and version of it are you using:
`@lion/select-rich@0.21.21`

Thanks in advance, I really like the idea of a style'less component library and I'm looking forward to using it!

jorenbroekema:
In your TS Config you can put `"skipLibCheck": true` as a workaround, this will silence errors if they come from types in your node_modules.

You will likely need this as we have some other @ts-expect-error statements in our source code that we haven't been able to work around yet (it's on our backlog), so the chances you will get other errors from lion's type definition files is substantial.

That said, we should try to fix it on our end, so I'll list it as a bug and try to fix it.. Lit-element's styles getter is always a little awkward because it accepts an array of CSSResult, or a single CSSResult, and it flattens if you nested arrays of CSSResults... @daKmoR  were you still planning on proposing that Lit-element always accepts only an array to simplify things?

MendyBerger:
@jorenbroekema thanks for the idea, I added `skipLibCheck` and I'm no longer getting the previous error.

However, I'm now getting a new error:
```
Error: ./node_modules/@lion/overlays/src/utils/inert-siblings.js 11:78
Module parse failed: Unexpected token (11:78)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|  */
| export function setSiblingsInert(element) {
>   const parentChildren = /** @type {HTMLCollection} */ (element.parentElement?.children);
|   for (let i = 0; i < parentChildren.length; i += 1) {
|     const sibling = parentChildren[i];
```

I think that webpack 4 can't handle optional chaining, but I'm not sure

jorenbroekema:
[...]

PR #1171, opened 2021-01-07
fix: type issues with regards to static styles

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1169

### DIFF
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -9,11 +9,10 @@ import { css, html } from '@lion/core';
 /**
  * LionSelectInvoker: invoker button consuming a selected element
  */
-// @ts-expect-error
 export class LionSelectInvoker extends LionButton {
   static get styles() {
     return [
-      super.styles,
+      ...super.styles,
       css`
         :host {
           justify-content: space-between;
--- a/.changeset/long-parents-divide.md
+++ b/.changeset/long-parents-divide.md
@@ -1,0 +1,7 @@
+---
+'@lion/combobox': patch
+'@lion/select-rich': patch
+'@lion/tooltip': patch
+---
+
+Fix some type issues with static get styles, CSSResultArray combines CSSResult and CSSResult[].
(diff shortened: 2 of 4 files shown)


gerjanvangeest (approved):

packages/select-rich/src/LionSelectInvoker.js
```diff
@@ -9,11 +9,10 @@ import { css, html } from '@lion/core';
 /**
  * LionSelectInvoker: invoker button consuming a selected element
  */
-// @ts-expect-error
 export class LionSelectInvoker extends LionButton {
   static get styles() {
     return [
-      super.styles,
+      ...super.styles,
```
tlouisse, line 15: So basically we should make sure we also use `CSSResult[]` inside our components, no matter what, but it would be nice if LitElement would enforce this in the next version.
Doesn't the example below result in a type of `Array<CSSResult|CSSResult[]>` ?

Shouldn't we put `...(super.styles || [])` everywhere?

Shouldnt

jorenbroekema: > Doesn't the example below (Tooltip) result in a type of Array<CSSResult|CSSResult[]> ?

Yep. But LitElement is smart enough to merge/flatten deeply nested arrays.. 

I explained in more detail in Mattermost team chat ;) it's a little complicated.. definitely agree Lit should make this consistent and enforce, but I'm a bit concerned it will go a bit beyond their breaking change budget for v3

tlouisse: I know LitElement allows, it even would allow `Array<CSSResult|Array<CSSResult|CSSResult[]>>` etc.
But if we force all implementation in our code base to only use `CSSResult[]`, wouldn't it be solved?

jorenbroekema: Hmm but `super.styles` is not always iterable. If you extend from LitElement like we do in LionTooltip, it can also be `CSSResult` or `CSSStyleSheet` instead of `CSSResultArray`. So in order to always get our own styles to use `CSSResult[]` everywhere, we need to typecast the super styles in a bunch of places to only "allow" `CSSResult[]`. I'm fine with doing that to get things consistent, maybe create an issue?

PR #1178, opened 2021-01-12
fix(overlays): typescript errors

bennypowers:
Corrects typescript from `@popperjs/core` for overlays package

## What I did

1. Corrects typescript from `@popperjs/core` for overlays package

### DIFF
--- a/.changeset/serious-candles-serve.md
+++ b/.changeset/serious-candles-serve.md
@@ -1,0 +1,5 @@
+---
+'@lion/overlays': patch
+---
+
+Fix type declaration for \_\_syncFromPopperState method.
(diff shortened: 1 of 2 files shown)


bennypowers:
Please verify that I picked the right param type (`State`)

jorenbroekema (approved):

jorenbroekema:
Hey thanks! I had to edit to `Partial<State>` since we call the method from Popper's `onFirstUpdate` and `afterWrite` methods which only passes partial state date. https://github.com/popperjs/popper-core/blob/master/src/types.js#L168. not sure what $Shape is exactly but yeah..

PR #1180, opened 2021-01-13
fix: types base constructor same return type

jorenbroekema:


packages/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -22,19 +22,16 @@ import { FormRegistrarPortalMixin } from '../src/registration/FormRegistrarPorta
  */
 export const runRegistrationSuite = customConfig => {
```
jorenbroekema, line 23: Changed this because our registration mixins have strict contract where superclass must be extending from LitElement, so doesn't make sense to test with UpdatingElement / HTMLElement. 

Alternatively we can look into making the registration mixins work with HTMLElement, but I don't really see the point as Lion is pretty locked into lit right now.

gerjanvangeest (approved):

PR #1181, opened 2021-01-15
chore(ajax cache): added ajax cache

yevgeniyvaleyev:
## What I did

1. Ajax cache interceptors
2. Documentation
3. Tests


jorenbroekema:
Really cool contribution!

If I understand correctly, this interceptor caches requests making it faster on the frontend, and less work for the backend because it will receive fewer requests (since they will be cached and resolved on client)?

In principle I really like that idea, and I like the options that come with it, and just supporting 'get' method for now is a good start in my opinion.

I could give it a full review but maybe I should start with a general architectural question first:
- Why is it a separate package? Could we maybe include this interceptor as an export inside `@lion/ajax`? For example, `@lion/ajax` currently exports 3 interceptors already, we could add this one perhaps?

I will have to dive a bit deeper into the source code / tests during a full review but from a first scan it definitely all makes sense to me 👍  so thanks for deciding to contribute this feature to lion 🦁 !

yevgeniyvaleyev:
@jorenbroekema I made it as separate package to keep caching functionality separate from ajax, and when a user exactly knows he or she needs caching they'd easily import it. IMO ajax has to take care about calls and extra work of caching has to be explicitly added. What do you think ?

yevgeniyvaleyev:
@jorenbroekema I've moved ajax cache interceptors form separate package to ajax package. Please check it when you have time

packages/ajax/README.md
```diff
@@ -187,6 +187,157 @@ export const cancelConcurrentRequests = () => html`
...
+  ajax,
+  lionCacheRequestInterceptorFactory,
+  lionCacheResponseInterceptorFactory,
+} from '@lion-web/ajax.js';
+
+const globalCacheOptions = {
+  useCache: 'always',
+  timeToLive: 1000 * 60 * 5, // 5 minutes
+};
+const getCacheIdentifier = () => getActiveProfile().refreshToken;
```
LarsDenBakker, line 219: This is probably very application specific. If we do show the profile, we shouldn't use the `refreshToken` because it will change each refresh. We want to use the profileId here.

packages/ajax/src/interceptors-cache.js
```diff
@@ -0,0 +1,361 @@
...
+    this.expiration = +new Date() + HOUR;
+    /**
+     * @type {{[url: string]: CacheConfig }}
+     */
+    this.cacheConfig = {};
+
+    /**
+     * @type {{[url: string]: {date: number, data: object} }}
+     */
+    this.cacheObject = {};
```
LarsDenBakker, line 21: Isn't this a private property? In that case we should prefix it with `__`. Or perhaps even store it with a symbol, so that devs don't try to take things from the cache manually.

Using a `Map` is faster when it's accessed frequently.

LarsDenBakker:
Is there some way we can prevent developers from including these interceptors in non-app projects?

packages/ajax/README.md
yevgeniyvaleyev: thanks @LarsDenBakker

ruphin:
This functionality should not be in the ajax package. If we migrate away from Axios, we want to be able to use this in combination with Fetch. If I understand correctly, these interceptors will also work with the upcoming Fetch solution.

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+      .then(() => {
+        expect(server.requests.length).to.equal(1);
+      })
+      .finally(() => {
+        ajaxGetSpy.restore();
+        removeCacheInterceptors(ajax, indexes);
+      });
+  });
+
+  it('all calls with non-default `timeToLive` are cached proactively', () => {
```
jorenbroekema, line 197:
```suggestion
  it('all calls with non-default `timeToLive` are cached on demand per request', () => {
```

Or maybe I don't understand the behaviour...

If timeToLive is kept as a default, but then we do:

```js
ajax.get('/test', {
  lionCacheOptions: {
    useCache: 'always',
  },
}),
```

Do we get a cached response with default timeToLive expiration time? Or do we not get a cached response because timeToLive is still default value?

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+      .then(() => {
+        expect(server.requests.length).to.equal(4);
+      })
+      .then(() => ajax.get('/foo-request-1'))
+      .then(() => {
+        expect(server.requests.length).to.equal(5);
+      })
+      .then(() => ajax.get('/foo-request-2'))
+      .then(() => {
+        expect(server.requests.length).to.equal(5);
```
jorenbroekema, line 289: I would expect 6 here because the `ajax.post('/test', {}, actionConfig)` has the invalidate regex rule, which invalidates cache right?

Does it only invalidate "global" cache for the next request to "any" url? Meaning that the foo-request-2 (which is a different url) gets you a cached response immediately?

In general, I guess I don't understand, why do we have to invalidate cache when invalidation rules change?

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+      .then(() => {
+        expect(server.requests.length).to.equal(4);
+      })
+      .then(() => ajax.get('/posts'))
+      .then(() => {
+        expect(server.requests.length).to.equal(5);
+      })
+      .then(() => ajax.get('/posts/1'))
+      .then(() => {
+        expect(server.requests.length).to.equal(5);
```
jorenbroekema, line 350: I would expect 6 here as well as I thought that because it's a nested url, it would not have a valid cached value for this url yet?

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+
+    return ajax
+      .get('/test')
+      .then(() => {
+        const clock = sinon.useFakeTimers();
+        expect(ajaxGetSpy.calledOnce).to.be.true;
+        expect(ajaxGetSpy.calledWith('/test')).to.be.true;
+        clock.tick(1);
+        clock.restore();
+      })
```
jorenbroekema, line 375: I don't really understand what this tests? We're spying ajax.get and we call it with '/test', so it seems quite obvious to me that the two expects would be be true, even without a fake timer? Is the fake timer only used so that the next request happens 1 ms later and gets a normal response because it is past the 0ms timeToLive expiration? if that's the case, shouldn't we restore the clock in line 378 after the expect?

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+          q: 'test',
+          page: 1,
+        },
+      })
+      .then(() => {
+        expect(ajaxGetSpy.calledOnce).to.be.true;
+        expect(ajaxGetSpy.calledWith('/test')).to.be.true;
+      })
+      .then(() =>
+        ajax.get('/test', {
```
jorenbroekema, line 408: Shall we add a test with different params, to see if the server requests length is then 2?

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+      .then(() => {
+        expect(server.requests.length).to.equal(2);
+      })
+      .finally(() => {
+        ajaxAlwaysGetSpy.restore();
+        removeCacheInterceptors(ajax, indexes);
+      });
+  });
+
+  it('allows direct ajax calls without cache interceptors configured', () => {
```
jorenbroekema, line 447: Maybe put this test at the top, since it's kinda a smoke test of the default ajax behavior without caching.

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+      })
+      .then(() => ajax.get('/test-invalid-url'))
+      .then(() => {
+        expect(server.requests.length).to.equal(2);
+      })
+      .then(() => ajax.post('/test-valid-url', {}, actionConfig))
+      .then(() => {
+        expect(ajaxPostSpy.calledOnce).to.be.true;
+        expect(ajaxPostSpy.calledWith('/test-valid-url')).to.be.true;
+        expect(server.requests.length).to.equal(3);
```
jorenbroekema, line 593: I thought this one would be `2` because it was cached on line 579 as it is a valid url? How come it is not returning a cached response here? Or is this a side effect from having a trigger of a new invalidateUrls rule invalidating also the cache for the valid url?

What I would expect this test to test is that you call the invalid url twice in a row, but never get a cached response because the url matches the invalidateUrl rule.

jorenbroekema: Oh I see now I think, it's because of the `post` method which is not supported w caching right now?

Either way, this test doesn't seem to be testing what it says it should be testing in the description

packages/ajax/test/AjaxClass.interceptors-cache.test.js
```diff
@@ -0,0 +1,713 @@
...
+        lionCacheOptions: validateOptions({
+          // @ts-ignore needed for test
+          invalidateUrls: 'not an array',
+        }),
+      });
+    }).to.throw(/Property `invalidateUrls` must be of type `Array` or `falsy`/);
+    removeCacheInterceptors(ajax, indexes);
+  });
+
+  it('throws error when invalidateUrls is applied to global config', () => {
```
jorenbroekema, line 623: Is there a specific reason why invalidateUrls / invalidateRegex can only be applied per action to force a "real" response?

Why not allow to set it on the global config too, e.g. because we never want cached response for certain urls? Then the user doesn't have to pass it as an action config every time the call it

jorenbroekema (changes requested):
Hey, looks good in general, just have a lot of questions on the details 😅  some of which are probably my own misunderstandings and not necessarily code imperfections.

The biggest change I would request is renaming some of the public stuff to not have `lion` prefixed in the var names:

- lionCacheRequestInterceptorFactory -> cacheRequestInterceptorFactory
- lionCacheResponseInterceptorFactory -> cacheResponseInterceptorFactory
- lionCacheOptions -> cacheOptions

I would also consider grouping some of the tests in describe statements. I really like your tests, they are quite thorough, but they don't seem organised logically which made it a bit harder to read them.

Lastly, could you add some real working demos to illustrate the behaviours?

- Show how a consecutive request gets cached
- Show how it doesn't if you invalidate the cache (show multiple ways that invalidate the cache)

Since a cached response is identical to a real response except for that we didn't fire an actual HTTP request, I suppose we could demo by showing a counter that increments for every "real" request, could we read out the "fromCache" private var perhaps to be able to make the distinction?

packages/ajax/README.md
```diff
@@ -187,6 +187,158 @@ export const cancelConcurrentRequests = () => html`
...
+frontend `services`.
+
+> Technical documentation and decisions can be found in
+> [./docs/technical-docs.md](./docs/technical-docs.md)
+
+### Getting started
+
+Consume the global `ajax` instance and add the interceptors to it, using a cache configuration
+which is applied on application level. If a developer wants to add specifics to cache behavior
+he or she has to provide a cache config per action (get etc) via `lionCacheOptions` field of local ajax config,
```
jorenbroekema, line 202:
```suggestion
they have to provide a cache config per action (`get`, `post`, etc.) via `lionCacheOptions` field of local ajax config,
```

jorenbroekema:
@ruphin  I do think it belongs in the ajax package. If we move to the fetch wrapper solution, we will move it to there. The same concepts of caching and intercepting/transforming apply, so porting it should be minimal effort.

packages/ajax/README.md
```diff
@@ -187,6 +187,158 @@ export const cancelConcurrentRequests = () => html`
...
+```
+
+### Ajax cache Options
+
+```js
+const cacheOptions = {
+  // `useCache`: determines wether or not to use the cache
+  // can be string 'always' | 'never'
+  // default: 'never'
+  useCache: 'always',
```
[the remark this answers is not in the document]
yevgeniyvaleyev: )I think we can change it to boolean, it was done like this (in the beginning) for readability but, I agree, it does not really give here any benefit

packages/ajax/test/AjaxClass.interceptors-cache.test.js
yevgeniyvaleyev: The intention was allow invalidation of the get responses which are owned by the user, but in fact you can invalidate other responses if you want. Since we cannot limit it anyhow probably better idea just to allow it from any config

packages/ajax/test/AjaxClass.interceptors-cache.test.js
yevgeniyvaleyev: I updated it, and added explanations. Yes, 'post' invalidates own cache and the one mentioned in config

packages/ajax/test/AjaxClass.interceptors-cache.test.js
yevgeniyvaleyev: The behaviour is to cache by default and use the cache only when needed, by using useCache=true.
So the app could collect responses in cache but not use it, and only when , for instance, a component does a call with useCache=true (and cache is not expired), the component will get a response from cache.
This behaviour is aligned with @LarsDenBakker some time ago

packages/ajax/src/interceptors-cache.js
```diff
@@ -0,0 +1,361 @@
...
+
+    const cacheId = lionCacheOptions.requestIdentificationFn(cacheRequest, searchParamSerializer);
+
+    // cacheIdentifier is used to bind the cache to the current session
+    const currentCache = getCache(getCacheIdentifier());
+    const cacheResponse = currentCache.get(cacheId, lionCacheOptions.timeToLive);
+
+    // don't use cache if the request method is not part of the configs methods
+    if (lionCacheOptions.methods.indexOf(method) === -1) {
+      // If it's NOT one of the config.methods, invalidate caches
```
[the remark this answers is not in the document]
yevgeniyvaleyev: Only GET is allowed, and if the request is non GET with the same url we clean it's cache

packages/ajax/src/interceptors-cache.js
```diff
@@ -0,0 +1,361 @@
...
+    ? Object.keys(params)
+        .map(key => `${key}=${params[key]}`)
+        .join('&')
+    : '';
+
+/**
+ * Returns the active cache instance for the current session
+ * @param {string} cacheIdentifier usually the refreshToken of the owner of the cache
+ */
+const getCache = cacheIdentifier => {
```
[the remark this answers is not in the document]
yevgeniyvaleyev: I'll add comments and docs

packages/ajax/src/interceptors-cache.js
```diff
@@ -0,0 +1,361 @@
...
+
+let caches = {};
+
+/**
+ * Serialize search parameters into url query string parameters.
+ * If params === null, returns ''
+ * @param {Params} params query string parameters object
+ * @returns {string} of querystring parameters WITHOUT `?` or empty string ''
+ */
+export const searchParamSerializer = params =>
```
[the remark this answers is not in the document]
yevgeniyvaleyev: Is Lion-web fine introducing something which requires polyfills ? For me this function is very small but hassle of adding extra dependencies is more risky IMO.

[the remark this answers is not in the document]
jorenbroekema: Yep if it's just IE11 we're cool with it. I'm finalising the fetch wrapper and as it uses Fetch and AbortControllers, it mandates IE11 polyfills as well. As long as we document it it's fine. People should try to ditch IE11 so I don't mind giving them extra reasons to do so >:)

packages/ajax/src/interceptors-cache.js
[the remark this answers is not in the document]
jorenbroekema: I got that part, but I don't get why we clean its cache? Why not ignore it, and keep the cache for consecutive calls with supported methods?

yevgeniyvaleyev:
> Hey, looks good in general, just have a lot of questions on the details 😅 some of which are probably my own misunderstandings and not necessarily code imperfections.
> 
> The biggest change I would request is renaming some of the public stuff to not have `lion` prefixed in the var names:
> 
> * lionCacheRequestInterceptorFactory -> cacheRequestInterceptorFactory
> * lionCacheResponseInterceptorFactory -> cacheResponseInterceptorFactory
> * lionCacheOptions -> cacheOptions
> 
> I would also consider grouping some of the tests in describe statements. I really like your tests, they are quite thorough, but they don't seem organised logically which made it a bit harder to read them.
> 
> Lastly, could you add some real working demos to illustrate the behaviours?
> 
> * Show how a consecutive request gets cached
> * Show how it doesn't if you invalidate the cache (show multiple ways that invalidate the cache, primarily TTL / invalidate rules)
> 
> Since a cached response is identical to a real response except for that we didn't fire an actual HTTP request, I suppose we could demo by showing a counter that increments for every "real" request, could we read out the "fromCache" private var perhaps to be able to make the distinction?

yevgeniyvaleyev:
@jorenbroekema I did not consider a demo to be a part of this PR.
and why do you mean about 'fromCache' ?

packages/ajax/src/interceptors-cache.js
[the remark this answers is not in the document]
yevgeniyvaleyev: I've added the function

packages/ajax/src/interceptors-cache.js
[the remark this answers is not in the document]
yevgeniyvaleyev: The idea was to rather clean it immediately than keep. Support for consecutive will be added (probably) later. Some time ago we were discussing it with @daKmoR

packages/ajax/test/AjaxClass.interceptors-cache.test.js
yevgeniyvaleyev: ah, the reason is `clock` for some reason does not work in asynchronous  function, I was checking for an example in lion-web but could not find, but this one works like this

packages/ajax/src/interceptors-cache.js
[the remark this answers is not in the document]
jorenbroekema: Let's put in the shape of a test:

```
it('returns the cached object on second call with `useCache: true`', () => {
  newCacheId();

  const indexes = addCacheInterceptors(ajax, {
    useCache: true,
  });
  const ajaxGetSpy = sinon.spy(ajax, 'get');

  return ajax
    .get('/test')
    .then(() => {
      expect(ajaxGetSpy.calledOnce).to.be.true;
      expect(ajaxGetSpy.calledWith('/test')).to.be.true;
      expect(server.requests.length).to.equal(1);
    })
    .then(() => ajax.get('/test'))
    .then(() => {
      expect(server.requests.length).to.equal(1);
    })
    .then(() => ajax.post('/test')) // ignore unsupported method
    .then(() => {
      expect(server.requests.length).to.equal(1);
    })
    .then(() => ajax.get('/test')) // return cached
    .then(() => {
      expect(server.requests.length).to.equal(1);
    })
    .finally(() => {
      ajaxGetSpy.restore();
      removeCacheInterceptors(ajax, indexes);
    });
});
```

This makes sense to me. But current behaviour would be that the last expect would equal 2, because doing unsupported method cleans the cache. I just don't see why that behaviour would be preferable. Can you explain?

jorenbroekema:
@yevgeniyvaleyev I usually consider demos to be part of documentation, but I am also okay with adding it later myself as this PR is plenty big already.

With 'fromCache' I mean that it would be nice that we can access some kind of metadata of the Response to check whether the Response is served from the API or served from the cache. Maybe as a protected or private property because the end user shouldn't care, but for demos it can certainly be useful, so we can show what is a cached response and what is a real response in our demos.

packages/ajax/src/interceptors-cache.js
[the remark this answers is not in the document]
daKmoR: if you run a "post" on an url then it usually means the data is new so you need to fetch it... for example

```js
// foo.js
ajax.get('/test') // returns "10"

// bar.js
ajax.get('/test') // returns "10" [cached]

// foo.js later
ajax.post('/test') 

// bar.js later
ajax.get('/test') // cache is very likely no longer accurate => do request // returns "20"
ajax.get('/test') // returns "20" [cached]
```

as in, doing a post usually mutates entries in the database through the API, so it should invalidate the cache? because the cache is out of date now

alternatives for cache invalidation we looked at
- time  =>  not gonna work.... as we have time-sensitive data
- manually => is gonna be a maintenance nightmare :see_no_evil:

I would rather have cache more often invalidated than needed than having to maintain all these cache "dependencies"

Conclusion: a post request should invalidate all cache for this particular url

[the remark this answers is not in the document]
jorenbroekema: Makes sense now!

packages/ajax/src/interceptors-cache.js
```diff
@@ -0,0 +1,361 @@
...
+    } catch (e) {
+      // We need console.error here to show error somehow since Errors are not emitted from Axios
+      // eslint-disable-next-line no-console
+      console.error(e.message);
+    }
+
+    const lionCacheOptions = {
+      ...validatedInitialLionCacheOptions,
+      ...actionCacheOptions,
+    };
```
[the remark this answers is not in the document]
yevgeniyvaleyev: changed

jorenbroekema:
Ported this to the new ajax package which wraps fetch https://github.com/ing-bank/lion/pull/1208

PR #1187, opened 2021-01-19
fix(form-core): dedupe click event label choice inputs

jorenbroekema:
Not sure if this is the right place to fix it, but doing

```html
<lion-checkbox 
  @click=${() => console.log('clicked')} 
  label="Francis Bacon" 
  .choiceValue=${'Francis Bacon'}
></lion-checkbox>
```

and then clicking the label creates 2 click events.

Clicking the graphic container only creates 1 click event. 

Therefore, I figured we just prevent the label click from propagating on the CE (lion-choice-input), so that you only get 1 click event on the CE.

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceInputMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceInputMixin.js
@@ -128,6 +128,7 @@ const ChoiceInputMixinImplementation = superclass =>
       super();
       this.modelValue = { value: '', checked: false };
       this.disabled = false;
+      this._preventDuplicateLabelClick = this._preventDuplicateLabelClick.bind(this);
       this.__toggleChecked = this.__toggleChecked.bind(this);
     }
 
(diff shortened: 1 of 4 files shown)


packages/form-core/src/choice-group/ChoiceInputMixin.js
```diff
@@ -189,14 +189,28 @@ const ChoiceInputMixinImplementation = superclass =>
...
       super.disconnectedCallback();
+      this._labelNode.removeEventListener('click', this._preventDuplicateLabelClick.bind(this));
       this.removeEventListener('user-input-changed', this.__toggleChecked);
     }
 
+    /**
+     * @param {Event} ev
+     */
+    // eslint-disable-next-line no-unused-vars
+    _preventDuplicateLabelClick(ev) {
```
tlouisse, line 206: Looks good. Maybe add some explanation which 'platform hack' we work around here

packages/form-core/src/choice-group/ChoiceInputMixin.js
```diff
@@ -189,14 +189,28 @@ const ChoiceInputMixinImplementation = superclass =>
 
     connectedCallback() {
       super.connectedCallback();
+      this._labelNode.addEventListener('click', this._preventDuplicateLabelClick.bind(this));
       this.addEventListener('user-input-changed', this.__toggleChecked);
     }
 
     disconnectedCallback() {
       super.disconnectedCallback();
+      this._labelNode.removeEventListener('click', this._preventDuplicateLabelClick.bind(this));
```
tlouisse, line 198: This will be a different instance than the one created at connected

tlouisse (approved):

tlouisse (approved):

PR #1192, opened 2021-01-22
Update OverlayController.js

barroudjo:
When the body element is configured with a `margin: 0 auto;` property (such as in the old ING styleguide, TheGuide), the overlay controller fails with the following error:
```
OverlayController.js:762 Uncaught (in promise) TypeError: Failed to execute 'px' on 'CSS': The provided double value is non-finite.
    at OverlayController._keepBodySize (OverlayController.js:762)
    at OverlayController.show (OverlayController.js:694)
```
This is due to the use of computedStyleMap(), which returns (rightfully so) `auto`, which is then used in line 762:
```javascript
document.body.attributeStyleMap.set('margin-right', CSS.px(newMarginRight));
``` 
And obviously `CSS.px('auto 10px')` will crash...

The obvious solution to use `getComputedStyle(...)`  instead of `computedStyleMap()` as it returns computed values (ie in `px` units), not values set in CSS such as `auto`.

### DIFF
--- a/packages/overlays/src/OverlayController.js
+++ b/packages/overlays/src/OverlayController.js
@@ -735,21 +735,17 @@ export class OverlayController extends EventTargetShim {
       case 'before-show':
         this.__bodyClientWidth = document.body.clientWidth;
         this.__bodyClientHeight = document.body.clientHeight;
-        this.__bodyMarginRight = 0;
-        this.__bodyMarginBottom = 0;
+        this.__bodyMarginRightInline = document.body.style.marginRight;
+        this.__bodyMarginBottomInline = document.body.style.marginBottom;
         break;
       case 'show': {
-        if (supportsCSSTypedObject) {
-          // @ts-expect-error types attributeStyleMap not available yet
-          this.__bodyMarginRight = document.body.computedStyleMap().get('margin-right').value;
-          // @ts-expect-error types computedStyleMap not available yet
-          this.__bodyMarginBottom = document.body.computedStyleMap().get('margin-bottom').value;
-        } else if (window.getComputedStyle) {
+        if (window.getComputedStyle) {
           const bodyStyle = window.getComputedStyle(document.body);
-          if (bodyStyle) {
-            this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
-            this.__bodyMarginBottom = parseInt(bodyStyle.getPropertyValue('margin-bottom'), 10);
-          }
+          this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
+          this.__bodyMarginBottom = parseInt(bodyStyle.getPropertyValue('margin-bottom'), 10);
+        } else {
+          this.__bodyMarginRight = 0;
+          this.__bodyMarginBottom = 0;
         }
         const scrollbarWidth =
           document.body.clientWidth - /** @type {number} */ (this.__bodyClientWidth);
@@ -769,15 +765,8 @@ export class OverlayController extends EventTargetShim {
         break;
       }
       case 'hide':
-        if (supportsCSSTypedObject) {
-          // @ts-expect-error types attributeStyleMap + CSS.px not available yet
-          document.body.attributeStyleMap.set('margin-right', CSS.px(this.__bodyMarginRight));
-          // @ts-expect-error types attributeStyleMap + CSS.px not available yet
-          document.body.attributeStyleMap.set('margin-bottom', CSS.px(this.__bodyMarginBottom));
-        } else {
-          document.body.style.marginRight = `${this.__bodyMarginRight}px`;
-          document.body.style.marginBottom = `${this.__bodyMarginBottom}px`;
-        }
+        document.body.style.marginRight = this.__bodyMarginRightInline || '';
+        document.body.style.marginBottom = this.__bodyMarginBottomInline || '';
         break;
       /* no default */
     }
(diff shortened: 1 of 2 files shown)


barroudjo:
The typescript error the checks detected has nothing to do with my changes. Should I fix or not ?

jorenbroekema:
I would expect errors here as well:

```js
this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
```

in the case that the value is 'auto'. Perhaps it is better to do keep the original code but add a guard that catches non-numeric values for margin?

tlouisse:
> I would expect errors here as well:
> 
> ```js
> this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
> ```
> 
> in the case that the value is 'auto'. Perhaps it is better to keep the original code but add a guard that catches non-numeric values for margin?

I think it would actually compute a value of 0px in this case, instead of returning the inline value.

Also, it's probably best to use `getPropertyValue ` for reading the value that needs to compensate for the scrollbar,
but just `document.body.style.marginRight` for _restoring_ value.

Let's assume we have:
```html
<body style="margin-left: auto; margin-right:auto;"/>
```

The margin property makes sure the body is centered in the page. When we now do our thing, we might end up with:
```html
<body style="margin-left: auto; margin-right:0px;"/>
```

Result: body is not centered anymore.


Maybe it's better to do:

**- on before-show**
```js
this.__bodyMarginRightInline = document.body.style.marginRight; // can be "" if no inline styles, or "auto" or "16px" etc...
this.__bodyMarginBottomInline = document.body.style.marginBottom;
this.__bodyMarginRightComputed = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
// etc...
```
**- on hide**
```js
document.body.style.marginRight = this.__bodyMarginRightInline; 
document.body.style.marginBottom = this.__bodyMarginBottomInline;
```

barroudjo:
> 
> 
> > I would expect errors here as well:
> > ```js
> > this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
> > ```
> > 
> > 
> > in the case that the value is 'auto'. Perhaps it is better to keep the original code but add a guard that catches non-numeric values for margin?
> 
> I think it would actually compute a value of 0px in this case, instead of returning the inline value.
> 
> Also, it's probably best to use `getPropertyValue ` for reading the value that needs to compensate for the scrollbar,
> but just `document.body.style.marginRight` for _restoring_ value.
> 
> Let's assume we have:
> 
> ```
> <body style="margin-left: auto; margin-right:auto;"/>
> ```
> 
> The margin property makes sure the body is centered in the page. When we now do our thing, we might end up with:
> 
> ```
> <body style="margin-left: auto; margin-right:0px;"/>
> ```
> 
> Result: body is not centered anymore.
> 
> Maybe it's better to do:
> 
> **- on before-show**
> 
> ```js
> this.__bodyMarginRightInline = document.body.style.marginRight; // can be "" if no inline styles, or "auto" or "16px" etc...
> this.__bodyMarginBottomInline = document.body.style.marginBottom;
> this.__bodyMarginRightComputed = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
> // etc...
> ```
> 
> **- on hide**
> 
> ```js
> document.body.style.marginRight = this.__bodyMarginRightInline; 
> document.body.style.marginBottom = this.__bodyMarginBottomInline;
> ```

Exactly right ! I think you have the cleanest solution. Let me see if I can find the time to work it into this PR.

barroudjo:
I finally found the time to implement the improvement proposed by @tlouisse !

packages/overlays/src/OverlayController.js
```diff
@@ -735,21 +735,19 @@ export class OverlayController extends EventTargetShim {
...
-          this.__bodyMarginBottom = document.body.computedStyleMap().get('margin-bottom').value;
-        } else if (window.getComputedStyle) {
+        if (window.getComputedStyle) {
           const bodyStyle = window.getComputedStyle(document.body);
           if (bodyStyle) {
             this.__bodyMarginRight = parseInt(bodyStyle.getPropertyValue('margin-right'), 10);
             this.__bodyMarginBottom = parseInt(bodyStyle.getPropertyValue('margin-bottom'), 10);
           }
+        } else {
+          this.__bodyMarginRight = 0;
```
jorenbroekema, line 749: Now that this is wrapped in an else here, this.__bodyMarginRight can be undefined, but we're using it in a computation a bit further down on line 756 and 757, so we get a type error now. Do you think it still makes sense to set these properties to 0 by default even if getComputedStyle is not on window?

barroudjo: bodyStyle is actually always defined (I don't see a case where it wouldn't be if `window.getComputedStyle` exists). So the `if (bodyStyle)` on line 744 is useless, which the compiler can't know, and hence throws an error because it believes this.__bodyMarginRight can be undefined.
I'll fix this right now by removing this `if (bodyStyle)` on line 744.

jorenbroekema: Hm yeah fair enough, I think getComputedStyle always returns a CSSStyleDeclaration. If not, a typecast could work in this case, I think it would be accceptable for document.body:

```js
const bodyStyle = /** @type {CSSStyleDeclaration} */ (window.getComputedStyle(document.body));
```

barroudjo:
Could someone approve this if it's OK ? Or tell me what should be adjusted.

jorenbroekema (approved):

jorenbroekema:
Was a little swamped this week, sorry it took a while!

Gonna merge and release this and when the pipeline is done :)

barroudjo:
> 
> 
> Was a little swamped this week, sorry it took a while!
> 
> Gonna merge and release this and when the pipeline is done :)

No problem, I understand. Your work is much appreciated !

Issue #1132, opened 2020-12-09
[choice-group] Radio/checkbox without group causes infinite loop


jorenbroekema:
## Expected behavior

```html
<lion-fieldset name="c-bar">
  <lion-radio name="c-foo" .choiceValue=${'foo'} label="I like foo"></lion-radio>
</lion-fieldset>
```

to work fine.

## Actual Behavior

Crashes browser tab due to infinite loop which I managed to track down to lit's enqueue update looping infinitely.

### Works just fine

```html
<lion-radio-group name="c-bar">
  <lion-radio name="c-foo" .choiceValue=${'foo'} label="I like foo"></lion-radio>
</lion-radio-group>
```

### Crashes
```html
<lion-fieldset name="c-bar">
  <lion-radio name="c-foo" .choiceValue=${'foo'} label="I like foo"></lion-radio>
</lion-fieldset>
```

### Removing the fieldset name attribute resolves the crash.

```html
<lion-fieldset>
  <lion-radio name="c-foo" .choiceValue=${'foo'} label="I like foo"></lion-radio>
</lion-fieldset>
```

So might be some infinite looping stuff going in the formElements / collection or something. @tlouisse  maybe you have ideas 😅 ?

## Root cause

Not sure yet, I tried to debug it but got pretty lost..

PR #1193, opened 2021-01-26
fix(form-core): fix infinite loop syncing name to parent

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1132

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceInputMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceInputMixin.js
@@ -119,8 +119,7 @@ const ChoiceInputMixinImplementation = superclass =>
         // @ts-expect-error
         this._parentFormGroup.name !== this.name
       ) {
-        // @ts-expect-error not all choice inputs have a name prop, because this mixin does not have a strict contract with form control mixin
-        this.name = changedProperties.get('name');
+        this._syncNameToParentFormGroup();
       }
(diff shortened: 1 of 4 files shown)


packages/form-core/src/choice-group/ChoiceInputMixin.js
```diff
@@ -232,6 +231,20 @@ const ChoiceInputMixinImplementation = superclass =>
...
 
+    /**
+     * Override this in case of extending ChoiceInputMixin and requiring
+     * to sync differently with parent form group name
+     * Right now it checks tag name match where the parent form group tagname
+     * should include the child field tagname ('checkbox' is included in 'checkbox-group')
+     */
+    _syncNameToParentFormGroup() {
+      // @ts-expect-error not all choice inputs have a name prop, because this mixin does not have a strict contract with form control mixin
+      if (this._parentFormGroup.tagName.includes(this.tagName)) {
```
gerjanvangeest, line 242: New code, new tests?

jorenbroekema: The passing condition is tested in the choice input mixin already (e.g. if I invert this condition, some tests fail)

I will also write a test for failing the condition (means that name does not get synced)

gerjanvangeest (approved):

PR #1195, opened 2021-01-28
fix(overlays): reinsert missing tab detect el, don't error on disconnect

jorenbroekema:
## What I did

1. Observe content node children, reinsert tab detection element if children have changed but tab detection element is missing 
2. Guard disconnect tab detection el to check if the el is a child of the root element. This happens only if content node children change is observed and the tab detection element is identified as missing. It will disconnect and reconnect a containFocus, essentially re-inserting the tab detection element

### DIFF
--- a/packages/overlays/src/utils/contain-focus.js
+++ b/packages/overlays/src/utils/contain-focus.js
@@ -77,6 +77,8 @@ export function containFocus(rootElement) {
   const initialFocus = focusableElements.find(e => e.hasAttribute('autofocus')) || rootElement;
   /** @type {HTMLElement} */
   let tabDetectionElement;
+  /** @type {MutationObserver} */
+  let rootElementMutationObserver;
 
   // If root element will receive focus, it should have a tabindex of -1.
   // This makes it focusable through js, but it won't appear in the tab order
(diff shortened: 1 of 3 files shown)


packages/overlays/src/utils/contain-focus.js
```diff
@@ -101,6 +101,7 @@ export function containFocus(rootElement) {
   function createHelpersDetectingTabDirection() {
     tabDetectionElement = document.createElement('div');
     tabDetectionElement.style.display = 'none';
+    tabDetectionElement.setAttribute('is-tab-detection-element', '');
```
tlouisse, line 104: Could we prefix with `data-`, so it's w3c validator compliant? => https://validator.w3.org/nu/#textarea
(non data attrs are only allowed on custom elements)

packages/overlays/src/OverlayController.js
```diff
@@ -1050,6 +1050,31 @@ export class OverlayController extends EventTargetShim {
     if (this.__hasActiveTrapsKeyboardFocus) {
       return;
     }
+
+    // Reset contain focus util when contentNode content has changed
+    // and tab detection element is missing
+    if (!this.trapsKeyboardObserver) {
```
tlouisse, line 1056: Could this logic be moved into `src/utils/contain-focus.js`, so that it's completely self contained as a util?

jorenbroekema: Updated

packages/overlays/src/utils/contain-focus.js
jorenbroekema: Ah true, updated

tlouisse (approved):

packages/overlays/src/OverlayController.js
tlouisse: Nice, test as well if I would be nitpicking. But I approved :)

jorenbroekema: I added one test in OverlayController test to kinda end to end test it by doing
```
ctrl.contentNode.innerHTML = `my content`;
```
and then verifying that the tab detection el is still there in the contentNode (reinserted)

Or do you mean a different test?

tlouisse: There was a test purely for the util if I'm correct

tlouisse: https://github.com/ing-bank/lion/blob/master/packages/overlays/test/utils-tests/contain-focus.test.js

jorenbroekema: Ahh hm okay I moved the test to contain-focus.test.js since we should indeed unit test the contain-focus part and not only check if it works in conjunction with OverlayController

PR #1198, opened 2021-02-03
fix: transpile out optional chaining on prepublish

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1169


jorenbroekema:
tons of type errors created by babel transform, requiring us to make our users do skipLibCheck in their TS config. This approach is not gonna work :\

PR #1199, opened 2021-02-05
refactor: add package exports entries

manolakis:
## What I did

1. Add the `exports` entry in each `package.json`, exposing the public parts of each package and hiding the private ones.
2. Adapt broken code that accessed private elements of other packages.
3. Unify the way of publishing and access to side-effects.

This commit does not change the behavior of the components but could cause some access issues in those consumers who are accessing internal implementation details.

This change also allows better bundle optimizations by library clients.

### DIFF
--- a/.changeset/red-steaks-brake.md
+++ b/.changeset/red-steaks-brake.md
@@ -1,0 +1,47 @@
+---
+'babel-plugin-extend-docs': minor
+'providence-analytics': minor
+'remark-extend': minor
+'@lion/accordion': minor
+'@lion/ajax': minor
+'@lion/button': minor
+'@lion/calendar': minor
+'@lion/checkbox-group': minor
+'@lion/collapsible': minor
+'@lion/combobox': minor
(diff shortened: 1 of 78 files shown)


manolakis:
I didn't include the changeset because I'm not sure if we should consider it as a breaking change or just a patch 🤷

jorenbroekema:
I think this is not a breaking change on lion, only a breaking change if lion is used in conjunction with tools that respect exports field as the single source of truth for imports. Meaning that some of your imports might be broken when upgrading. I think most developers using @web/dev-server are not aware of this, so to be safe I would appreciate doing a minor release for all packages and explaining how this could be breaking in the changelog.

jorenbroekema:
Gonna wait a bit until Ajax PRs are done as that changes some of the exports, I will then rebase and fix this branch and getthis merged :)

jorenbroekema (approved):

jorenbroekema:
Thanks 🎉

PR #1207, opened 2021-02-11
docs: add custom elements manifest

thepassle:
## What I did

1. Add `custom-elements.json` for packages. There might still be quirks in the output `custom-elements.json`, but as discussed with @daKmoR : lets iterate 🙂  Please let me know if you do find any issues/bugs/weird output


package.json
```diff
@@ -10,6 +10,7 @@
     "build:docs": "wca analyze \"packages/tabs/**/*.js\"",
     "build:types": "tsc -p tsconfig.build.types.json",
     "bundlesize": "rollup -c bundlesize/rollup.config.js && bundlesize",
+    "custom-elements-manifest": "node scripts/workspaces-scripts.mjs run custom-elements-manifest",
```
thepassle, line 13: ☝️ this should be: `custom-elements-manifest analyze`, otherwise it only outputs the CLI options 😅

thepassle: Also you might have to exclude the `docs/` folder, I forget which package it was exactly, but `.js` files from the docs folder ended up in the `custom-elements.json` file, and for some it even errored on those. I can take a look later (tomorrow morning) if you want

you can exclude like so: `custom-elements-manifest analyze --exclude "docs/**/*.js"`

jorenbroekema: Ah yeah so this runs `npm run custom-elements-manifest` from the `/packages/package/..` root folder, and there we have
```json
"scripts": {
  "custom-elements-manifest": "custom-elements-manifest analyze --exclude \"docs/**/*.js\""
}
```
 unless I messed something up.

jorenbroekema (approved):

package.json
thepassle: Ah yes ofcourse, my bad!

PR #1212, opened 2021-02-15
Delete superfluous code

erikkroes:
## What I did

1. Remove a superfluous role
2. Remove a superfluous span

### DIFF
--- a/packages/calendar/src/utils/dataTemplate.js
+++ b/packages/calendar/src/utils/dataTemplate.js
@@ -21,7 +21,7 @@ export function dataTemplate(
             aria-labelledby="month year"
           >
             <thead>
-              <tr role="row">
+              <tr>
                 ${weekdaysShort.map(
                   (header, i) => html`
                     <th
--- a/packages/calendar/src/utils/dayTemplate.js
+++ b/packages/calendar/src/utils/dayTemplate.js
@@ -72,7 +72,7 @@ export function dayTemplate(day, { weekdays, monthsLabels = defaultMonthLabels }
         ?current-month=${day.currentMonth}
         ?next-month=${day.nextMonth}
       >
-        <span class="calendar__day-button__text"> ${day.date.getDate()} </span>
+        ${day.date.getDate()}
       </button>
     </td>
   `;
--- a/packages/calendar/test/lion-calendar.test.js
+++ b/packages/calendar/test/lion-calendar.test.js
@@ -671,9 +671,7 @@ describe('<lion-calendar>', () => {
               aria-pressed="false"
               past=""
               current-month="">
-              <span class="calendar__day-button__text">
                 30
-              </span>
             </button>
           `);
         });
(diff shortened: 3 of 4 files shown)


jorenbroekema:
- [ ] Fix failing tests (DOM diffs now fail, which makes sense)
- [ ] Create a changeset for a patch
- [ ] Squash into 1 commit?

packages/calendar/src/utils/dayTemplate.js
```diff
@@ -72,7 +72,7 @@ export function dayTemplate(day, { weekdays, monthsLabels = defaultMonthLabels }
         ?current-month=${day.currentMonth}
         ?next-month=${day.nextMonth}
       >
-        <span class="calendar__day-button__text"> ${day.date.getDate()} </span>
```
gerjanvangeest, line 75: We use this span internally for styling.

tlouisse: Indeed. The sole reason for this span is a styling hook and we need it ourselves in our extension layer :)

packages/calendar/src/utils/dataTemplate.js
```diff
@@ -21,7 +21,7 @@ export function dataTemplate(
             aria-labelledby="month year"
           >
             <thead>
-              <tr role="row">
```
tlouisse, line 24: I can understand the confusion: the role is not needed when we don't apply css. 
By adding these roles,  we allow the extension layer to style without breaking semantics:
if the `display` property of table cells are changed (to flex or whatever), semantics are lost. 

@erikkroes do you think we should document this somewhere?

erikkroes: We don't change the `display` anywhere? I think it's the responsibility of who does.

tlouisse: We don't do it ourselves, but we aim for a dev experience like this:
- only add styles on top, no further logic. All a11y is handled by Lion 

This styling could either be added via an extension layer or via shadow parts. In the shadow parts case, there would be no way to change the html.
I experienced this myself: when I made the Material Design datepicker, I actually needed to put flex properties here:an external dev not knowing these implementation details, would have broken a11y without knowing when there wouldn't have been a role in the markup. Tools like axe probably don't spot these errors (since they would have to get computed styles for every element in the dom).

Changing html would be tricky, error prone and should be avoided when there's no need for it.

@erikkroes I hope you agree with the above? Shall I add a comment in the source code explaining this? I'm in a 'cleanup mode' wrt open issues and pull requests  :)

Some comments like:

```
## Why put [role="row"] on a tr element?

By default the platform offers these semantics implicitly via the tr element, but overriding display styles - imagine an extension layer putting 'display:flex' on this element (a very likely scenario) - break those semantics.
```

daKmoR:
I'm going to close this as it seems there is a need to discuss this more upfront. Feel free to open an issue for each individual topic 🤗

PR #1213, opened 2021-02-15
feat: add preprocessor hook and example amount preprocessor

jorenbroekema:


packages/form-core/test-suites/FormatMixin.suite.js
```diff
@@ -401,6 +410,38 @@ export function runFormatMixinSuite(customConfig) {
...
+          processedVal = val.replace(/[o]/g, '');
+        }
+
+        const el = /** @type {FormatClass} */ (await fixture(html`
+          <${elem} .preprocessor=${preprocessorSpy}>
+            <input slot="input" value="${val}">
+          </${elem}>
+        `));
+
+        if (cfg.tagString?.startsWith('lion-input-date')) {
```
tlouisse, line 432: Would fail if the suite would be run for an extension

packages/form-core/test-suites/FormatMixin.suite.js
```diff
@@ -300,6 +307,8 @@ export function runFormatMixinSuite(customConfig) {
 
         el.formattedValue = 'raw';
         expect(parserSpy.called).to.equal(true);
+        el.dispatchEvent(new CustomEvent('user-input-changed'));
+        expect(preprocessorSpy.called).to.equal(true);
```
tlouisse, line 311: Maybe it would be nice to test the order of function calls here as well, since - especially for the preprocessor - it's vital that it runs before the rest

packages/form-core/test-suites/FormatMixin.suite.js
```diff
@@ -401,6 +410,38 @@ export function runFormatMixinSuite(customConfig) {
...
+        `));
+
+        if (cfg.tagString?.startsWith('lion-input-date')) {
+          // For date and datepicker the stringified modelValue will be quite different
+          // from the original view value, due to date parsing etc.
+          // So for such inputs we don't test this preprocessor.
+          return;
+        }
+        expect(preprocessorSpy.callCount).to.equal(1);
+        expect(`${el.modelValue}`).to.equal(`${processedVal}`);
```
tlouisse, line 439: So we basically want to know whether modelValue did not become `Unparseable`?

It is not something that inherently is true for preprocessors, right? (it could be the outcome though, depending on the parser)

Maybe we should test the viewValue (`.value`) instead?

Like:

```js
let val = generateValueBasedOnType({ viewValue: true }) || 'init-value';
val = val + '-'

const preprocessorSpy = sinon.spy(value => value.replace(/\-$/g, ''));

```

packages/form-core/test-suites/FormatMixin.suite.js
jorenbroekema: Hm thats true, I wonder what a good way to do this is in the suite here.. Any ideas?

tlouisse: Maybe check for `modelValue instanceof Date`?
Or use the test below that focuses on viewValues and the fact that the preprocessed value is passed on to the parser (which in turn determines whether modelValue will be invalid or not, depending on its implementation)

tlouisse (approved):

Issue #1209, opened 2021-02-11
[combobox] check `_showOverlayCondition` on keyup instead of keydown


tlouisse:
## Expected behavior

`_showOverlayCondition` can be dependent on currentValue.length. This is updated on keyup, not on keydown. I expect currentValue.length to be correct

## Actual Behavior

currentValue.length has the previous value.

## Fix

In LionCombobox.js, change keydown to keyup:

```js
  _setupOpenCloseListeners() {
    super._setupOpenCloseListeners();
    this._inputNode.addEventListener('keyup', this.__requestShowOverlay);
  }
```

## Test

Add test like:

```js
  MyCb extends LionCombobox {
    _showOverlayCondition({ currentValue }) {
      return currentValue > 3;
    }
  }
  
  mimicUserInput(el, 'aaa');
  expect(el.opened).to.be.false;

  mimicUserInput(el, 'aaaa');
  expect(el.opened).to.be.true;
```

PR #1214, opened 2021-02-15
fix(combobox): displayoverlay on keydown event

palash2601:
## What I did

1. fix for https://github.com/ing-bank/lion/issues/1209

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -389,8 +389,11 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    */
   // eslint-disable-next-line class-methods-use-this
   _showOverlayCondition({ lastKey }) {
+    if (!lastKey) {
+      return this.opened;
+    }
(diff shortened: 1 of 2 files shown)


tlouisse:
Nice 👍  Could you also add mimicKeypress for listboxMixin suite tests (because the combobox runs them)?

packages/combobox/src/LionCombobox.js
```diff
@@ -812,6 +815,8 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
       })
     ) {
       this.opened = true;
+    } else {
+      this.opened = false;
```
tlouisse, line 819: For brevitiy, could you change this method to:
```js
this.opened = this._showOverlayCondition({
        lastKey: ev && ev.key,
        currentValue: this._inputNode.value,
      })
```

tlouisse:
Finished pr here: https://github.com/ing-bank/lion/pull/1288

Issue #1219, opened 2021-02-16
No data tag name on lion-validation-feedback


rvt:
## Expected behavior

`<lion-validation-feedback-14964 data-tag-name="lion-validation-feedback" aria-live="polite" slot="feedback"</lion-validation-feedback-14964>
`
## Actual Behavior

`<lion-validation-feedback-14964 aria-live="polite" slot="feedback"</lion-validation-feedback-14964>
`

## Additional context

When I want to run tests using 
```
expect(element).shadowDom.to.equalSnapshot({
       ignoreTags: ['lion-validation-feedback']
     });
```

I cannot make them work because the test framework cannot find a `data-tag-name` on the elements

PR #1223, opened 2021-02-17
fix: add data-tag-name attribute on scoped slottables

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1219


tlouisse:
"Nice" :)
But we will only have this temporarily until we use the polyfill, I guess...

I'll approve, but maybe make two commits instead of one?

tlouisse (approved):

jorenbroekema:
Yep it will probably all change a fair bit when we're on lit 3. But that will take a while I think

PR #1232, opened 2021-02-22
feat: introduce transfer component to allow users

naveen2593kumar:
## What I did
* introducing a brand new component which allows users to shift/transfer one or more items between lists
Idea initiated at https://github.com/ing-bank/lion/discussions/1149

## A working screenshot of storybook
Screenshot 2021-02-19 at 4 11 39 PM

## Test coverages
Screenshot 2021-02-19 at 4 11 01 PM

### DIFF
--- a/.changeset/twelve-terms-knock.md
+++ b/.changeset/twelve-terms-knock.md
@@ -1,0 +1,5 @@
+---
+'@lion/transfer': patch
+---
+
+introducing transfer component which allows the user to shift/transfer one or more items between lists
--- a/packages/transfer/CHANGELOG.md
+++ b/packages/transfer/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Change Log
+
+# 0.0.1 (2021-02-22)
+
+### Features
+
+- introducing transfer component which allows the user to shift/transfer one or more items between lists ([7de1b96](https://github.com/ing-bank/lion/commit/7de1b96))
--- a/packages/transfer/README.md
+++ b/packages/transfer/README.md
@@ -1,0 +1,41 @@
+# Transfer
+
+`lion-transfer` component allows the user to shift/transfer one or more items between lists.
+
+You cannot use interactive elements inside the options. Avoid very long names to
+facilitate the understandability and perceivability for screen reader users. Sets of options
+where each option name starts with the same word or phrase can also significantly degrade
+usability for keyboard and screen reader users.
+
+```js script
+import { html } from 'lit-html';
+import { Required } from '@lion/form-core';
+import '@lion/listbox/lion-option.js';
+
+import './lion-transfer.js';
+
+export default {
+  title: 'Others/Transfer',
+};
+```
+
+```js preview-story
+export const main = () => html`
+  <lion-transfer>
+    <div slot="leftList">
+      <lion-option .choiceValue="Apple">Apple</lion-option>
+      <lion-option .choiceValue="Artichoke">Artichoke</lion-option>
+      <lion-option .choiceValue="Asparagus">Asparagus</lion-option>
+    </div>
+    <button slot="transferToLeftActionSlot">To Left &lt;</button>
+    <button slot="transferToRightActionSlot">To Right &gt;</button>
+    <div slot="rightList">
+      <lion-option .choiceValue="Banana">Banana</lion-option>
+      <lion-option .choiceValue="Beets">Beets</lion-option>
+      <lion-option .choiceValue="Bell pepper">Bell pepper</lion-option>
+      <lion-option .choiceValue="Broccoli">Broccoli</lion-option>
+      <lion-option .choiceValue="Brussels sprout">Brussels sprout</lion-option>
+    </div>
+  </lion-transfer>
+`;
+```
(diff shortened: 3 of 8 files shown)


naveen2593kumar:
@jorenbroekema  can you review it please...

jorenbroekema:
Hey @naveen2593kumar , since I was a bit preoccupied the last 2 days, I believe @tlouisse  is going to post some feedback here today, he has also discussed this PR with 2 other people in our team :) so the review is coming!

tlouisse:
Hi @naveen2593kumar,

Thanks a lot for helping us out 💪 
I think this component would be a really valuable addition, especially since it brings a lot of value to consumers if all accessibility described here would be automatically handled: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html

We have a few remarks below that would make it a better fit with our current systems, conventions and Lion philosophies in general.
Hopefully you agree with them and want to do some finetuning :)


### Extendibility / reusability

Since lion components are built with the mindset that they should be highly configurable and extendible and serve as the fundament for as many subcomponents / design patterns as possible, we would like to propose a few api adjustments.

The below api tries to cater for the following situations:
    1. How do I build https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html
    2. What if I want to configure behavior of the listbox? (single select example in link above)
    3. What If I want to change the look and feel of both listboxes?
    4. What if I want to add more buttons? 

```html
<lion-transfer>
  <lion-listbox slot=“source”>
      ${repeat(…)}
  </lion-listbox>
  <lion-listbox slot=“target” name=“thingIwantToSubmit”>
      ${repeat(…)}
  </lion-listbox>             
  <button slot=“to-target”></button>
  <button slot=“to-source”></button>
</my-transfer>
```

An extending party could then easily do:

```html
<my-transfer>
  <my-transfer-listbox-source slot=“source”>
      ${repeat(…)}
  </my-transfer-listbox-source>
  <my-transfer-listbox-target slot=“target” name=“thingIwantToSubmit”>
      ${repeat(…)}
  </my-transfer-listbox-target>             
  <button slot=“to-target”></button>
  <button slot=“to-source”></button>
</my-transfer>
```

### A11
See https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html
 - live region
 - keyboard interactions up/dow/add/delete
 - keyboard shortcuts for buttons + `aria-keyshortcuts`
 - nice to have: group with role=toolbar when 4 buttons present
 - Basically everything else mentioned here under Accessibility :)  
 https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html

### Form system compatibility
- No `modelValue` needed, since lion-transfer is not a `FormControl`. Users can interact directly with the listboxes.
- users can add a name attribute to the listbox (see  `name=“thingIwantToSubmit”`). Usually the right listbox is the 'result' listbox and of interest to the backend. Currently the name attr is required on all FormControls, in the future, we should (just like in the platform allow to omit it, so only those with the name attr are registered (and submitted)).
                
### UX (nice to haves)
- Can I rearrange items? (Could be future addition)
- Should drag and drop work (nice to have)?
- Add multi select with keyboard (nice to have)

### Other things to think about
- Do click listeners on slots catch space/enter on buttons? If it works it's ok, otherwise you could add listeners programatically to the buttons
- slot names are kebab-cased        
- performance overhead: Moving around options is not a good practice. We currently do it ourselves only to create a forward compatible api as long as AOM is not in place. With the api above we would not need to move around options.
- Interoperability lit: did not test this myself, but will it work with the `repeat` directive and `map`?
- it would be nice to support (as can be found in demos):

```html
  <button slot=“all-to-target”></button>
  <button slot=“all-to-source”></button>
```

naveen2593kumar:
thanks @tlouisse,

That's alot to fix, but I am on it..... I may ask if I need some help or better way of implementation.

tlouisse:
Yeah it was quite a list, but a lot could also be added as future improvements.

List looks smaller if you formulate it like this :)
- change the template so that the listbox is provided by the end user (and remove the moving around of options)
- Implement the a11y requirements from the spec: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html. (This might be most of the work. Let us know if you need help with that)
- remove modelValue property

tlouisse:
For the live region(see https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/js/listbox-rearrangeable.js), you could for now hardcode the English texts:

- `Added ${addedOption.innerText} to ${this.targetName || this._targetListbox.fieldName}.`

- `Removed ${addedOption.innerText} from ${this.targetName || this._targetListbox.fieldName}.`


- `Added all options to ${this.targetName || this._targetListbox.fieldName}.`


- `Removed all options from ${this.targetName || this._targetListbox.fieldName}.`

We can in the meantime request translations for multiple languages

tlouisse:
hi @naveen2593kumar any progress on this? Let us know if we can help out :)

gerjanvangeest:

Issue #1233, opened 2021-02-22
modelValue setter not working for choice-groups


jorenbroekema:
## Expected behavior

Setting .modelValue on radio-group or other choice-group fields should set the child checked properly, even if the modelValue is an object or array

## Actual Behavior

Setting the modelValue for these fields doesn't do anything.

## Additional context

```js
const checkCondition = (el, val) => el.choiceValue === val;
```

This condition fails if the value is object or array because the equality check becomes a reference value check, and will fail. We should check if the type is 'object' and do a JSON.stringify comparison instead for these cases.

Issue #1058, opened 2020-10-28
[lion-switch] value within form is parsed in modelValue under "undefined" 


davidoff31:
## Expected behavior

When lion-switch is used within form its value should be parsed in modelValue with the name as key. This happened until last release.

## Actual Behavior

The value is still parsed, but with key "undefined".

## Additional context

Looking at documentation doesn't seem as if the switch is expected to be used in forms, still if this is the case it should not pollute the form modelValue.

tlouisse:
The switch is designed to work with and without form (just like every FormControl), so this is a 'valid bug' :)
For the switch it's just more likely to be used outside of the form, that's why it's in the Buttons section in docs.

It should actually be added to our 'umbrella form' here: https://lion-web-components.netlify.app/?path=/docs/forms-features-overview--main 

I will mark this as a bug, thanks for reporting.

jorenbroekema:
Added it in the umbrella form demo in the PR mentioned above. I am not able to reproduce the bug there so I will assume it got fixed. If not, feel free to reopen with a reproduction

PR #1234, opened 2021-02-22
Fix/mvalue choice group

jorenbroekema:
## What I did

1. add switch demo
2. Fix choice group model value setter for choice children with complex model values

fixes https://github.com/ing-bank/lion/issues/1233
closes https://github.com/ing-bank/lion/issues/1058

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -44,7 +44,12 @@ const ChoiceGroupMixinImplementation = superclass =>
        * @param {ChoiceInputHost} el
        * @param {any} val
        */
-      const checkCondition = (el, val) => el.choiceValue === val;
+      const checkCondition = (el, val) => {
+        if (typeof el.choiceValue === 'object') {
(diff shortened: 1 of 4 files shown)


packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -44,7 +44,12 @@ const ChoiceGroupMixinImplementation = superclass =>
        * @param {ChoiceInputHost} el
        * @param {any} val
        */
-      const checkCondition = (el, val) => el.choiceValue === val;
+      const checkCondition = (el, val) => {
+        if (typeof el.choiceValue === 'object') {
```
tlouisse, line 48: When is this an object again? In our extension layer SelectRich?

tlouisse (approved):

packages/form-core/src/choice-group/ChoiceGroupMixin.js
jorenbroekema: Yep :) our select rich extension indeed!

tlouisse (approved):

Issue #1211, opened 2021-02-11
[switch-button] syncing checked attr differs from property


tlouisse:
switch-button has a dirty check for `checked` in `requestUpdateInternal`:

```js
if (this.isConnected && name === 'checked' && this.checked !== oldValue) {
```

This works fine: 

```js
<lion-switch-button .checked="${this.themes.includes('dark')}" class="tool-button" @checked-changed=${({ currentTarget}) => { this.themes = currentTarget.checked ? ['dark'] : ['light']; }} >
```
✅ This returns false, since `this.checked === oldValue`

```js
<lion-switch-button checked="${this.themes.includes('dark')}" class="tool-button" @checked-changed=${({ currentTarget}) => { this.themes = currentTarget.checked ? ['dark'] : ['light']; }} >
```
❌  This returns true, because `this.checked !== oldValue`  ...?

Notice the subtle difference between `.checked` and `checked`.
This might be a more generic lit-el problem.

tlouisse:
> This should be reconsidered after migration to LitEl 3

Unfortunately this is still an issue in  lit 2/3 (c.q lit-el 3/4). 
I updated the opening post with more solution directions

PR #1240, opened 2021-02-23
chore: update to latest eslint

gerjanvangeest:
## What I did

1. Update to latest eslint versions
2. Fixed eslint errors


.storybook/main.js
```diff
@@ -4,7 +4,7 @@ const path = require('path');
 module.exports = {
   stories: [
     '../{packages,packages-node}/!(ajax)*/README.md',
-    '../{packages,packages-node}/*/docs/*.md',
+    '../{packages,packages-node}/!(ajax)*/docs/*.md',
```
jorenbroekema, line 7: fixed https://github.com/ing-bank/lion/pull/1228 so maybe leave this out otherwise we get merge conflict

packages/form-core/test-suites/FormatMixin.suite.js
```diff
@@ -367,7 +367,7 @@ export function runFormatMixinSuite(customConfig) {
 
         const el = /** @type {FormatClass} */ (await fixture(html`
           <${elem} .formatter=${formatterSpy}>
-            <input slot="input" value="${generatedViewValue}">
+            <input slot="input" .value="${generatedViewValue}">
```
tlouisse, line 370: I think it's fine in this particular pr, but in general we should be careful with blindly replacing props and attrs.
Especially for the input, `[value]` is only synced at the initialization of a component (future attr updates don't take effect).

There's also slight discrepancies in behavior in LitElement: https://github.com/ing-bank/lion/issues/1211

jorenbroekema: Yeah I would argue we should use this rule to "warn", not "error", because sometimes it's actually intended and totally fine to set an attribute in your template to a JS variable. For `.value` it works because value is an actual property on HTMLInputElement, but that's a coincidence. If you did it for `.foo` you would not see attribute foo in the DOM.

Updated the PR

jorenbroekema: Oh wait this linting rule only applies to "value" attribute, never mind, then I think the rule makes sense.

jorenbroekema (approved):

Issue #1237, opened 2021-02-23
Success feedback cycles on blur and focus


thatdudemanguy:
## Expected behavior

A success notification that either does not change on focus/blur or does not get displayed again after initial success. 

## Actual Behavior

When the success feedback is displayed, the component cycles through its translations when a user focuses/blurs the input field without changing its input value.

Issue #1236, opened 2021-02-23
Success feedback showing on initial field input, not only after adjusting. 


thatdudemanguy:
## Expected behavior

When a user enters a faulty value into an input; the validation displays an error notification. 
Thereafter the client corrects the input and the validation shows a success notification instead of the previous error notification. 

When a user enters a correct value into an input; the validation displays nothing.

Example 1: 
case: input number
1. user inputs "foo"
2. feedback displays "please use a number"
3. user inputs 1
4. feedback displays "correct"

Example 2:
case: input number
1. user inputs 1
2. feedback displays nothing

## Actual Behavior

The user enters a correct value into an input and the success notification displays.
A success notification should not show here, as it is the first entry the user did. 

Example 2:
case: input number
1. user inputs 1
2. feedback displays "correct"

PR #1254, opened 2021-03-10
fix(form-core): only show success after recovery from shown feedback

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1236
fixes https://github.com/ing-bank/lion/issues/1237

### DIFF
--- a/packages/form-core/src/validate/resultValidators/DefaultSuccess.js
+++ b/packages/form-core/src/validate/resultValidators/DefaultSuccess.js
@@ -17,18 +17,17 @@ export class DefaultSuccess extends ResultValidator {
    *
    * @param {Object} context
    * @param {Validator[]} context.regularValidationResult
-   * @param {string} context.prevShownValidationFeedback
+   * @param {Validator[]} context.prevValidationResult
+   * @param {Validator[]} context.prevShownValidationResult
    * @returns {boolean}
    */
   // eslint-disable-next-line class-methods-use-this
-  executeOnResults({ regularValidationResult, prevShownValidationFeedback }) {
+  executeOnResults({ regularValidationResult, prevShownValidationResult }) {
     const errorOrWarning = /** @param {Validator} v */ v =>
       v.type === 'error' || v.type === 'warning';
     const hasErrorOrWarning = !!regularValidationResult.filter(errorOrWarning).length;
+    const hasShownErrorOrWarning = !!prevShownValidationResult.filter(errorOrWarning).length;
 
-    return (
-      !hasErrorOrWarning &&
-      (prevShownValidationFeedback === 'error' || prevShownValidationFeedback === 'warning')
-    );
+    return !hasErrorOrWarning && hasShownErrorOrWarning;
   }
 }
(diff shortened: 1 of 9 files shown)


packages/form-core/src/validate/resultValidators/DefaultSuccess.js
```diff
@@ -17,15 +17,18 @@ export class DefaultSuccess extends ResultValidator {
...
+  executeOnResults({ regularValidationResult, prevShownValidationFeedback }) {
     const errorOrWarning = /** @param {Validator} v */ v =>
       v.type === 'error' || v.type === 'warning';
     const hasErrorOrWarning = !!regularValidationResult.filter(errorOrWarning).length;
-    const prevHadErrorOrWarning = !!prevValidationResult.filter(errorOrWarning).length;
-    return !hasErrorOrWarning && prevHadErrorOrWarning;
+
+    return (
+      !hasErrorOrWarning &&
+      (prevShownValidationFeedback === 'error' || prevShownValidationFeedback === 'warning')
```
tlouisse, line 31: Nice fix. Definitely makes sense to supply prevShownValidationFeedback and check for that indeed 👍 

But:
- could we keep it an array, like `prevShownValidationResult`?

In this way, we make the presentation of the types independent from the ResultValidator logic.
(imagine __prioritizedResult was ordered by custom logic that didn't put error on top, but some custom type), then below would fail.

```js
if (this.__prioritizedResult.length > 0) {
            this.__prevShownValidationFeedback = this.__prioritizedResult[0].type;
          }
```

Also, it can be nice to have more detailed info about the previous validation results, in case a ResultValidator needs to do a comparison between latest and previous results (more use cases than just SucessValidators). 
Also for SuccessValidators it would allow you to provide more directed feedback.

- for backwards compatibility, keep `prevValidationResult`? This will be almost the same as `prevShownValidationResult` of course, except it contains info about error states 'before blur'. We could consider to remove it in v1

jorenbroekema: prevValidationResult is a private property of the validate mixin right? Don't really see why we should pass it to the ResultValidator if it doesn't use it.

As for prevShownValidationResult, you never show more than 1 type of feedback at once, so we can stick to just a string of the most recently shown feedback type right? I'm okay with making it a type Validator so we have more information, rather than just "type", but currently we don't have a use case for it and I don't want to overengineer things. We can change it later if we find that we need it?

jorenbroekema: Ah and I don't see why that code would fail if the validator on top has a custom type?? Your result validator may not show then, but that's intended. the result validator only shows after warning / error, if you make a custom type go on top, then don't show the result validator.

tlouisse: - The current logic is based on the assumption there is one result in `__prioritizedResult `, but if you configure `_prioritizeAndFilterFeedback()` for your custom case it can also be:

  [0] ValidatorX with type 'custom'
  [1] ValidatorX with type 'error'
  [2] ValidatorX with type 'warning'
  And now the validation feedback shows all three of them, but you miss the fact there was an error or warning.

- A use case for needing info about prevValidations:
You want a success message that gives specific feedback about the fact that you resolved the error for maxLength. Based on just the current result, you would never know...

tlouisse: > prevValidationResult is a private property of the validate mixin right? Don't really see why we should pass it to the ResultValidator if it doesn't use it.

But I think we should keep all signatures intact. Because whatever we pass on to the `execute` or `executeOnResults, is considered public

tlouisse (approved):

Issue #920, opened 2020-09-04
[dialog] overlay closed on mouseup


MathieuPuech:
## Describe the bug

When you release the mouse click outside the dialog with the option to hide on outside click, the dialog is closing.

## To Reproduce

1. Go to https://lion-web-components.netlify.app/?path=/story/overlays-dialog--styling-content
2. Click on Button
3. Click to select text but release the mouse click outside the dialog
4. Dialog is closed

## Expected behavior

Dialog should not close when the click was initiated in the dialog

## Screenshots

![2020-09-04_14-12-02](https://user-images.githubusercontent.com/9367152/92273054-b23da900-eeb8-11ea-9207-632739d2f0a9.gif)

daKmoR:
you are really good and finding all these edge cases 💪

keep them coming 🤗

not sure how fast we will be in resolving those 🙈
but knowing about them is the first step 👌

MathieuPuech:
It’s ok if it’s not fixed fast. I just document issues I found when using the components. If I have time to fix them I will do of course.

tlouisse:
> you are really good and finding all these edge cases 
> 
> keep them coming 

Amen. Well spotted :muscle: 

So what should be the behavior, you think? Only close the dialog if both mousedown and mouseup are catched outside?

MathieuPuech:
We have 2 choices:
- close on mousedown
- close on mouseup but with the position of the mouse on mousedown

I've no idea what is best, is there any issue to close on mousedown directly?

tlouisse:
I thought that was not default platform behavior, apparently it is :) (tested `<select>` and `<datalist>` on Chrome Linux)
Then it probably will be a matter of swapping the current click handler in OverlayController with a mousedown handler :+1:

jorenbroekema:
Yeah I agree, mousedown seems like the most practical choice 👍

PR #1257, opened 2021-03-11
fix: overlay closed on mouseup

Knorgias:
## What I did

1. Updated Overlaycontroller to add/remove mousedown events instead of click for "outside clicks"
2. Added & updated relevant tests

fixes https://github.com/ing-bank/lion/issues/920

### DIFF
--- a/packages/input-datepicker/test/lion-input-datepicker.test.js
+++ b/packages/input-datepicker/test/lion-input-datepicker.test.js
@@ -2,6 +2,7 @@ import { LionCalendar, isSameDate } from '@lion/calendar';
 import { html, LitElement } from '@lion/core';
 import { IsDateDisabled, MaxDate, MinDate, MinMaxDate } from '@lion/form-core';
 import { aTimeout, defineCE, expect, fixture as _fixture, nextFrame } from '@open-wc/testing';
+import { mimicClick } from '@lion/overlays/test-helpers';
 import sinon from 'sinon';
 import { setViewport } from '@web/test-runner-commands';
 import '@lion/input-datepicker/define';
@@ -98,13 +99,13 @@ describe('<lion-input-datepicker>', () => {
       expect(elObj.overlayController.isShown).to.equal(false);
     });
 
-    it('closes the calendar via outside click', async () => {
+    it('closes the calendar via outside click event', async () => {
       const el = await fixture(html`<lion-input-datepicker></lion-input-datepicker>`);
       const elObj = new DatepickerInputObject(el);
       await elObj.openCalendar();
       expect(elObj.overlayController.isShown).to.equal(true);
 
-      document.body.click();
+      mimicClick(document.body);
       await aTimeout(0);
       expect(elObj.overlayController.isShown).to.be.false;
     });
--- a/packages/overlays/test/OverlayController.test.js
+++ b/packages/overlays/test/OverlayController.test.js
@@ -8,6 +8,7 @@ import { OverlayController } from '../src/OverlayController.js';
 import { overlays } from '../src/overlays.js';
 import { keyCodes } from '../src/utils/key-codes.js';
 import { simulateTab } from '../src/utils/simulate-tab.js';
+import { mimicClick } from '../test-helpers.js';
 
 /**
  * @typedef {import('../types/OverlayConfig').OverlayConfig} OverlayConfig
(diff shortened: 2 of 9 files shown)


MathieuPuech:
I thought it was already done, here is the issue related: #920

jorenbroekema:
Not sure what made you think it was fixed already @MathieuPuech ?

MathieuPuech:
I just thought I already have done a PR for it. I remember the issue but I think I forget to open a PR for it

packages/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -98,14 +98,19 @@ describe('<lion-input-datepicker>', () => {
...
 
-    it('closes the calendar via outside click', async () => {
+    it('closes the calendar via outside mousedown event', async () => {
       const el = await fixture(html`<lion-input-datepicker></lion-input-datepicker>`);
       const elObj = new DatepickerInputObject(el);
       await elObj.openCalendar();
       expect(elObj.overlayController.isShown).to.equal(true);
 
       document.body.click();
       await aTimeout(0);
```
jorenbroekema, line 108:
```suggestion
```
Nitpick:

Can these two lines be removed if we're testing with a mouse down event instead of `.click()`?

packages/overlays/test/OverlayController.test.js
```diff
@@ -452,8 +452,14 @@ describe('OverlayController', () => {
         });
         await ctrl.show();
 
+        // doesn't hide on click, ie doesn't hide on mousedown followed by mouseup event
         document.body.click();
         await aTimeout(0);
+        expect(ctrl.isShown).to.be.true;
```
jorenbroekema, line 458:
```suggestion
```
Nitpick: 

Same as my suggestion above

packages/overlays/test/OverlayController.test.js
```diff
@@ -479,6 +485,13 @@ describe('OverlayController', () => {
 
         expect(ctrl.isShown).to.be.true;
 
+        // Don't hide on inside mousedown & outside mouseup
```
jorenbroekema, line 488: Blocker:

I think we want to test here that:
- Does not hide on inside (content) mousedown ✅ 
- Does not hide on invoker mousedown <-- missing right now, this is the blocker

Nitpick:

We could remove the .click() checks because they wouldn't really have any effect now that we changed the event type.

Not sure where mouseup comes from, I don't see a reason to test that one but I also don't see the harm in it so we might as well, up to you!

packages/select-rich/test/lion-select-rich.test.js
```diff
@@ -254,10 +254,14 @@ describe('lion-select-rich', () => {
       ));
 
       expect(el.opened).to.be.true;
-      // a click on the button will trigger hide on outside click
+      // a mousedown (a full click won't work) on the button will trigger hide on outside click
       // which we then need to sync back to "opened"
       outerEl.click();
```
jorenbroekema, line 259: Nitpick:

Feel free to remove the .click, I don't think we have to verify in the test that it doesn't work :P more important to test what does work (mousedown), but it's up to you as well, I don't mind keeping it.

jorenbroekema (changes requested):
Hey, only 1 blocker and some nitpicks.

Can you also add a changeset to this commit? yarn changeset --> do the stuff (patch for `@lion/overlays`, leave the rest since we dont need a release for datepicker or rich select to update tests)

Also, can you add `fixes https://github.com/ing-bank/lion/issues/920` to the description so it auto closes the related issue? Also makes it easier to see what issue the PR addresses.

I verified the demo and the bug is fixed indeed 👍

packages/overlays/test/OverlayController.test.js
tlouisse: Added a test for this @jorenbroekema . You might want to have a look?
Then we can close this one

tlouisse:
@MathieuPuech @jorenbroekema: we aligned the behavior with iOS and others: the overlay will hide only when both mousedown and mouseup occur outside of the dialog. Seemed like the most common and sensible default and it's how a click behaves as well.

We could go so far to make this configurable, but only if there is demand for it.

jorenbroekema (approved):

tlouisse (approved):

Issue #1140, opened 2020-12-15
[types] improve docs and make _ properties private/protected


daKmoR:
as we now have types you do get autocomplete for all properties.

private/protected properties we have with `_` and `__` should add `@protected`, `@private`

also adding more explainations and docs to functions/classes will now benefit users directly 👍

tlouisse:
I think by doing this, we serve 3 goals:

1. extending parts would be notified when they use a private/protected method/prop in contexts they shouldn't
2. auto generated api tables/docs would benefit from this meta info
3. We get a nice method/prop description in our code editor

The usual way to go would be jsdoc (needed for 1, 2 and 3):

```js
constructor() {
  super();
  /**
   * The description a dev would see
   * @protected
   */  
   this._x = '';
  /**
   * The description a dev would see
   * @private
   */
  this.__y = true;
}

/**
 * The description a dev would see
 * @protected
 */
_methodX() {...}

/**
 * The description a dev would see
 * @private
 */
__methodY() {...}
```

In case of mixins (needed for 1 and 3, maybe 2 as well):

```js
export declare class MixinHost {
  /**
   * The description a dev would see
   */
  public a: string;
  /**
   * The description a dev would see
   */
  protected _x: string;
  /**
   * The description a dev would see
   */
  private __y: boolean;

  /**
   * The description a dev would see
   */
  protected _methodX(): void;
  /**
   * The description a dev would see
   */
  private __methodY(): void;
}
```

tlouisse:
A large part of point 2 is covered already via: https://github.com/ing-bank/lion/pull/1263

tlouisse:
Point 1 (`.d.ts` files for mixins of form-core) is covered here: https://github.com/ing-bank/lion/pull/1328/commits/43e4bb810c22555127ba3d110378202eb65548fa

daKmoR:
@tlouisse what is still open here? 🤔

tlouisse:
**Private/protected types**

@dakmor I think `form-core`, `overlays`  and `core` have all types solved (private and protected). These packages are the fundament for a lot of other packages (those other packages contain less code and usually no mixins).

I think all other packages are in good state as well, didn't explicitly check it though. I think `LocalizeMixin` could use an update maybe, `ListboxMixin` seems to be done, but please check. And I think
My experience is that the current state of Lion solves at least all our types issues in our extension layer (no need anymore for `skipLibCheck:true`).

**Descriptions**

[...]

PR #1263, opened 2021-03-15
Chore/add private protected types

Knorgias:
## What I did

1. Added @protected & @private to _ & __ properties and methods

### DIFF
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -142,7 +142,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
 
   constructor() {
     super();
-    /** @type {{months: Month[]}} */
+    /** @type {{months: Month[]}}
+     * @private
+     */
     this.__data = { months: [] };
     this.minDate = new Date(0);
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
@@ -156,17 +158,27 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
 
     this.firstDayOfWeek = 0;
     this.weekdayHeaderNotation = 'short';
+    /** @private */
     this.__today = normalizeDateTime(new Date());
     /** @type {Date} */
     this.centralDate = this.__today;
-    /** @type {Date | null} */
+    /**
+     * @type {Date | null}
+     * @private
+     */
     this.__focusedDate = null;
+    /** @private */
     this.__connectedCallbackDone = false;
+    /** @private */
     this.__eventsAdded = false;
     this.locale = '';
+    /** @private */
     this.__boundKeyboardNavigationEvent = this.__keyboardNavigationEvent.bind(this);
+    /** @private */
     this.__boundClickDateDelegation = this.__clickDateDelegation.bind(this);
+    /** @private */
     this.__boundFocusDateDelegation = this.__focusDateDelegation.bind(this);
+    /** @private */
     this.__boundBlurDateDelegation = this.__focusDateDelegation.bind(this);
   }
 
@@ -312,6 +324,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     }
   }
 
+  /**
+   * @private
+   */
   __calculateInitialCentralDate() {
     if (this.centralDate === this.__today && this.selectedDate) {
       // initialised with selectedDate only if user didn't provide another one
@@ -324,6 +339,7 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
   /**
    * @param {string} month
    * @param {number} year
+   * @private
    */
   __renderMonthNavigation(month, year) {
     const nextMonth =
@@ -348,6 +364,7 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
   /**
    * @param {string} month
    * @param {number} year
+   * @private
    */
   __renderYearNavigation(month, year) {
     const nextYear = year + 1;
@@ -362,6 +379,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     `;
   }
 
+  /**
+   * @private
+   */
   __renderNavigation() {
     const month = getMonthNames({ locale: this.__getLocale() })[this.centralDate.getMonth()];
     const year = this.centralDate.getFullYear();
@@ -372,6 +392,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     `;
   }
 
+  /**
+   * @private
+   */
   __renderData() {
     return dataTemplate(this.__data, {
       monthsLabels: getMonthNames({ locale: this.__getLocale() }),
@@ -393,6 +416,7 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
    * @param {string} type
    * @param {string} previousMonth
    * @param {number} previousYear
+   * @private
    */
   __getPreviousDisabled(type, previousMonth, previousYear) {
     let disabled;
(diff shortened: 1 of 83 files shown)


packages/calendar/src/LionCalendar.js
```diff
@@ -130,11 +130,13 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
 
       /**
        * The currently focused date (if any)
+       * @private
```
jorenbroekema, line 133: I don't believe TS looks at static get properties for class fields, so these @private type annotations probably need to be moved to where the class field is initialised (usually in the constructor).

packages/combobox/src/LionCombobox.js
```diff
@@ -253,22 +266,36 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
     this._ariaVersion = browserDetection.isChromium ? '1.1' : '1.0';
 
     /**
      * @configure ListboxMixin
+     * @protected
      */
     this._listboxReceivesNoFocus = true;
 
+    /**
+     * @protected
```
jorenbroekema, line 280: private right?

packages/combobox/src/LionCombobox.js
```diff
@@ -253,22 +266,36 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
+     */
     this.__prevCboxValueNonSelected = '';
+    /**
+     * @protected
```
jorenbroekema, line 284: private

packages/combobox/src/LionCombobox.js
```diff
@@ -523,6 +562,7 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    * - complete: completes the textbox value inline (the 'missing characters' will be added as
    * selected text)
    *
+   * @private
```
jorenbroekema, line 565: protected

packages/form-core/src/registration/FormRegistrarMixin.js
```diff
@@ -38,6 +38,7 @@ const FormRegistrarMixinImplementation = superclass =>
          * it should act like an array (see ChoiceGroupMixin).
          * Usually, when false, we deal with a choice-group (radio-group, checkbox-group,
          * (multi)select)
+         * @protected
```
jorenbroekema, line 41: static get prop so this @protected won't do much, annotate in initialisation instead.

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -83,6 +83,7 @@ export const ValidateMixinImplementation = superclass =>
         /**
          * Subclassers can enable this to show multiple feedback messages at the same time
          * By default, just like the platform, only one message (with highest prio) is visible.
+         * @protected
```
jorenbroekema, line 86: static get prop, see above comment

jorenbroekema (changes requested):

jorenbroekema:
Reviewed all commits since my previous review. Looks good 👍 

Now we just need to:
- Rebase on latest master --> fix merge conflict
- Add a changeset (patch) for all changed packages
- Squash all commits into one

Then we can merge :)

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -145,12 +158,15 @@ describe('<lion-checkbox-indeterminate>', () => {
...
     elIndeterminate._inputNode.click();
     await elIndeterminate.updateComplete;
+    const { subCheckboxes } = getProtectedMembers(elIndeterminate);
 
     // Assert
-    expect(elIndeterminate?.hasAttribute('indeterminate')).to.be.false;
-    expect(elIndeterminate?._subCheckboxes[0].hasAttribute('checked')).to.be.true;
-    expect(elIndeterminate?._subCheckboxes[1].hasAttribute('checked')).to.be.true;
-    expect(elIndeterminate?._subCheckboxes[2].hasAttribute('checked')).to.be.true;
+    if (elIndeterminate) {
```
tlouisse, line 164: I think the test should fail if `elIndeterminate` is null (is it possible to remove this if? (according to lin 154, TS shouldn't complain)

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -251,13 +273,21 @@ describe('<lion-checkbox-indeterminate>', () => {
...
+
     // Assert - the second sibling should not be affected
-    expect(elFirstIndeterminate?.hasAttribute('indeterminate')).to.be.false;
-    expect(elFirstIndeterminate?._subCheckboxes[0].hasAttribute('checked')).to.be.true;
-    expect(elFirstIndeterminate?._subCheckboxes[1].hasAttribute('checked')).to.be.true;
-    expect(elFirstIndeterminate?._subCheckboxes[2].hasAttribute('checked')).to.be.true;
-    expect(elSecondIndeterminate?._subCheckboxes[0].hasAttribute('checked')).to.be.false;
-    expect(elSecondIndeterminate?._subCheckboxes[1].hasAttribute('checked')).to.be.false;
+
+    if (elFirstIndeterminate) {
```
tlouisse, line 281: see comments above

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -251,13 +273,21 @@ describe('<lion-checkbox-indeterminate>', () => {
...
+      expect(elFirstIndeterminate.hasAttribute('indeterminate')).to.be.false;
+      expect(elFirstSubCheckboxes.subCheckboxes[0].hasAttribute('checked')).to.be.true;
+      expect(elFirstSubCheckboxes.subCheckboxes[1].hasAttribute('checked')).to.be.true;
+      expect(elFirstSubCheckboxes.subCheckboxes[2].hasAttribute('checked')).to.be.true;
+    }
+    if (elSecondIndeterminate) {
```
tlouisse, line 287: see comments above

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -301,18 +331,20 @@ describe('<lion-checkbox-indeterminate>', () => {
     const elParentIndeterminate = /**  @type {LionCheckboxIndeterminate} */ (el.querySelector(
       '#parent-checkbox-indeterminate',
     ));
+    const ElNestedSubCheckboxes = getProtectedMembers(elNestedIndeterminate);
```
tlouisse, line 334: Instance or 'regular variable' names should not be capitalized. So usually we have the convention:

```js
class MyClass {}

const myClass = new MyClass();

```

So in the example above `MyClass` is the constructor, `myClass` is the instance. As a rule of thumb, we can say that only constructor names should be capitalized, all other variable names should start with a lowercase char.

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -301,18 +331,20 @@ describe('<lion-checkbox-indeterminate>', () => {
...
+    const ElParentSubCheckboxes = getProtectedMembers(elParentIndeterminate);
 
     // Act - check a nested checkbox
-    elNestedIndeterminate?._subCheckboxes[0]._inputNode.click();
+    if (elNestedIndeterminate) ElNestedSubCheckboxes.subCheckboxes[0]._inputNode.click();
```
tlouisse, line 338: We have the convention to always use parentheses and newlines for an if for readability (should be captured by linting though...)

tlouisse:
Hi, great start 💪 

If we look at [this](https://github.com/ing-bank/lion/issues/1140) issue, not everything is tackled. (we focused mainly on point 2 now, I think.)
Next step would be to also add all the info to the mixin type declarations and add all the descriptions there as well (especially for public methods and protected methods)

tlouisse (approved):

Issue #1160, opened 2020-12-29
[form-core] Required validator should only apply aria-required on supported controls


tlouisse:
## Expected behavior

`aria-required=true` should not be set on non supported roles (like ‘group’). For all supported roles, see: https://www.w3.org/TR/wai-aria/#aria-required. Note that the native elements input, select and textarea are supported as well.

## Actual Behavior

It's always set, without the check described above: https://github.com/ing-bank/lion/blob/master/packages/form-core/src/validate/validators/Required.js#L24

For `role=group` for instance, axe marks this as invalid markup

PR #1272, opened 2021-03-17
Fix/update required validator

Knorgias:
## What I did

1. Updated required validator to check for compatible roles & tags when adding 'aria-required' attribute.
2. Added tests for Required.js validator

closes https://github.com/ing-bank/lion/issues/1160

### DIFF
--- a/packages/form-core/src/validate/validators/Required.js
+++ b/packages/form-core/src/validate/validators/Required.js
@@ -10,6 +10,33 @@ export class Required extends Validator {
   }
 
   /**
+   * In order to prevent accessibility violations, the aria-required attribute will
+   * be combined with compatible aria roles: https://www.w3.org/TR/wai-aria/#aria-required
+   */
+  static get _compatibleRoles() {
+    return [
+      'combobox',
+      'gridcell',
+      'input',
+      'listbox',
+      'radiogroup',
+      'select',
+      'spinbutton',
+      'textarea',
+      'textbox',
+      'tree',
+    ];
+  }
+
+  /**
+   * In order to prevent accessibility violations, the aria-required attribute will
+   * be combined with compatible platform input elements
+   */
+  static get _compatibleTags() {
+    return ['input', 'select', 'textarea'];
+  }
+
+  /**
    * We don't have an execute function, since the Required validator is 'special'.
    * The outcome depends on the modelValue of the FormControl and
    * FormControl.__isEmpty / FormControl._isEmpty.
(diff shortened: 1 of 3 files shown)


jorenbroekema:
Is there a linked issue? Do you want to add a changeset, I assume you want this released as a patch?

jorenbroekema:
Also you may wanna rebase a bit, there's an unrelated commit attached :P

packages/form-core/test/validate/Required.test.js
```diff
@@ -0,0 +1,74 @@
+import { expect, fixture, html, unsafeStatic, defineCE } from '@open-wc/testing';
+// import { LitElement } from '@lion/core';
```
tlouisse, line 2: can you remove this line?

packages/form-core/test/validate/Required.test.js
```diff
@@ -0,0 +1,74 @@
...
+import { LionField } from '@lion/form-core';
+
```
tlouisse, line 4: and this one? :)

packages/form-core/test/validate/Required.test.js
```diff
@@ -0,0 +1,74 @@
...
+import { Required } from '../../src/validate/validators/Required.js';
+
+/**
+ * @typedef {import('../../types/FormControlMixinTypes.js').FormControlHost} FormControlHost
+ */
```
tlouisse, line 9: and stitch these together?

packages/form-core/test/validate/Required.test.js
```diff
@@ -0,0 +1,74 @@
...
+    ));
+
+    Required._compatibleTags.forEach(tagName => {
+      inputNodeTag = /** @type {HTMLElementWithValue} */ (document.createElement(tagName));
+
+      validator.onFormControlConnect(el);
+      expect(el._inputNode).to.have.attribute('aria-required', 'true');
+    });
+
+    // @ts-ignore
```
tlouisse, line 48: Here it would be nice to add a comment about the fact that we don't add an attr for incompatible tags

packages/form-core/test/validate/Required.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+    Required._compatibleRoles.forEach(role => {
+      // @ts-ignore
+      inputNodeTag = /** @type {HTMLElementWithValue} */ (document.createElement('div'));
+      inputNodeTag.setAttribute('role', role);
+
+      validator.onFormControlConnect(el);
+      expect(el._inputNode).to.have.attribute('aria-required', 'true');
+    });
+    // @ts-ignore
```
tlouisse, line 67: Here it would be nice to add a comment about the fact that we don't add an attr for incompatible elems

packages/form-core/src/validate/validators/Required.js
```diff
@@ -9,6 +9,25 @@ export class Required extends Validator {
     return 'Required';
   }
 
+  static get _compatibleRoles() {
```
tlouisse, line 12: Here it would be nice to explain how these relate to: https://www.w3.org/TR/wai-aria/#aria-required

tlouisse (approved):

PR #1273, opened 2021-03-18
chore: add generic docs to core

gerjanvangeest:
## What I did

1. Added `/docs/components/index.md` to `@lion/core`
2. Added `/docs/docs/index.md` to `@lion/core`

### DIFF
--- a/.changeset/wet-rules-help.md
+++ b/.changeset/wet-rules-help.md
@@ -1,0 +1,5 @@
+---
+'@lion/core': patch
+---
+
+Added some default documentation pages that needs to be exported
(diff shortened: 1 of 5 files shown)


Knorgias (approved):
Looks good.

daKmoR (changes requested):
let's try to avoid adding docs to `core` as every release will release EVERY other package... and we don't want to have a docs change needing to force that...

most of the "generic" stuff (which was all form related so far) I currently put into `form-integrations` we could even think about adding a new package `docs` or so? in it we could add then guides and stuff

Issue #1275, opened 2021-03-18
[lion-checkbox-group]: Defining .serializer does not works


palash2601:
## Expected behavior

When serializer is defined on `checkbox-group` it should get called

## Actual Behavior

serializer defined on `checkbox-group` it NOT getting called 

## Additional context
This issue is there for all the components which extends ChoiceGroupMixin, like check-box-group and combobox.

The serlialedValue getter defined in this Mixin does not makes call to serializer
https://github.com/ing-bank/lion/blob/master/packages/form-core/src/choice-group/ChoiceGroupMixin.js#L68

Proposed solution: Add below code in serializedValue getter
```js
  if(this.serializer) {
    return this.serializer(this.modelValue);
  }
```

tlouisse:
Nice @palash2601 . You could do the proposal we discussed.
What we should also keep in mind is that we need to support a separate `deserializer` as well.

So modelValue < = > serializedValue

tlouisse:
Closing, see: https://github.com/ing-bank/lion/pull/1277

PR #1277, opened 2021-03-22
feat(form-core): serializer support choicegroupmixin

palash2601:
## What I did

1. Fix for https://github.com/ing-bank/lion/issues/1275

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -71,6 +71,10 @@ const ChoiceGroupMixinImplementation = superclass =>
       // user are relevant, choice values are always defined by the Application Developer
       // and known by the backend.
 
+      if (this.serializer) {
+        return this.serializer(this.modelValue);
+      }
+
       // Assuming values are always defined as strings, modelValues and serializedValues
       // are the same.
       const elems = this._getCheckedElements();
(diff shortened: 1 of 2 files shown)


packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -71,6 +71,10 @@ const ChoiceGroupMixinImplementation = superclass =>
       // user are relevant, choice values are always defined by the Application Developer
```
tlouisse, line 71: Type issues above can be solved in https://github.com/ing-bank/lion/blob/master/packages/form-core/types/choice-group/ChoiceGroupMixinTypes.d.ts

packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -71,6 +71,10 @@ const ChoiceGroupMixinImplementation = superclass =>
...
       // and known by the backend.
 
+      if (this.serializer) {
```
tlouisse, line 74: I think it makes sense to also add a deserializer check in the setter below

tlouisse:
Nice, thanls 👍  Could you also add these checks in FormGroupMixin? 
And add a small explanation that we allow form groups to define serialized formats as well at the bottom of this doc: https://lion-web.netlify.app/docs/systems/form/formatting-and-parsing/

tlouisse:
Will close this one, since the use case leading to this request was solved in a different way... 
Feel free to reopen at any time

PR #1280, opened 2021-03-23
Chore/overview remove lion namings

Knorgias:
## What I did

1. Removed lion-specific component namings from overview.md files


Knorgias:
Already merged

Issue #1173, opened 2021-01-07
[form-core] enhance isTriggeredByUser for arrow navigation


tlouisse:
## Expected behavior

When I press arrow up/down/left/right in listbox (descendants), `isTriggeredByUser` should be true.
Enter and space should be covered as well (I think they trigger a click and therefore should be covered already, but we should add tests)

jorenbroekema:
Not having it here can probably be considered as a bug, I think most users would expect this meta info to also work for other interactions than just 'click'.

PR #1291, opened 2021-03-25
isTriggeredByUser for listbox (descendants)

tlouisse:
closes https://github.com/ing-bank/lion/issues/1173


jorenbroekema (approved):
Needs a rebase though because it has an old trailing commit that's already merged to master

PR #1299, opened 2021-03-30
Fix super.styles in the docs

abdonrd:
## What I did

Fix `super.styles` in the docs

To avoid this kind of TS errors:

Screenshot 2021-03-30 at 16 53 56


abdonrd:
Superseded by #1300

PR #1302, opened 2021-03-31
fix(combobox): fix google combobox demo

jorenbroekema:


daKmoR:
why do you want to do a release?

seem to be only docs changes? 🤔

tlouisse (approved):

tlouisse:
> why do you want to do a release?
> 
> seem to be only docs changes? 🤔

I think it broke our extension layer

Issue #1298, opened 2021-03-30
[form-core][input][textarea] The preprocessor feature causes bugs for input composition in iOS


alangdm:
## Expected behavior

User input in languages that require composition (e.g. Japanese, Chinese, Korean, etc.) works as with native inputs

https://user-images.githubusercontent.com/6799159/112948664-4cc24a00-9173-11eb-8093-a01b2f3fca9c.mov

## Actual Behavior

https://user-images.githubusercontent.com/6799159/112948762-6a8faf00-9173-11eb-9b31-7ee37bf93f83.mov

As you can see, lion seems to be adding every keystroke to the input instead of letting composition do its job

I think the bug lies [here](https://github.com/ing-bank/lion/blob/806017a5b1f93ca528e47495b3b0776b0b21f6c3/packages/form-core/src/FormatMixin.js#L351), lion seems to be just overriding the value on every keystroke instead of waiting for the composition to be over

I couldn't find out why though, lion-field supposedly listens to the [`change` event](https://github.com/ing-bank/lion/blob/fa9cdffc8f4f2d65d395fd6c2e16e072a1eefa3b/packages/form-core/src/LionField.js#L60) not the `input` event, if it was doing it on change the bug shouldn't really occur

## Additional context

This only happens for components depending in `@lion/form-corev0.10.0` or upper
I've tested in Mac OS Chrome, Android Chrome, and iOS 14.4 Safari and the issue only occurs on iOS

## Possible Solutions

I think one way of handling this would be to add listeners for the `compositionstart` and `compositionend` events.

The `compositionstart` handler could set a flag/promise that prevents `this.value` being overwritten and said flag/promise could be settled on `compositionend`

jorenbroekema:
Yeah the preprocessor is a layer between the user filling the value and the value actually being reflected in the input, so it makes sense it would break things by doing this manually without thinking about composition. 

I agree that it sounds like a good fix to make it a little smarter:
- If compositionstart happens set `__userIsComposing` to true.
- If `__userIsComposing` is true, don't preprocess the value
- If compositionend happens set `__userIsComposing` to false
- If `__userIsComposing` is false, do preprocess the value

Although perhaps changing `formatOn` property to `input` instead of `change` is also a good fix? I'm not sure why we chose one over the other. @tlouisse  mentioning you for that :P

tlouisse:
Thanks for the bug report and fix suggestion.
[...]

PR #1304, opened 2021-03-31
fix(form-core): do not preprocess during composition

tlouisse:
closes https://github.com/ing-bank/lion/issues/1298

The test structure was a bit rearranged (to avoid merge conflicts with an upcoming merge request), the added test for this merge request: 

```js
        it('does not preprocess during composition', async () => {
          const el = /** @type {FormatClass} */ (await fixture(html`
            <${tag} .preprocessor=${(/** @type {string} */ v) => v.replace(/\$$/g, '')}>
              <input slot="input">
            </${tag}>
          `));
          const preprocessorSpy = sinon.spy(el, 'preprocessor');
          el._inputNode.dispatchEvent(new Event('compositionstart', { bubbles: true }));
          mimicUserInput(el, '`');
          expect(preprocessorSpy.callCount).to.equal(0);
          // "à" would be sent by the browser after pressing "option + `", followed by "a"
          mimicUserInput(el, 'à');
          el._inputNode.dispatchEvent(new Event('compositionend', { bubbles: true }));
          expect(preprocessorSpy.callCount).to.equal(1);
        });
```

### DIFF
--- a/packages/form-core/src/FormatMixin.js
+++ b/packages/form-core/src/FormatMixin.js
@@ -234,13 +234,6 @@ const FormatMixinImplementation = superclass =>
     }
 
     /**
-     * @param {string} value
-     */
-    __callPreprocessor(value) {
-      return this.preprocessor(value);
-    }
-
-    /**
      * @param {string|undefined} value
      * @return {?}
      */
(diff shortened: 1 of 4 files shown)


jorenbroekema (approved):

packages/form-core/src/FormatMixin.js
```diff
@@ -435,6 +451,8 @@ const FormatMixinImplementation = superclass =>
           /** @type {EventListenerOrEventListenerObject} */ (this
             ._reflectBackFormattedValueDebounced),
         );
+        this._inputNode.addEventListener('compositionstart', this.__onCompositionEvent);
```
tlouisse, line 454: whoops, need to remove here... :)

PR #1305, opened 2021-03-31
Feat/ajax concurrent fetch

LarsDenBakker:
## What I did

1. Added support for caching when multiple requests are made concurrently. When two requests have caching enabled, the second will wait for the first to finish and then serve the response from cache. This fulfills a major use case for caching where multiple web components on the page need to fetch the same data. They can be implemented independently and rendered on the page at the same time, fetch data from the same endpoint with a short cache time to live (for instance 1 minute) and only result in one actual call to the backend when they are rendered on the page.

2. While implementing I noticed a bug where only the response text was cached. I fixed this by also caching things like status and headers.

### DIFF
--- a/packages/ajax/src/interceptors-cache.js
+++ b/packages/ajax/src/interceptors-cache.js
@@ -16,11 +16,14 @@ class Cache {
     this.cacheConfig = {};
 
     /**
-     * @type {{[url: string]: {expires: number, data: object} }}
-     * @protected
+     * @type {{[url: string]: {expires: number, response: CacheResponse} }}
+     * @private
      */
     this._cacheObject = {};
-    /** @type {{ [url: string]: { promise: Promise<void>, resolve: (v?: any) => void } }} */
+    /**
+     * @type {{ [url: string]: { promise: Promise<void>, resolve: (v?: any) => void } }}
+     * @private
+     */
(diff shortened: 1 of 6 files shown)


rodrigo-garcia-leon:
I checked the MR and it looks good to me. Good addition @LarsDenBakker 💯

packages/ajax/src/interceptors-cache.js
```diff
@@ -267,24 +302,29 @@ export const cacheRequestInterceptorFactory = (getCacheIdentifier, globalCacheOp
...
       }
 
-      const init = /** @type {LionRequestInit} */ ({
-        status,
-        statusText,
-        headers,
-      });
-
-      const response = /** @type {CacheResponse} */ (new Response(cacheResponse, init));
+      const response = /** @type {CacheResponse} */ cacheResponse.clone();
```
jorenbroekema, line 318: Ah neat I didn't know you could clone those 👍

jorenbroekema (approved):
Cool, so the main use case I assume is when you load the page, and a bunch of components connect at once and all start doing calls to endpoints at the same time, but those concurrent calls to the same endpoint are not reused even with caching turned on? And now they are :)?

Looks good to me!

PR #1306, opened 2021-03-31
chore: update search highlight color - for a11y

Knorgias:
## What I did

1. Updated search highlight color to lion's primary-text-color - for accessibility purposes
2. P: Purple on white gives a contrast ratio of 4.3:1. On grey it's 3.7:1. That's less than the required 4.5:1 for small text P: Found text is using color only to be recognized. S: Make it bold
3. P: Image in search has no alt (invalid HTML) S: Set alt to function of image (search)
4. P: Focus on logo-link practically invisible S: Add padding 2px to sidebar-nav.
5. P: The follow guides-link (CTA) has small text and a contrast ratio of less than 4.5:1 with the background. S: Pick a different color combination.
6. P: "see components" does not have a visual focus state S: Change styling to have a visual focus state.


docs/_assets/style.css
```diff
@@ -12,12 +12,37 @@ mdjs-preview {
...
 
-#main-header a[href='/blog/'] {
-  display: none;
+#main-header a {
+  font-weight: normal;
 }
 
-@media screen and (min-width: 1024px) {
+#main-header a.current.active {
+  font-weight: bold;
```
gerjanvangeest, line 20: You should keep them bold even when it is not the `current` page, as in when you are on a sub-page.

docs/_assets/style.css
```diff
@@ -12,12 +12,37 @@ mdjs-preview {
...
+}
+
+#main-header .current.active {
+  display: block;
+}
+
+@media only screen and (max-width: 600px) {
+  #main-header a {
+    font-size: 12px;
```
gerjanvangeest, line 29: By making the anchors this small the click-area will not be big enough.

I guess you tried to keep all elements in the header to be sure every page is still reachable, but the blog and our github are also reachable via the footer. Preferably they are also available under the hamburger menu.

docs/_assets/style.css
```diff
@@ -12,12 +12,37 @@ mdjs-preview {
...
+  }
+  #main-header a.social-link {
+    display: block;
+    width: 20px;
+  }
+}
+
+@media only screen and (min-width: 1024px) {
```
gerjanvangeest, line 37: If you don't set the blog on `display: none` in the first place, then this is also not needed.

docs/_assets/style.css
Knorgias: Yes, I tried to keep all anchors on the header for consistency across devices/viewports.
"but the blog and our github are also reachable via the footer"
In that case, is it better to remove the blog and github anchors from the header altogether?

docs/_assets/variables.css
```diff
@@ -15,6 +15,9 @@ html {
   --footer-background: rgba(0, 0, 0, 0.1);
 
   --text-color: black;
+
+  /* search colors*/
+  --rocket-search-highlight-color: #2c3e50;
 }
 
 html.dark {
```
gerjanvangeest, line 23: To my knowledge we don't support dark mode for lion atm. All these values look like original rocket values, please remove them all, instead of adding one for the search.

docs/_assets/style.css
```diff
@@ -1,23 +1,62 @@
 body[layout='layout-home'] .markdown-body .call-to-action:nth-of-type(2) {
-  --primary-color: #222;
-  --primary-color-lighter: #333;
+  --primary-color: #333;
```
gerjanvangeest, line 2: Overriding variables here will break the usage of theming (dark mode).

docs/_assets/style.css
```diff
@@ -1,23 +1,62 @@
...
+  --primary-color-lighter: #444;
   --primary-color-darker: #000;
 }
 
+body[layout='layout-home'] .markdown-body .call-to-action:nth-of-type(1) {
+  --primary-color: #8f7325;
+  --primary-color-lighter: #8f7325;
```
gerjanvangeest, line 9: As far as i can find the `--primary-color-lighter` is only used in call-to-actions. Don't override the value here, but in variables.css.

docs/_assets/style.css
gerjanvangeest: If i remove the github anchor and change the padding left and right of the header, i was able to increase the font-size to 14px. Which looks acceptable:

image

If you play around with the padding you can get click areas of 44px or bigger.

Knorgias: Then it's still not consistent among all viewports (4 links VS 5), which defeats the purpose. so I'd rather leave it as is fot the smaller viewports, since the blog & github are reachable via hamburger menu and footer.
Alternative: remove blog & github links from header altogether - in order to achieve consistency.

PR #1309, opened 2021-04-01
@lit/localize POC

jorenbroekema:
I included this discussion https://github.com/ing-bank/lion/discussions/664 as well to establish requirements, as well as our current implementation of localize and its most important features.


jorenbroekema:
https://github.com/jorenbroekema/lit-localize-test

PR #1311, opened 2021-04-02
chore(docs): remove lion references for easier extending

gerjanvangeest:
## What I did

1.

### DIFF
--- a/.changeset/dry-fishes-shake.md
+++ b/.changeset/dry-fishes-shake.md
@@ -1,0 +1,16 @@
+---
+'@lion/checkbox-group': patch
+'@lion/collapsible': patch
+'@lion/fieldset': patch
+'@lion/form': patch
+'@lion/input-amount': patch
+'@lion/input-date': patch
+'@lion/input-datepicker': patch
+'@lion/input-email': patch
+'@lion/input-iban': patch
+'@lion/input-range': patch
+'@lion/input-stepper': patch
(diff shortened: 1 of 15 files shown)


tlouisse:
@gerjanvangeest @daKmoR I think using `our` everywhere changes the tone of voice and in some contexts, it doesn't really apply (for instance in telegraphed contexts, like listing a set of features). Wouldn't `the` be more neutral and in line with the original tone of voice?

Also, we now solve a technical problem (something we tackled before) via language constructs. I can imagine that in extending contexts, it has value to know the exact name of a certain webcomponent.
Is this really the only way to do this?

daKmoR:
our or the is something designers can decide apon once they join the content party - imho our is good enough for now

You do get the exact web component in the actual code and if you want to add it in extending layer you can do so. But it's a cost you opt into. While right now we force it on everyone and it will lead to lots of not-needed overrides and complexity.

tlouisse:
Alright, I don't want to be in the way of your progress, so as a short term solution I would agree. 

But it would be a pity if we loose this opportunity. If we ditch the prefix, it won't be possible anymore via a find and replace to detect when a web component is meant and when it wasn't.
The sentence 'our `button` is a wrapper for the native `button`' cannot be converted anymore to '`my-button` is a wrapper for the native `button`'

daKmoR (approved):

Issue #1241, opened 2021-02-24
ing-form - prefilled radiogroup is set to dirty


tush:
## Expected behavior

dirty flag in ing-form should be 'false' even when initialised with pre-filled value in lion-checkbox-group or  lion-radio-group via serializedValue. 

## Actual Behavior

dirty flag in ing-form is set to 'true' even when initialised with pre-filled value in lion-checkbox-group or  lion-radio-group via serializedValue. 

https://webcomponents.dev/edit/kpZmz1CJN580OaXsk56f?pm=1

* In example above, notice that the form has **dirty** flag set to 'true'. This happens if we pass pre-filled data (serializedValue) to form especially when using **lion-checkbox-group** or **lion-radio-group**.

* In example above, if we remove 'serializedValue', notice that the form has dirty flag set to 'false'

```javascript
<lion-form .serializedValue=${this.initialData}>
```

## Additional context

Link above can be used to view the behavior.

tlouisse:
Hi @tush is this still an issue?

In the link, `.serializedValue` was already removed. When I add: 
```html
<lion-form .serializedValue="${{"full_name":{"first_name":"","last_name":""},"datepicker":"2020-12-12","bio":"","money":"","iban":"","email":"","checkers":[],"dinosaurs":"","lyrics":"1","range":2.3}}"
```
I still see that the form does not initialize as `dirty`...

I also added this test: https://github.com/ing-bank/lion/pull/1313/files#diff-e3ef11b14acc5fb8afe1ae5515dfc5d2c409766985fa544cc4060a3fa6be7978R674

tlouisse:
After speaking to @tush and further investigation: it fails for radio (groups): https://github.com/ing-bank/lion/pull/1313/files#diff-ed5b84d0b04e88508b30acb7e205f20c2609f7e36896c8e178a386f7316aca90R72

tlouisse:
Fixed by https://github.com/ing-bank/lion/pull/1313

PR #1313, opened 2021-04-06
chore(form-core): fix prefilled FormGroups

tlouisse:
see https://github.com/ing-bank/lion/issues/1241

## What happened?

Assume this code:
`<lion-form .serializedValue="${{ x: 'y' }}">
      <lion-radio-group name="x">
        <lion-radio>`
        
`lionForm.dirty` became `true` on init, whereas it should be `false`...

This was because the initial `.serializedValue` was scheduled until after `registrationComplete`, just like `__initInteractonStates()`.
Since the latter was scheduled and executed first, the inital radio value caused the `dirty` state of the form to change.

## How to solve this?
- first, wait for registrationComplete to resolve.
- then resolve `initComplete` that guarantees all initial logic affecting children has run. The `__initInteractonStates` can wait for this Promise and everything works as expected.

## More changes in this mr:
- all registration logic is put in one mixin: FormRegistrarMixin
- ChoiceGroupMixin should not have InteractionStateMixin; it was designed to be compatible with Fields, and ChoiceGroupMixin can be applied on top of both Fields (listbox) and FormGroups (radio/cb-group)

### DIFF
--- a/packages/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -11,6 +11,12 @@ import { InteractionStateMixin } from '../InteractionStateMixin.js';
  */
 
 /**
+ * ChoiceGroupMixin applies on both Fields (listbox/select-rich/combobox)  and FormGroups
+ * (radio-group, checkbox-group)
+ * TODO: Ideally, the ChoiceGroupMixin should not depend on InteractionStateMixin, which is only
+ * designed for usage with Fields, in other words: their interaction states are not derived from
+ * children events, like in FormGroups
+ *
  * @type {ChoiceGroupMixin}
  * @param {import('@open-wc/dedupe-mixin').Constructor<import('@lion/core').LitElement>} superclass
  */
@@ -53,8 +59,8 @@ const ChoiceGroupMixinImplementation = superclass =>
       };
 
       if (this.__isInitialModelValue) {
-        this.__isInitialModelValue = false;
         this.registrationComplete.then(() => {
+          this.__isInitialModelValue = false;
           this._setCheckedElements(value, checkCondition);
           this.requestUpdate('modelValue', this.__oldModelValue);
         });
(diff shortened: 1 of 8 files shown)


packages/form-integrations/test/form-integrations.test.js
```diff
@@ -37,23 +37,51 @@ describe('Form Integrations', () => {
...
+          .serializedValue="${{
+            full_name: { first_name: '', last_name: '' },
+            date: '2000-12-12',
+            datepicker: '2020-12-12',
+            bio: '',
+            money: '',
+            iban: '',
+            email: '',
+            checkers: ['foo', 'bar'],
+            dinosaurs: 'brontosaurus',
```
tlouisse, line 72: The prefilled value for 'brontosaurus' currently breaks the test. ('' results in a value of `false` for dirty).
Obviously, solution should be looked for in radioGroup

packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -53,8 +52,8 @@ const ChoiceGroupMixinImplementation = superclass =>
       };
 
       if (this.__isInitialModelValue) {
-        this.__isInitialModelValue = false;
         this.registrationComplete.then(() => {
+          this.__isInitialModelValue = false;
```
tlouisse, line 56: These changes will prevent merge conflicts with https://github.com/ing-bank/lion/pull/1312

packages/form-core/src/form-group/FormGroupMixin.js
```diff
@@ -146,32 +146,33 @@ const FormGroupMixinImplementation = superclass =>
...
-        },
-        () => {
-          this.registrationComplete.done = true;
-          throw new Error(
-            'Registration could not finish. Please use await el.registrationComplete;',
-          );
-        },
-      );
+
+      // TODO: shouldn't we put all this registration logic in FormRegistrarMixin?
```
tlouisse, line 150: comments will be deleted...

packages/form-integrations/test/helpers/umbrella-form.js
```diff
@@ -23,9 +25,16 @@ export class UmbrellaForm extends LitElement {
...
+  set serializedValue(v) {
+    this.__serializedValue = v;
+  }
+
   render() {
     return html`
-      <lion-form>
+      <lion-form .serializedValue="${this.__serializedValue}">
         <form>
           <lion-fieldset name="full_name">
```
tlouisse, line 39: I took the same umbrell-form as in the docs, so that it contains all our FormControls

jorenbroekema (approved):
Looks cleaner having all this logic in the FormRegistrarMixin

PR #1314, opened 2021-04-07
fix(ajax): bug fixes and readability improvements

narzac:
## What I did

1. Fix options expansion
2. Fix removing request interceptors
3. Use 1 hour as the default time to live
4. Check for null when serializing the search params

### DIFF
--- a/packages/ajax/src/AjaxClient.js
+++ b/packages/ajax/src/AjaxClient.js
@@ -27,11 +27,11 @@ export class AjaxClient {
       xsrfCookieName: 'XSRF-TOKEN',
       xsrfHeaderName: 'X-XSRF-TOKEN',
       jsonPrefix: '',
+      ...config,
       cacheOptions: {
         getCacheIdentifier: () => '_default',
         ...config.cacheOptions,
       },
-      ...config,
     };
 
     /** @type {Array.<RequestInterceptor|CachedRequestInterceptor>} */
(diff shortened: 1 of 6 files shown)


packages/ajax/src/AjaxClient.js
```diff
@@ -27,11 +27,11 @@ export class AjaxClient {
       xsrfCookieName: 'XSRF-TOKEN',
       xsrfHeaderName: 'X-XSRF-TOKEN',
       jsonPrefix: '',
+      ...config,
```
jorenbroekema, line 30: Ah good catch

packages/ajax/src/AjaxClient.js
```diff
@@ -123,11 +115,11 @@ export class AjaxClient {
       // In this instance we actually do want to await for each sequence
       // eslint-disable-next-line no-await-in-loop
       const interceptedRequestOrResponse = await intercept(interceptedRequest);
-      if (interceptedRequestOrResponse instanceof Request) {
-        interceptedRequest = interceptedRequestOrResponse;
-      } else {
+      if (interceptedRequestOrResponse instanceof Response) {
```
jorenbroekema, line 118: Lars refactored this code a bit in his PR that I just merged, so please rebase and check whether your change is still required.

jorenbroekema (approved):
Looks good to me! There will be some merge conflicts for sure with the refactors from https://github.com/ing-bank/lion/pull/1305 but hopefully nothing too annoying.

packages/ajax/src/AjaxClient.js
jorenbroekema: I would see this as a bugfix, if you have time could you add a regression test for this fix? And then we should release it as a patch as well and make clear what was fixed in the changelog, by running `npx run changeset` :)

narzac: Sure, planning to :)

packages/ajax/src/AjaxClient.js
narzac: Cool, I will rebase

narzac:
I think it is ready to merge

jorenbroekema (approved):

PR #1321, opened 2021-04-08
test(form-registration): failing test for form registration in dialog

AymenBenAmor:
## What I did

1. Created a failing test for registration of fields when inside a dialog with the form being outside the dialog.


tlouisse:
As discussed: for this to work the `FormRegistrarPortalMixin` needs to be a applied (see `LionOptions` in `listbox` package)

Issue #1322, opened 2021-04-08
[form-core][validate] allow override feedback visibility condition for Application Developers


tlouisse:
see https://github.com/ing-bank/lion/discussions/1159

Issue #1204, opened 2021-02-10
[switch] submitted interaction state


tlouisse:
## Expected behavior

When no form interaction has taken place, `submitted` is not set. (it should not be set by anyone except generic Form mixins, actually)

## Actual Behavior

submitted is set to true in connectedCallback. Probably to enable error/info messages outside forms. We should look for a cleaner way of doing this (?)

tlouisse:
Remove the submitted states, then implement this:
```js
    _showFeedbackConditionFor(type) {
      const originalCondition = (type) => (this.touched && this.dirty) || this.prefilled || this.submitted;
      if (typeof this.showFeedbackConditionFor === 'function') {
        return this.showFeedbackConditionFor(type, originalCondition)
      }
      return originalCondition(type);
    }    
```

```js
<first-input .showFeedbackConditionFor=$"(type, originalCondition) => originalCondition(type) && second.showsFeedbackFor.includes(type)">
```

Then, create a demo implementing that info messages should always be shown.

See https://github.com/ing-bank/lion/discussions/1159 as well

Issue #1203, opened 2021-02-10
[switch] focus input on label click


tlouisse:
## Expected behavior

When I click the label of a switch, I want to see the _inputNode (the switchbutton) focused

## Actual Behavior

checked state is toggled but focus is not set.
Maybe it can trigger a click on the input? (feels hacky, but this is exactly what the platform does for native inputs)

PR #1323, opened 2021-04-08
fix(switch|form-core): enhanced feedback visibility configuration / applied on switch / a11y fix

tlouisse:
closes https://github.com/ing-bank/lion/issues/1322
closes https://github.com/ing-bank/lion/issues/1203
closes https://github.com/ing-bank/lion/issues/1204

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -652,11 +652,13 @@ export const ValidateMixinImplementation = superclass =>
     /**
      * Allows the end user to specify when a feedback message should be shown
      * @example
-     * showFeedbackConditionFor(type, defaultCondition) {
+     * showFeedbackConditionFor(type, meta, defaultCondition) {
      *   if (type === 'info') {
-     *     return true;
+     *     return return;
+     *   } else if (type === 'prefilledOnly') {
+     *     return meta.prefilled;
      *   }
-     *   return defaultCondition(type);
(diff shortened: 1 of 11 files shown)


packages/switch/test/lion-switch.test.js
```diff
@@ -39,6 +50,12 @@ describe('lion-switch', () => {
     expect(el.checked).to.be.false;
   });
 
+  it('clicking the label should focus the toggle button', async () => {
+    const el = await fixture(html`<lion-switch disabled label="Enable Setting"></lion-switch>`);
```
gerjanvangeest, line 54: Not when its disabled

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -628,15 +628,53 @@ export const ValidateMixinImplementation = superclass =>
...
      * @protected
      */
     // eslint-disable-next-line no-unused-vars
-    _showFeedbackConditionFor(type) {
+    _showFeedbackConditionFor(type, meta) {
       return true;
     }
 
+    /**
+     * Allows super classes tp add meta info for showFeedbackConditionFor
```
gerjanvangeest, line 645: tp?

packages/switch/test/lion-switch.test.js
tlouisse: Good catch 💪

gerjanvangeest (approved):

Issue #1329, opened 2021-04-12
[form-core] revalidate when child modelValue changes


tlouisse:
When a child of a fieldset changes and validity state of fieldset is affected, this is not reflected in `fieldset.hasFeedbackFor('error')`.

Fix by listening for @model-value-changed of child and call `.validate()` in ValidateMixin

Issue #1324, opened 2021-04-08
[form-core][validate] docs for custom and complex validation and interaction states


tlouisse:
Add an example that shows:
- how to use `.showFeedbackFor`, allowing every user to customize feedback per FormControl, based on types and interaction states
- how to write Validators where multiple inputs depend on each other

Basically what is explained here: https://github.com/ing-bank/lion/discussions/1159
closes https://github.com/ing-bank/lion/discussions/1159

PR #1330, opened 2021-04-12
fix(form-core): validate on child change + {feedbackType}StateChanged

tlouisse:
closes https://github.com/ing-bank/lion/discussions/1222
closes https://github.com/ing-bank/lion/issues/1329
closes https://github.com/ing-bank/lion/discussions/1159
closes https://github.com/ing-bank/lion/issues/1324

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -85,6 +85,8 @@ export const ValidateMixinImplementation = superclass =>
          * By default, just like the platform, only one message (with highest prio) is visible.
          */
         _visibleMessagesAmount: { attribute: false },
+
+        __childModelValueChanged: { attribute: false },
       };
     }
 
@@ -181,6 +183,12 @@ export const ValidateMixinImplementation = superclass =>
       this.__onValidatorUpdated = this.__onValidatorUpdated.bind(this);
       /** @protected */
       this._updateFeedbackComponent = this._updateFeedbackComponent.bind(this);
+
+      /**
+       * This will be used for FormGroups that listen for `model-value-changed` of children
+       * @private
+       */
+      this.__childModelValueChanged = false;
     }
 
     connectedCallback() {
@@ -200,6 +208,11 @@ export const ValidateMixinImplementation = superclass =>
       super.firstUpdated(changedProperties);
       this.__validateInitialized = true;
       this.validate();
+      if (this._repropagationRole !== 'child') {
+        this.addEventListener('model-value-changed', () => {
+          this.__childModelValueChanged = true;
+        });
+      }
     }
 
     /**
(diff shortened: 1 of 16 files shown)


Knorgias (approved):

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -181,6 +183,12 @@ export const ValidateMixinImplementation = superclass =>
       this.__onValidatorUpdated = this.__onValidatorUpdated.bind(this);
       /** @protected */
       this._updateFeedbackComponent = this._updateFeedbackComponent.bind(this);
+
+      /**
+       * This will be used for FormGroups that listen for `model-valuc-changed` of children
```
jorenbroekema, line 188:
```suggestion
       * This will be used for FormGroups that listen for `model-value-changed` of children
```

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -200,6 +208,11 @@ export const ValidateMixinImplementation = superclass =>
       super.firstUpdated(changedProperties);
       this.__validateInitialized = true;
       this.validate();
+      if (this._repropagationRole !== 'child') {
+        this.addEventListener('model-value-changed', () => {
+          this.__childModelValueChanged = true;
```
jorenbroekema, line 213:
```suggestion
        this.validate({ clearCurrentResult: true });
```

Then we don't need the __childModelValueChanged lit property right?

jorenbroekema (approved):
Small comment where I think the code could be simpler, but not a blocker

packages/form-core/src/validate/ValidateMixin.js
tlouisse: Yeah, that's what I suggested initially to Ben Amor as well :) But I thought it would be better to reduce the amount of sync handlers in the form and let it 'breath' 1 microtask

tlouisse: For the reusability of mixins it might even be better to listen for this in FormControlMixin and do a requetsUpdate for modelValue (for groups that only have setters/getters for children)...
I think we can optimize this later (for now it's  a private value that can easiliy be refactored/improved upon)

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -85,6 +85,8 @@ export const ValidateMixinImplementation = superclass =>
          * By default, just like the platform, only one message (with highest prio) is visible.
          */
         _visibleMessagesAmount: { attribute: false },
+
+        __childModelValueChanged: { attribute: false, type: Boolean },
```
jorenbroekema, line 89: Ah small detail but `type: Boolean` is essentially redundant as all it does it tell lit-html how to reflect to an attribute, but we don't reflect to attribute.

packages/form-core/src/validate/ValidateMixin.js
jorenbroekema: True! I also felt like maybe ValidateMixin wouldn't be the most logical place for the prop indeed

packages/form-core/src/validate/ValidateMixin.js
tlouisse: Ah, true indeed 👍

Issue #759, opened 2020-06-07
[select-rich] focus after collapsing


MathieuPuech:
## Describe the bug

When a select-rich is blurred, it closes and after closing, it focus it. It's an issue when you open another select-rich.

## To Reproduce

1. Go to [Select-rich storybook demo](https://lion-web-components.netlify.app/?path=/docs/forms-select-rich--main)
2. Click on first select-rich to opened it
3. Scroll down to next select-rich
4. click on it to open

## Expected behavior

Should not scroll to the first select-rich opened and should focus the new select-rich.

## Screenshots

![2020-06-07_14-14-51](https://user-images.githubusercontent.com/9367152/83976758-04772a00-a8ca-11ea-9b25-75398fca0ddf.gif)

## Additional context

The issue is the same if you scroll and click on anything else on screen.

tlouisse:
It would probably be fixed if line 492 would be enabled again:
https://github.com/ing-bank/lion/blob/master/packages/overlays/src/OverlayController.js#L493
```
      // We only are allowed to move focus if we (still) 'own' it.
      // Otherwise we assume the 'outside world' has, purposefully, taken over
=>    // if (this._contentWrapperNode.activeElement) {
```
@daKmoR or @jorenbroekema: do you know if this was purposefully commented out and if it's safe to enable again?

jorenbroekema:
I'm not aware of why it's commented out, maybe git blame it :P? If it was me, then I'd say uncomment it to fix this

daKmoR:
that is really an interesting one... and possible 2 fold...

the native input
1) prevents scrolling when a select is open 😱  (never thought about that 🙈)
2) it closes before you can open another one

Note: once I open the select I try to scroll with my mouse wheel but nothing happens
![NativeSelectBlocksScrolling](https://user-images.githubusercontent.com/24378/84036075-052cc080-a99d-11ea-9cb2-1917a74b033f.gif)

probably better to align with the native browser behaviour in that case right? 🤔
opinions?

MathieuPuech:
It's not the same behavior on Mac or on Windows. On windows, scrolling is closing the select

MathieuPuech:
![2020-06-08_10-02-04](https://user-images.githubusercontent.com/9367152/84039361-237bc780-a96f-11ea-935d-7bbf1aa2c25a.gif)

daKmoR:
oh boy 😅

[...]

PR #1331, opened 2021-04-12
fix(overlays): only focus controlled elements after hide

tlouisse:
closes https://github.com/ing-bank/lion/issues/759

### DIFF
--- a/.changeset/lemon-shoes-doubt.md
+++ b/.changeset/lemon-shoes-doubt.md
@@ -1,0 +1,5 @@
+---
+'@lion/checkbox-group': patch
+---
+
+recover possibility to registrer without []
--- a/.changeset/loud-rocks-teach.md
+++ b/.changeset/loud-rocks-teach.md
@@ -1,0 +1,5 @@
+---
+'@lion/overlays': patch
+---
+
+only focus elements after hide when OverlayController has ownership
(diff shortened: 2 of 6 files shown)


.changeset/lemon-shoes-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/checkbox-group': patch
+---
+
+recover possibility to registrer without []
```
daKmoR, line 5: huh why? 
this enforces a best practice... we no longer want to enforce it? 🤔

tlouisse: @daKmoR this apparently already didn't work. When doing the types I reenabled it, but it breaks our extension layer (so a breaking change and reverting again).

tlouisse: Somehow it never hit this line before: https://github.com/ing-bank/lion/blob/master/packages/checkbox-group/src/LionCheckboxGroup.js#L16
I actually think it also conflicts with https://github.com/ing-bank/lion/blob/master/packages/form-core/src/registration/FormRegistrarMixin.js#L155

So we probably should rediscuss before Lion v1

daKmoR: ok, makes sense

daKmoR (approved):

PR #1333, opened 2021-04-13
fix: expose members as protected for extension compat. till v1

tlouisse:
-

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -240,8 +240,10 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    * of the form, and firing click on this button. This will fire the form submit
    * without side effects caused by the click bubbling back up to lion-button.
    * @param {Event} ev
-   * @private
+   * @protected
+   * @returns {Promise<void>}
    */
+  // TODO: rename to _clickDelegationHandler in v1
   async __clickDelegationHandler(ev) {
     // Wait for updateComplete if form is not yet available
     if (!this._form) {
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -466,8 +466,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
    * @param {string} type
    * @param {string} previousMonth
    * @param {number} previousYear
-   * @private
+   * @protected
    */
+  // TODO: rename to _previousButtonTemplate in v1. Also see: https://github.com/ing-bank/lion/discussions/591
   __renderPreviousButton(type, previousMonth, previousYear) {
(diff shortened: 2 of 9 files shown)


packages/button/src/LionButton.js
```diff
@@ -240,8 +240,10 @@ export class LionButton extends DisabledWithTabIndexMixin(SlotMixin(LitElement))
    * of the form, and firing click on this button. This will fire the form submit
    * without side effects caused by the click bubbling back up to lion-button.
    * @param {Event} ev
-   * @private
+   * @protected
```
Knorgias, line 243: __ so it's still private instead, right?

packages/calendar/src/LionCalendar.js
```diff
@@ -466,8 +466,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
    * @param {string} type
    * @param {string} previousMonth
    * @param {number} previousYear
-   * @private
+   * @protected
```
Knorgias, line 469: same a previous comment

packages/calendar/src/LionCalendar.js
```diff
@@ -496,8 +497,9 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
    * @param {string} type
    * @param {string} nextMonth
    * @param {number} nextYear
-   * @private
+   * @protected
```
Knorgias, line 500: same comment

packages/select-rich/src/LionSelectRich.js
```diff
@@ -462,8 +462,9 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
 
   /**
    * @param {KeyboardEvent} ev
-   * @private
+   * @protected
```
Knorgias, line 465: same comment

packages/button/src/LionButton.js
tlouisse: We override this in our extension layer, and overriding private methods is not possible. So this method should actually be called '_clickDelegationHandler'

Since it's likely those methods are used by others atm, we delay the (breaking) name change until v1. Hence the comment `// TODO: rename to _clickDelegationHandler in v1`

packages/calendar/src/LionCalendar.js
tlouisse: see prev reply :)

packages/calendar/src/LionCalendar.js
tlouisse: see prev reply

packages/select-rich/src/LionSelectRich.js
tlouisse: see prev reply

Knorgias (approved):

PR #1337, opened 2021-04-14
Apply all a11y improvements on new branch

Knorgias:
## What I did

1. [x] P: Purple on white gives a contrast ratio of 4.3:1. On grey it's 3.7:1. That's less than the required 4.5:1 for small text P: Found text is using color only to be recognized. S: Make it bold
2. [x] P: Image in search has no alt (invalid HTML) S: Set alt to function of image (search)
3. [x] P: Accessible name: "rocket search". S: Change name to search
4. [x] P: There's no indication of the current page. S: Use aria-current and different styling on the current page
5. [x] P: Focus on logo-link practically invisible S: Add padding 2px to sidebar-nav.
6. [x] P: The follow guides-link (CTA) has small text and a contrast ratio of less than 4.5:1 with the background. S: Pick a different color combination.
7. [x] P: "see components" does not have a visual focus state S: Change styling to have a visual focus state.
8. [x] P: Link in the page are only recognizable by color. S: Underline links - links are actually underlined on hover

### DIFF
--- a/docs/_assets/style.css
+++ b/docs/_assets/style.css
@@ -1,7 +1,9 @@
 body[layout='layout-home'] .markdown-body .call-to-action:nth-of-type(2) {
-  --primary-color: #222;
-  --primary-color-lighter: #555;
-  --primary-color-darker: #000;
+  background: linear-gradient(to right, #555, #222);
+}
+
+body[layout='layout-home'] .markdown-body .call-to-action:nth-of-type(2):hover {
+  background: #000;
 }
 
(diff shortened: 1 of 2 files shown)


docs/_assets/style.css
```diff
@@ -1,23 +1,53 @@
 body[layout='layout-home'] .markdown-body .call-to-action:nth-of-type(2) {
   --primary-color: #222;
```
gerjanvangeest, line 2: As i said in your previous MR: "Overriding variables here will break the usage of theming (dark mode)." 
Don't use variables, but set the values directly.

gerjanvangeest (approved):

Issue #1340, opened 2021-04-16
[listbox|select-rich|combobox] not removing its options properly when using the 'nice' api


remcovaes:
## repro:
1. Add options to a listbox using the lit directive repeat
1. clear all the options

### Expected behavior
Listbox to contain no options

### Actual Behavior
Listbox containing the initially added options

## Additional context

I created an isolated example on Github: https://github.com/remcovaes/listbox-bug

## Non working code:
```
<lion-listbox>
    ${repeat(this.options, option => option, option => html`
        <lion-option>${option}</lion-option>
    `)}
</lion-listbox>
```
## Workaround
Use the syntax that is but more verbose:
```
<lion-listbox>
    <lion-options slot="input">
        ${repeat(this.options, option => option, option => html`
            <lion-option>${option}</lion-option>
        `)}
    </lion-options>
</lion-listbox>
```

PR #1341, opened 2021-04-16
fix(listbox): teleported options compatible with map/repeat

tlouisse:
closes https://github.com/ing-bank/lion/issues/1340


packages/combobox/src/LionCombobox.js
```diff
@@ -413,54 +428,62 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
    * @param {KeyboardEvent} ev
    * @protected
    */
-  _textboxOnKeydown(ev) {
-    if (ev.key === 'Tab') {
-      this.opened = false;
-    }
-  }
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
+  _textboxOnKeydown(ev) {}
```
gerjanvangeest, line 473: Before the overlay closed when tabbing out of it, why did you remove this feature?

tlouisse: It was already checked here: https://github.com/ing-bank/lion/pull/1341/files/22678f13ce450e43fcdc94c49456ad37073abd43#diff-03152156fa1a67197eb600827fa4d55e5cf2571242b96a7333935109c12cbe33R438

tlouisse: Correction: you were right indeed 🙂 Great catch. (apparently we need to check this on keydown and not on keyup and the test wrongly sent a keyup event, which in non apparent in the browser)

tlouisse: See: https://github.com/ing-bank/lion/pull/1341/commits/788f500a2ca38855ebcbf70db30b015fb667fa69

gerjanvangeest (approved):

PR #1347, opened 2021-04-20
Feat/rocket simulator

Knorgias:
## What I did

1. Set up rocket simulator
2. Use simulator on all demos

### DIFF
--- a/.changeset/slow-spoons-carry.md
+++ b/.changeset/slow-spoons-carry.md
@@ -1,0 +1,46 @@
+---
+'@lion/accordion': minor
+'@lion/ajax': minor
+'@lion/button': minor
+'@lion/calendar': minor
+'@lion/checkbox-group': minor
+'@lion/collapsible': minor
+'@lion/combobox': minor
+'@lion/core': minor
+'@lion/dialog': minor
+'@lion/fieldset': minor
+'@lion/form': minor
+'@lion/form-core': minor
+'@lion/form-integrations': minor
+'@lion/helpers': minor
+'@lion/icon': minor
+'@lion/input': minor
+'@lion/input-amount': minor
+'@lion/input-date': minor
+'@lion/input-datepicker': minor
+'@lion/input-email': minor
+'@lion/input-iban': minor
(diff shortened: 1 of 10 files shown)


tlouisse (dismissed):
Wrt the commit messages: we usually do `feat/fix/chore(package-name)` in commit messages. When there is no code changes that affect the end consumer, then it should be `chore` (because the simulator feat lives in the rocket repo). 
In this case, all your changes could be squashed into one commit and have the `chore` prefix (that's the default for docs)

Update: with this particular docs change (that heavily changes the appearance of all demo viewers) a `feat` is justified :)

tlouisse (changes requested):
Looks good. Few things @dakmor:

- can we have the simulator unstyled by default (lion approach)?
- can we opt out of themes and platform?
- can we get rid of the green border around the demo?

daKmoR:
@tlouisse 

> can we have the simulator unstyled by default (lion approach)?

feel free add css variables via a PR (please don't unstyle it but make it possible to style it how we need it)

> can we opt out of themes and platform?

currently, is has

```js
    this.themes = [
      { key: 'light', name: 'Light' },
      { key: 'dark', name: 'Dark' },
    ];
    
    this.platforms = [
      { key: 'web', name: 'Web' },
      // { key: 'web-windows', name: 'Windows' },
      // { key: 'web-mac', name: 'Mac' },
      { key: 'android', name: 'Android' },
      { key: 'ios', name: 'iOS' },
    ];
```

we could default this to an empty array `[]` and then hide empty arrays? 🤔

> can we get rid of the green border around the demo?

let's add a css variable

tlouisse:
> feel free add css variables via a PR (please don't unstyle it but make it possible to style it how we need it)

Sounds good. Shadow part could work as well. 

> we could default this to an empty array [] and then hide empty arrays? 

Good idea. Or override in Lion to an empty array would work as well

> let's add a css variable

Sounds good. Shadow part could work again here


@Knorgias, could you look into the above?

tlouisse (approved):

PR #1352, opened 2021-04-27
Typo in `mulipleChoice`

ffoodd:
I was seeking for exactly this example, just noticed it. Thanks for the Umbrella form, it's really valuable :)


tlouisse:
> I was seeking for exactly this example, just noticed it. Thanks for the Umbrella form, it's really valuable :)

Way better with the 't' :) Thanks 👍

tlouisse (approved):

Issue #1354, opened 2021-04-28
Update documentation for creating a custom LionField extension


tlouisse:
https://lion-web.netlify.app/guides/how-to/creating-a-custom-field/ contains a few mistakes. It would be nice to add a working/running example as well.

A correct example can be found here: https://webcomponents.dev/edit/QEdq4Jm01GPGR8sS7zEd/src/index.js
More context: https://lit-and-friends.slack.com/archives/CJGFWJN9J/p1619619066121200

PR #1359, opened 2021-05-01
Doc/field and input tutorials

tlouisse:
closes https://github.com/ing-bank/lion/issues/1354


packages/input-date/src/LionInputDate.js
```diff
@@ -34,6 +34,8 @@ export class LionInputDate extends LocalizeMixin(LionInput) {
     this.parser = value => (value === '' ? undefined : parseDate(value));
     this.formatter = formatDate;
     this.defaultValidators.push(new IsDate());
+    // Just explicitly make clear we shouldn't use type 'date'
+    this.type = 'text';
```
tlouisse, line 38: When it was set in connectedCallback, initial parsers / serializers would not get the right value (because they run before connectedC.)

daKmoR (approved):

Issue #195, opened 2019-07-22
[calendar] Date bounds not clear for SR user.


erikkroes:
When using a screen reader, it is not clearly communicated when a minimum or maximum date is reached. The keyboard navigation for these dates seems to be partially undefined.
For example, using VO+Safari, the screen reader stops using the keyboard interaction that lion offers. It tries to navigate the dates out of range as if they were normal text.

Following the pattern mentioned in #194 would be a solution for navigation. 

This would still leave an issue. It would be hard for a screen reader user to know whether they are browsing a disabled date, or are outside the selectable date range. A text like ", before selectable date range" could be appended to the accessible name. Although I think there might be better phrasing. Suggestions welcome.

ovidiu1:
#468

daKmoR:
1. use `div` instead of button
2. put states (aria-disabled, aria-pressed, aria-current) on td role="gridcell"
3. for disabled dates put only ` aria-disabled` (no change to aria-label)
4. for disabled dates before minDate => aria-describe-by... element with `before first available date`
5. for disabled dates after maxDate => aria-describe-by... element with `after last available date`
6. Do not force any "automatic" assistance for finding available dates to keyboard users 

What to test
- VO + safari
- NVDIA + Firefox
- JAWS + chrome

daKmoR:
if unclear - I'm picking this and the other calendar a11y bugs up

Issue #194, opened 2019-07-22
[calendar] Disabled dates get skipped by keyboard


erikkroes:
When a calendar has disabled dates and a user navigates by keyboard, these dates get skipped. This is logical behaviour if we would see the calendar as just a set of buttons. Usually interactive elements that are disabled can't be navigated to.

In a calendar however, what we are mainly navigating is not the buttons, but the grid of dates. The buttons just happen to be in those cells. There is value in the coherence and structure of the grid, and being able to browse these disabled dates. 

For example, a screen reader user goes through a calendar. They would like to select a Monday, but for several weeks, all Mondays are unavailable. In the current pattern, they would have to use right or left arrow and go through all weekdays before they find a Monday. If the disabled dates were navigable, they would be able to use the down and up arrow to go through just the Mondays. 

I would suggest changing the keyboard navigation to include disabled buttons. For added convenience, I would suggest prepending something like "Unavailable, " to the accessible name of these dates.

ovidiu1:
#468

daKmoR:
As we are using native buttons and native buttons can not be focused if they are disabled we basically "can not" do this...

If we need a possibility to focus disable dates then these dates should NOT be buttons... we could make a fake `div` button - however, I assume this will have other a11y implications?

anyone else see a better solution?

daKmoR:
sooo we had a few discussion around this - quite complicated 🙈 

it all starts with that having focusable disabled native browser buttons is impossible.
Spec Highlight: `disabled controls are entirely non-interactive`
Spec Paragraph:

> The difference between disabled and readonly is that read-only controls are still focusable, so the user can still select the text and interact with it, whereas disabled controls are entirely non-interactive. (For this reason, only text controls can be made read-only: it wouldn't make sense for checkboxes or buttons, for instances.)

https://www.w3.org/TR/html50/forms.html#attr-input-readonly

Therefore we need to tackle this in a different way 🙈 

We think this is the most promising solution
- Use a `div` with role=“button“ instead of a native button
- Use `disabled` attribute for styling
- Use `aria-disabled` for screen readers
- Manage click/enter/space from within the Calendar

daKmoR:
an example in react with lots of divs 🤣
http://react-day-picker.js.org/examples/disabled-interaction/

PR #1360, opened 2021-05-03
wip: allow disabled dates in current month view

tlouisse:
Just a poc branch for https://github.com/ing-bank/lion/issues/194
Feel free to implement differently

It does:
- based on date boundaries of current view, makes disabled dates navigable
- keep buttons for enabled nav (because they give keyboard selection for free and backwards compatibility in case extension layers styled pseudo states)  
- add div[role=buttton][aria-disabled] for disabled days 
- normalize styling for the two types of buttons

Plus:
- normalize styling of headings (adopt font-size of context)
- doesn't put 2 h2s next to eachother

TODO:
- still buggy, it adds one day too much atm when navigating
- request translations for the "Unavailable prefix"
- While we're at it, also do: https://github.com/ing-bank/lion/issues/195

### DIFF
--- a/packages/calendar/src/LionCalendar.js
+++ b/packages/calendar/src/LionCalendar.js
@@ -171,6 +171,11 @@ export class LionCalendar extends LocalizeMixin(LitElement) {
     this.__connectedCallbackDone = false;
     /** @private */
     this.__eventsAdded = false;
+    /**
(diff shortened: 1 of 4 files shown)


gerjanvangeest:
Issues 194 and 195 are already solved via another PR

PR #1370, opened 2021-05-06
feat: update to lit 2

daKmoR:
## What I did

1. Update to lit 2

## Todo

- [x] Update all reexports in `@lion/core`
- [x] `this.constructor.render` no longer exists => find a way to render scoped elements 🤔
- [x] All tests pass
- [x] Remove usage of deprecated methods

### DIFF
--- a/.changeset/fuzzy-snails-attack.md
+++ b/.changeset/fuzzy-snails-attack.md
@@ -1,0 +1,40 @@
+---
+'babel-plugin-extend-docs': minor
+'providence-analytics': minor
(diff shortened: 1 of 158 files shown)


tlouisse:
`constructor.render` is just the render function, right (`LitElement.render = render`)?

You would just have to call it with the right args? 
(`render(templateResult, this.renderRoot, { scopeName: this.localName, eventContext: this });` by default)

Or is it more complicated?

tlouisse (approved):

PR #1371, opened 2021-05-06
Update public API of ajax & refactor package

ruphin:
This PR changes the Ajax API to `ajax.fetch()` instead of `ajax.request()`


ruphin:
TODO: add changeset

narzac (approved):

packages/ajax/src/interceptors/createXsrfRequestInterceptor.js
```diff
@@ -33,7 +18,7 @@ export async function acceptLanguageRequestInterceptor(request) {
  * @param {Document | { cookie: string }} _document overwriteable for testing
  * @returns {RequestInterceptor}
  */
-export function createXSRFRequestInterceptor(cookieName, headerName, _document = document) {
+export function createXsrfRequestInterceptor(cookieName, headerName, _document = document) {
```
jorenbroekema, line 21: Nitpick: I tend to prefer to keep abbreviations all-capsed for better readability, but this is acceptable for me too.

packages/ajax/src/cache.js
```diff
@@ -147,7 +146,7 @@ export const searchParamSerializer = (params = {}) =>
  * and proactively clean it
  * @param {string} cacheIdentifier usually the refreshToken of the owner of the cache
  */
-const getCache = cacheIdentifier => {
+export const getCache = cacheIdentifier => {
```
jorenbroekema, line 149: If we export this now, we should have some tests for it I think that test this import and its functionality in isolation

jorenbroekema (changes requested):

packages/ajax/index.js
```diff
@@ -1,15 +1,22 @@
...
+ * All interactions with {@link:ajax} after the call to setAjax will use this new instance
+ * (so make sure to call this method before dependant code using {@link:ajax} is ran and this
+ * method is not called by any of your (indirect) dependencies.)
+ * @param {Ajax} newAjax the globally used instance of {@link:ajax}.
+ */
+export function setAjax(newAjax) {
+  ajax = newAjax;
+}
+
+export { acceptLanguageRequestInterceptor } from './src/interceptors/acceptLanguageRequestInterceptor.js';
```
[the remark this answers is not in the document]
narzac: Thanks

- Keeping the interceptors  then
- Changing createXSRFRequestInterceptor to  createXsrfRequestInterceptor, followed the camel case convention we use,  like ING -> `getIngSth`, i can revert it if this is not consistent with the rest of the code.
- i think getCookie is private in this context, and used in createXsrfRequestInterceptor internally, so consumers don't need to use getCookie to be able to use  createXsrfRequestInterceptor. Here:

```js
export function createXsrfRequestInterceptor(cookieName, headerName, _document = document) {
  /**
   * @type {RequestInterceptor}
   */
  async function xsrfRequestInterceptor(request) {
    const xsrfToken = getCookie(cookieName, _document);
    if (xsrfToken) {
      request.headers.set(headerName, xsrfToken);
    }
    return request;
  }

  return xsrfRequestInterceptor;
}
```

packages/ajax/CHANGELOG.md
```diff
@@ -39,7 +39,7 @@
 
 ### Minor Changes
 
-- 2cd7993d: Set fromCache property on the Response, for user consumption. Allow setting cacheOptions on the AjaxClient upon instantiation. Create docs/demos.
+- 2cd7993d: Set fromCache property on the Response, for user consumption. Allow setting cacheOptions on the Ajax upon instantiation. Create docs/demos.
```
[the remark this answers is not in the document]
narzac: Good catch thanks, yeah it is a find/replace gone wild 🤦‍♂️

packages/ajax/src/interceptors/createXsrfRequestInterceptor.js
narzac: Changing `createXSRFRequestInterceptor` to `createXsrfRequestInterceptor`, followed the camel case convention we use,  like ING -> `getIngSth`, I can revert it if this is not consistent with the rest of the code.

packages/ajax/index.js
```diff
@@ -1,22 +1,8 @@
-import { Ajax } from './src/Ajax.js';
-
 export { Ajax } from './src/Ajax.js';
 export { AjaxFetchError } from './src/AjaxFetchError.js';
-
-export let ajax = new Ajax(); // eslint-disable-line import/no-mutable-exports
```
jorenbroekema: Maybe:
```js
export const ajax = new Ajax();
```

So people still have this easy way of making sure they always import the same instance, since ES modules are singletons in essence. Optionally, they can import the Ajax (manager/class) instead if they want to create their own instance, but the most common use case is probably that people will just import the singleton instance.

packages/ajax/test/index.test.js
```diff
@@ -15,16 +13,6 @@ describe('public interface', () => {
     expect(Ajax).to.exist;
   });
 
-  it('exports an instance of Ajax', () => {
```
jorenbroekema: With my previous comment, this test would stay

jorenbroekema (changes requested):

packages/ajax/index.js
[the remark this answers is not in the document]
jorenbroekema: Ah so it's just a private utility used for the XSRF request interceptor "factory", then yeah keep it private 👍

packages/ajax/index.js
narzac: done

packages/ajax/test/index.test.js
narzac: done

jorenbroekema (approved):
Let's squash commit history, make it a feat instead of chore commit and perhaps adjust the changeset a little bit to make clear that there's lots of **breaking** changes due to public API changes

PR #1375, opened 2021-05-10
Fix documentation typo in attribute name

vnq:
## What I did

1. Fix documentation typo in attribute name (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly)


daKmoR:
thank you 🤗

PR #1377, opened 2021-05-11
feat: add _getTextboxValueFromOption method

MatthieuLebigre:
## What I did

This PR aims to fix this discussion: https://github.com/ing-bank/lion/discussions/1247

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -320,7 +320,10 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
     if (name === 'modelValue' && this.modelValue && this.modelValue !== oldValue) {
       if (this._syncToTextboxCondition(this.modelValue, this._oldModelValue)) {
         if (!this.multipleChoice) {
-          this._setTextboxValue(this.modelValue);
+          const textboxValue = this._getTextboxValueFromOption(
+            this.formElements[/** @type {number} */ (this.checkedIndex)],
+          );
+          this._setTextboxValue(textboxValue);
         } else {
           this._syncToTextboxMultiple(this.modelValue, this._oldModelValue);
         }
@@ -337,8 +340,12 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    */
   __unsyncCheckedIndexOnInputChange() {
     const autoselect = this._autoSelectCondition();
-    if (!this.multipleChoice && !autoselect && !this._inputNode.value.startsWith(this.modelValue)) {
-      this.checkedIndex = -1;
+    const checkedElement = this.formElements[/** @type {number} */ (this.checkedIndex)];
+    if (!this.multipleChoice && !autoselect && checkedElement) {
+      const textboxValue = this._getTextboxValueFromOption(checkedElement);
+      if (!this._inputNode.value.startsWith(textboxValue)) {
+        this.checkedIndex = -1;
+      }
     }
   }
 
@@ -396,8 +403,9 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    */
   matchCondition(option, textboxValue) {
     let idx = -1;
-    if (typeof option.choiceValue === 'string' && typeof textboxValue === 'string') {
-      idx = option.choiceValue.toLowerCase().indexOf(textboxValue.toLowerCase());
+    const inputValue = this._getTextboxValueFromOption(option);
+    if (typeof inputValue === 'string' && typeof textboxValue === 'string') {
+      idx = inputValue.toLowerCase().indexOf(textboxValue.toLowerCase());
     }
 
     if (this.matchMode === 'all') {
@@ -450,6 +458,17 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
       return /** @type {boolean} */ (this.opened);
     }
     return true;
+  }
+
+  /**
+   * Return the value to be used for the input value
+   * @overridable
+   * @param {LionOption} option
+   * @returns {string}
+   */
+  // eslint-disable-next-line class-methods-use-this
+  _getTextboxValueFromOption(option) {
+    return option.choiceValue;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/combobox/src/LionCombobox.js
```diff
@@ -850,7 +870,11 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    */
   _syncToTextboxMultiple(modelValue, oldModelValue = []) {
     const diff = modelValue.filter(x => !oldModelValue.includes(x)).toString();
-    this._setTextboxValue(diff); // or last selected value?
+    const newValue = this.formElements
+      .filter(option => diff.includes(option.choiceValue))
+      .map(option => this._getTextboxValueFromOption(option))
+      .join(' ');
```
jorenbroekema, line 876: So multiple choices are concatenated with a space? Can we add a test for that?

packages/combobox/src/LionCombobox.js
```diff
@@ -450,6 +460,17 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
...
   }
 
+  /**
+   * Return the value to be used for the input value
+   * @overridable
+   * @param {LionOption} option
+   * @returns {string}
+   */
+  // eslint-disable-next-line class-methods-use-this
+  _getTextboxValueFromOption(option) {
```
jorenbroekema, line 470: Can we add some tests for this method and the expected behaviors when you override this method?

packages/combobox/src/LionCombobox.js
```diff
@@ -335,8 +340,12 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    */
   __unsyncCheckedIndexOnInputChange() {
     const autoselect = this._autoSelectCondition();
-    if (!this.multipleChoice && !autoselect && !this._inputNode.value.startsWith(this.modelValue)) {
-      this.checkedIndex = -1;
+    const checkedElement = this.formElements[/** @type {number} */ (this.checkedIndex)];
+    if (!this.multipleChoice && !autoselect && checkedElement) {
```
jorenbroekema, line 344: If checkedElement does not exist, should the checkedIndex be set to -1 as well? That edge case doesn't seem covered by the logic here, but I could be wrong as no tests are failing.

jorenbroekema (dismissed):
Mostly just adding tests and running `npm run format` so prettier doesn't complain :). Thanks for creating the PR!

packages/combobox/src/LionCombobox.js
MatthieuLebigre: I believe if checkedElement doesn't exist, checkedIndex should already be set to -1. I don't think that's the place where it should set ?

packages/combobox/src/LionCombobox.js
MatthieuLebigre: done

packages/combobox/src/LionCombobox.js
MatthieuLebigre: I've added some tests. But I'm not so sure to understand what is to be done here. I've copy/paste the code from the discussion https://github.com/ing-bank/lion/discussions/1247

MatthieuLebigre:
> Mostly just adding tests and running `npm run format` so prettier doesn't complain :). Thanks for creating the PR!

For prettier, when I run `npm run format` it modifies a bunch of files, but not the one I've changed :D Have you already experienced that?

tlouisse:
> > Mostly just adding tests and running `npm run format` so prettier doesn't complain :). Thanks for creating the PR!
> 
> For prettier, when I run `npm run format` it modifies a bunch of files, but not the one I've changed :D Have you already experienced that?

It appears that different configs of the Prettier vscode plugin can explain those differences (even though they take .eslintrc, prettier config in package.json into consideration.) 
Not sure what 'wins' and how we can best mitigate this problem

MatthieuLebigre:
> > > Mostly just adding tests and running `npm run format` so prettier doesn't complain :). Thanks for creating the PR!
> > 
> > 
> > For prettier, when I run `npm run format` it modifies a bunch of files, but not the one I've changed :D Have you already experienced that?
> 
> It appears that different configs of the Prettier vscode plugin can explain those differences (even though they take .eslintrc, prettier config in package.json into consideration.)
> Not sure yet what 'wins' and when and how we can best mitigate this problem

I tried a lot of different options/configuration but it doesn't work. I tried turning off the VSCode plugin, changing the config to a `.prettierrc` file, but it changes nothing :(

packages/combobox/src/LionCombobox.js
tlouisse: Before, I think setting an array to the textbox implicitly had the same behavior (concatting with spaces). This makes or  more specific and type compatible. But an extra test never hurts ;)
Maybe you can disable Prettier plugin (on autosave) and add your new tests again?

packages/combobox/src/LionCombobox.js
tlouisse: Should indeed be the case that checkedEl will only not be found if index is already -1

tlouisse (changes requested):
Maybe you can reapply the tests with prettier disabled? Then we can review the tests better.
The formatting inconsistencies could be tackled in a separate issue.

packages/combobox/src/LionCombobox.js
MatthieuLebigre: Should be good now for the format. I did some investigation and will open an issue to discuss about it :)

packages/combobox/test/lion-combobox.test.js
```diff
@@ -1575,6 +1575,211 @@ describe('lion-combobox', () => {
       });
     });
 
+    describe('Subclassers', () => {
```
tlouisse, line 1578: Nice :+1: 
Thanks a lot.

Could you maybe rename this describe to `'Override _getTextboxValueFromOption'`  and put it inside the Subclassers describe above?

If @jorenbroekema also agrees, it only needs:
- a reformat of this file, so that it passes the ci
- a changeset 
- a squash of the commits

MatthieuLebigre: - `describe` is moved (be aware there are already several `describe('Subclassers'` in this file)
- changeset added
- squash done
- For the format, unfortunately I can't do it, I'm not able to find the proper way to make prettier work (see https://github.com/ing-bank/lion/issues/1388)

tlouisse: Good catch about the multiple 'Subclassers' describes :) We will wrap up this mr and fix that later.

Prettier seems to work now (after a rebase with latest master, which pobably uses a later Prettier version).

tlouisse:
Everything is solved 👍 Thanks for this contribution

tlouisse (approved):

PR #1379, opened 2021-05-12
fix(switch): fire model-value-changed with isTriggeredByUser: true

jorenbroekema:


tlouisse:
Nice!

Ideally it should also be covered in form-integrations for completeness.
(it should actually fail there atm, but it doesn't -> my bad...)
But you can also make a separate task for that

tlouisse (approved):

tlouisse (approved):

Issue #1376, opened 2021-05-11
[localize] Race condition in LocalizeManager


dereli:
## Expected behavior

After resetting the LocalizeManager, I would expect the `.__storage` to be empty and have no locales.

## Actual Behavior

If a pending promise is fulfilled after a `localize.reset()`, it will `addData` into the storage. This causes unwanted side effects later on.

Here is the failing test:

```javascript
it('should have empty storage after reset', async () => {
    manager = new LocalizeManager();

    let deferredResolve;

    manager.loadNamespace({
        generic: () => new Promise(resolve => {
            deferredResolve = () => resolve({ title: 'Example' });
        }),
    });
    const { loadingComplete } = manager;

    manager.reset();
    expect(manager.__storage).to.be.empty;

    deferredResolve();
    await loadingComplete;

    // storage still needs to be empty
    expect(manager.__storage).to.be.empty;
});
```

## Possible Solution
As a workaround, a reset call can wait for `.loadingComplete` (inside `.reset`  or before being called).

For a better solution, when the promise is fulfilled, it needs to be checked if it's still in the `.__namespaceLoaderPromisesCache` before calling addData.

jorenbroekema:
Hey, good catch and nice test to show the issue! 

I agree with your second proposal, that promises should do a check before calling addData to see if no reset has been done and it is still in namespaceLoaderPromisesCache. Care to create a pull request for this fix?

GuiAmPm:
The test doesn't fail on the most recent version. Perhaps it was fixed through another update?

PR #1387, opened 2021-05-19
Race condition in LocalizeManager

dereli:
Long running imports were polluting newly reset storage (#1376)

## What I did

1.  When the promise is fulfilled, it is checked if it's still in the .__namespaceLoaderPromisesCache before calling addData.

### DIFF
--- a/.changeset/mighty-cars-grow.md
+++ b/.changeset/mighty-cars-grow.md
@@ -1,0 +1,5 @@
+---
+'@lion/localize': patch
+---
+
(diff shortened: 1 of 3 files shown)


jorenbroekema:
Hey could you run `npm run changeset` in your local branch and then select the localize package and do "patch"? Then it will release with a patch version bump and the right changelog message :). Feel free to squash that changeset in your current commit

jorenbroekema (approved):

jorenbroekema:
Added a changelog entry ;) thanks for the fix!

PR #1392, opened 2021-05-20
Fix typo under Private heading of definitions

dbowling:
Fix documentation to specify private methods and properties, rather than protected, with two underscores.

## What I did

1. Changed the word `protected` under the `Private` heading to the word `private`.


tlouisse (approved):

tlouisse:
Good catch! Thanks

PR #1396, opened 2021-05-26
fix: select rich to reuse LionFieldBase (no FormatMixin)

jorenbroekema:
## What I did

1. Create LionFieldBase that doesn't include FormatMixin
2. Make select rich use LionFieldBase to share a common ancestor with other fields


gerjanvangeest:

Issue #1395, opened 2021-05-25
[Combobox] Autocomplete not working fine with dynamic options


MatthieuLebigre:
In a server interaction scenario (data being fetch from a server and populate dynamically the options), some autocomplete features are not working fine

It's easily reproducible on this demo: https://lion-web.netlify.app/components/inputs/combobox/features/#server-interaction

## Expected behavior

In the demo, type the character "l"
=> The `lorem` option should be selected

Then type the character "o"
=> Both options `lorem` and `dolor` are displayed and `lorem` is selected

## Actual Behavior

In the demo, type the character "l"
=> No option is selected

Then type the character "o"
=> Before the "server" returns data, both options `lorem` and `dolor` are displayed and `lorem` is selected - that's correct
      As soon as the "server" returns a response, only `lorem` is displayed

MatthieuLebigre:
I already have a fix locally, I'll push a PR very soon

PR #1398, opened 2021-05-28
fix: better support when options change dynamically

MatthieuLebigre:
## What I did

Added some logic to better handle the autocompletion when options changed (use case of a server interaction).

Fixes https://github.com/ing-bank/lion/issues/1395

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -386,6 +386,7 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
       // Only update list in render cycle
       this._handleAutocompletion();
       this.__shouldAutocompleteNextUpdate = false;
+      this.__listboxContentChanged = false;
     }
 
     if (typeof this._selectionDisplayNode?.onComboboxElementUpdated === 'function') {
@@ -479,6 +480,7 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
   _onListboxContentChanged() {
     super._onListboxContentChanged();
     this.__shouldAutocompleteNextUpdate = true;
+    this.__listboxContentChanged = true;
   }
 
   // eslint-disable-next-line no-unused-vars
@@ -609,7 +611,11 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
       prevValue.length &&
       curValue.length &&
       prevValue[0].toLowerCase() !== curValue[0].toLowerCase();
-    return userIsAddingChars || userStartsNewWord;
+    return (
+      userIsAddingChars ||
+      userStartsNewWord ||
+      (this.__listboxContentChanged && this.__hadUserIntendsInlineAutoFill)
+    );
   }
 
   /* eslint-enable no-param-reassign, class-methods-use-this */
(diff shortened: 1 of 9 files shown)


jorenbroekema:
Hey, thanks for making improvements to this, I actually was facing some non-ideal behavior as well when fetching options dynamically, not sure if your fix is fixing my issue too.

Can you elaborate a little bit on what this fixes, maybe a before & after GIF or a test that shows it?

MatthieuLebigre:
More detailed on what I changed in this PR:

- When the option list changed (and after the user typed something in the textbox), the `_handleAutocompletion` is called again. Before my change, the logic in this method doesn't take into account that the user actually didn't changed anything, so a side effect is the `doAutoSelect()` is called, which leads to a very bad UX experience. I've changed the logic to take this into account and don't select any options if the user didn't trigger it.
- Also in `_handleAutocompletion()` method, for the current value if was taking as current value the whole textbox value, without removing the `selected` part of it, which again was causing trouble after the option list are changing
- Similirarly, I've added a check in `__textboxInlineComplete()` for not loosing inputNode.selectionStart and inputNode.selectionEnd

MatthieuLebigre:
## Before

At the second letter entered, the autocomplete is too "aggressive" and autocomplete the word

lion-combobox-before

## After

Now it's working as intended:

lion-combobox-after

jorenbroekema:
Alright makes sense I think. This was indeed the issue I was facing as well in my combobox implementation.

Would it be possible to create a regression test for this so we don't break this in the future?

MatthieuLebigre:
> Alright makes sense I think. This was indeed the issue I was facing as well in my combobox implementation.
> 
> Would it be possible to create a regression test for this so we don't break this in the future?

done

jorenbroekema:
I noticed there is a commit that shouldn't be in there for localize. This was a change done by Zubeyr, and it's already in master.

Maybe try rebasing / fixing the git problems.

Another thing, why a changeset to release all packages with a minor? We only need to release combobox right?

packages/combobox/src/LionCombobox.js
```diff
@@ -629,7 +631,11 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
   _handleAutocompletion() {
     const hasSelection = this._inputNode.value.length !== this._inputNode.selectionStart;
 
-    const curValue = this._inputNode.value;
+    const inputValue = this._inputNode.value;
+    const inputSelectionStart = this._inputNode.selectionStart;
+    const curValue =
```
tlouisse, line 636: Nice. Makes sense to limit this variable just to the 'non selected part' 👍

packages/combobox/src/LionCombobox.js
```diff
@@ -670,7 +676,10 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
           hasAutoFilled = true;
         };
 
-        if (userIntendsInlineAutoFill) {
+        if (
+          userIntendsInlineAutoFill ||
+          (this.__listboxContentChanged && this.__hadUserIntendsInlineAutoFill)
```
tlouisse, line 681: Maybe we can add this check inside `__computeUserIntendsAutoFill` method?

tlouisse:
Very valuable contribution @MatthieuLebigre, thanks :)

It looks good, only had a small remark to move 1 line of code to a different method.

And then indeed the commit history (we only need the last one) and changeset need to be fixed as @jorenbroekema  pointed out

MatthieuLebigre:
I fixed the latest comments, but didn't managed to restore properly the git history ... :(
I've opened a new one instead: https://github.com/ing-bank/lion/pull/1421
Sorry for the trouble...

PR #1402, opened 2021-05-31
chore: improve lit 2 release info and only release if needed

daKmoR:
## What I did

1. lit update does not affect most node-packages
2. add more info about lit 2 update

### DIFF
--- a/.changeset/fuzzy-snails-attack.md
+++ b/.changeset/fuzzy-snails-attack.md
@@ -1,8 +1,5 @@
 ---
-'babel-plugin-extend-docs': minor
 'providence-analytics': minor
-'publish-docs': minor
-'remark-extend': minor
 '@lion/accordion': minor
 '@lion/button': minor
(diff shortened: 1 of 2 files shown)


.changeset/fuzzy-snails-attack.md
```diff
@@ -37,4 +34,46 @@
...
+```
+
+```html
+<!-- before (ScopedElements 1.x) -->
+<my-element>
+  #shadow-root
+  <my-button-23243424>click me</my-button-23243424>
+</my-element>
+
+<!-- after (ScopedElements 2.x) -->
```
Knorgias, line 74: Before and after look the same? 🤔

Knorgias (approved):

.changeset/fuzzy-snails-attack.md
daKmoR: oups fixed... should be without the hash at the end of the tag

PR #1403, opened 2021-05-31
fix: try awaiting render for missing name attributes

jorenbroekema:
## What I did

Draft fix for issue described by @alangdm https://lit-and-friends.slack.com/archives/CJGFWJN9J/p1622445542027800

Apparently React or lit-labs/react first initialises the component with default values and assigns the properties on a second pass, so when the form element is first connected it still doesn't have a name

Before merging this we should add a test that specifically tests that a form element is able to register when its `name` attribute is set during render (`update` lifecycle in Lit perhaps), so that it can be expected after updateComplete but not before.

### DIFF
--- a/packages/form-core/src/registration/FormRegistrarMixin.js
+++ b/packages/form-core/src/registration/FormRegistrarMixin.js
@@ -4,12 +4,13 @@ import { FormControlsCollection } from './FormControlsCollection.js';
 import { FormRegisteringMixin } from './FormRegisteringMixin.js';
 
 /**
+ * @typedef {import('@lion/core').LitElement} LitElement
  * @typedef {import('../../types/FormControlMixinTypes').FormControlHost} FormControlHost
  * @typedef {import('../../types/registration/FormRegistrarMixinTypes').FormRegistrarMixin} FormRegistrarMixin
  * @typedef {import('../../types/registration/FormRegistrarMixinTypes').FormRegistrarHost} FormRegistrarHost
  * @typedef {import('../../types/registration/FormRegistrarMixinTypes').ElementWithParentFormGroup} ElementWithParentFormGroup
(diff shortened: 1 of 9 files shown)


jorenbroekema:
Updated the other addFormElement override methods to be async and await super.addFormElement. One test still failing tho, couldn't figure it out in a timely manner

lpoulter:
I'm also experiencing this issue. Any update on what is needed to merge this?

jorenbroekema:
There was one failing test and I couldn't really figure out why it was failing, and I dont think anyone from the team reviewed this yet. I am not sure if I can find time to dive deeper into that test so if possible I would delegate it to maybe @tlouisse  who is our form system maestro. Or if someone else figures it out, also works. I do think this should have some prio because as it is lion form doesn't work very well with lit react wrapper, although afaik it does with the other wrapper that I forgot the name of, "reactify" or something?

packages/form-core/src/registration/FormRegistrarMixin.js
```diff
@@ -154,10 +155,16 @@ const FormRegistrarMixinImplementation = superclass =>
...
       // 2. Add children as object key
       if (this._isFormOrFieldset) {
-        const { name } = child;
+        let { name } = child;
         if (!name) {
-          console.info('Error Node:', child); // eslint-disable-line no-console
-          throw new TypeError('You need to define a name');
+          // Give child a chance to render if property bindings are done later
+          // E.g. for @lit-labs/react
+          await child.updateComplete;
```
tlouisse, line 162: Sorry for the late feedback (totally occupied with smth else back then and forgot about this one)....

So the goal is to prevent React from throwing name errors? 
Maybe we can always wait for the updateComplete (via .then) and afterwards throw the error.
Allows us to keep the method sync (since I'm afraid it can lead to nasty timing errors in our extension layers otherwise)

gerjanvangeest:
This PR is abandoned, please reopen if you want to update it.

PR #1410, opened 2021-06-02
Add legacy branch for release

Knorgias:
Created a new "legacy" branch from last commit pre lit2 dependency update.
With this PR the goal is to: set up github actions workflows to also trigger on PR/merge to legacy branch

## What I did

1. Added new "legacy" branch in workflow for releasing legacy patch updates


Knorgias:
I cannot access the “Settings” of ing-bank/lion in order to protect branch "legacy" with the same protection rules as master.
I guess I'll either have to be granted permissions or one of the repository owners can do this instead.

gerjanvangeest (approved):

.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
```
jorenbroekema, line 6: What if we remove master here, and just merge this branch to legacy instead of to master?

jorenbroekema (changes requested):

jorenbroekema:
I'll fix the settings when this is merged :)

jorenbroekema:
Sorry for all the formatting changes, is a result of https://github.com/prettier/prettier/blob/main/CHANGELOG.md#231  Preserve attributes order for element node + Fix incorrectly wrapped arrow functions with return types change

PR #1418, opened 2021-06-10
fix(switch): stop checked-changed event from propagating too far

jorenbroekema:


packages/switch/src/LionSwitchButton.js
```diff
@@ -114,7 +114,6 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
   __checkedStateChange() {
     this.dispatchEvent(
       new Event('checked-changed', {
-        composed: true,
         bubbles: true,
```
tlouisse, line 117: can we also not bubble (and thus not stop propagation?)

tlouisse (approved):

Issue #1413, opened 2021-06-08
[ListboxMixin] Miscalculated offset value in isInView function


AndreeaMoise:
## Expected behavior
I have a container with multiple headers elements with "position-sticky" and a lion listbox.  When I navigate through the listbox using the arrow keys, I expect the focused listbox option to be fully visible (in view).

## Actual Behavior
When I navigate through the listbox using the arrow keys, the focused listbox option is not fully visible (in view).

This seems to be caused by the way the top offset of an element is calculated in the isInView function when the container includes elements with position sticky. This incorrect behaviour is documented [here](https://medium.com/@alexcambose/js-offsettop-property-is-not-great-and-here-is-why-b79842ef7582) and the reproduction is [here](https://webcomponents.dev/edit/yyQ1R6gG3fMdEdwLjP3v/src/index.js). 

## Additional context
Adding scroll-margin to the lion-option does not fully solve the problem.

 "@lion/core": "0.17.0",
 "@lion/listbox": "0.9.0",

PR #1419, opened 2021-06-10
fix(listbox): take into account sticky/fixed el for offsetTop calc

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1413

https://webcomponents.dev/edit/cE9kPezoQwaAPfRbdQEd/ as shown here, our isInView util remains a bit naive of random elements blocking view of the option. 

Might be better after all to just call scrollIntoView, always, and let the browser decide if it is necessary to scroll, seems like browser is smarter in this regard?

### DIFF
--- a/.changeset/dull-paws-prove.md
+++ b/.changeset/dull-paws-prove.md
@@ -1,0 +1,5 @@
+---
+'@lion/listbox': patch
+---
+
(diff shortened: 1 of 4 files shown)


jorenbroekema:
Superseded by https://github.com/ing-bank/lion/pull/1427

PR #1427, opened 2021-06-17
fix(listbox): always scroll active item into view

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1413

### DIFF
--- a/packages/listbox/src/ListboxMixin.js
+++ b/packages/listbox/src/ListboxMixin.js
@@ -58,31 +58,6 @@ function uuid() {
 }
 
 /**
- * @param {HTMLElement} container
- * @param {HTMLElement} element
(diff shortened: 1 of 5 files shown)


packages/listbox/src/ListboxMixin.js
```diff
@@ -714,6 +689,16 @@ const ListboxMixinImplementation = superclass =>
       this._listboxNode.focus();
     }
 
+    /**
+     * @protected
```
tlouisse, line 693: TODO: add comment on why we have `_scrollIntoView ` method (to allow Subclassres to adjust the animation: like non smooth behavior, different timing etc.)

gerjanvangeest (approved):

PR #1433, opened 2021-06-30
fix(input-datepicker): set isTriggeredByUser on change via calendar

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/input-datepicker/src/LionInputDatepicker.js
@@ -193,6 +193,8 @@ export class LionInputDatepicker extends ScopedElementsMixin(
     this._hideOnUserSelect = true;
     /** @protected */
     this._syncOnUserSelect = true;
+    /** @protected */
+    this._isHandlingCalendarUserInput = false;
 
     /** @private */
     this.__openCalendarOverlay = this.__openCalendarOverlay.bind(this);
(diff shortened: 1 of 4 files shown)


packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -374,10 +377,34 @@ export class LionInputDatepicker extends ScopedElementsMixin(
     }
     if (this._syncOnUserSelect) {
       // Synchronize new selectedDate value to input
+      this._isCalendarTriggeredByUser = true;
```
tlouisse, line 380: Nice improvement 👍 

Could you configure the parent FormatMixin here (use variable `_isHandlingUserInput` instead of `_isCalendarTriggeredByUser`)?

Then we can also leave out the _dispatchModelValueChangedEvent, because it uses the one from FormatMixin

https://github.com/ing-bank/lion/blob/2b73c50f6e7ff9ef96122a74536e7c4df9b8553b/packages/form-core/src/FormatMixin.js#L307

tlouisse:
Looks good (see comments).
Fro completeness, I would maybe make a separate issue to also add an integration test here: https://github.com/ing-bank/lion/blob/master/packages/form-integrations/test/model-value-consistency.test.js#L343
(unless you want to do it right away, of course)

packages/input-datepicker/src/LionInputDatepicker.js
gerjanvangeest: I tried, but it broke the input. The test worked, but the value didn't get displayed in the demo.

packages/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -12,6 +12,7 @@ import { LionCalendarOverlayFrame } from './LionCalendarOverlayFrame.js';
 
 /**
  * @typedef {import('@lion/core').RenderOptions} RenderOptions
+ * @typedef {import('@lion/form-core/types/FormControlMixinTypes.js').ModelValueEventDetails} ModelValueEventDetails
```
tlouisse, line 15: Nice. Only thing is that this line could be lef out now

gerjanvangeest: done

tlouisse (approved):

PR #1436, opened 2021-07-02
Fix/button toggling

maartenst:
## What I did

1.fix the button toggling so it is always toggled if relevant values change


gerjanvangeest (approved):
Hi, thanks for the fix.
I added another one on top of it, hope you don't mind :)

Issue #1446, opened 2021-07-07
[validate] Validation Feedback "fieldName" timing issue with localization


jorenbroekema:
## Expected behavior

If I change the locale, and my label is localized and used as `fieldName` inside my validation feedback, the validation feedback should update to render the fieldName in the validation feedback, in the new locale.

## Actual Behavior

![155SyWx](https://user-images.githubusercontent.com/36734656/124769233-97551e80-df39-11eb-8ca9-7d0b7294de29.gif)

It seems like there is a timing issue, where the label gets updated properly, and the fieldName also gets updated, but not in time for the validation feedback re-render, since it renders the old fieldName in the previous locale.

I will try to create a failing test for this in form-integrations package (as it is an integration issue between validate/form-core & localize) and update this issue.

PR #1447, opened 2021-07-07
fix(localize,form-core): wait for localize before updating feedback

jorenbroekema:
fixes https://github.com/ing-bank/lion/issues/1446

### DIFF
--- a/packages/form-integrations/test/form-validation-integrations.test.js
+++ b/packages/form-integrations/test/form-validation-integrations.test.js
@@ -1,6 +1,7 @@
 import { expect, fixture, defineCE } from '@open-wc/testing';
 import { html, unsafeStatic } from 'lit/static-html.js';
 import { Required, DefaultSuccess, Validator } from '@lion/form-core';
+import { localize } from '@lion/localize';
 import { loadDefaultFeedbackMessages } from '@lion/validate-messages';
 import { LionInput } from '@lion/input';
 import sinon from 'sinon';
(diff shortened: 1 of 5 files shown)


packages/form-integrations/test/form-validation-integrations.test.js
```diff
@@ -117,5 +118,63 @@ describe('Form Validation Integrations', () => {
...
+      await el.updateComplete;
+      await el.feedbackComplete;
+      const { _feedbackNode } = getFormControlMembers(el);
+      expect(_feedbackNode.feedbackData?.[0].message).to.equal('Please enter a(n) Text.');
+
+      localize.locale = 'nl-NL';
+      await el.updateComplete;
+      await el.feedbackComplete;
+      expect(el.label).to.equal('Tekst');
+      expect(_feedbackNode.feedbackData?.[0].message).to.equal('Vul een Tekst in.');
```
jorenbroekema, line 171: This was failing because "Tekst" was still "Text" from the old locale. Now that it awaits localize.loadingComplete, it will get the right localized label in the feedback message

tlouisse:
Nice catch and fix.

Maybe we can solve it in a way such that ValidateMixin (and with that the complete form-core package) becomes i18n system agnostic?
See: https://github.com/ing-bank/lion/blob/master/packages/form-core/src/validate/ValidateMixin.js#L3  😬 

We should actually get rid of the lines here: https://github.com/ing-bank/lion/blob/master/packages/form-core/src/validate/ValidateMixin.js#L256

Maybe we can solve it with a controller, like this?
### form-core: localization system agnostic 
```js
// ValidateManager
export class ValidateManager {     
  addLocalizeController(ctrl) {        
    this.localizeControllers.push(ctrl);     
  } 
}
ValidateManager.localizeControllers = [];

 // ValidateMixin
 import { ValidateManager } from ‘./ValidateManager.js’;  
 
 …. constructor() {   
 ValidateManager.localizeControllers.forEach(ctrl => {
    ctrl(this);
  });
 }
  ….   
```
  
### validate-messages
  
  ```js
 import { ValidateManager } from ‘@lion/form-core’;
 import {localize} from ‘@lion/localize;

  ValidateManager.addLocalizeController(class LionLocalizeController {  
  …
 hostConnected() {    
    localize.addEventListener('localeChanged', this._updateValidateMixinFeedbackComponent);   
 }   
 …   
 async _updateValidateMixinFeedbackComponent(ev) { 	
   await localize.loadingComplete; 	
   this.host._updateFeedbackComponent(ev);   
 }
  });
 ```

tlouisse:
If the proposal is a bit too much and you need the fix on a short term, we can also merge this fix and add the proposal as an issue/idea 👍

tlouisse (approved):

PR #1449, opened 2021-07-08
Add required peer dependencies

ebabel-eu:
Out of the box, installing a @lion package in a new project doesn't include the required peer dependencies lit-html and lit-element. I suggest they are mentioned in this documentation.

## What I did

1. I added a section in the documentation.

### DIFF
--- a/docs/guides/how-to/get-started.md
+++ b/docs/guides/how-to/get-started.md
@@ -24,7 +24,7 @@ npm init @open-wc
 ### Install peer dependencies lit-html and lit-element
 
 ```bash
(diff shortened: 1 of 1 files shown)


daKmoR:
that seems strange... as every component depends on `@lion/core` which depends on `lit` which depends on `lit-html` and `lit-element`.

there should be no need to installing peer dependencies? 

could you open an issue with an example repo or clear instructions on how to reproduce the error you are seeing?

PS: closing this PR as it sounds more like an issue (until proven otherwise) 😅

Issue #1415, opened 2021-06-09
[input][textarea][Safari] Can't move focus to a different input after writing some text


alangdm:
## Expected behavior

When using lion-input/textarea in Safari, trying to tap on a different form element will move the focus to said form element.

## Actual Behavior

When viewing a page with code like the following in either Mac OS or iOS Safari:

```html
<lion-form>
  <form>
    <lion-input label="Name" name="name"></lion-input>
    <lion-input label="Avatar" name="avatar"></lion-input>
  </form>
</lion-form>
```

If you do the following actions the focus doesn't move as expected:

1. Click on the first input
2. Write something
3. Try to click on a different input in the same form (the focus doesn't move)
4. Try to click a second time (the focus will move)

https://user-images.githubusercontent.com/6799159/121306449-cdeb4980-c939-11eb-923c-0241692aa92a.mov

As you can see in the video, if you click one input but don't write anything, the focus will move properly.

Here's a simple code reproduction:

https://webcomponents.dev/edit/dpk7kxtGiNZF4GnT07gE/src/index.js

## Additional context

I've tested this both on the latest versions (based on form-core v0.14.0) and a past version (based on form-core v0.12.0)

This only happens in Safari, it doesn't happen in Android Chrome nor in Firefox/Chrome for Mac OS (haven't tested on Windows)

### Event logs

I tried logging the focusin, focusout, focus, blur, change, and input events for the reproduction above. Here are some results:

I think the cause for this bug might be the extra `focus` and `focusin` events that are getting fired in Safari when clicking on another form element after the text input.

#### Mac OS Safari - with text input

> Red: Click on first input
> Yellow: text input
> Green: First attempt at moving focus - failed
> Blue: Second attempt at moving focus - success

<img width="749" alt="Safari Logs - with input" src="https://user-images.githubusercontent.com/6799159/121307850-62a27700-c93b-11eb-9302-214bbbfe1519.png">

#### Mac OS Chrome - with text input

> Red: Click on first input
> Yellow: text input
> Green: First attempt at moving focus - success

![Chrome Logs - with input](https://user-images.githubusercontent.com/6799159/121307934-7948ce00-c93b-11eb-9b3d-c7a3d35d1345.png)

#### Mac OS Safari - without text input

> Red: Click on first input
> Yellow: First attempt at moving focus - success

[...]

PR #1450, opened 2021-07-08
fix(form-core): only preserve caret if value changed

gerjanvangeest:
## What I did

- [x] fix: https://github.com/ing-bank/lion/issues/1415
- [x] create test suite for NativeTextFieldMixin

### DIFF
--- a/packages/form-core/src/NativeTextFieldMixin.js
+++ b/packages/form-core/src/NativeTextFieldMixin.js
@@ -67,16 +67,25 @@ const NativeTextFieldMixinImplementation = superclass =>
       }
     }
 
+    /**
+     * The view value. Will be delegated to `._inputNode.value`
+     * @override FormatMixin
+     */
     get value() {
       return (this._inputNode && this._inputNode.value) || this.__value || '';
     }
 
-    // We don't delegate, because we want to preserve caret position via _setValueAndPreserveCaret
-    /** @param {string} value */
+    /**
+     * @param {string} value
+     * @override FormatMixin - We don't delegate, because we want to preserve caret position via _setValueAndPreserveCaret
+     */
     set value(value) {
       // if not yet connected to dom can't change the value
       if (this._inputNode) {
-        this._setValueAndPreserveCaret(value);
+        // Only set if newValue is new, fix for Safari bug: https://github.com/ing-bank/lion/issues/1415
+        if (this._inputNode.value !== value) {
+          this._setValueAndPreserveCaret(value);
+        }
         /** @type {string | undefined} */
         this.__value = undefined;
       } else {
--- a/packages/form-core/test-helpers/getFormControlMembers.js
+++ b/packages/form-core/test-helpers/getFormControlMembers.js
@@ -12,7 +12,7 @@ export function getFormControlMembers(el) {
   // eslint-disable-next-line
   const { _inputNode, _helpTextNode, _labelNode, _feedbackNode, _allValidators } = el;
   return {
-    _inputNode,
+    _inputNode: /** @type {* & FormControlHost} */ (el)._inputNode,
     _helpTextNode,
     _labelNode,
     _feedbackNode,
--- a/packages/form-core/test-suites/NativeTextFieldMixin.suite.js
+++ b/packages/form-core/test-suites/NativeTextFieldMixin.suite.js
@@ -1,0 +1,69 @@
+import { LitElement } from '@lion/core';
+import { getFormControlMembers } from '@lion/form-core/test-helpers';
+import { defineCE, expect, fixture, html, triggerFocusFor, unsafeStatic } from '@open-wc/testing';
+import { sendKeys } from '@web/test-runner-commands';
+import { spy } from 'sinon';
+import { NativeTextFieldMixin } from '../src/NativeTextFieldMixin.js';
+
+/**
+ * @typedef {import('../types/FormControlMixinTypes').FormControlHost} FormControlHost
+ * @typedef {ArrayConstructor | ObjectConstructor | NumberConstructor | BooleanConstructor | StringConstructor | DateConstructor | 'iban' | 'email'} modelValueType
+ */
+
+/**
+ * @param {{tagString?: string, modelValueType?: modelValueType}} [customConfig]
+ */
+export function runNativeTextFieldMixinSuite(customConfig) {
+  const cfg = {
+    tagString: null,
+    ...customConfig,
+  };
+
+  describe('NativeTextFieldMixin', () => {
+    class NativeTextFieldClass extends NativeTextFieldMixin(LitElement) {
+      get slots() {
+        return {
+          // NativeTextFieldClass needs to have an _inputNode defined in order to work...
+          input: () => document.createElement('input'),
+        };
+      }
+    }
+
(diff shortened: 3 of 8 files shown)


packages/form-core/test-suites/NativeTextFieldMixin.suite.js
```diff
@@ -0,0 +1,63 @@
...
+      await el.updateComplete;
+      _inputNode.value = 'hello world';
+      _inputNode.selectionStart = 2;
+      _inputNode.selectionEnd = 2;
+      el.value = 'hey there universe';
+      expect(_inputNode.selectionStart).to.equal(2);
+      expect(_inputNode.selectionEnd).to.equal(2);
+    });
+
+    it('move focus to a next focusable element after writing some text', async () => {
```
gerjanvangeest, line 48: This test also succeeds without the fix in all browsers, not yet sure how to solve it.

packages/form-core/src/NativeTextFieldMixin.js
```diff
@@ -90,27 +96,33 @@ const NativeTextFieldMixinImplementation = superclass =>
...
+        // Only preserve caret if focused (changing selectionStart will move focus in Safari)
+        if (this.focused) {
+          // Not all elements might have selection, and even if they have the
+          // right properties, accessing them might throw an exception (like for
+          // <input type=number>)
+          try {
+            // SelectElement doesn't have selectionStart/selectionEnd
+            if (
+              !(this._inputNode instanceof HTMLSelectElement) &&
+              this._inputNode.value !== newValue
```
tlouisse, line 110: I think the check abocve (line 100) already covered this?

packages/form-core/src/NativeTextFieldMixin.js
```diff
@@ -90,27 +96,33 @@ const NativeTextFieldMixinImplementation = superclass =>
...
-      if (this.focused) {
-        // Not all elements might have selection, and even if they have the
-        // right properties, accessing them might throw an exception (like for
-        // <input type=number>)
-        try {
-          // SelectElement doesn't have selectionStart/selectionEnd
-          if (!(this._inputNode instanceof HTMLSelectElement)) {
-            const start = this._inputNode.selectionStart;
+      // Only set if newValue is new, fix for Safari bug: https://github.com/ing-bank/lion/issues/1415
+      if (this._inputNode.value !== newValue) {
```
tlouisse, line 100: This could be combined into one if statement

tlouisse, line 100: Maybe move to the top of the function to make the nexted if structure more readable:
```js
if (this._inputNode.value === newValue) { 
  return;
}
```

packages/form-core/test-helpers/getFormControlMembers.js
```diff
@@ -12,7 +12,7 @@ export function getFormControlMembers(el) {
   // eslint-disable-next-line
   const { _inputNode, _helpTextNode, _labelNode, _feedbackNode, _allValidators } = el;
   return {
-    _inputNode,
+    _inputNode: /** @type {* & HTMLInputElement} */ (el)._inputNode,
```
tlouisse, line 15: I think it should be `/** @type {* & HTMLInputElement} */ (el._inputNode)`?

packages/textarea/test/lion-textarea-integrations.test.js
```diff
@@ -1,11 +1,34 @@
...
+import { LionTextarea } from '../src/LionTextarea.js';
 
-const tagString = 'lion-textarea';
+const fieldTagString = defineCE(
+  class extends LionTextarea {
+    get slots() {
+      return {
+        ...super.slots,
+        // LionTextarea needs to have an _inputNode defined in order to work...
+        input: () => document.createElement('textarea'),
```
tlouisse, line 16: this is already covered by LionTextArea, or not?

packages/textarea/test/lion-textarea-integrations.test.js
```diff
@@ -1,11 +1,34 @@
-import { runFormatMixinSuite } from '@lion/form-core/test-suites';
+import { defineCE } from '@open-wc/testing';
+import {
+  runInteractionStateMixinSuite,
+  runFormatMixinSuite,
+  runNativeTextFieldMixinSuite,
+} from '@lion/form-core/test-suites';
 
-import '@lion/textarea/define';
+import { LionTextarea } from '../src/LionTextarea.js';
```
tlouisse, line 8: Is there a reason to not use textarea/define anymore for this test?

packages/form-core/test-helpers/getFormControlMembers.js
gerjanvangeest: Nope, the type was for the `el` element. But i changed it into `FormControlHost`

packages/textarea/test/lion-textarea-integrations.test.js
gerjanvangeest: It didn't work before, but now it works again. Thanks for letting me double check it :)

packages/form-core/src/NativeTextFieldMixin.js
gerjanvangeest: The return made lots of tests of the `textarea` fail, i now moved the if statement one function up.

packages/form-core/test-suites/NativeTextFieldMixin.suite.js
```diff
@@ -0,0 +1,63 @@
...
+      // TODO: this test did not fail before the fix for Safari bug: https://github.com/ing-bank/lion/issues/1415, so its not a good test.
+      const el = /** @type {NativeTextFieldClass} */ (await fixture(html`<${tag}></${tag}>`));
+      const { _inputNode } = getFormControlMembers(el);
+      await triggerFocusFor(el);
+      await el.updateComplete;
+      expect(document.activeElement).to.equal(_inputNode);
+      await sendKeys({
+        press: 'h',
+      });
+      await sendKeys({
```
tlouisse, line 57: Maybe we can enable this test and instead try to check for a spy on `_setValueAndPreserveCaret` (that should not be called after the Tab)

packages/form-core/test-suites/NativeTextFieldMixin.suite.js
tlouisse: Maybe we can enable this test and instead try to check for a spy on _setValueAndPreserveCaret (that should not be called after the Tab)

packages/form-core/test-suites/NativeTextFieldMixin.suite.js
```diff
@@ -0,0 +1,69 @@
...
+        press: 'Tab',
+      });
+      expect(document.activeElement).to.not.equal(_inputNode);
+      expect(setValueAndPreserveCaretSpy.calledOnce).to.be.false;
```
tlouisse, line 66: Hmm, it should have callCount of 1 after the h key and after tab it should still have that callCount, right?
If calledOnce is true at the end, it seems to be called twice somehow?

gerjanvangeest: Yes i thought so too, but it didn't get called...

tlouisse (approved):

PR #1455, opened 2021-07-13
fix(input-amount): convert serializedValue to corrct amount of decimals

gerjanvangeest:
## What I did

1.


packages/input-amount/src/LionInputAmount.js
```diff
@@ -90,6 +91,20 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
     }
   }
 
+  /**
+   * Converts `.modelValue` to `.serializedValue`, which needs to be the same amount of decimals as the formattedValue
+   * @param {number} modelValue
+   * @returns {any} formattedValue
```
tlouisse, line 97: serializedValues shoild always return a string

packages/input-amount/src/LionInputAmount.js
```diff
@@ -90,6 +91,20 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
...
+   * @override FormatMixin
+   */
+  serializer(modelValue) {
+    if (modelValue === undefined) {
+      return '';
+    }
+    const formatted = formatAmount(modelValue, this.formatOptions);
+    return parseAmount(formatted) || '';
```
tlouisse, line 105: parseAmount returns `{number}` , so please cast to string.

packages/input-amount/src/LionInputAmount.js
gerjanvangeest: So add a `toString()` ?

packages/input-amount/test/lion-input-amount.test.js
```diff
@@ -119,6 +119,29 @@ describe('<lion-input-amount>', () => {
     ).to.equal('TL');
   });
 
+  it('sets serializedValue in correct currency decimals', async () => {
```
tlouisse, line 122: Maybe formulate it as 'the serializedValue should reflect what the end user sees on his screen (so serializedValue should be aligned with viewValue)'. 

Also, the goal of serializedValue is to be a serialized version of the modelValue. That means serializedValue <=> modelValue...

If we want to achieve that as well, it might be better to cut the number of decimals already in the parsing phase (serialize function could probably stay as was before, then).

Sorry for the confusion, I thought serializer function above was going to solve it, forgetting modelValue would be out of sync now.... :)

packages/input-amount/src/LionInputAmount.js
```diff
@@ -90,6 +91,20 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
...
+  /**
+   * Converts `.modelValue` to `.serializedValue`, which needs to be the same amount of decimals as the formattedValue
+   * @param {number} modelValue
+   * @override FormatMixin
+   */
+  serializer(modelValue) {
+    if (modelValue === undefined) {
+      return '';
+    }
+    const formatted = formatAmount(modelValue, this.formatOptions);
```
tlouisse, line 103: By doing it like this, extension that override the formatter may get out of sync... but also see remarks below. We might need to solve it on modelValue level and make the serializedValue benefit from that

packages/input-amount/test/lion-input-amount.test.js
gerjanvangeest: No problem, good to keep ourselves sharp

tlouisse: So that would probably mean adding `decimals`to options here: https://github.com/ing-bank/lion/blob/master/packages/input-amount/src/parsers.js#L124

As a final step in the method, we can cut the number of decimals when provided as option.
To prevent rounding errors, we could probably use Intl (like we do in the formatter) to get the right decimals.

packages/input-amount/test/parsers.test.js
```diff
@@ -1,165 +1,198 @@
...
+    });
+
+    it('uses locale to parse amount if there is only one separator e.g. 1.234', () => {
+      localize.locale = 'en-GB';
+      expect(parseNumber('12.34')).to.equal(12.34);
+      expect(parseNumber('12,34')).to.equal(1234);
+      expect(parseNumber('1.234')).to.equal(1.234);
+      expect(parseNumber('1,234')).to.equal(1234);
+
+      localize.locale = 'nl-NL';
```
tlouisse, line 100: hmm, is this related to just locales or specifically to currencies (amounts)?

packages/input-amount/index.js
```diff
@@ -1,4 +1,4 @@
 export { LionInputAmount } from './src/LionInputAmount.js';
 export { formatAmount } from './src/formatters.js';
-export { parseAmount } from './src/parsers.js';
+export { parseAmount, parseNumber } from './src/parsers.js';
```
tlouisse, line 3: If we want to align formatNumber/Amount and parseNumber/Amount fully, I think we should put parseNumber inside `localize`.

We can also do it at a later point, but then parseNumber should be private (not exported) inside `input-amount` package

packages/input-amount/src/parsers.js
```diff
@@ -148,3 +165,36 @@ export function parseAmount(value, options) {
...
+    return number;
+  }
+
+  /** @type {FormatOptions} */
+  const options = {
+    ...givenOptions,
+  };
+
+  if (options.currency && typeof options.minimumFractionDigits === 'undefined') {
+    options.minimumFractionDigits = getFractionDigits(options.currency);
```
tlouisse, line 194: apparently we don't use this below?

tlouisse (approved):
Nice :)

Issue #1445, opened 2021-07-06
 Could not resolve import "@lion/accordion/define" 


sammitom1983:
npm i --save @lion/core
import { dedupeMixin, LitElement } from '@mdjs/mdjs-preview';

I am tring to install '@mdjs/mdjs-preview', and it does't work.

![image](https://user-images.githubusercontent.com/55822232/124572590-5565a300-de7b-11eb-8627-3c2228f5f163.png)

daKmoR:
what are you trying to do?

if you create a component that works well with lion you would need to do 
```js
import { dedupeMixin, LitElement } from '@lion/core';
```

you only need to use `@mdjs/mdjs-preview` if you want to display demos on the documentation page

sammitom1983:
Thanks for your reply,  yes, I did this before, and a few days ago, I notice the online document has been changed..
https://lion-web.netlify.app/guides/principles/scoped-elements/ .

gerjanvangeest:
I believe our regex to change it into `@mdjs/mdjs-preview` was a bit too aggressive, it shouldn't be updated in the examples, only on real imports.

Reverted the change in: https://github.com/ing-bank/lion/pull/1457

PR #1457, opened 2021-07-14
chore: fix documentation of how to import core elements

gerjanvangeest:
## What I did

In https://github.com/ing-bank/lion/issues/1445 it was mentioned that our documentation was changed.
This PR is to revert that in the cases we document it.


tlouisse (approved):

daKmoR:
good catch 👍

PR #1460, opened 2021-07-19
feat(localize): await loadingComplete before sending locale changed

jorenbroekema:
## What I did

See changeset, it explains it in more detail. Hopefully the test I added also shows why it is more user-friendly this way. Mostly, `localize.msg()` calls may fail when called directly after locale switches where the new locale is still loading. now this is fixed :)

### DIFF
--- a/.changeset/unlucky-cameras-joke.md
+++ b/.changeset/unlucky-cameras-joke.md
@@ -1,0 +1,5 @@
+---
+'@lion/localize': minor
+---
+
+BREAKING: Fires localeChanged event (and as a result, invokes onLocaleChanged / onLocaleUpdated) after localize loading has completed. This means if the user switches the locale to a locale which has not loaded yet, it will load it first before sending the event. This will allow users to immediately call localize.msg and get the right output, without having to await localize.loadingComplete themselves. This is slightly breaking with regards to timing and might break tests in extensions. In that case, you probably need a `await localize.loadingComplete` statement in front of the failing assertion.
--- a/packages/calendar/test/lion-calendar.test.js
+++ b/packages/calendar/test/lion-calendar.test.js
@@ -1393,6 +1393,7 @@ describe('<lion-calendar>', () => {
       ]);
 
       localize.locale = 'nl-NL';
+      await localize.loadingComplete;
       await el.updateComplete;
       expect(elObj.nextMonthButtonEl?.getAttribute('aria-label')).to.equal(
         'Volgende maand, december 2019',
--- a/packages/calendar/test/lion-calendar.test.js
+++ b/packages/calendar/test/lion-calendar.test.js
@@ -1393,7 +1393,7 @@ describe('<lion-calendar>', () => {
       ]);
 
       localize.locale = 'nl-NL';
-      await localize.loadingComplete;
+      await el.localizeNamespacesLoaded;
       await el.updateComplete;
       expect(elObj.nextMonthButtonEl?.getAttribute('aria-label')).to.equal(
         'Volgende maand, december 2019',
--- a/packages/localize/src/LocalizeManager.js
+++ b/packages/localize/src/LocalizeManager.js
@@ -490,6 +490,8 @@ export class LocalizeManager {
    * @protected
    */
   _onLocaleChanged(newLocale, oldLocale) {
+    // Event firing immediately, does not wait for loading the translations
+    this.dispatchEvent(new CustomEvent('__localeChanging'));
     if (newLocale === oldLocale) {
       return;
     }
--- a/packages/localize/src/LocalizeManager.js
+++ b/packages/localize/src/LocalizeManager.js
@@ -41,7 +41,7 @@ export class LocalizeManager {
     this.__namespaceLoadersCache = {};
 
     /**
-     * @type {Object.<string, Object.<string, Promise.<Object>>>}
+     * @type {Object.<string, Object.<string, Promise.<Object|void>>>}
      * @private
      */
     this.__namespaceLoaderPromisesCache = {};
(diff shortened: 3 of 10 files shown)


packages/localize/src/LocalizeManager.js
```diff
@@ -495,32 +495,30 @@ export class LocalizeManager {
...
-    const promises = [];
     Object.keys(oldLocaleNamespaces).forEach(namespace => {
       const newNamespaceData = newLocaleNamespaces[namespace];
       if (!newNamespaceData) {
-        promises.push(
-          this.loadNamespace(namespace, {
-            locale: newLocale,
-          }),
-        );
+        this.loadNamespace(namespace, {
```
jorenbroekema, line 517: calling loadNamespace will ensure the promise is added to the __namespaceLoaderPromisesCache which enables the user to then await the loaders through `localize.loadingComplete`. Therefore, it is not necessary here to store those promises separately and return them, loadingComplete should be used for this instead as it uses the central promises cache

packages/calendar/test/lion-calendar.test.js
```diff
@@ -1393,6 +1393,8 @@ describe('<lion-calendar>', () => {
       ]);
 
       localize.locale = 'nl-NL';
+      // await localize.loadingComplete;
```
tlouisse, line 1396: why would  `localize.loadingComplete` not work here?

.changeset/unlucky-cameras-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/localize': minor
+---
+
+BREAKING: Fires localeChanged event (and as a result, invokes onLocaleChanged / onLocaleUpdated) after localize loading has completed. This means if the user switches the locale to a locale which has not loaded yet, it will load it first before sending the event. This will allow users to immediately call localize.msg and get the right output, without having to await localize.loadingComplete themselves. This is slightly breaking with regards to timing and might break tests in extensions. In that case, you probably need a `await localize.loadingComplete` statement in front of the failing assertion.
```
tlouisse, line 5: I have a hard time estimating what the breaking impact would be for people consuming LocalizeMixin.

Is there a difference between awaiting `el.localizeNamespacesLoaded` and `localize.loadingComplete`? They all return the same promise in the end, right?
(so for people that used `el.localizeNamespacesLoaded` before, nothing should change). 

Only people that would rely on synchronous (usually mocked) namespaceCache, would need to await an extra promise, now?

jorenbroekema: Yep pretty much. And yeah I do believe `await localize.loadingComplete` or `await el.localizeNamespacesLoaded` do pretty much the same thing, one for the manager and one on the element and whatever namespaces are local to it.

packages/calendar/test/lion-calendar.test.js
jorenbroekema: Ah right, that was a remnant of me trying to check for all failing tests whether they are timing related. Updated it.

jorenbroekema:
I added another commit where I fixed a few things and await el.localizeNamespacesLoaded instead of localize.loadingComplete which is the more "global" version, and usually a bit less performant if you only care about a specific custom element's localize namespaces.

I ran into one issue though, as you can see in the tests I had to add `await nextFrame()` in some instances. This is because the order of execution is something along the lines of:

1. `element.localizeNamespacesLoaded` resolves
2. `localize.loadingComplete` is awaited for missing translations when the locale was switched, resolves slightly later apparently
3. `localeChanged` event is fired as a result

So a nextFrame needs to be awaited otherwise you are slightly earlier than the `localeChanged` event is fired. Not sure if we can improve that one, `localize.locale` is set on a global level so it makes sense that `localize.loadingComplete` is awaited before the event is dispatched, but it's not really clear to me why that promise takes longer than `element.localizeNamespacesLoaded` promise

jorenbroekema:
Should be fixed. We were setting the `localizeNamespacesLoaded promise` on locale changed. With my change, locale changed fires only **after** missing translations are done, so if you're awaiting `localizeNamespacesLoaded` in that case, it will resolve immediately because it had not yet been initialized to await missing translations. 

Fixed it with an event `__localeChanging` that the `LocalizeMixin` now listens to instead for initializing the `localizeNamespacesLoaded` promise, and as expected, I could then remove the `await nextFrame()` in the tests, and this promise is nicely in sync with the `localize.loadingComplete` promise

jorenbroekema:
Added another commit which reverts the change in https://github.com/ing-bank/lion/pull/1460

Now that localeChanged is sent **after** localize loading complete, it is no longer necessary to await this in the ValidateMixin, which removes at least part of the localize dependency in validate 🎉

packages/localize/src/LocalizeMixin.js
```diff
@@ -38,6 +37,12 @@ const LocalizeMixinImplementation = superclass =>
           this.__localizeOnLocaleChanged(event);
         };
 
+      this.__boundLocalizeOnLocaleChanging =
```
tlouisse, line 40: Is there a difference with `this.__boundLocalizeOnLocaleChanging = this.__localizeOnLocaleChanging.bind(this)`?
(or `this.__localizeOnLocaleChanging = this.__localizeOnLocaleChanging.bind(this)` to reduce complexity (and already high) mental load)?

.changeset/unlucky-cameras-joke.md
tlouisse: Cool. I think the scope is relatively limited then and we can consider it a breaking bugfix 👍

tlouisse (approved):

packages/localize/src/LocalizeMixin.js
jorenbroekema: Probably identical, I just copied from the bound locale changed method above 😅 so it's at least legible in the sense that they do the same thing. I suppose I could have changed them both in retrospect 🤷

PR #1461, opened 2021-07-19
feat(form-core): add overridable filter method for form-group children

jorenbroekema:


packages/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -207,11 +207,34 @@ const ChoiceGroupMixinImplementation = superclass =>
...
+     * TODO: Consider making this explicit by extracting this method to its own mixin and
+     * using it in both FormGroupMixin and ChoiceGroupMixin, then override it here
+     * This also makes it more DRY as we have same method with similar implementation
+     * in FormGroupMixin. I (@jorenbroekema) think the abstraction is worth it here..
+     *
      * @param {string} property
      * @protected
      */
-    _getFromAllFormElements(property, filterCondition = () => true) {
+    _getFromAllFormElements(property) {
```
tlouisse, line 237: Maybe `_getFromAllFormElements(property, filterCondition = this. _getFromAllFormElementsFilter) {` to keep it backw. compatible?

jorenbroekema: Not sure how it would make it backward compatible? Compatible with what exactly?

Also, are we concerned with backwards compatibility for lion, which is v0? As far as I know this should not be problematic for the ING lion extension

tlouisse: Basically with all Subclassers that use this protected api (calling this method directly, regardless of what is configured on prototype level(could still be a use case)). Problem is that we don't have a good overview of all Subclassers and what they do. 
But I think breaking changes should be avoided as much as possible, unless unavoidable: lit major version, shadow parts (even these could be implemented in a backwards compatible way).

jorenbroekema: Hmm I guess it could be possible that a subclasser relies on this method and relies on that default parameter "filterCondition"

I definitely disagree that we should avoid breaking changes as minor as this one. Lion should be kept clean, no unnecessary backwards compatibility please. This is a breaking change that can be easily alleviated on the subclasser's part, we just need to release it as a minor and mark it as breaking in the changelog, and the subclasser can deal with it.

daKmoR: I'm also for keeping it clean... if an extension layer really needs it they can override it in that extension layer

packages/form-core/src/form-group/FormGroupMixin.js
```diff
@@ -317,21 +317,39 @@ const FormGroupMixinImplementation = superclass =>
...
         if (elem instanceof FormControlsCollection) {
-          result[name] = elem.filter(el => filterFn(el)).map(el => el[property]);
-        } else if (filterFn(elem)) {
+          result[name] = elem
+            .filter(el => this._getFromAllFormElementsFilter(el, property))
+            .map(el => el[property]);
+        } else if (this._getFromAllFormElementsFilter(elem, property)) {
           if (typeof elem._getFromAllFormElements === 'function') {
-            result[name] = elem._getFromAllFormElements(property, filterFn);
+            result[name] = elem._getFromAllFormElements(property);
```
tlouisse, line 352: This kind of assumes that the filter is always there for one single purpose (serialization). We would loose the ability to filtern (on form/fieldset level) to filter for readonly for instance.

We will look for a way to keep that option, while still providing a default serialize condition for every FormControl.

Maybe it's also an idea to rename `_getFromAllFormElementsFilter()` to `_serializeCondition()` ?

jorenbroekema: I disagree. You are passing the property to the filter function, and inside the filter function you can do different things for different properties. So it could be serializeCondition, it could be modelValueCondition, initialModelValueCondition, etc. etc. and in the filter function you can specify what you want to filter for on the element, e.g. disabled, readonly, checked, etc.

The test should make that a bit clearer

tlouisse: I think the problem is that on fieldset or form level you, generecially speaking, don't know anyhting about your children FormControls (they could be added dynamically by 3drd parties that consume your form). You should still be able to get all readonly (or non readonly or any other criterium) elements of all your children and their children etc. without being dependent on implementation details (the condition in `_getFromAllFormElementsFilter` in this case) you don't know about.

jorenbroekema: I don't understand what this PR changes about any of that. The only thing it does is pull out the filter condition function into a protected method which makes it easier to override without having to override the entire `_getFromAllFormElements` method.

tlouisse: I mean that if the form below (that wouldn't know anything about the implementation of `_getFromAllFormElementsFilter` calls `_getFromAllFormElements('modelValue')`, it gets a kind of unexpected result.
I think the default condition could highly depend on the first argument (modelValue in example above, but could be any prop). If a FormControl wants to totally be independent of its parent fieldset/form, it should probably not provide a name.

Also, I want to be able to do this in my form: `_getFromAllFormElements('modelValue', (el) => !el.getAttribute('field-type') === 'helper')`


(since the public attr field-type is one I provided myself.)

```html
<lion-form id="myForm">

  <my-field></my-field>
  <!-- has . _getFromAllFormElementsFilter="${(el) => !el.disabled}" -->
  
  <my-other-field field-type="helper""></my-other-field>
    <!-- has . _getFromAllFormElementsFilter="${(el) => !el.readonly}" -->


</lion-form>
```

jorenbroekema: > Also, I want to be able to do this in my form: `_getFromAllFormElements('modelValue', (el) => !el.getAttribute('field-type') === 'helper')`

Yeah that makes sense, we can keep the 2nd argument filter condition function that takes precedence over the protected filter method? Would that suffice?

So you can do

```js
_getFromAllFormElements('modelValue'); // uses this._getFromAllFormElementsFilter();
_getFromAllFormElements('modelValue', (el) => !el.getAttribute('field-type') === 'helper'); // uses the 2nd arg as the filter function
```

jorenbroekema: Pushed the change in a separate commit @tlouisse . Let me know if it's alright, then I can rebase / squash commits

tlouisse (approved):

PR #1462, opened 2021-07-20
feat(input-iban): allow array for IsCountryIBAN validator

jorenbroekema:
Accepting an array for IsNotCountryIBAN was a contribution accepted a while ago. We never got to implementing the same feature for the reverse validator, and I ran into a use case where I actually wanted to show how to enforce IBANs to only BeNeLux countries which might be a common use case, and wasn't able to do it without creating my own validator.

So I implemented it :)

### DIFF
--- a/.changeset/chatty-clocks-sleep.md
+++ b/.changeset/chatty-clocks-sleep.md
@@ -1,0 +1,5 @@
+---
+'@lion/input-iban': patch
+---
+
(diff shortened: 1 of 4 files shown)


daKmoR:
sweet - mind adding docs as well?

gerjanvangeest (approved):

PR #1464, opened 2021-07-20
fix(validate-messages): correct description of MaxNumber, MinNumber, …

### LINKED ISSUE #1428: MaxDate error message is not clear if the boundary date can be filled or not
If one is using component "lion-input-date" and add validation for MaxDate.`.validators=${[new MaxDate(new Date())]`.
The error message shown is "Please enter a(n) MaxDate before 06/17/2021". It is not clear to user if 17th June is a valid date or not.
The issue can be seen at the below link "https://lion-web.netlify.app/docs/systems/form/validate/"

varun-aggarwal:
…MinDate and MaxDate

fix #1428

## What I did

change info and error messages for MaxNumber, MinNumber, MinDate and MaxDate


daKmoR:
hey, adding a negation into an error message seems a harder to understand message?

why would you like to change it?

varun-aggarwal:
Hey @daKmoR , could you check the issue #1428?

tlouisse:
We could consider formulating it like'higher than or equal to', because (we think) it would better meet this criterion: https://www.w3.org/WAI/WCAG22/Understanding/reading-level.html

varun-aggarwal:
I agree that the negation is not the best way forward, I also want to consider the space this error message will take.

I prefer to have the old error message but with date changed by validator to "T+1" in case of max date or "T-1" in case of min date.

What will be your preference?

tlouisse:
For min/maxDate it makes sense I think (when unit is one day). 

For an amount of max=60000, it wouldn't make sense: 'smaller than 60.000,01'  (unit is one cent?)

varun-aggarwal:
I have changed the messages.

packages/validate-messages/translations/en.js
```diff
@@ -8,12 +8,12 @@ export default {
...
-    MinNumber: 'Please enter a(n) {fieldName} higher than {params}.',
-    MaxNumber: 'Please enter a(n) {fieldName} lower than {params}.',
+    MinNumber: 'Please enter a(n) {fieldName} higher than or equal to {params}.',
+    MaxNumber: 'Please enter a(n) {fieldName} lower than or equal to {params}.',
     MinMaxNumber: 'Please enter a(n) {fieldName} between {params.min} and {params.max}.',
     IsDate: 'Please enter a valid date (DD/MM/YYYY).',
-    MinDate: 'Please enter a(n) {fieldName} after {params, date, YYYYMMDD}.',
-    MaxDate: 'Please enter a(n) {fieldName} before {params, date, YYYYMMDD}.',
+    MinDate: 'Please enter a(n) {fieldName} higher than or equal to {params, date, YYYYMMDD}.',
+    MaxDate: 'Please enter a(n) {fieldName} lower than or equal to {params, date, YYYYMMDD}.',
```
jorenbroekema, line 16:
```suggestion
    MinDate: 'Please enter a(n) {fieldName} after or on {params, date, YYYYMMDD}.',
    MaxDate: 'Please enter a(n) {fieldName} before or on {params, date, YYYYMMDD}.',
```
Dates aren't really units that can go high or low, but rather they can go ahead or back in time, so I think this would make more sense.

varun-aggarwal: somehow `after or ON` is not making sense to me. could you share someplace where I can validate this?
```suggestion
    MinDate: 'Please enter a(n) {fieldName} after or same as {params, date, YYYYMMDD}.',
    MaxDate: 'Please enter a(n) {fieldName} before or same as {params, date, YYYYMMDD}.',
```

jorenbroekema: I think the latter suggestion is good, "enter a date equal to 21-12-2020" sounds better than "enter a date on 21-12-2020", "on" is indeed a bit weird here.

varun-aggarwal: I have done the necessary changes, can you review again?

jorenbroekema:
Looks good to me. Can you rebase and squash your commits into 1?

Also, in the same commit, add a changeset by running:
```sh
yarn changeset
```
selecting the package you changed, and selecting "patch", then add a description of the change in plain English so the changelog looks human readable :)

varun-aggarwal:
@jorenbroekema , commits squashed and added changeset

jorenbroekema (approved):

PR #1472, opened 2021-07-27
fix(input-amount): Reflect more currency changes

### LINKED ISSUE #1430: [lion-input-amount] Passing a currency value to lion-input-amount once input is initialized does not get reflected
## Expected behavior

When trying to pass a currency when lion input amount is initialized with an undefined value, the currency is shown, and alternatively, setting the currency to undefined when it is initialized with a value should remove the currency.

## Actual Behavior

I believe it is caused by only modifying the currency slot in the connectedCallback function

`  connectedCallback() {
    // eslint-disable-next-line wc/guard-super-call
    super.connectedCallback();
    this.type = 'text';
    this._inputNode.setAttribute('inputmode', 'decimal');

    if (this.currency) {
      this.__setCurrencyDisplayLabel();
    }
  }`

This can be observed in demo as well.
https://user-images.githubusercontent.com/8704542/123062274-f21d4f00-d43e-11eb-85f2-e3055855d2f8.mp4

## Additional context

Which component and version of it are you using.
lion-input-amount

enqueue:
Switching between defined / undefined currency yields better currency
label updates

Closes #1430

### DIFF
--- a/packages/input-amount/src/LionInputAmount.js
+++ b/packages/input-amount/src/LionInputAmount.js
@@ -110,34 +110,15 @@ export class LionInputAmount extends LocalizeMixin(LionInput) {
   _onCurrencyChanged({ currency }) {
     this.formatOptions.currency = currency || undefined;
     if (this.currency) {
-      if (!this._currencyDisplayNode) {
-        this._currencyDisplayNode = this._createCurrencyDisplayNode();
+      if (!this.__currencyDisplayNode) {
+        this.__currencyDisplayNode = this._createCurrencyDisplayNode();
       }
-      this._currencyDisplayNode.textContent = this.__currencyLabel;
+      this.__currencyDisplayNode.textContent = this.__currencyLabel;
       this._calculateValues({ source: null });
     } else {
-      this._currencyDisplayNode = undefined;
+      this.__currencyDisplayNode = undefined;
     }
     this.__setCurrencyDisplayLabel();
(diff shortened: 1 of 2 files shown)


jorenbroekema:
It should be noted that even though this PR adds the feature of removing a redundant currency node which I think is good, it also removes the coupling between this currency node and the after slot. Meaning, if someone puts in their own after slottable, it will not override the currency node, because they aren't the same node anymore. Since this after slot is public API there should have been a test for it, but there isn't. Correct me if I'm wrong on this btw!

Would it be perhaps possible to add this feature in a way where the currency node will still utilize the after slot, unless of course the component-user puts in their own "after" slottable? At least, I think the intention is that the component-user can do this to override the currencyNode being populated in the "after" slot through public API, rather than having to jump into the subclasser-role and use private API to do that.

enqueue:
@jorenbroekema  thank you for your review comment. I am not sure I fully understand the issue, though. The element is still coupled to the `after` node, as you can see from the getter `get __currencyDisplayNode()`. I added a test to show that a component user may still supply her own element for the currency label. However, this element will be lost if the user changes from "some currency" to "no currency" (which is intended as per issue comments), and it will *not* be restored if  the user changes again to some valid currency subsequently. Should we remember this node privately and restore it in this scenario?

enqueue:
Superseded by #1477

PR #1477, opened 2021-08-10
Amount reset improvements

jorenbroekema:
Hey @enqueue 

I was reviewing your PR and I wanted to refactor some of it, also to include the improvement of reusing the after slottable by storing it, removing/restoring when needed. 

However, when the user passes their own slottable, it is safe to assume it is unrelated to the currency. 
To change the after slottable markup but still use the after slot as a currency display, you would use protected API (subclassers) and probably override the slots getter in your lion-input-amount extension class. 
Therefore it is safe to assume we should keep custom passed slottable intact even if the currency becomes invalid. I added a test for that.

Instead of lots of comments I figured it would be simpler to refactor it and send a PR to your branch :) so here it is.


enqueue:
@jorenbroekema  thank you! Yes, it is probably more "polite" to remember and restore the node when toggling between defined and undefined currencies. Sorry if my PR has caused you more work than a direct fix by you or other, more experienced contributors.

jorenbroekema:
Oh no, no worries at all, sometimes it just happens that a fix seems pretty trivial on the surface, but then when you dive in, it turns out to be tricky with multiple possible approaches, this was one of those :) thanks for starting the initiative to fix the problem!

gerjanvangeest (approved):
LGTM @tlouisse any objections?

PR #1484, opened 2021-08-19
feat(switch): make key handlers protected, enable arrow keys

jorenbroekema:
## What I did

1. Make key handler methods protected. People may want to disable some of the keys, maybe add their own keys, etc.
2. By default also allow arrow keys to toggle the switch (but now overridable by subclassers), I think this is a better default especially given the visual cue of a switch which goes from one direction to another, therefore arrow keys make sense as interaction handlers for that. Often, ArrowUp gets paired with ArrowRight (on) in this case, and ArrowDown with ArrowLeft (off)

### DIFF
--- a/.changeset/perfect-bags-compare.md
+++ b/.changeset/perfect-bags-compare.md
@@ -1,0 +1,5 @@
+---
+'@lion/switch': minor
+---
+
+Make keyup handlers protected, enables subclassers to switch using alternative keys.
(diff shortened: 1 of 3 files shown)


tlouisse (approved):

.changeset/perfect-bags-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/switch': minor
+---
+
+BREAKING: Make keyup handlers protected, enables subclassers to switch using alternative keys.
```
tlouisse, line 5: It should actually not be considered breaking, right?

jorenbroekema: Ah true, I think this was still from adding the arrow key behaviors, changed

tlouisse (approved):

tlouisse (approved):

PR #1489, opened 2021-08-21
Update Lit to v2

abdonrd:
## What I did

1. Update Lit to v2.0.0

### DIFF
--- a/.changeset/twelve-zoos-turn.md
+++ b/.changeset/twelve-zoos-turn.md
@@ -1,5 +1,5 @@
 ---
-"providence-analytics": patch
+'providence-analytics': patch
 ---
 
 Correct default target in Windows if no -t
--- a/packages/input/test/lion-input.test.js
+++ b/packages/input/test/lion-input.test.js
@@ -1,8 +1,7 @@
 import { Validator } from '@lion/form-core';
-import { expect, fixture, triggerFocusFor, aTimeout } from '@open-wc/testing';
-import { html, unsafeStatic } from 'lit/static-html.js';
-import { getInputMembers } from '../test-helpers/index.js';
 import '@lion/input/define';
+import { aTimeout, expect, fixture, html, triggerFocusFor, unsafeStatic } from '@open-wc/testing';
+import { getInputMembers } from '../test-helpers/index.js';
 
 /**
  * @typedef {import('../src/LionInput').LionInput} LionInput
--- a/packages/overlays/test-suites/OverlayMixin.suite.js
+++ b/packages/overlays/test-suites/OverlayMixin.suite.js
@@ -1,4 +1,5 @@
 import { expect, fixture, html, nextFrame, aTimeout } from '@open-wc/testing';
+
 import sinon from 'sinon';
 import { overlays } from '../src/overlays.js';
 import { OverlayController } from '../src/OverlayController.js';
--- a/packages/select-rich/src/LionSelectInvoker.js
+++ b/packages/select-rich/src/LionSelectInvoker.js
@@ -13,7 +13,8 @@ import { css, html, SlotMixin } from '@lion/core';
 export class LionSelectInvoker extends SlotMixin(LionButton) {
   static get styles() {
     return [
-      ...super.styles,
+      // TODO switch back to ...super.styles once fixed https://github.com/lit/lit.dev/pull/535
+      ...LionButton.styles,
       css`
         :host {
           justify-content: space-between;
--- a/.changeset/fair-mails-smell.md
+++ b/.changeset/fair-mails-smell.md
@@ -1,0 +1,5 @@
+---
+'@lion/core': minor
+---
+
+Update Lit to v2
--- a/.changeset/few-books-provide.md
+++ b/.changeset/few-books-provide.md
@@ -1,0 +1,17 @@
+---
+'@lion/button': patch
+'@lion/combobox': patch
+'@lion/core': patch
+'@lion/dialog': patch
+'@lion/form': patch
+'@lion/form-core': patch
+'@lion/form-integrations': patch
+'@lion/input': patch
+'@lion/listbox': patch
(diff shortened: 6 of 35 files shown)


tlouisse:
Awesome  @abdonrd 👍 Would you like to add a changeset please as well? (`npm run changeset` and then a patch bump (assuming there's nothing breaking of course))

tlouisse:
Hmm, build fails atm... You can set this limit to 13 for now: https://github.com/ing-bank/lion/blob/master/package.json#L105

abdonrd:
> Hmm, build fails atm... You can set this limit to 13 for now: https://github.com/ing-bank/lion/blob/master/package.json#L105
Done!

> Could you also squash all commits into one?

Could you do it with the GitHub merge button?

Screenshot 2021-08-24 at 13 13 52

tlouisse:
> Could you do it with the GitHub merge button?

Yes, possible as well of course :)

Still two problems:
- https://github.com/ing-bank/lion/blob/master/package.json#L109 -> easy fix
- https://github.com/ing-bank/lion/pull/1489/checks?check_run_id=3416181654#step:8:7114 

The latter is something we most likely need to look at I think (although contributions are always welcome of course). 
Hopefully we will find some time for this soon.

abdonrd:
@tlouisse updated, squashed and rebased!

tlouisse:
Updated to v2.0.0-rc.4. Same errors still :(

abdonrd:
Probably related: https://github.com/lit/lit.dev/pull/535

.changeset/twelve-zoos-turn.md
```diff
@@ -1,5 +1,5 @@
 ---
-"providence-analytics": patch
+'providence-analytics': patch
```
tlouisse, line 2: finally 👍

packages/select-rich/src/LionSelectInvoker.js
```diff
@@ -13,7 +13,7 @@ import { css, html, SlotMixin } from '@lion/core';
 export class LionSelectInvoker extends SlotMixin(LionButton) {
   static get styles() {
     return [
-      ...super.styles,
+      ...LionButton.styles,
```
tlouisse, line 16: can't we use super here?

tlouisse (approved):

packages/select-rich/src/LionSelectInvoker.js
gerjanvangeest: Let me check

tlouisse: Thanks, Would be hard though when the parent is a Mixin. 
Does the transpilation issue also occur in non Typescript code (https://github.com/lit/lit.dev/pull/535#discussion_r719905974 mentions TS explicitly)?

tlouisse: Ok, apparently it triggers TS linting errors (which we get via jsdoc annotations), so should be good as a workaround 👍

packages/overlays/test-suites/OverlayMixin.suite.js
```diff
@@ -19,10 +27,21 @@ function getGlobalOverlayNodes() {
...
+    tagString: null,
+    suffix: '',
+    ...customConfig,
+  };
+
+  describe(`OverlayMixin${cfg.suffix}`, () => {
+    class TestOverlay extends OverlayMixin(LitElement) {}
+
+    cfg.tagString = cfg.tagString ? cfg.tagString : defineCE(TestOverlay);
+    const tag = unsafeStatic(cfg.tagString);
```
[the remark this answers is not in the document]
tlouisse: In other words: everywhere we supply both tagString and tag: https://github.com/ing-bank/lion/search?q=runOverlayMixinSuite

So I don't really get how adding `TestOverlay` would make a difference...?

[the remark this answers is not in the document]
gerjanvangeest: The tag somehow doesn't get applied and it shows `lit$316162849$` instead of `test-0`.

All other test-suites have backup class inside the test suite, so this way its also fully aligned.
I will remove the class creation in `./test/OverlayMixin.test.js`, since that is now double.

The slots are also in the test itself, so not needed in the render.

packages/select-rich/src/LionSelectInvoker.js
gerjanvangeest: I added a TODO to change it back once fixed.

packages/overlays/test-suites/OverlayMixin.suite.js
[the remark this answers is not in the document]
tlouisse: Ok, so the difference was explained by incompatible versions of unsafeStatic/staticHtml (from @open-wc and from lit)

gerjanvangeest:
@abdonrd Thanks for all the updates, helps us a lot :)

gerjanvangeest (approved):

packages/input/test/lion-input.test.js
```diff
@@ -1,8 +1,8 @@
 import { Validator } from '@lion/form-core';
-import { expect, fixture, triggerFocusFor, aTimeout } from '@open-wc/testing';
-import { html, unsafeStatic } from 'lit/static-html.js';
-import { getInputMembers } from '../test-helpers/index.js';
 import '@lion/input/define';
+import { aTimeout, expect, fixture, html, triggerFocusFor, unsafeStatic } from '@open-wc/testing';
+import {} from 'lit/static-html.js';
```
tlouisse, line 4: ?

tlouisse: Only this line needs to be removed. Apart from that it looks great.

Maybe also flatten the commits?
- Update to Lit v2.0.2 (on abdonrd's name)
- fix: use html & unsafeStatic from @open-wc/testing (on your name)

PR #1493, opened 2021-09-06
remove errant semicolon from tabs npm install step

thescientist13:
## What I did

1. Removed an errant `;` from the Tabs npm install instructions


thescientist13:
Let me know if you would like a changeset generated for this.  👍

gerjanvangeest (approved):

gerjanvangeest:
Thanks for the fix :)

PR #1496, opened 2021-09-13
feat(core): support default slot for SlotMixin

jorenbroekema:


packages/core/src/SlotMixin.js
```diff
@@ -49,7 +49,13 @@ const SlotMixinImplementation = superclass =>
     _connectSlotMixin() {
       if (!this.__isConnectedSlotMixin) {
         Object.keys(this.slots).forEach(slotName => {
-          if (!Array.from(this.children).find(el => el.slot === slotName)) {
+          // default slot means, we just check if there's child elements, otherwise check for slottable by slot-name
+          const hasSlottableFromUser =
+            slotName === ''
+              ? this.children.length > 0
```
tlouisse, line 55: If there's a different named slot already it would unrightfully pass I thunk

jorenbroekema: Ohh true

jorenbroekema: Amended, added a test as well which failed before the update due to the edge case you pointed out 👍

packages/core/src/SlotMixin.js
```diff
@@ -49,7 +49,15 @@ const SlotMixinImplementation = superclass =>
     _connectSlotMixin() {
       if (!this.__isConnectedSlotMixin) {
         Object.keys(this.slots).forEach(slotName => {
-          if (!Array.from(this.children).find(el => el.slot === slotName)) {
+          const hasSlottableFromUser =
+            slotName === ''
+              ? // for default slot (''), we can't use el.slot because polyfill for IE11
+                // will do .querySelector('[slot=]') which produces a fatal error
+                // therefore we check if there's children that do not have a slot attr
+                Array.from(this.children).find(el => !el.hasAttribute('slot'))
```
tlouisse, line 57: Following that reasoning, you should also check via getAttribute in the line below? :P

tlouisse: Ah nvm, didn't read it correctly

tlouisse (approved):

lpoulter:
I was experiencing this issue, I tested this change locally and it works as expected for me.

PR #1497, opened 2021-09-15
feat: Adds prefix and suffix slots to LionButton

lpoulter:
Enables subclassers to target part of the template rather than overriding the whole render function.

### DIFF
--- a/packages/button/src/LionButton.js
+++ b/packages/button/src/LionButton.js
@@ -35,7 +35,13 @@ export class LionButton extends DisabledWithTabIndexMixin(LitElement) {
   }
 
   render() {
-    return html` <div class="button-content" id="${this._buttonId}"><slot></slot></div> `;
+    return html`
+      <div class="button-content" id="${this._buttonId}">
+        <slot name="prefix"></slot>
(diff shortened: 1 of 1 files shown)


packages/button/src/LionButton.js
```diff
@@ -35,7 +35,13 @@ export class LionButton extends DisabledWithTabIndexMixin(LitElement) {
   }
 
   render() {
-    return html` <div class="button-content" id="${this._buttonId}"><slot></slot></div> `;
+    return html`
+      <div class="button-content" id="${this._buttonId}">
+        <slot name="prefix"></slot>
+        <slot></slot>
```
gerjanvangeest, line 41: Sorry for not responding for so long, we had a button story on our backlog, which made sense to wait upon since it touched this topic. But the story got postponed a few times.

How about making use of a template, which can easily be overridden to your taste?

```js
_buttonContentTemplate() {
  return html`<slot></slot>`;
}
    
render() {
    return html` <div class="button-content" id="${this._buttonId}">this._buttonContentTemplate()</div> `;
```

PR #1510, opened 2021-10-07
chore: remove typing inside import statement

tlouisse:
See: https://lit-and-friends.slack.com/archives/CJGFWJN9J/p1633609846134500


gerjanvangeest:
Should this be a chore or a fix?

PR #1519, opened 2021-10-28
Correctly parse exports like const x=3;export {x};

sposh:
## What I did

1. Based on conversations with @tlouisse


sposh:
Hi @tlouisse I signed the contributor licence but it still comes up as pending

tlouisse (approved):

tlouisse:
See https://github.com/ing-bank/lion/pull/1535

PR #1524, opened 2021-11-04
fix(overlays): fix condition in is-equal-config util and add null guard

Alex-Radu:
## How I stumbled upon the issue

I wanted to add a dynamic `elementToFocusAfterHide` to a dialog config object. The reason is that I have a dialog that can be opened from multiple places and I want to return focus to where it was opened from.  While the solution worked, I noticed `Cannot convert undefined or null to object` errors in the console, coming from **line 17** in the `is-equal-config` file. 

I saw that the conditional was checking the same thing twice and also that **null** values would go through and cause an error after the conditional. In my case, it was an **HTMLButtonElement** which has a `form` property that is read-only and can be **null**.

## What I did

1. Fixed the initial condition and added checks for **null**
2. Added a test for this scenario


tlouisse:
@Alex-Radu makes sense, get the null check now :) Could you add a changeset (`npm run changeset` (patch/fix)) and add that to the current commit?)

Alex-Radu:
Done. Hope the message in the changeset is ok 😊

tlouisse (approved):

PR #1526, opened 2021-11-05
Fix black bars on dialogs with preventsScroll enabled on ios 15

Joel-Levi:
## What I did

1. Found that the global-overlays-scroll-lock-ios-fix sets position: fixed on the body. 
2. Found out that iOS does not like that, and how to fix it : [Stackoverflow](https://stackoverflow.com/questions/69589924/ios-15-minimized-address-bar-issue-with-fixed-position-full-screen-content)
3. Fix it by applying height: 100vh; to the <html> root element in the same way the other iOS fix is applied.

### DIFF
--- a/.changeset/short-horses-run.md
+++ b/.changeset/short-horses-run.md
@@ -1,0 +1,6 @@
+---
+"@lion/overlays": patch
+---
+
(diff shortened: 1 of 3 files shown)


tlouisse (approved):

tlouisse:
Thanks @Joel-Levi 💪

PR #1530, opened 2021-11-08
fix: avoid use of innerHTML to please Checkmax

Knorgias:
## What I did

1. Removed use of innerHTML for LionCombobox to make Checkmarx happy (false positive finding)

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -2,6 +2,54 @@ import { html, css, browserDetection } from '@lion/core';
 import { OverlayMixin, withDropdownConfig } from '@lion/overlays';
 import { LionListbox } from '@lion/listbox';
 
+const matchReverseFns = new WeakMap();
+const matchA11ySpanReverseFns = new WeakMap();
+
+function checkForTextMatch(root, matchingString, option) {
+  Array.from(root.childNodes).forEach(childNode => {
+    if (childNode.nodeName === '#text') {
+      // check for match based on nodeValue
+
+      const re = new RegExp(`^(.*)(${matchingString})(.*)$`, 'i');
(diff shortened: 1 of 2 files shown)


packages/combobox/src/LionCombobox.js
```diff
@@ -2,6 +2,60 @@ import { html, css, browserDetection } from '@lion/core';
...
 
+const matchReverseFns = new WeakMap();
+const matchA11ySpanReverseFns = new WeakMap();
+
+/**
+ * @param {Node} root
+ * @param {string} matchingString
+ * @param {object} option
+ */
+function checkForTextMatch(root, matchingString, option) {
```
tlouisse, line 13: The name `checkForTextMatch` doesn't fully cover what it does anymore. Maybe call it `makeMatchingTextBold`
  
We can change it later into a util function that can be reused in contexts that need to traverse shadow dom (in which case root would have been option.shadowRoot).  For reference: see our demos.
Then we could call it checkForTextMatch again and allow a callback function (in case you want to do smth different than makling text bold.)

packages/combobox/src/LionCombobox.js
```diff
@@ -2,6 +2,60 @@ import { html, css, browserDetection } from '@lion/core';
 import { OverlayMixin, withDropdownConfig } from '@lion/overlays';
 import { LionListbox } from '@lion/listbox';
 
+const matchReverseFns = new WeakMap();
+const matchA11ySpanReverseFns = new WeakMap();
+
+/**
+ * @param {Node} root
+ * @param {string} matchingString
+ * @param {object} option
```
tlouisse, line 11: This is a Node as well

tlouisse (approved):

PR #1532, opened 2021-11-09
add lit1 as release branch

Knorgias:
For lit1 patch releases

## What I did

1. Added lit1 as release branch for lit1 patches


tlouisse:
doesn't need to be here, since it will not be triggered via lit1 when pushed to master

tlouisse:
Closing this one, since solved on lit1 branch

PR #1533, opened 2021-11-09
fix: Removed use of innerHTML from LionComboBox

Knorgias:
## What I did

1. Removed use of innerHTML from LionComboBox to make Checkmarx happy

### DIFF
--- a/packages/combobox/src/LionCombobox.js
+++ b/packages/combobox/src/LionCombobox.js
@@ -2,6 +2,60 @@ import { html, css, browserDetection } from '@lion/core';
 import { OverlayMixin, withDropdownConfig } from '@lion/overlays';
 import { LionListbox } from '@lion/listbox';
 
+const matchReverseFns = new WeakMap();
+const matchA11ySpanReverseFns = new WeakMap();
+
+/**
+ * @param {Node} root
+ * @param {string} matchingString
+ * @param {Node} option
+ */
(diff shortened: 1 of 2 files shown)


packages/combobox/src/LionCombobox.js
```diff
@@ -218,8 +272,10 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    * @protected
    */
   get _listboxNode() {
-    return /** @type {LionOptions} */ ((this._overlayCtrl && this._overlayCtrl.contentNode) ||
-      Array.from(this.children).find(child => child.slot === 'listbox'));
+    return /** @type {LionOptions} */ (
```
tlouisse, line 275: I think you should do a commit with the correct prettier (remove node_modules and do a fresh install based on lock file of that time)

packages/combobox/src/LionCombobox.js
```diff
@@ -284,6 +340,16 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
      * @private
      */
     this.__prevCboxValue = '';
+    /**
```
tlouisse, line 343: where do these lines come from ?

tlouisse: I think you should manually override the _onFilterMatch and _onFilterUnmatch methods (and add the methods and vars on top)

tlouisse (approved):

Issue #1539, opened 2021-11-12
[providence-analytics] tests incompatible with Node >=16


tlouisse:
Currently, it's needed to switch to Node < 16 to run the tests. Possibly related to `mock-fs` version...

Issue #1538, opened 2021-11-12
[providence-analytics] support export maps in match analyzers


tlouisse:
## Expected behavior

When I run `match-imports`, `match-paths` or `match-classes`, projects that use export maps should be supported.

## Actual Behavior

We used a custom resolve mechanism that only accounted for 'legacy' exports:
https://github.com/ing-bank/lion/blob/master/packages-node/providence-analytics/src/program/analyzers/helpers/from-import-to-export-perspective.js#L20

## Solution

Use `node-resolve` mechanism from rollup. It covers all edge cases, works in older node versions and it can easily be mocked with a virtual fs.

PR #1541, opened 2021-11-12
fix: support export maps for match-* analyzers

tlouisse:
## What I did

Use `node-resolve` mechanism from rollup.

closes https://github.com/ing-bank/lion/issues/1538
closes https://github.com/ing-bank/lion/issues/1539


packages-node/providence-analytics/test-node/program/analyzers/match-imports.test.js
```diff
@@ -253,58 +253,206 @@ describe('Analyzer "match-imports"', () => {
...
       });
     });
   });
 
   describe('Matching', () => {
     it(`produces a list of all matches, sorted by project`, async () => {
-      mockTargetAndReferenceProject(searchTargetProject, referenceProject);
-      await providence(matchImportsQueryConfig, _providenceCfg);
+      /**
+       * N.B. output structure could be simplified, since there is
```
gerjanvangeest, line 415: "Since there is ..."?

tlouisse: Changed to:

```
* N.B. output structure could be simplified, since there is no need to order results by
* target project (there's only one target project per run).
* For now we keep it, so integration with dashboard stays intact.
```

gerjanvangeest (approved):

PR #1557, opened 2021-11-22
feat(localize): add option to show the key when a locale is missing a translation

pndewit:
## What I did

Whenever a language is missing a translation for a key, it currently shows an empty string. I would like the option to show the translation key instead. This is not only making it visible for developers that some translation is missing, but could potentially even be descriptive enough for users to figure out what it should say (at least more than an empty string) so they can continue their work for the time being.

### DIFF
--- a/.changeset/fast-trees-prove.md
+++ b/.changeset/fast-trees-prove.md
@@ -1,0 +1,5 @@
+---
+'@lion/localize': minor
+---
+
+add option to show the key as a fallback when a locale is missing a translation
--- a/packages/localize/src/LocalizeManager.js
+++ b/packages/localize/src/LocalizeManager.js
@@ -14,13 +14,19 @@ import isLocalizeESModule from './isLocalizeESModule.js';
  */
 export class LocalizeManager {
   // eslint-disable-line no-unused-vars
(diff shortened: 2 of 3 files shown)


tlouisse:
@daKmoR @jorenbroekema @gerjanvangeest any opinions on this feature?

Since it's backwards compatible, it could be a helpful debug feature imho. 
If my colleagues think as well, could you:
- rename `fallbackTranslations` to `showKeyAsFallback`
- add tests
- sign license/cla

jorenbroekema:
Yep, I like this! Makes it easier for users to report problems, easier to continue using the component/site by figuring out what it's meant to say roughly, so definitely agree it's good for local dev & also prod

pndewit:
@tlouisse I've:
- added 2 tests
- changed it to a minor update (as it is a feature and not a fix IMHO)
- squashed all commits into 1

I've also signed the license agreement about a dozen times, also used the `recheck` link, but it doesn't seem to pick it up :(

tlouisse:
> @tlouisse I've:
> 
> * added 2 tests
> * renamed the flag as requested
> * changed it to a minor update (as it is a feature and not a fix IMHO)
> * squashed all commits into 1
> * signed the CLA

Feature makes sense indeed. Looks good, thanks!

tlouisse (approved):

PR #1572, opened 2021-12-06
chore: try execution of tests

daKmoR:
## What I did

1. checking why tests are still not executed in release task


daKmoR:
reopening the original PR made it work

PR #1573, opened 2021-12-06
fix(tabs): prevent selection of disabled tabs.

jrobind:
## What I did

1. Added a function to retrieve the next available index of a not-disabled tab. Depending on the keyboard direction, the next available not-disabled tab will be selected.  
3. Added tests to ensure disabled tabs are skipped with [arrow-right]/[arrow-down] and [arrow-left]/[arrow-up].

## Context
Currently, any disabled tabs within a tab list are still selectable. A user shouldn't be able to interact with a disabled tab – please see a minimal reproduction [here](https://codesandbox.io/s/lion-tabs-disabled-attr-issue-nwrzu?file=/src/index.js). 

Unfortunately,  adding an early return if the next tab element is disabled will not cover all edge cases (as originally mentioned in the Slack channel). We could have instances where there are disabled tabs in varying orders within the tab list.

## Visual

https://user-images.githubusercontent.com/30320536/144885840-cd804b7a-4ef7-4f02-b3ba-67c11a1c0ca7.mov

### DIFF
--- a/packages/tabs/src/LionTabs.js
+++ b/packages/tabs/src/LionTabs.js
@@ -164,6 +164,21 @@ export class LionTabs extends LitElement {
   firstUpdated(changedProperties) {
     super.firstUpdated(changedProperties);
     this.__setupSlots();
+    if (this.tabs[0].disabled) {
+      this.selectedIndex = this.tabs.findIndex(tab => !tab.disabled);
+    }
+  }
+
+  get tabs() {
+    return /** @type {HTMLButtonElement[]} */ (Array.from(this.children)).filter(
+      child => child.slot === 'tab',
+    );
+  }
+
+  get panels() {
+    return /** @type {HTMLElement[]} */ (Array.from(this.children)).filter(
+      child => child.slot === 'panel',
+    );
   }
 
   /** @private */
@@ -186,22 +201,16 @@ export class LionTabs extends LitElement {
   __setupStore() {
     /** @type {StoreEntry[]} */
     this.__store = [];
-    const buttons = /** @type {HTMLElement[]} */ (Array.from(this.children)).filter(
-      child => child.slot === 'tab',
-    );
-    const panels = /** @type {HTMLElement[]} */ (Array.from(this.children)).filter(
-      child => child.slot === 'panel',
-    );
-    if (buttons.length !== panels.length) {
+    if (this.tabs.length !== this.panels.length) {
       // eslint-disable-next-line no-console
       console.warn(
-        `The amount of tabs (${buttons.length}) doesn't match the amount of panels (${panels.length}).`,
+        `The amount of tabs (${this.tabs.length}) doesn't match the amount of panels (${this.panels.length}).`,
       );
     }
 
-    buttons.forEach((button, index) => {
+    this.tabs.forEach((button, index) => {
       const uid = uuid();
-      const panel = panels[index];
+      const panel = this.panels[index];
 
       /** @type {StoreEntry} */
       const entry = {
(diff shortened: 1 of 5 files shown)


gerjanvangeest:
Looks good, it only still breaks when the first tab is disabled, since the selectedIndex is standard set to the first tab.

packages/tabs/src/LionTabs.js
```diff
@@ -267,11 +268,11 @@ export class LionTabs extends LitElement {
       return newIndex;
     }
     if (direction === 'ArrowRight' || direction === 'ArrowDown') {
-      const nextNotDisabledTab = this.__getNextNotDisabledTab(tabs, currentTab);
+      const nextNotDisabledTab = this.__getNextNotDisabledTab(tabs, currentTab, 'right');
       return tabs.findIndex(tab => nextNotDisabledTab === tab);
     }
     if (direction === 'ArrowLeft' || direction === 'ArrowUp') {
-      const nextNotDisabledTab = this.__getNextNotDisabledTab([...tabs].reverse(), currentTab);
```
jorenbroekema: This trick doesn't work. If you have let's say 7 tabs, and you're on number 5, pressing arrow left would reverse that to be number 2 and focus would shift to 1. Fixed in my commit

packages/tabs/test/lion-tabs.test.js
```diff
@@ -15,6 +15,8 @@ const basicTabs = html`
     <div slot="panel">panel 2</div>
     <button slot="tab">tab 3</button>
     <div slot="panel">panel 3</div>
+    <button slot="tab">tab 4</button>
+    <div slot="panel">panel 4</div>
```
jorenbroekema, line 19: Added another tab, in case some of the test expects being adjusted to +1, this explains those diffs ;)

jorenbroekema:
@gerjanvangeest  nice catch, added a test + fix for when first tab is disabled.

packages/tabs/test/lion-tabs.test.js
```diff
@@ -254,6 +256,92 @@ describe('<lion-tabs>', () => {
...
+          </lion-tabs>
+        `)
+      );
+      el.selectedIndex = 2;
+      const tabs = el.querySelectorAll('[slot=tab]');
+      tabs[2].dispatchEvent(new KeyboardEvent('keyup', { key: 'ArrowLeft' }));
+      expect(el.selectedIndex).to.equal(0);
+    });
+
+    it('stays on current tab if previous or next tab is last tab in that direction and also disabled', async () => {
```
gerjanvangeest, line 317: If tab 0 is active and we press ArrowLeft we go to the last tab.
So why is the behavior different for disabled tabs?

packages/tabs/src/LionTabs.js
```diff
@@ -238,18 +238,19 @@ export class LionTabs extends LitElement {
...
+  __getNextNotDisabledTab(tabs, currentTab, dir) {
+    let nextNotDisabledTabs = /** @type {HTMLButtonElement[]} */ ([]);
+    if (dir === 'right') {
+      nextNotDisabledTabs = tabs.filter((tab, i) => !tab.disabled && i > this.selectedIndex);
+      return nextNotDisabledTabs[0];
     }
-    return notDisabledTabs.find((tab, i) => i >= this.selectedIndex && tab !== currentTab);
+    // if dir is left, we filter for previous tabs instead
+    nextNotDisabledTabs = tabs.filter((tab, i) => !tab.disabled && i < this.selectedIndex);
+    return nextNotDisabledTabs[nextNotDisabledTabs.length - 1];
```
jrobind: I think the only issue with this approach is that we won't be able to cycle through tabs. If we're on the last not-disabled tab, pressing the arrow right key will not shift focus to the start of the tab list.

packages/tabs/test/lion-tabs.test.js
jrobind: Thanks 😄

jorenbroekema:
@gerjanvangeest  @jrobind 

Wasn't aware that our Tabs cycled, my bad. I'll update this PR to reflect that behavior for disabled tabs as well.

jorenbroekema:
Fixed it, so if it's all good now, just need to squash the 4 commits into a single fix commit :)

jrobind:
Thanks @jorenbroekema, looks great! I've just squashed into a single fix commit 👍

gerjanvangeest (approved):

PR #1581, opened 2021-12-14
fix: type info in OverlayMixinTypes

remcovaes:
## What I did

1.  The type for `open`, `close` and `toggle` of the overlayMixin were incorrect, because the functions are async. That means that the return type should be `Promise<void>` instead of `void`.
1. Removed type import as it was not used.  
1. Removed the `toggle` from `LionSelectRich` as this can be handled by the `overlayMixin` now (as the TODO states).

There are 2 failing tests that already failed before I started, so I assume my changes do not have effect on those tests. Let me know if I have to make them work in order to get this pull request merged. Also, let me know if I messed something up :)

### DIFF
--- a/packages/select-rich/src/LionSelectRich.js
+++ b/packages/select-rich/src/LionSelectRich.js
@@ -247,11 +247,6 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     this._onFormElementsChanged();
   }
 
-  // TODO: move to overlayMixin and offer open and close
-  toggle() {
-    this.opened = !this.opened;
-  }
-
   /**
    * In the select disabled options are still going to a possible value for example
    * when prefilling or programmatically setting it.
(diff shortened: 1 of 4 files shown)


packages/select-rich/src/LionSelectRich.js
```diff
@@ -247,11 +247,6 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     this._onFormElementsChanged();
   }
 
-  // TODO: move to overlayMixin and offer open and close
-  toggle() {
```
jorenbroekema, line 251: Yeah toggle is already in OverlayMixin so no need to override it here in LionSelectRich and therefore not a breaking change to remove this.

jorenbroekema (approved):
Looks good to me, sometimes we forget that we manually type our mixins because TS doesn't play nice with them, so we probably just forgot making these async in the .d.ts file

tlouisse (approved):

Issue #1589, opened 2021-12-30
Wrong link to custom-elemnts manifest in package.json


jogibear9988:
## Expected behavior

In your package.json of your components it should look like this:

"customElements": "custom-elements.json",

## Actual Behavior

at th emoment it looks like this:
"customElementsManifest": "custom-elements.json",

## Additional context

I tried @lion/switch.

I'm currently workin on a designer application wich uses custom elements, by parsing it's manifest.
So it would be nice if this could be fixed.
see: https://node-projects.github.io/web-component-designer-demo/index.html
see the spec: https://github.com/webcomponents/custom-elements-manifest#referencing-manifests-from-npm-packages

PR #1590, opened 2022-01-04
fix: rename customElementsManifest to customElements in package.json

gerjanvangeest:
## What I did

1. fix: rename customElementsManifest to customElements in package.json

closes https://github.com/ing-bank/lion/issues/1589

### DIFF
--- a/.changeset/nasty-tables-visit.md
+++ b/.changeset/nasty-tables-visit.md
@@ -1,0 +1,36 @@
+---
+'rocket-preset-extend-lion-docs': patch
+'@lion/accordion': patch
+'@lion/button': patch
+'@lion/calendar': patch
+'@lion/checkbox-group': patch
+'@lion/collapsible': patch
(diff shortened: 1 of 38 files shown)


.changeset/nasty-tables-visit.md
```diff
@@ -0,0 +1,36 @@
+---
+'rocket-preset-extend-lion-docs': patch
```
daKmoR, line 2: this package does not export a custom element manifest as it's not a custom element (the found cases are test fixtures)

```suggestion

```

tlouisse (approved):

.changeset/nasty-tables-visit.md
tlouisse: ah, so no bump needed indeed

gerjanvangeest: Ai, didn't see that

PR #1597, opened 2022-01-19
chore(docs): correct event-name to close overlay

sbolz:
## What I did

The event-name to close an overlay was wrong in a docs page. This PR just corrects the name.

I also checked for other occurrences where this name might be wrong, but I couldn't find any.


gerjanvangeest (approved):

gerjanvangeest:
@sbolz Thanks for the fix

PR #1600, opened 2022-02-02
chore: upgrade CEM

jorenbroekema:


jorenbroekema:
@gerjanvangeest Hmm locally with this edited yarn.lock upgrading CEM works okay for me now without the postinstall error but in CI it still produces that postinstall error ;( too bad, can't really see what's causing this atm

PR #1603, opened 2022-02-02
feat: remove fixed versions @lion and upgrade TSC

jorenbroekema:
## What I did

1. Upgrade some deps that were bothering me
2. Unfix lion dependencies and use caret
3. Lint-versions script allows carets and tildes now, but still requires always having the latest version number for internal dependencies
4. Upgrade to latest TSC, fixed 120 or so type issues. Some of those had to be `@ts-ignore`'d due to how bad TS is with JS mixins, [the maintainer himself mentioned this is a classic example of where you should `@ts-ignore`](https://github.com/microsoft/TypeScript/issues/36821#issuecomment-588375051), so that also means is that we should tell our users that `skipLibCheck` in their TS configs is unavoidable. Hopefully TS becomes better at this stuff....

WIP, I'm going to run some tests with changesets to see what happens if:
- [x] `@lion/core` does a patch bump --> I expect no bumps anywhere else, because in the other places we now use `^` which means end users will get latest `@lion/core` with a fresh install of a dependent (let's say `@lion/button`), the patch bump is non-breaking and the dependents will be semver compatible
- [x] `@lion/core` does a breaking change minor bump --> I expect bumps everywhere else because this is a breaking bump and thus semver incompatible bump for dependents (e.g. `@lion/button`). Changesets should patch the button so that our users will get the latest core with the latest button (but not the `@lion/button` before the bump!). If the `@lion/core` change was actually breaking for `@lion/button`, of course since `@lion/button` is part of this repo we would manually bump `@lion/button` as we fix the breakage, if that fix leads to a breaking change in `@lion/button`, that will be a minor. Changesets only does a patch bump to ensure latest `@lion/core` can end up at the end user if they only consume e.g. `@lion/button`, which would otherwise never happen because the old `@lion/button` has `@lion/core` stuck at 1 minor before.

We should be seeing way less bloat in our changelogs with this, since we'll only need dependents to bump for semver incompatible bumps which are minors `<1.0.0` and majors `>1.0.0`. 

Broader discussion here: https://lit-and-friends.slack.com/archives/CJGFWJN9J/p1643202849020900 and also discussed at length with @tlouisse


jorenbroekema:
I checked how changeset will deal with the non-fixed versions and my hypotheses were correct.

While reviewing, feel free to double check by running:
```sh
yarn changeset # do a patch for lion core, only core is bumped, do a minor for lion core, all other packages are bumped with patch
yarn changeset version # see the diff
```

jorenbroekema:
@tlouisse  this PR is basically ready but we still need to add a changeset. 

In my opinion, unfixing the dependencies and upgrading to latest TSC, it's probably safe to say that this is bound to be a breaking change for someone at some point, do you agree doing a minor bump for all involved packages (that's basically everything in `packages` folder)? I don't think the TSC upgrade in providence is going to be breaking, no type issues there, so maybe a patch for that one?

packages/listbox/src/LionOption.js
```diff
@@ -14,7 +14,9 @@ import { css, DisabledMixin, html, LitElement } from '@lion/core';
  * Element gets state supplied externally, reflects this to attributes,
  * enabling SubClassers to style based on those states
  */
-export class LionOption extends DisabledMixin(ChoiceInputMixin(FormRegisteringMixin(LitElement))) {
+export class LionOption extends DisabledMixin(
+  ChoiceInputMixin(FormRegisteringMixin(SlotMixin(LitElement))),
```
jorenbroekema, line 18: LionOption uses slots getter, so SlotMixin should be applied

packages/select-rich/index.js
```diff
@@ -1,10 +1,11 @@
 /**
- * @deprecated
+ * !deprecated
```
jorenbroekema, line 2: Latest typescript is complaining about this for some dumb reason so I just prefix it with something else.. just FYI in case you're searching for these.

gerjanvangeest (approved):
Looks good to me, but i believe @tlouisse also wanted to review

packages/core/src/DelegateMixin.js
```diff
@@ -18,6 +18,7 @@ import { dedupeMixin } from '@open-wc/dedupe-mixin';
  */
 const DelegateMixinImplementation = superclass =>
   // eslint-disable-next-line
+  // @ts-ignore https://github.com/microsoft/TypeScript/issues/36821#issuecomment-588375051
   class extends superclass {
```
tlouisse, line 22: Will there be an alternative way, so we can become "skipLibCheck free" again?

daKmoR: TS and Mixins will probably not be friends any time soon 😅

for `DelegateMixin` in particular, it could be a lit controller... and this pattern is supported by TS
IMHO we should convert all mixins we can to lit controllers

I expect quite some Mixins will remain unless we find a nice way to add properties via controllers... so then it's either skipLibCheck or do specific inheritance...

packages/overlays/src/OverlaysManager.js
```diff
@@ -113,7 +113,7 @@ export class OverlaysManager {
       .reverse()
       .forEach((ctrl, i) => {
         // eslint-disable-next-line no-param-reassign
-        ctrl.elevation = i + 1;
+        ctrl.elevation = `${i + 1}`;
```
tlouisse, line 116: can it be fixed here: https://github.com/ing-bank/lion/blob/2b73c50f6e7ff9ef96122a74536e7c4df9b8553b/packages/overlays/src/OverlayController.js#L420 
?

packages/core/src/DelegateMixin.js
jorenbroekema: Yeah I think the only way to be skipLibCheck free again is for us to migrate to something typescript can understand, like Lit controllers. I'm somewhat skeptical that this will magically solve our typed inheritance issue that surfaced with the latest typescript, but who knows.

My interpretation of the comments by Ryan Cavanaugh are that @ts-ignore is not a sin and not uncommon, which means that skipLibCheck as an extension of that isn't either. So when users ask about it we can just explain that or have it in our README -> Rationale somewhere?

packages/overlays/src/OverlaysManager.js
jorenbroekema: Yep, good idea, since the getter is already type number too.

.changeset/odd-suns-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/overlays': minor
+---
+
+BREAKING: elevation property setter on OverlayController accepts numbers only, previously this was a number as a string. This syncs it with the getter which returns a number.
```
tlouisse, line 5: Nice, thanks.

I think we can regard it a bugfix, though? Since getter/setter should have been the same type (that's a bug) and providing a string will still work in runtime (not breaking in that regard).

Apart from that, thanks for all the work and I will approve it :)

jorenbroekema: Hmm it's a public API type change, I'd say a breaking bugfix. I guess it doesn't matter much since it's bundled in the same breaking minor release of @lion/overlays anyway, so might as well clarify that this could potentially break a user's code (types-wise).

tlouisse: Depends on how you look at it, indeed. Since we need the minor release anyways, it's not a huge deal/difference :)

jorenbroekema:
Added some rationale docs about versioning which we can link in our Slack channel when we release this
localhost_8000_docs_rationales_versioning_

jorenbroekema:
Added some docs about TypeScript in our Rationale section as well which we can also link in the future when people ask questions about type issues etc.
localhost_8000_docs_rationales_TypeScript_

tlouisse (approved):

daKmoR (approved):

narzac (approved):

PR #1605, opened 2022-02-07
fix(collapsible): moved binding to constructor

joquitch:
## What I did

1. Moved `.bind(this)` to constructor

### DIFF
--- a/packages/collapsible/test/lion-collapsible.test.js
+++ b/packages/collapsible/test/lion-collapsible.test.js
@@ -100,6 +100,14 @@ describe('<lion-collapsible>', () => {
       await collapsible.updateComplete;
       expect(isCollapsibleOpen).to.equal(false);
     });
+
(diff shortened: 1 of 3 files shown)


packages/collapsible/test/lion-collapsible.test.js
```diff
@@ -100,7 +100,6 @@ describe('<lion-collapsible>', () => {
       await collapsible.updateComplete;
       expect(isCollapsibleOpen).to.equal(false);
     });
-  });
```
gerjanvangeest, line 103: This line should stay.

gerjanvangeest (approved):

packages/collapsible/test/lion-collapsible.test.js
gerjanvangeest: Since your obviously busy, i fixed it for you :)

gerjanvangeest:
Thanks for the fix 💪

PR #1612, opened 2022-02-28
Fix: Prevent default behavior while scrolling using arrow keys in a h…

Jellepostma:
…orizontal listbox

## What I did

1. Added ev.preventDefault in the left and right arrow key handlers in the listboxmixin, to prevent the listbox from scrolling unintended when using these keys to navigate through the list when it is horizontal. The same fix has been already applied for vertical scrolling.

## How to test
1. On the current master, check the horizontal listbox example using a browser window that is smaller than the width of the listbox. Using arrow keys, it will start scrolling even if the next option is completely in view due to default scrolling behavior of these arrow keys.

### DIFF
--- a/.changeset/tasty-socks-hammer.md
+++ b/.changeset/tasty-socks-hammer.md
@@ -1,0 +1,5 @@
+---
+'@lion/listbox': patch
+---
+
+Fix: Prevent default behavior while scrolling using arrow keys in a horizontal listbox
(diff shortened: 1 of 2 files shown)


gerjanvangeest:
Hi Jelle, thanks for the contribution 💪 

Can you add a changeset (`yarn changeset`) to make the change also ready for release?

gerjanvangeest (approved):

Jellepostma:
@jorenbroekema Done! Thank you for the rebase :)

PR #1647, opened 2022-04-01
feat: update to @open-wc/scoped-element not needing a polyfill

daKmoR:
## What I did

BREAKING CHANGE: Work without polyfill if possible

When using [component composition](https://lit.dev/docs/composition/component-composition/) in a Lion Component we always made it very explicit which sub-components are used.
On top of that we scoped these [sub components](https://open-wc.org/docs/development/scoped-elements/) to the [current shadow root](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) allowing multiple version to be used simultaneously.

To enable this features we relied on the fact that the `ScopedElementsMixin` did loaded the needed polyfill for us.

We however over time got feedback from multiple consumers that lion components "break the app as soon as you load them".
The reasons is/was that not everyone is always using `ScopedElementsMixin` or in full control of the app (or its load order).

To quote the release notes of `ScopedElementsMixin` v2.1.0:

> ScopedElementsMixin 2.x tried to be as convenient as possible by automatically loading the scoped custom elements registry polyfill.
> This however led to a fatal error whenever you registered any component before ScopedElementsMixin was used.

And this was the case.

With the upgrade to `@open-wc/scoped-elements` v2.1.1 Lion now no longer automatically loads the polyfill through `ScopedElementsMixin`.

This essentially means the polyfill became optional which results in the following behavior

1. If polyfill is not loaded it will use the global registry as a fallback
2. Log error if actually scoping is needed and polyfill is not loaded
3. If you manually create elements you will need to handle polyfilled and not polyfilled cases now

```diff
-  const myButton = this.shadowRoot.createElement('my-button');
+  const myButton = this.createScopedElement('my-button');
```

This also removes `@webcomponents/scoped-custom-element-registry` as a production dependency.

If you need scoping be sure to load the polyfill before any other web component gets registered.

It may look something like this in your HTML

```html
<script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script>
```

or if you have an SPA you can load it at the top of your app shell code

```js
import '@webcomponents/scoped-custom-element-registry';
```

You need scoping if you want to

use 2 major versions of a web component (e.g. in an SPA pageA uses 1.x and pageB uses 2.x of color-picker)
or you want to use the same tag name with different implementations (use tag color-picker from foo here and from bar here)

See more details at

- [Lion release blog post](https://lion-web.netlify.app/blog/lion-without-polyfills/)
- [@open-wc/scoped-elements release blog post](https://open-wc.org/blog/scoped-elements-without-polyfill/)
- [Change log of ScopedElementsMixin](https://github.com/open-wc/open-wc/blob/master/packages/scoped-elements/CHANGELOG.md#210)

### DIFF
--- a/docs/blog/lion-without-polyfills.md
+++ b/docs/blog/lion-without-polyfills.md
@@ -1,0 +1,138 @@
+---
+title: Lion without polyfills
+published: true
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-05
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+The only reason Lion always loaded a polyfill was because of its usage of [@open-wc/scoped-elements](https://open-wc.org/docs/development/scoped-elements/). From today on this polyfill became optional.
+
+When using [component composition](https://lit.dev/docs/composition/component-composition/) in a Lion Component we always made it very explicit which sub-components are used.
+On top of that we scoped these [sub components](https://open-wc.org/docs/development/scoped-elements/) to the [current shadow root](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) allowing multiple version to be used simultaneously.
+
+This means that you can use a Lion Component like `lion-listbox` (which uses component composition) and never have to worry about if the internally used components clash with others you are already using.
+
+## How does it work?
+
+1. Within Lion classes we only import other classes (e.g. in class `MyCard` we use `MyCardHeader` via composition)
+2. We define them as `scopedElements` (`my-card-header: MyCardHeader`) and let the ScopedElementsMixin handle the rest
+
+To clarify: within Lion class files we never import files that run `customElement.define`
+
+```js
+import { LitElement, html, ScopedElementsMixin } from '@lion/core';
+import { MyCardHeader } from './MyCardHeader.js';
+
+export class MyCard extends ScopedElementsMixin(LitElement) {
+  static scopedElements = {
+    'my-card-header': MyCardHeader,
+  };
+
+  render() {
+    return html`
+      <div>
+        <my-card-header></my-card-header>
+        <slot></slot>
+      </div>
+    `;
+  }
+}
+```
+
+## Known challenges in previous releases
+
+The code above totally makes sense - however we always assumed that a scoped registry will be available.
+Which was somewhat of a valid assumption as all our components are using the `ScopedElementsMixin` and it in turn loads a polyfill for the scoped registry.
+
+We however over time got feedback from multiple consumers that lion components "break the app as soon as you load them".
+The reasons is/was that not everyone is always using `ScopedElementsMixin` or in full control of the app (or its load order).
+
+To quote the release notes of the latest version of `ScopedElementsMixin`:
+
+> ScopedElementsMixin 2.x tried to be as convenient as possible by automatically loading the scoped custom elements registry polyfill.
+> This however led to a fatal error whenever you registered any component before ScopedElementsMixin was used.
+
+And this was the case.
+
+## How do we fix it?
+
+With the latest release of Lion we now updated to the latest version of `ScopedElementsMixin` which means Lion now works in all apps as long as there is no need for actual scoping.
+
+To rephrase it:
+
+> Lion works without loading any polyfills
+
+If you extend Lion components and you imperatively create scoped custom elements, you should now use a helper function that will work in scoped and unscoped cases.
+
+```diff
+-  const myButton = this.shadowRoot.createElement('my-button');
++  const myButton = this.createScopedElement('my-button');
+```
+
+## Be explicit and stay forward compatible
+
+Be sure to always define **ALL** the sub elements you are using in your template within your `scopedElements` property.
+
+```js
+import { LitElement, html, ScopedElementsMixin } from '@lion/core';
+import { MyCardHeader } from './MyCardHeader.js';
+
+export class MyCard extends ScopedElementsMixin(LitElement) {
+  static scopedElements = {
+    'my-card-header': MyCardHeader,
+  };
+
+  render() {
+    return html`
+      <div>
+        <my-card-header></my-card-header>
+        <slot></slot>
+        <my-card-footer></my-card-footer>
+      </div>
+    `;
+  }
+}
+```
+
+☝️ here we are missing a definition for `my-card-footer` in `scopedElements`.
+
+This means as soon as there is support for the scoped registry (be it native of via a polyfill) this component will not be available anymore because every new scoped registry starts off empty (there is no inheritance of a global or parent registry).
+
+Therefore **always** define all your sub elements.
+
+## How to get scoping
+
+You need scoping if you want to:
+
+- use 2 major versions of a web component (e.g. in an SPA pageA uses 1.x and pageB uses 2.x of color-picker)
+- use the same tag name with different implementations (use tag color-picker from foo here and from bar here)
+
+This usually is only needed in bigger Single Page Applications.
+In smaller applications or static sites (like 11ty, wordpress, ...) these tag name clashes are unlikely.
+
+If you need scoping and the browser you are using does not support a [scoped registry](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md) yet (which is none in April 2022) then you need to install and load a polyfill first thing in your HTML.
+
+```bash
+npm i @webcomponents/scoped-custom-element-registry
+```
+
+It could look something like this:
+
+```html
+<script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script>
+```
+
+or if you have an SPA you can load it at the top of your app shell code
+
+```js
+import '@webcomponents/scoped-custom-element-registry';
+```
+
+## Learn more
+
+If you want to learn more please check the
+
+- [Release blog post](https://open-wc.org/blog/scoped-elements-without-polyfill/)
+- [Change log of ScopedElementsMixin](https://github.com/open-wc/open-wc/blob/master/packages/scoped-elements/CHANGELOG.md#210)
--- a/.changeset/real-yaks-float.md
+++ b/.changeset/real-yaks-float.md
@@ -1,0 +1,98 @@
+---
+'@lion/core': minor
+'@lion/form-core': minor
+'@lion/listbox': minor
+'@lion/select-rich': minor
+'@lion/switch': minor
+'@lion/accordion': minor
+'@lion/ajax': minor
+'@lion/button': minor
+'@lion/calendar': minor
+'@lion/checkbox-group': minor
+'@lion/collapsible': minor
+'@lion/combobox': minor
+'@lion/dialog': minor
+'@lion/fieldset': minor
+'@lion/form': minor
+'@lion/form-integrations': minor
+'@lion/helpers': minor
+'@lion/icon': minor
+'@lion/input': minor
+'@lion/input-amount': minor
+'@lion/input-date': minor
+'@lion/input-datepicker': minor
+'@lion/input-email': minor
+'@lion/input-iban': minor
+'@lion/input-range': minor
+'@lion/input-stepper': minor
+'@lion/input-tel': minor
+'@lion/input-tel-dropdown': minor
+'@lion/localize': minor
+'@lion/overlays': minor
+'@lion/pagination': minor
+'@lion/progress-indicator': minor
+'@lion/radio-group': minor
+'@lion/select': minor
+'@lion/steps': minor
+'@lion/tabs': minor
+'@lion/textarea': minor
+'@lion/tooltip': minor
+'@lion/validate-messages': minor
+---
+
+BREAKING CHANGE: Work without polyfill if possible
+
+When using [component composition](https://lit.dev/docs/composition/component-composition/) in a Lion Component we always made it very explicit which sub-components are used.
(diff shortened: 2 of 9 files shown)


docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,97 @@
...
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-04
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
+This also scoped these sub components to the current shadow root.
+
+This means that you can use a lion components like `lion-listbox` and never have to worry about if the internally used components clash with others you are already using.
```
tlouisse, line 13: component*

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,130 @@
...
+title: Lion without poylfills
+published: true
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-04
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
+This also scoped these sub components to the current shadow root.
```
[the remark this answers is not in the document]
daKmoR: good idea - added 👍

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,130 @@
+---
+title: Lion without poylfills
+published: true
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-04
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
```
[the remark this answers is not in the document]
daKmoR: added

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+This is commonly revered to as [component composition](https://lit.dev/docs/composition/component-composition/).
+On top of that we scoped these [sub components](https://open-wc.org/docs/development/scoped-elements/) to the [current shadow root](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md).
+
+This means that you can use a lion component like `lion-listbox` and never have to worry about if the internally used components clash with others you are already using.
```
tlouisse, line 14: Maybe we should mention here that `lion-listbox` uses LionOptions internally as 'lion-options'/

Even better: give a more hypothetical example, since most of our internal compositions are workarounds for the to be implemented [a11y spec](https://wicg.github.io/aom/notification-api.html).

So the most simple and understandable use case for a consumer would be shadow dom usage like `MyCard/my-card` that uses `MyCardHeader/my-card-header`

daKmoR: the example below is for Listbox? 🤔

not clear enough? or worth repeating here?

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+title: Lion without poylfills
+published: true
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-04
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
+This is commonly revered to as [component composition](https://lit.dev/docs/composition/component-composition/).
```
tlouisse, line 11: referred*

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Lion without poylfills
+published: true
+description: Lion has been a long user of the scoped registry - always requiring a polyfill - but no more. Load the polyfill only if you need it.
+date: 2022-04-04
+tags: [javascript, polyfills]
+cover_image: /blog/images/introducing-lions-website-cover-image.jpg
+---
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
```
tlouisse, line 10: components*

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+      <div>
+        <!-- example code - not the real implementation -->
+        <lion-option></lion-option>
+      </div>
+    `;
+  }
+}
+```
+
+## What is wrong with it?
```
tlouisse, line 44: The title seems to suggest smth is wrong with the above implementation, whereas it was the prev version of ScopedElementsMixin was wrong.

Maybe make it "## Known challenges in previous releases"

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+On top of that we scoped these [sub components](https://open-wc.org/docs/development/scoped-elements/) to the [current shadow root](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md).
+
+This means that you can use a lion component like `lion-listbox` and never have to worry about if the internally used components clash with others you are already using.
+
+## How does it work?
+
+1. Within lion class files we only import other class files
+2. We define them as `scopedElements` and let the mixin handle the rest
+
+To clarify within lion class files we never import files that run `customElement.define`
```
tlouisse, line 21:
```suggestion
To clarify: within lion class files we never import files that run `customElement.define`
```

docs/blog/lion-without-polyfills.md
tlouisse, line 11:
```suggestion
This is commonly referred to as [component composition](https://lit.dev/docs/composition/component-composition/).
```

docs/blog/lion-without-polyfills.md
tlouisse, line 10:
```suggestion
From the start we have been focused on making it very explicit which sub components are used within a given custom element.
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+```
+
+## What is wrong with it?
+
+The code above totally makes sense - however we always assumed that a scoped registry will be available.
+Which was somewhat of a valid assumption as all our components are using the `ScopedElementsMixin` and it in turn loads a polyfill for the scoped registry.
+
+We however over time we got multiple feedbacks that lion components "break the app as soon as you load them".
```
tlouisse, line 49:
```suggestion
We however over time we got feedback from multiple consumers that lion components "break the app as soon as you load them".
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+The reasons is/was that not everyone is always using `ScopedElementsMixin` or in full control of the app (or it's load order).
```
tlouisse, line 50:
```suggestion
The reasons is/was that not everyone is always using `ScopedElementsMixin` or in full control of the app (or its load order).
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+> ScopedElementsMixin 2.x tried to be as convenient as possible by automatically loading the scoped custom elements registry polyfill.
+> This however led to a fatal error whenever you registered any component before ScopedElementsMixin was used.
+
+And this was the case.
+
+## How do we fix it?
+
+With the latest release of lion we now updated to the latest version of `ScopedElementsMixin` which means lion now works in all apps as long as there is no need for actual scoping.
+
+If you extends lion components and you imperatively create scoped custom elements then you should now use a helper function that will work in scoped and unscoped cases.
```
tlouisse, line 63:
```suggestion
If you extend lion components and you imperatively create scoped custom elements, you should now use a helper function that will work in scoped and unscoped cases.
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+
+```diff
+-  const myButton = this.shadowRoot.createElement('my-button');
++  const myButton = this.createScopedElement('my-button');
+```
+
+## How to get scoping
+
+You need scoping if you want to
```
tlouisse, line 72:
```suggestion
You need scoping if you want to:
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+
+- use 2 major versions of a web component (e.g. in an SPA pageA uses 1.x and pageB uses 2.x of color-picker)
+- or you want to use the same tag name with different implementations (use tag color-picker from foo here and from bar here)
```
tlouisse, line 75:
```suggestion
- use the same tag name with different implementations (use tag color-picker from foo here and from bar here)
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+
+From the start we have been focused on making it very explicit which sub component are used within a given custom element.
+This is commonly revered to as [component composition](https://lit.dev/docs/composition/component-composition/).
+On top of that we scoped these [sub components](https://open-wc.org/docs/development/scoped-elements/) to the [current shadow root](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/Scoped-Custom-Element-Registries.md).
+
+This means that you can use a lion component like `lion-listbox` and never have to worry about if the internally used components clash with others you are already using.
+
+## How does it work?
+
+1. Within lion class files we only import other class files
```
tlouisse, line 18:
```suggestion
1. Within Lion class files we only import other class files (e.g. in class `MyLionCard` we use `MyLionCardHeader` via composition)
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+2. We define them as `scopedElements` and let the mixin handle the rest
```
tlouisse, line 19:
```suggestion
2. We define them as `scopedElements` (`my-lion-card-header: MyLionCardHeader `) and let the ScopedElmentsMixin handle the rest
```

docs/blog/lion-without-polyfills.md
tlouisse: I would go with a purely hypothetical one consumers could immediately relate to without knowing our codebase and without the complexity of needing to be rendered to light dom (apart from the fact that inside our codebase it's implemented way different).
(MyLionCard and MyLionCardHeader would have a really low 'cognitive load')

patches/@open-wc+scoped-elements+2.0.1.patch
```diff
@@ -0,0 +1,113 @@
...
+     }
+ 
+     /**
+@@ -128,7 +135,24 @@ const ScopedElementsMixinImplementation = superclass =>
+      * @param {typeof HTMLElement} klass
+      */
+     defineScopedElement(tagName, klass) {
+-      return this.registry.get(tagName) || this.registry.define(tagName, klass);
++      const registeredClass = this.registry.get(tagName);
++      if (registeredClass && supportsScopedRegistry === false && registeredClass !== klass) {
```
tlouisse, line 81: Wouldn't this break already for every instance of LionInput that creates LionValidationFeedback (at instance level and not per constructor?), a.k.a. as soon as we don't load the polyfill ourselves internally in Lion?

docs/blog/lion-without-polyfills.md
daKmoR, line 10:
```suggestion
From the start, we have been focused on making it very explicit which sub-components are used within a given custom element.
```

docs/blog/lion-without-polyfills.md
daKmoR, line 11:
```suggestion
This is commonly referred to as [component composition](https://lit.dev/docs/composition/component-composition/).
```

docs/blog/lion-without-polyfills.md
daKmoR, line 21:
```suggestion
To clarify, within lion class files we never import files that run `customElement.define`
```

docs/blog/lion-without-polyfills.md
```diff
@@ -0,0 +1,98 @@
...
+```html
+<script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script>
+```
+
+or if you have an SPA you can load it at the top of your app shell code
+
+```js
+import '@webcomponents/scoped-custom-element-registry';
+```
+
```
tlouisse, line 94: In this scoping section, we should emphasize this is desired behavior and that as soon as scoped registries are supported by the platform, the behavior that you get with loading, will be the default.

tlouisse: Then I would expect a section here that refers to the title... (w/o polyfills)
Explain the need for it (perf in small apps) and how it is backw. compatible with enabling the registry (either via browser or polyfill)?

patches/@open-wc+scoped-elements+2.0.1.patch
daKmoR: this is comparing the `class` and the `LionValidationFeedback` will always be the same no matter how many `<lion-validate-feedback>` instances you create...

also I guess at least one test should fail if that would. not be the case

tlouisse: Exactly, looked over the `registeredClass !== klass` part :)

tlouisse:
So we only need to squash and then we can merge, right?

tlouisse (approved):

Issue #1630, opened 2022-03-16
[form-core] add documentation example for [label-sr-only]


tlouisse:

Knorgias:
Create a PR for this: https://github.com/ing-bank/lion/pull/1648

PR #1648, opened 2022-04-04
chore: add eg for label-sr-only attribute on input element

Knorgias:
## What I did

1. Added a section under `components/input/use-cases/` that displays an example on how to use the `label-sr-only` attribute on an input element.

closes https://github.com/ing-bank/lion/issues/1630

### DIFF
--- a/docs/components/input/use-cases.md
+++ b/docs/components/input/use-cases.md
@@ -13,6 +13,19 @@ Can be provided via the `label` attribute or via the `slot="label"`.
 export const label = () => html` <lion-input label="Input" name="input"></lion-input> `;
 ```
 
+## Label sr-only (a11y)
+
+Can be provided via the `label-sr-only` boolean attribute.
+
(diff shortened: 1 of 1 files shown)


gerjanvangeest (approved):

docs/components/input/use-cases.md
```diff
@@ -13,6 +13,17 @@ Can be provided via the `label` attribute or via the `slot="label"`.
 export const label = () => html` <lion-input label="Input" name="input"></lion-input> `;
 ```
 
+## Label sr-only (a11y)
+
+Can be provided via the `label-sr-only` boolean attribute.
+
+The label will be hidden, but still readable by screen readers.
```
tlouisse, line 20: It might be worth adding here that as soon as we support ElementInternals api, the equivalent would be `<lion-input aria-label="Input" name="input"></lion-input>`

PR #1649, opened 2022-04-04
chore: add eg for validator class that returns enum

Knorgias:
## What I did

1. Added an example of a new Validator class that returns an enum instead of a boolean value.

closes [#1631](https://github.com/ing-bank/lion/issues/1631)

### DIFF
--- a/docs/fundamentals/systems/form/validate.md
+++ b/docs/fundamentals/systems/form/validate.md
@@ -110,7 +110,7 @@ As you can see the 'validators' property expects a map (an array of arrays). So,
 
 ### Validator classes
 
-All validators extend from the default `Validator` class. Below example is an example of a validator could look like:
+All validators extend from the default `Validator` class. Below is an example of what a validator could look like:
 
 ```js
 class EqualsText extends Validator {
@@ -126,6 +126,40 @@ class EqualsText extends Validator {
 
   static getMessage({ fieldName, modelValue, formControl }) {
     return `Please make sure values are equal`;
+  }
+}
+```
+
+The `execute()` method returns a boolean by default, but it can also be customized to return enums instead. In this example below, we customize the method to return one among some possible enums as strings.
+
+```js
+class EnumOutComeValidator extends Validator {
+  static validatorName = 'EnumOutCome';
+
+  /**
+   * @param {string} modelValue
+   * @param {number} maxLength
+   * @returns {false|'overload'|'error'}
+   */
+  execute(modelValue, maxLength) {
+    if (modelValue.length > maxLength) {
+      return 'overload';
+    } else if (modelValue.length < 0) {
+      return 'error';
+    }
+    return false;
+  }
+
+  /**
+   * @param {{outcome: 'overload'|'error'|false}} opts
+   */
+  static async getMessage({ outcome }) {
+    const results = {
+      overload: 'Too many characters!',
+      error: 'Something went wrong..',
+      false: 'All good!',
+    };
+    return results[outcome];
   }
 }
 ```
--- a/packages/form-core/src/validate/Validator.js
+++ b/packages/form-core/src/validate/Validator.js
@@ -50,9 +50,9 @@ export class Validator extends EventTarget {
   static async = false;
 
   /**
-   * The function that returns a validity outcome. When we need to shpw feedback,
+   * The function that returns a validity outcome. When we need to show feedback,
    * it should return true, otherwise false. So when an error\info|warning|success message
-   * needs to be shown, return true. For async Validators, the function canretun  a Promise.
+   * needs to be shown, return true. For async Validators, the function can return a Promise.
    * It's also possible to return an enum. Let's say that a phone number can have multiple
    * states: 'invalid-country-code' | 'too-long' | 'too-short'
    * Those states can be retrieved in the getMessage


docs/fundamentals/systems/form/validate.md
```diff
@@ -110,7 +110,7 @@ As you can see the 'validators' property expects a map (an array of arrays). So,
 
 ### Validator classes
 
-All validators extend from the default `Validator` class. Below example is an example of a validator could look like:
+All validators extend from the default `Validator` class. Below is an example of what a validator could look like:
```
tlouisse, line 113: I hope I didn't write this :)

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,33 @@ class EqualsText extends Validator {
...
 ```
 
+The `execute()` method returns a boolean by default, but it can also be customized to return enums instead. In this example below, we customize the method to return one among some possible enums as strings.
+
+```js
+class EnumOutComeValidator extends Validator {
+  static validatorName = 'EnumOutCome';
+
+  execute(modelValue, foodCalories) {
+    return new Promise(resolve => {
```
tlouisse, line 140: this is a bit overhead and complex for the example, but it would be ok if you rmv this execute method and rename `mealSize ` below to `execute` 👍

packages/form-core/src/validate/Validator.js
```diff
@@ -50,9 +50,9 @@ export class Validator extends EventTarget {
   static async = false;
 
   /**
-   * The function that returns a validity outcome. When we need to shpw feedback,
+   * The function that returns a validity outcome. When we need to show feedback,
    * it should return true, otherwise false. So when an error\info|warning|success message
-   * needs to be shown, return true. For async Validators, the function canretun  a Promise.
+   * needs to be shown, return true. For async Validators, the function can retun a Promise.
```
tlouisse, line 55:
```suggestion
   * needs to be shown, return true. For async Validators, the function can return a Promise.
```

docs/fundamentals/systems/form/validate.md
tlouisse: Then (see in the tests), we should also show how to consume the enum outcome in `static async getMessage`

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,32 @@ class EqualsText extends Validator {
...
+      return 'Overload';
+    } else if (modelValue.length < 0) {
+      return 'Error';
+    }
+    return 'OK';
+  }
+
+  static async getMessage({ outcome }) {
+    const results = {
+      Overload: 'Too many characters!',
```
tlouisse, line 150: Looks good! Only thing I would change is lowercase obj keys, that is a good practice. Also we don't have to typecast the outcome

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,32 @@ class EqualsText extends Validator {
...
+      Error: 'Something went wrong..',
+      OK: 'All good!',
+    };
+    return results[/** @type {string} */ (outcome)];
```
tlouisse, line 154:
```suggestion
      overload: 'Too many characters!',
      error: 'Something went wrong..',
      ok: 'All good!',
    };
    return results[outcome];
```

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,35 @@ class EqualsText extends Validator {
...
+class EnumOutComeValidator extends Validator {
+  static validatorName = 'EnumOutCome';
+
+  execute(modelValue, maxLength) {
+    if (modelValue.length > maxLength) {
+      return 'overload';
+    } else if (modelValue.length < 0) {
+      return 'error';
+    }
+    return 'ok';
```
tlouisse, line 145: In case there's no error, it should return false: https://github.com/ing-bank/lion/blob/master/packages/input-tel/src/validators.js#L34 
(not really clear from this code, so that's why it's good we have an explicit demo ;))

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,35 @@ class EqualsText extends Validator {
 }
 ```
 
+The `execute()` method returns a boolean by default, but it can also be customized to return enums instead. In this example below, we customize the method to return one among some possible enums as strings.
+
+```js
+class EnumOutComeValidator extends Validator {
+  static validatorName = 'EnumOutCome';
+
+  execute(modelValue, maxLength) {
```
tlouisse, line 139:
```suggestion
 /**
  * @param {string} modelValue
  * @param {number} maxLength
  * @returns {false|'overload'|'error'} 
  */
  execute(modelValue, maxLength) {
```

docs/fundamentals/systems/form/validate.md
tlouisse, line 145:
```suggestion
    return false;
```

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,35 @@ class EqualsText extends Validator {
...
+  }
+
+  /**
+   * @param {{outcome: 'overload'|'error'|'ok'}} opts
+   */
+  static async getMessage({ outcome }) {
+    const results = {
+      overload: 'Too many characters!',
+      error: 'Something went wrong..',
+      ok: 'All good!',
```
tlouisse, line 155:
```suggestion
```

docs/fundamentals/systems/form/validate.md
```diff
@@ -130,6 +130,35 @@ class EqualsText extends Validator {
...
+    if (modelValue.length > maxLength) {
+      return 'overload';
+    } else if (modelValue.length < 0) {
+      return 'error';
+    }
+    return 'ok';
+  }
+
+  /**
+   * @param {{outcome: 'overload'|'error'|'ok'}} opts
```
tlouisse, line 149:
```suggestion
   * @param {{outcome: 'overload'|'error'}} opts
```

tlouisse (approved):

PR #1666, opened 2022-04-14
fix: do not show ResultValidator feedback message if input field is empty

Knorgias:
Co-authored by: Gerjan van Geest <gerjan.van.geest@ing.com>

## What I did

1. Check if the modeValue is empty to reset the validators after a form `reset` button click
2. Add test for this case
3. Refactor code to keep __isEmpty from being called more times, causing existing tests to fail

### DIFF
--- a/packages/form-core/src/validate/ValidateMixin.js
+++ b/packages/form-core/src/validate/ValidateMixin.js
@@ -499,7 +499,7 @@ export const ValidateMixinImplementation = superclass =>
 
     /**
      * step b (as explained in `validate()`), called by __finishValidation
-     * @param {{validator:Validator; outcome: boolean|string;}[]} regularValidationResult result of steps 1-3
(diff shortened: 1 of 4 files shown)


tlouisse:
Nice 👍

packages/form-core/src/validate/ValidateMixin.js
```diff
@@ -499,8 +499,8 @@ export const ValidateMixinImplementation = superclass =>
 
     /**
      * step b (as explained in `validate()`), called by __finishValidation
-     * @param {{validator:Validator; outcome: boolean|string;}[]} regularValidationResult result of steps 1-3
-     * @private
```
tlouisse, line 503: was `@private` removed by accident?

Knorgias: yes, reinstated

tlouisse (approved):

PR #1669, opened 2022-04-26
feat: add tri-state indeterminate checkbox feature

jorenbroekema:
[See WAI-ARIA docs on this pattern](https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html)

- [x] Add test for tri-state feature
- [x] Implement feature

GIF

### DIFF
--- a/packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
+++ b/packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
@@ -432,7 +432,7 @@ describe('<lion-checkbox-indeterminate>', () => {
 
   // https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html
   describe('mixed-state', () => {
-  it('can have a mixed-state (using mixed-state attribute), none -> indeterminate -> all, cycling through', async () => {
+    it('can have a mixed-state (using mixed-state attribute), none -> indeterminate -> all, cycling through', async () => {
       const el = await fixture(html`
         <lion-checkbox-group name="scientists[]">
           <lion-checkbox-indeterminate mixed-state label="Favorite scientists">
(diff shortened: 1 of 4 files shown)


gerjanvangeest:
In the wai-aria example, when all/no child boxes are checked the indeterminate only toggles between checked and unchecked. In your case it remembers the indeterminate state before.

Might be a good addition.

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -428,4 +429,79 @@ describe('<lion-checkbox-indeterminate>', () => {
     expect(elIndeterminate?.hasAttribute('indeterminate')).to.be.false;
     expect(elIndeterminate?.checked).to.be.true;
   });
+
+  // https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html
+  describe('Tri-state', () => {
```
gerjanvangeest, line 434:
```suggestion
  describe('mixed-state', () => {
  it('can have a mixed-state (using mixed-state attribute), none -> indeterminate -> all, cycling through', async () => {
```

jorenbroekema:
Oooh yeah you're right, I think that's a bit awkward, the WAI ARIA one situation makes more sense indeed. Will change it and add a test for that case tomorrow, shouldn't be too difficult :)

packages/checkbox-group/test/lion-checkbox-indeterminate.test.js
```diff
@@ -428,4 +429,80 @@ describe('<lion-checkbox-indeterminate>', () => {
     expect(elIndeterminate?.hasAttribute('indeterminate')).to.be.false;
     expect(elIndeterminate?.checked).to.be.true;
   });
+
+  // https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-2/checkbox-2.html
+  describe('mixed-state', () => {
+  it('can have a mixed-state (using mixed-state attribute), none -> indeterminate -> all, cycling through', async () => {
+    it('can have a tri-state (using tri-state attribute), none -> indeterminate -> all, cycling through', async () => {
```
gerjanvangeest, line 436: Now there is a double "it"

```suggestion
```

jorenbroekema: Giving me bad suggestions to make CI fail, I seeeee hehe

jorenbroekema:
I was super hungover yesterday so I couldn't get to it, trainings today and tomorrow, so hopefully this weekend I can get to it and finish it

jorenbroekema:
Done :) was actually slightly harder to implement than I thought initially but it works great now, also for nested indeterminate groups with mixed state 🎉

gerjanvangeest:
The WAI aria example and your implementation are different:

jorenbroekema:
Huh, weird, I can't reproduce your first gif @gerjanvangeest , for me it works (toggling between 3 states like in WAI ARIA)

gerjanvangeest:
> Huh, weird, I can't reproduce your first gif @gerjanvangeest , for me it works (toggling between 3 states like in WAI ARIA)

Ok, let me double check why i have it.

gerjanvangeest:
Oke, my bad. I was looking at the wrong example 🙈

gerjanvangeest (approved):

PR #1670, opened 2022-04-26
feat(ajax): add an option to initialize the cache interceptors...

whenlit:
Add an option to initialize the cache interceptors even when useCache is turned off in the global options

## What I did

1. Add an option `addCaching` to the regular options of the Ajax library, which causes the cache interceptors to be added even when the global `useCache` option is turned off
2. Remove the `useCache` check from the tests when storing things in the cache, effectively storing everything proactively. It could be argued that this should be configurable as well, in which case we could change the `addCaching` option to a `cacheOptions.proactive` option, or something.

### DIFF
--- a/.changeset/silent-ravens-smell.md
+++ b/.changeset/silent-ravens-smell.md
@@ -1,0 +1,5 @@
+---
+'@lion/ajax': minor
+---
+
+Add an option "addCaching" to the Ajax config, in order to add cache interceptors when useCache is turned off. In this situation, all requests are cached proactively.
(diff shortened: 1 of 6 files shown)


LarsDenBakker:
Looks good to me :)

LarsDenBakker (approved):

LarsDenBakker:
@whenlit  @daKmoR @tlouisse this is an important change for broader adoption within ING, could we move this forward?

tlouisse (approved):

PR #1677, opened 2022-05-03
fix: fixes issue 1483

DannyMoerkerke:
replaced `getElementById` with `querySelector` so correct element  is also selected when size = inline

## What I did

1. replaced `getElementById` with `querySelector` so correct element  is also selected when size = inline


docs/components/select-rich/use-cases.md
```diff
@@ -188,15 +188,15 @@ export const checkedIndexAndValue = ({ shadowRoot }) => html`
...
       }}
     />
   </div>
   <button
     class="log-button"
     @click=${() => {
-      const selectEl = shadowRoot.getElementById('checkedRichSelect');
+      const selectEl = shadowRoot.querySelector('#checkedRichSelect');
       console.log(`checkedIndex: ${selectEl.checkedIndex}`);
       console.log(`checkedValue: ${selectEl.checkedValue}`);
```
gerjanvangeest, line 201: CheckedValue is something from the past, should be `modelValue`.

DannyMoerkerke: fixed.

PR #1678, opened 2022-05-04
fix(input-tel-dropdown): do not sync regionCode if countryCode is the…

gerjanvangeest:
## What I did

1. Do not sync regionCode from input to dropdown if the countryCode didn't change
2. Only focus input on dropdownValue change if dropdown was opened. Otherwise it blocks user from use type ahead on the dropdown.

### DIFF
--- a/packages/input-tel-dropdown/src/LionInputTelDropdown.js
+++ b/packages/input-tel-dropdown/src/LionInputTelDropdown.js
@@ -300,9 +300,6 @@ export class LionInputTelDropdown extends LionInputTel {
       setTimeout(() => {
         this._inputNode.focus();
       });
-    } else {
-      // For native select
-      this._inputNode.focus();
     }
   }
 
@@ -333,9 +330,23 @@ export class LionInputTelDropdown extends LionInputTel {
     if (!dropdownElement || !regionCode) {
       return;
     }
+    const inputCountryCode = this._phoneUtil?.getCountryCodeForRegionCode(regionCode);
+
     if ('modelValue' in dropdownElement) {
+      const dropdownCountryCode = this._phoneUtil?.getCountryCodeForRegionCode(
+        dropdownElement.modelValue,
+      );
+      if (dropdownCountryCode === inputCountryCode) {
+        return;
+      }
       /** @type {* & FormatHost} */ (dropdownElement).modelValue = regionCode;
     } else {
+      const dropdownCountryCode = this._phoneUtil?.getCountryCodeForRegionCode(
+        dropdownElement.value,
+      );
+      if (dropdownCountryCode === inputCountryCode) {
+        return;
+      }
       /** @type {HTMLSelectElement} */ (dropdownElement).value = regionCode;
     }
   }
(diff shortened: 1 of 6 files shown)


packages/input-tel-dropdown/src/LionInputTelDropdown.js
```diff
@@ -301,9 +301,6 @@ export class LionInputTelDropdown extends LionInputTel {
       setTimeout(() => {
         this._inputNode.focus();
       });
-    } else {
```
tlouisse, line 304: hmm, unfortunately this will give weird behavior in browser

tlouisse: Maybe we should explicitly test this actually, so that it's captured in our tests as well

gerjanvangeest: Yes, was already thinking about it

packages/input-tel-dropdown/src/LionInputTelDropdown.js
```diff
@@ -334,9 +331,23 @@ export class LionInputTelDropdown extends LionInputTel {
     if (!dropdownElement || !regionCode) {
       return;
     }
+    const inputCountryCode = this._phoneUtil?.getCountryCodeForRegionCode(regionCode);
+
     if ('modelValue' in dropdownElement) {
```
tlouisse, line 336: 👍

packages/input-tel-dropdown/src/LionInputTelDropdown.js
```diff
@@ -227,6 +227,9 @@ export class LionInputTelDropdown extends LionInputTel {
     this._onDropdownValueChange = this._onDropdownValueChange.bind(this);
     /** @type {EventListener} */
     this.__syncRegionWithDropdown = this.__syncRegionWithDropdown.bind(this);
+
+    /** Set to true, since the input-tel will set the correct activeRegion */
+    this.hasNoDefaultSelected = true;
```
tlouisse, line 232: This was a prop of RichSelect, right?
Do you want to pass it along or smth, since it's not on InputTel prototype?
And if so, what's the goal of not setting/showing the default in the dropdown?

gerjanvangeest: It's a property of listbox.

Without this the input and dropdown will both define their own modelValue, and then decide together which one wins. This goes correct in lion, but not in our extended layer.

By setting it default to none, les code needs to run and it also prevents bugs (as in our extended layer).

tlouisse: But then it would need to be applied on refs.dropdown?

gerjanvangeest: Let me double check

gerjanvangeest: Removed it again

tlouisse (approved):

PR #1679, opened 2022-05-04
Fix/indeterminate checkbox combined with disabled

DannyMoerkerke:
## What I did

1.  fixed issue where disabled children would be checked when parent was checked
2. fixed issue where parent would not become checked when all children except disabled ones were checked
3. added tests


jorenbroekema:
Ah yes this is definitely a bug, we don't take into account disabled children in indeterminate checkbox.

However, I think the behavior should be slightly different:
Screenshot 2022-05-06 at 09 52 10

In my opinion when a child is disabled and unchecked, the indeterminate parent can never reach checked state, it will always be either unchecked or indeterminate.

jorenbroekema:
https://github.com/w3c/aria-practices/issues/2311 created an issue here as well, since the WAI ARIA docs on this pattern seem to not include the case where child checkboxes can be disabled.

tlouisse:
> Ah yes this is definitely a bug, we don't take into account disabled children in indeterminate checkbox.

I agree that a disabled unchecked child should not cause the parent to be checked (i.e. they should take part in the parent state computation). Feels better intuitively.

DannyMoerkerke:
So the correct behaviour should be that when the parent is checked, any non-disabled children should be checked but the parent should remain indeterminate?

jorenbroekema:
Yep, if there are disabled unchecked children, then checking the parent will check any non-disabled children but the parent remains indeterminate, meaning you can only ever toggle that one between indeterminate and unchecked.

DannyMoerkerke:
Fixed the edge cases.

jorenbroekema:
Nice! Did you sign the CTA? The check is still pending. I'll be able to do a thorough review tomorrow :)

DannyMoerkerke:
> Nice! Did you sign the CTA? The check is still pending. I'll be able to do a thorough review tomorrow :)

Should be done now.

gerjanvangeest:
You forgot to add a changeset (run: `yarn changeset`), so can be be released.

jorenbroekema (approved):

jorenbroekema:
https://github.com/ing-bank/lion/pull/1698

PR #1683, opened 2022-05-09
chore: ensure CEM order via wireIt

daKmoR:
## What I did

1. Add [wireit](https://github.com/google/wireit)
2. Ensure order of execution for CEM (for input-amount only) so CEM can get data from "dependency"
3. e.g. when running `yarn custom-elements-manifest` in input-amount it will actually execute `core -> localize -> form-core -> input -> input-amount`

TODO:
1. After testing a new version of `custom elements manifest` analyzer (with some added features and bugfixes) and we have all needed data for `input-amount` (including all dependencies) we should wire up all the other packages

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,9 @@
     "packages/*",
     "packages-node/*"
   ],
+  "dependencies": {
+    "wireit": "^0.4.0"
+  },
   "devDependencies": {
     "@babel/core": "^7.10.1",
     "@bundled-es-modules/fetch-mock": "^6.5.2",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -36,6 +36,9 @@
     "packages/*",
     "packages-node/*"
   ],
+  "dependencies": {
+    "wireit": "^0.4.0"
```
tlouisse, line 40: should be devDep?

gerjanvangeest:

gerjanvangeest:
Since we only have 1 CEM this setup is not needed any more.

PR #1689, opened 2022-05-10
fix(lion-field): remove name property exception to align with platform

palash2601:
## What I did

1. Remove check for name property as discussed with @tlouisse


.changeset/hungry-mirrors-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/form-core': patch
```
tlouisse, line 2:
```suggestion
'@lion/form-core': minor
```

tlouisse (approved):

PR #1691, opened 2022-05-10
fix: export lit-html2 directive

narzac:
## What I did

1. Export lit-html 2.0 directive in separate file following the lit-html convention

### DIFF
--- a/.changeset/popular-grapes-drop.md
+++ b/.changeset/popular-grapes-drop.md
@@ -1,0 +1,5 @@
+---
+'@lion/core': patch
+---
+
+Export forward compatible class based Directive in separate file like lit-html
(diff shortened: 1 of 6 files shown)


.changeset/popular-grapes-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/core': patch
+---
+
+Export lit-html 2.0 directive in separate file like lit-html
```
tlouisse, line 5:
```suggestion
Export forward compatible class based Directive in separate file like lit-html
```
```suggestion
Export lit-html 2.0 directive in separate file like lit-html
```

packages/core/directive.js
```diff
@@ -0,0 +1 @@
+export { directive, Directive } from 'lit-html/directive.js';
```
tlouisse, line 1: think this points to https://github.com/lit/lit/blob/lit-html-1.x/src/lit-html.ts#L36 and could be exported from index.js as well?

tlouisse (approved):

PR #1699, opened 2022-05-17
feat(ajax): add a `maxCacheSize` cache option

whenlit:
add a `maxCacheSize` cache option to specify a max size for the whole cache; also a `replacementPolicy` option that operates on the cache when it gets too large

## What I did

1. Add `maxCacheSize` option to `cacheOptions`
2. Add `replacementPolicy` option to `cacheOptions`
3. Add default FIFO_REPLACEMENT_POLICY
4. Store size with requests in the cache
5. Add size field to cache
6. Remove entries from the cache according to the policy when it's full


whenlit:
Changed this to draft in favor of https://github.com/ing-bank/lion/pull/1711, let's do the `maxCacheSize` itself first.

PR #1705, opened 2022-05-22
Feat(rating): add star rating element

MHMALEK:
## What I did
With the Rating element, users can share their opinions and experiences and submit a rating of their own. this element, provide an easy UI to implement a rating system.

- show different numbers of stars (5 stars is the default )
- pre-defined rate or showing an empty state that the user can change the rate
- an on-rate-change-event as a callback to outside world when the user rate changes
- easy to use with `currentRate` prop to handle current rate

lion-rate-element


gerjanvangeest:

PR #1709, opened 2022-05-24
fix: Allow elements with display:contents to be focusable in keyboard event trapping

stefanpearson:
I noticed an issue in overlays, where keyboard event trapping would skip certain focusable elements. After debugging for a while, I discovered that certain elements were deemed 'not visible' and would be skipped from the traversal if `display: contents` had been set. This is because of the check that is in place to explicitly query the box model as a last check, which would give a false negative.

This is problematic for when `display: contents` is important, and definitely has visible focusable elements within it!

## What I did

1. Added a check to see if `display: contents` is in the computed styles, and resolve `true`.

If anyone can think of a smarter check then I'm happy to change this. Perhaps querying the box model is not the right approach here anyway?

### DIFF
--- a/.changeset/eight-years-pump.md
+++ b/.changeset/eight-years-pump.md
@@ -1,0 +1,5 @@
+---
+"@lion/overlays": patch
+---
+
+fix: isVisible check on elements with display:contents
--- a/packages/overlays/src/utils/is-visible.js
+++ b/packages/overlays/src/utils/is-visible.js
@@ -5,6 +5,11 @@ const hasStyleVisibility = ({ visibility, display }) =>
   visibility !== 'hidden' && display !== 'none';
 
 /**
+ * @param {CSSStyleDeclaration} styles
+ */
+const isDisplayContents = ({ display }) => display === 'contents';
+
+/**
  * @param {HTMLElement} element
  * @returns {boolean} Whether the element is visible
  */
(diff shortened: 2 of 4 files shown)


jorenbroekema:
Looks good to me, thanks for contributing a fix. Probably wasn't too easy to find the code that was handling the focus trap.

Do you think it would make sense to add another test in `/pacakages/overlays/test/utils-test/get-focusable-elements.test.js` just as a smoke test, making sure that if there is a focusable element inside a parent element that has display: contents, it indeed caught by this util?

stefanpearson:
> Do you think it would make sense to add another test in `/pacakages/overlays/test/utils-test/get-focusable-elements.test.js` just as a smoke test, making sure that if there is a focusable element inside a parent element that has display: contents, it indeed caught by this util?

Yep seems sensible; I've added this now.

stefanpearson:
Thanks for the approval! What is the merge/publish process for these PRs? I've managed to use a workaround for now (removing `display: contents` for a certain component), but would be great to see this land soon.

jorenbroekema:
I'm one of the maintainers but no longer working at ING so one of my ex-colleagues needs to put their approval as well in order to be able to merge. After that it can be released pretty much right away unless something else is holding up the release for some reason. @tlouisse @gerjanvangeest 

tomorrow is also Ascension Day so I suspect a fair few people are taking a long weekend off :)

gerjanvangeest (approved):

PR #1713, opened 2022-05-25
feat(progress-indicator): add support for progress bar

hardikpthv:
## What I did

1. Added progress bar support in `progress-indicator`

### DIFF
--- a/docs/components/progress-indicator/assets/my-indeterminate-progress-spinner.js
+++ b/docs/components/progress-indicator/assets/my-indeterminate-progress-spinner.js
@@ -1,27 +1,22 @@
 import { html, css } from '@lion/core';
 import { LionProgressIndicator } from '@lion/progress-indicator';
 
-export class CustomProgressIndicator extends LionProgressIndicator {
+export class MyIndeterminateProgressSpinner extends LionProgressIndicator {
   static get styles() {
     return [
       css`
-        :host {
-          display: block;
-        }
-
-        .progress--icon {
+        .progress__icon {
           display: inline-block;
           width: 48px;
           height: 48px;
           animation: spinner-rotate 2s linear infinite;
         }
 
-        .progress--icon--circle {
+        .progress__filled {
           animation: spinner-dash 1.35s ease-in-out infinite;
           fill: none;
           stroke-width: 6px;
           stroke: var(--primary-color);
-          stroke-dasharray: 100, 28; /* This is a fallback for IE11 */
         }
 
         @keyframes spinner-rotate {
@@ -50,11 +45,11 @@ export class CustomProgressIndicator extends LionProgressIndicator {
 
   _graphicTemplate() {
     return html`
-      <svg class="progress--icon" viewBox="20 20 47 47">
-        <circle class="progress--icon--circle" cx="44" cy="44" r="20.2" />
+      <svg class="progress__icon" viewBox="20 20 47 47">
+        <circle class="progress__filled" cx="44" cy="44" r="20.2" />
       </svg>
     `;
   }
 }
 
-customElements.define('custom-progress-indicator', CustomProgressIndicator);
+customElements.define('my-indeterminate-progress-spinner', MyIndeterminateProgressSpinner);
--- a/docs/components/progress-indicator/overview.md
+++ b/docs/components/progress-indicator/overview.md
@@ -2,19 +2,26 @@
 
 A web component that implements accessibility requirements for progress indicators.
 
-```html
-<lion-progress-indicator></lion-progress-indicator>
+```js script
+import { html } from '@mdjs/mdjs-preview';
+import '@lion/progress-indicator/define';
 ```
 
-Note: You don't see a live demo here as it would be empty, since there is no styling. Check out the [examples](./examples.md) if you want to see a possible implementation.
+```html
+<lion-progress-indicator aria-label="Interest rate" value="50"></lion-progress-indicator>
+```
 
 ## Features
 
 This component is designed to be extended in order to add visuals.
 
+- Can be indeterminate or determinate, depending on whether it has a value.
 - Accessibility compliant
-- Localized "Loading" label
+- Localized "Loading" label in case of an indeterminate progress-indicator
 - Implementation independent of visuals
+- `value`: progress value, setting this makes the progress-indicator determinate.
+- `min`: progress min value
+- `max`: progress max value
 
 ## Installation
 
--- a/packages/progress-indicator/src/LionProgressIndicator.js
+++ b/packages/progress-indicator/src/LionProgressIndicator.js
@@ -1,9 +1,27 @@
-/* eslint-disable class-methods-use-this, import/no-extraneous-dependencies */
-
-import { nothing, LitElement } from '@lion/core';
+/* eslint-disable import/no-extraneous-dependencies */
+import { LitElement, nothing } from '@lion/core';
 import { localize, LocalizeMixin } from '@lion/localize';
 
+/**
+ * @typedef {import('@lion/core').TemplateResult} TemplateResult
+ */
 export class LionProgressIndicator extends LocalizeMixin(LitElement) {
+  static get properties() {
+    return {
+      value: {
+        type: Number,
+      },
+      min: {
+        type: Number,
+      },
+      max: {
+        type: Number,
+      },
+      _ariaLabel: { attribute: 'aria-label', type: String },
+      _ariaLabelledby: { attribute: 'aria-labelledby', type: String },
+    };
+  }
+
   static get localizeNamespaces() {
     return [
       {
@@ -67,7 +85,41 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
     ];
   }
 
+  /**
+   * @readonly
+   * @type {boolean}
+   */
+  get indeterminate() {
+    return !this.hasAttribute('value');
+  }
+
+  /**
+   * In case of a determinate progress-indicator it returns the progress percentage
+   * based on value, min & max.
+   * Could be used for styling inside the _graphicTemplate
+   *
+   * @example
+   * style="width: ${this._progressPercentage}%"
+   */
+  get _progressPercentage() {
+    if (this.indeterminate) {
+      return undefined;
+    }
+    return ((this.value - this.min) / (this.max - this.min)) * 100;
+  }
+
+  constructor() {
+    super();
+    this.value = 0;
+    this.min = 0;
+    this.max = 100;
+    this._ariaLabel = '';
+    this._ariaLabelledby = '';
+    this.__hasDefaultLabelSet = false;
+  }
+
   /** @protected */
+  // eslint-disable-next-line class-methods-use-this
   _graphicTemplate() {
     return nothing;
   }
@@ -78,12 +130,76 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
 
   connectedCallback() {
     super.connectedCallback();
-    this.setAttribute('role', 'status');
-    this.setAttribute('aria-live', 'polite');
+    this.setAttribute('role', 'progressbar');
+  }
+
+  /**
+   * Update aria labels on state change.
+   * @param {import('@lion/core').PropertyValues } changedProperties
+   */
+  updated(changedProperties) {
+    super.updated(changedProperties);
+
+    if (this.indeterminate) {
+      if (changedProperties.has('_ariaLabel') || changedProperties.has('_ariaLabelledby')) {
+        this._setDefaultLabel();
+      }
+      if (changedProperties.has('value')) {
+        this._resetAriaValueAttributes();
+        this._setDefaultLabel();
+      }
+    } else {
+      if (changedProperties.has('value')) {
+        if (!this.value || typeof this.value !== 'number') {
+          this.removeAttribute('value');
+        } else if (this.value < this.min) {
+          this.value = this.min;
+          this.setAttribute('aria-valuenow', this.min.toString());
+        } else if (this.value > this.max) {
+          this.value = this.max;
+          this.setAttribute('aria-valuenow', this.max.toString());
+        } else {
+          this.setAttribute('aria-valuenow', this.value.toString());
+        }
+        if (this.__hasDefaultLabelSet === true) {
+          this.removeAttribute('aria-label');
+        }
+      }
+      if (changedProperties.has('min')) {
+        this.setAttribute('aria-valuemin', this.min.toString());
+        if (this.value < this.min) {
+          this.value = this.min;
+        }
+      }
+      if (changedProperties.has('max')) {
+        this.setAttribute('aria-valuemax', this.max.toString());
+        if (this.value > this.max) {
+          this.value = this.max;
+        }
+      }
+    }
   }
 
   onLocaleUpdated() {
-    const label = localize.msg('lion-progress-indicator:loading');
-    this.setAttribute('aria-label', label);
+    super.onLocaleUpdated();
+    // only set default label for indeterminate
+    if (this.indeterminate) {
+      this._setDefaultLabel();
+    }
+  }
+
+  _resetAriaValueAttributes() {
+    this.removeAttribute('aria-valuenow');
+    this.removeAttribute('aria-valuemin');
+    this.removeAttribute('aria-valuemax');
+  }
+
+  _setDefaultLabel() {
+    if (this._ariaLabelledby) {
+      this.removeAttribute('aria-label');
+    } else if (!this._ariaLabel) {
+      this.setAttribute('aria-label', localize.msg('lion-progress-indicator:loading'));
+      this.__hasDefaultLabelSet = true;
+    }
   }
 }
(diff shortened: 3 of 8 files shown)


packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +139,85 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
 
   connectedCallback() {
     super.connectedCallback();
-    this.setAttribute('role', 'status');
+    const uid = uuid();
+    if (this.getAttribute('variant') === 'indeterminate') {
```
gerjanvangeest, line 143: Maybe we could leave out the variant attribute?

```suggestion
    if (!this.hasAttribute('value')) {
```

hardikpthv: Yeah, that sounds good.

hardikpthv: Yeah, that looks good.

docs/components/progress-indicator/overview.md
```diff
@@ -15,6 +24,11 @@ This component is designed to be extended in order to add visuals.
 - Accessibility compliant
 - Localized "Loading" label
 - Implementation independent of visuals
+- `value`: progress value, setting this makes the progress-bar determinate.
+- `min`: progress min value
+- `max`: progress max value
+- `--bar-color`: css custom property to set the bar color
```
gerjanvangeest, line 30: Should still remove  these

docs/components/progress-indicator/overview.md
```diff
@@ -15,6 +24,11 @@ This component is designed to be extended in order to add visuals.
...
+- `--placeholder-color`: css custom property to set the bar's placeholder color
```
gerjanvangeest, line 31:
```suggestion
```

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +126,42 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+
   onLocaleUpdated() {
-    const label = localize.msg('lion-progress-indicator:loading');
-    this.setAttribute('aria-label', label);
+    if (!this.hasAttribute('value') && !this.label) {
+      this.setAttribute('aria-label', localize.msg('lion-progress-indicator:loading'));
+    }
+  }
+
+  get _customStyles() {
```
tlouisse, line 159: this could be added in a demo component since this is an unopinionated base component

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,18 +1,65 @@
...
+      expect(el.getAttribute('aria-valuenow')).to.be.eq('25');
+    });
 
-    it('sets the right role', async () => {
+    it('allows to set min & max values', async () => {
+      const el = await fixture(
+        html` <lion-progress-indicator value="25"></lion-progress-indicator> `,
+      );
+      expect(el.getAttribute('aria-valuemin')).to.be.eq('0');
+      expect(el.getAttribute('aria-valuemax')).to.be.eq('100');
```
tlouisse, line 41: all aria attrs can go inside an a11y describe

tlouisse: I would actually expect tests here that allow to define min and max. Also after init

tlouisse:
Hey @hardikpthv. 

Thanks for all the effort 💪 And sorry for the late reply.

Despite all my picky comments, looks good 👍 

Main improvement points I think: 
- make the base/primitive component unopinionated about appearance and determinate state
- only use [role=progressbar] for simplicity
- test whether the default loading label is still needed and apply always (regardless of determinate state). If not needed, .label can be deleted and the consumer can decide to use aria-label

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +126,42 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
     this.setAttribute('aria-live', 'polite');
   }
 
+  /**
+   * Update aria labels on state change.
+   * @param {import('@lion/core').PropertyValues } changedProperties
+   */
+  updated(changedProperties) {
+    if (changedProperties.has('value') && this.hasAttribute('value')) {
+      this.setAttribute('aria-valuenow', this.value.toString());
```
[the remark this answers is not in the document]
tlouisse: Why do we test for reflection of the value attr?

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +126,42 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+    }
+  }
+
   onLocaleUpdated() {
-    const label = localize.msg('lion-progress-indicator:loading');
-    this.setAttribute('aria-label', label);
+    if (!this.hasAttribute('value') && !this.label) {
```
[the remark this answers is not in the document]
gerjanvangeest: Remove the label for determinate will make the a11y test fail

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +125,45 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
-    this.setAttribute('aria-live', 'polite');
+    this.setAttribute('role', 'progressbar');
+  }
+
+  /**
+   * Update aria labels on state change.
+   * @param {import('@lion/core').PropertyValues } changedProperties
+   */
+  updated(changedProperties) {
+    if (changedProperties.has('value') && !this.indeterminate) {
```
tlouisse, line 136: You forgot to call `super.updated(changedProperties);`  ;)

Could we add at the top: 
```js
if (this.indeterminate) {
  return;
}
```

(and then drop all `!this.indeterminate` checks below)

docs/components/progress-indicator/assets/my-indeterminate-progress-spinner.js
```diff
@@ -50,11 +45,11 @@ export class CustomProgressIndicator extends LionProgressIndicator {
 
   _graphicTemplate() {
     return html`
-      <svg class="progress--icon" viewBox="20 20 47 47">
-        <circle class="progress--icon--circle" cx="44" cy="44" r="20.2" />
+      <svg class="progress__icon" viewBox="20 20 47 47">
```
tlouisse, line 48: better :)

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +125,49 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+  }
+
+  /**
+   * Update aria labels on state change.
+   * @param {import('@lion/core').PropertyValues } changedProperties
+   */
+  updated(changedProperties) {
+    super.updated(changedProperties);
+
+    if (!this.indeterminate) {
```
tlouisse, line 138: Yeah better, although I personally still prefer the early return:) 
- https://github.com/Shopify/web-configs/blob/main/packages/eslint-plugin/docs/rules/prefer-early-return.md
- https://medium.com/swlh/return-early-pattern-3d18a41bba8

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,125 @@
...
+      );
+      // sets to default max: 100
+      expect(el.getAttribute('aria-valuenow')).to.be.eq('100');
+      el.value = -20;
+      await el.updateComplete;
+      // sets to default min: 0
+      expect(el.getAttribute('aria-valuenow')).to.be.eq('0');
+    });
+
+    it('has a _progressPercentage getter to get the progress percentage', async () => {
```
tlouisse, line 71: Maybe make a separate describe 'Subclassers' and put it in there

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,125 @@
...
+  });
+
+  describe('Accessibility', () => {
+    describe('indeterminate', () => {
+      it('passes a11y test', async () => {
+        const el = await fixture(html` <lion-progress-indicator></lion-progress-indicator> `);
+        await expect(el).to.be.accessible();
+      });
+
+      it('by default', async () => {
```
tlouisse, line 88: sets the progressbar role? (and we should actually also allow to override it, but could be a later action point for the whole codebase)

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,125 @@
...
+        const el = await fixture(html` <lion-progress-indicator></lion-progress-indicator> `);
+        expect(el.getAttribute('role')).to.equal('progressbar');
+      });
+
+      it('adds a label by default', async () => {
```
tlouisse, line 93: what happens when the user defines his own aria label (or aria-describedby), after init?

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +125,49 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+          this.value = this.max;
+        }
+      }
+    }
   }
 
   onLocaleUpdated() {
-    const label = localize.msg('lion-progress-indicator:loading');
-    this.setAttribute('aria-label', label);
+    if (
```
tlouisse, line 166: what happens when the user defines his own aria label (or aria-describedby), after init?

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,125 @@
...
+        const el = await fixture(html` <lion-progress-indicator></lion-progress-indicator> `);
+        await el.localizeNamespacesLoaded;
+        expect(el.getAttribute('aria-label')).to.equal('Loading');
+      });
+    });
+
+    describe('indeterminate', () => {
```
tlouisse, line 100: can this be merged with the same describe of line 14?

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,125 @@
...
+      await el.updateComplete;
+      expect(el.getAttribute('aria-valuemax')).to.be.eq('70');
+    });
+
+    it('min & max limits value', async () => {
+      const el = await fixture(
+        html` <lion-progress-indicator value="150" aria-label="foo"></lion-progress-indicator> `,
+      );
+      // sets to default max: 100
+      expect(el.getAttribute('aria-valuenow')).to.be.eq('100');
```
tlouisse, line 64: I think all `to.be.eq` in this test can be replaced with `to.equal` (since eq is deep equal (meant for objects/arrays) and 'be' is redundant)

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +125,49 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+      this.indeterminate &&
```
tlouisse, line 167: Maybe do smth like:

```js
onLocaleUpdated() {
   super.onLocaleUpdated();
   
   if (!this.indeterminate) {
     return;
   }
   
   this._setDefaultLabel();  
}

_setDefaultLabel() {
   const hasLabelDefined = this.hasAttribute('aria-label') || this.hasAttribute('aria-labelledby');
  if (hasLabelDefined) {
    return;
  }
  this.setAttribute('aria-label', localize.msg('lion-progress-indicator:loading'));
}

```

docs/components/progress-indicator/overview.md
```diff
@@ -2,19 +2,26 @@
...
-Note: You don't see a live demo here as it would be empty, since there is no styling. Check out the [examples](./examples.md) if you want to see a possible implementation.
+```html
+<lion-progress-indicator aria-label="Interest rate" value="50"></lion-progress-indicator>
+```
 
 ## Features
 
 This component is designed to be extended in order to add visuals.
 
+- Can be indeterminate or determinate, depending if it has a value.
```
tlouisse, line 18:
```suggestion
- Can be indeterminate or determinate, depending on whether it has a value.
```

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,155 @@
...
+      expect(el.getAttribute('aria-label')).to.equal('Loading');
+    });
+
+    it('can override a label with "aria-labelledby"', async () => {
+      const el = await fixture(
+        html` <lion-progress-indicator aria-labelledby="foo-id"></lion-progress-indicator> `,
+      );
+      await el.localizeNamespacesLoaded;
+      expect(el.getAttribute('aria-labelledby')).to.equal('foo-id');
+      // expect(el.hasAttribute('aria-label')).to.be.false;
```
tlouisse, line 45: can this be enabled, or is it true? (otherwise aria-label would take precedence in computing the accessible name)

tlouisse: > can this be enabled, or is it true? (if it is still there, aria-label would take precedence in computing the accessible name)

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,155 @@
...
+      );
+      // sets to default max: 100
+      expect(el.getAttribute('aria-valuenow')).to.equal('100');
+      el.value = -20;
+      await el.updateComplete;
+      // sets to default min: 0
+      expect(el.getAttribute('aria-valuenow')).to.equal('0');
+    });
+  });
+
```
tlouisse, line 106: I think we only need a test with:
- 
- a TODO for aria-valuetext => allow a function like `_computeAriaValueText({ value:number; min:number; max:number;}): string;`

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,155 @@
...
+      expect(el.hasAttribute('aria-label')).to.be.false;
+      el.removeAttribute('aria-labelledby');
+      await el.updateComplete;
+      expect(el.hasAttribute('aria-labelledby')).to.be.false;
+      expect(el.getAttribute('aria-label')).to.equal('Loading');
+    });
+  });
+
+  describe('determinate', async () => {
+    it('is determinate when has value attribute', async () => {
```
tlouisse, line 57:
```suggestion
    it('is determinate when has a value', async () => {
```
(because .value prop is SSOT, and attr is a reflection of that or means to set SSOT)

tlouisse (approved):
Looks good. Would be nice to commit the suggestions and add tests for:

I think we only need tests with:

- changing indeterminate state should remove aria-label as well
- a TODO for aria-valuetext => allow a function like _computeAriaValueText({ value:number; min:number; max:number;}): string;

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +130,64 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+          this.value = this.min;
+          this.setAttribute('aria-valuenow', this.min.toString());
+        } else if (this.value > this.max) {
+          this.value = this.max;
+          this.setAttribute('aria-valuenow', this.max.toString());
+        } else {
+          this.setAttribute('aria-valuenow', this.value.toString());
+        }
+        if (this.__hasDefaultLabelSet === true) {
+          this.removeAttribute('aria-label');
```
tlouisse, line 159:
```suggestion
          this.removeAttribute('aria-label');
          this.__hasDefaultLabelSet = false;
```

And then a test that makes sure that

packages/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -78,12 +130,64 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
...
+   */
+  updated(changedProperties) {
+    super.updated(changedProperties);
+
+    if (this.indeterminate) {
+      if (changedProperties.has('_ariaLabel') || changedProperties.has('_ariaLabelledby')) {
+        this._setDefaultLabel();
+      }
+    } else {
+      if (changedProperties.has('value')) {
```
tlouisse, line 148: This can also mean value just became undefined, right? And then it would be indeterminate and the checks below would fail. 

It works of course since indeterminate is a static getter that reads the value, but it's hard to read/comprehend 😅 
And would most likely break if we would update status of indeterminate asynchronously (c.q. inside the update cycle).

packages/progress-indicator/test/lion-progress-indicator.test.js
```diff
@@ -1,23 +1,187 @@
...
+    });
+
+    it('keeps own aria-label when switch to determinate state', async () => {
+      const el = await fixture(
+        html` <lion-progress-indicator aria-label="foo"></lion-progress-indicator> `,
+      );
+      expect(el.getAttribute('aria-label')).to.equal('foo');
+      el.setAttribute('value', '30');
+      await el.updateComplete;
+      expect(el.getAttribute('aria-label')).to.equal('foo');
```
tlouisse, line 71:
```suggestion
      expect(el.getAttribute('aria-label')).to.equal('foo');
      
      el.removeAttribute('aria-label');
      await el.updateComplete;
      expect(el.getAttribute('aria-label')).to.equal('Loading');
```

packages/progress-indicator/src/LionProgressIndicator.js
gerjanvangeest: You prefer to change it to:

```js
if (changedProperties.has('value') && typeof this.value === 'number') {
```

tlouisse (approved):

PR #1725, opened 2022-06-07
fix(form-core): maintain correct formElements order with non-form elements in dom

daKmoR:
## What I did

1. Make sure form elements are added in the correct order even when non-form elements are in between.


packages/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -71,6 +71,28 @@ export const runRegistrationSuite = customConfig => {
       expect(el.formElements.length).to.equal(1);
     });
 
+    it('maintains dom order even if there are arbitrary none-form elements between registering elements', async () => {
+      const el = /** @type {RegistrarClass} */ (
+        await fixture(html`
+          <${parentTag}>
+            <${childTag} pos="1"></${childTag}>
+            <something-other></something-other>
```
tlouisse, line 79: does this also take into account lazy loading of children?

daKmoR: adding the element "later" is like "lazy loading"?

or what do you mean?

tlouisse: if you have:
```
<child-1/>
<child-2/>
<child-3/>
```
And they upgrade (meaning .isFormElement true and firing their reg events) in order 2,3,1...
will .formElements order then be [1,2,3]?

tlouisse (approved):

PR #1726, opened 2022-06-07
fix(core): use partial renderOptions in SlotMixin (for Safari)

tlouisse:
Makes sure that safari doesn't throw an error when ScopedElementsMixin renderOptions are used for "offline" render of nodes.
(was not possible to test this without loading the polyfill)


daKmoR:
strange - but ok... it's safari 🤷‍♂️

daKmoR (approved):

PR #1729, opened 2022-06-12
feat: toast message component

vahidm90:
## What I did

Using the Overlays mixin, I added a toast component as requested in https://github.com/ing-bank/lion/discussions/1274

## Todos
1. a better approach for dismissing toasts
2. tests


gerjanvangeest:

PR #1733, opened 2022-06-14
Fix/input tel format options

gerjanvangeest:
## What I did

1. Substract and export getFlagSymbol function
2. Add option to style the country-code with parentheses in the formatter
3. Remove unwanted characters in input-tel parser

### DIFF
--- a/packages/input-tel/src/parsers.js
+++ b/packages/input-tel/src/parsers.js
@@ -18,10 +18,11 @@ export function parsePhoneNumber(viewValue, { regionCode }) {
 
   // eslint-disable-next-line prefer-destructuring
   const PhoneNumber = /** @type {AwesomePhoneNumber} */ (PhoneUtilManager.PhoneUtil);
-
+  const regex = /[+0-9]+/gi;
+  const strippedViewValue = viewValue.match(regex)?.join('');
   let pn;
   try {
-    pn = PhoneNumber(viewValue, regionCode);
+    pn = PhoneNumber(strippedViewValue, regionCode);
     // eslint-disable-next-line no-empty
   } catch (_) {}
 
--- a/packages/input-tel/test/preprocessors.test.js
+++ b/packages/input-tel/test/preprocessors.test.js
@@ -16,7 +16,7 @@ describe('liveFormatPhoneNumber', () => {
         prevViewValue: '+36123',
         currentCaretIndex: 2,
       }),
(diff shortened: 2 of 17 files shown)


packages/input-tel/src/parsers.js
```diff
@@ -18,10 +18,11 @@ export function parsePhoneNumber(viewValue, { regionCode }) {
 
   // eslint-disable-next-line prefer-destructuring
   const PhoneNumber = /** @type {AwesomePhoneNumber} */ (PhoneUtilManager.PhoneUtil);
-
+  const regex = /[+0-9]+/gi;
```
tlouisse, line 21: Nice.

Only having a quick look atm, but do we still catch all cases, like:

pn.getNumber( 'national' );        // -> '070-712 34 56'
pn.getNumber( 'rfc3966' );         // -> 'tel:+46-70-712-34-56'

packages/input-tel/test/preprocessors.test.js
```diff
@@ -29,4 +29,54 @@ describe('liveFormatPhoneNumber', () => {
       }),
     ).to.eql({ caretIndex: 12, viewValue: '+31 6 12345678' });
   });
+
+  describe('with formatCountryCodeStyle is set to parantheses', () => {
```
tlouisse, line 33:
```suggestion
  describe('with formatCountryCodeStyle is set to parentheses', () => {
```

packages/input-tel/src/parsers.js
gerjanvangeest: I added the following tests:

it('removes unwanted characters', () => {
    expect(parsePhoneNumber('(+31)707123456', { regionCode: 'NL' })).to.equal('+31707123456');
    expect(parsePhoneNumber('+31 70 7123456', { regionCode: 'NL' })).to.equal('+31707123456');
    expect(parsePhoneNumber('+31-70-7123456', { regionCode: 'NL' })).to.equal('+31707123456');
    expect(parsePhoneNumber('+31|70|7123456', { regionCode: 'NL' })).to.equal('+31707123456');
  });
  
  Could add one extra for `rfc3966`

tlouisse: Most likely catches it like this, but would be nice to have the test for rfc3966 indeed

tlouisse (approved):

PR #1745, opened 2022-07-11
fix: fixed focus handling, focus now jumps from header to panel when …

DannyMoerkerke:
…tabbing

## What I did

1. after the slotted invokers and content sections are set up they are reordered into `<div class="accordion">` to fix the focus handling. Focus now jumps from header to panel.

todo: fix tests

### DIFF
--- a/packages/accordion/src/LionAccordion.js
+++ b/packages/accordion/src/LionAccordion.js
@@ -61,6 +61,25 @@ export class LionAccordion extends LitElement {
           visibility: visible;
           display: block;
         }
+
+        .accordion [slot='invoker'] {
+          margin: 0;
+        }
+
+        .accordion [slot='invoker'][expanded] {
+          font-weight: bold;
+        }
+
+        .accordion [slot='content'] {
+          margin: 0;
+          visibility: hidden;
+          display: none;
+        }
+
+        .accordion [slot='content'][expanded] {
+          visibility: visible;
+          display: block;
+        }
       `,
     ];
   }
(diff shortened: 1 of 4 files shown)


packages/accordion/src/LionAccordion.js
```diff
@@ -180,12 +202,33 @@ export class LionAccordion extends LitElement {
...
+   *  @private
+   */
+  __rearrangeInvokersAndContent() {
+    const invokers = /** @type {HTMLElement[]} */ (
+      Array.from(this.querySelectorAll('[slot="invoker"]'))
+    );
+    const contents = /** @type {HTMLElement[]} */ (
+      Array.from(this.querySelectorAll('[slot="content"]'))
+    );
+    const accordion = this.shadowRoot?.querySelector('slot[name=accordion]');
```
tlouisse, line 222: why would/could this be null? Maybe you can do `/** @type {ShadowRoot} */ (this.shadowRoot)`

DannyMoerkerke: I can but I followed the convention in this file which is currently `this.shadowRoot?`.

packages/accordion/src/LionAccordion.js
```diff
@@ -43,21 +43,40 @@ export class LionAccordion extends LitElement {
...
           margin: 0;
           visibility: hidden;
           display: none;
         }
 
-        .accordion ::slotted([slot='content'][expanded]) {
+        .accordion ::slotted([slot='accordion'][expanded]) {
+          visibility: visible;
+          display: block;
+        }
```
gerjanvangeest, line 63: All of this styling  can be gone if i'm correct.

packages/accordion/src/LionAccordion.js
```diff
@@ -129,6 +148,7 @@ export class LionAccordion extends LitElement {
       <div class="accordion">
         <slot name="invoker"></slot>
         <slot name="content"></slot>
+        <slot name="accordion"></slot>
```
gerjanvangeest, line 151: Am i correct that this slot should not be used by our users, so its basically a protected thing. Maybe call it `_accordion`?

packages/accordion/src/LionAccordion.js
DannyMoerkerke: Correct, I removed it.

packages/accordion/src/LionAccordion.js
DannyMoerkerke: Updated.

gerjanvangeest (approved):

Issue #1753, opened 2022-08-04
[form][validation] getMessage() returns only a Promise


AndreeaMoise:
### Expected behaviour

I can use the Validator getMessage as exemplified in the lion docs to return a string: `Required.getMessage = () => 'Please enter a value';`

### Actual behaviour
Because of the function's signature `static getMessage(data?: Partial<import("../../types/validate").FeedbackMessageData> | undefined): Promise<string | Element>;`, I can only return a promise.

tlouisse:
Hey @AndreeaMoise.

So your goal is to change the type to `Partial<import("../../types/validate").FeedbackMessageData> | undefined): Promise<string | Element> string | Element;`

Makes sense to support this api. Do you want to do a pull request for this?

tlouisse:
closed by https://github.com/ing-bank/lion/pull/1756

PR #1756, opened 2022-08-10
fix: getMessage return type supports string and Element

AndreeaMoise:
## What I did

I have extended the return type of getMessage in the validator to support string and Element types.


gerjanvangeest:
Thanks for the fix 💪 

fix: https://github.com/ing-bank/lion/issues/1753

gerjanvangeest (approved):

PR #1770, opened 2022-09-01
fix(LionSwitchButton): do not dispatch checked-change event 

jdevega:
I have a use case where I need to set the initial value for the switch and the event should not be triggered. As the behavior to not trigger the event is bounded to the disabled attribute, I've used the same approach on the requestUpdate method.
Not sure if there is another better approach to set the initial checked value without triggering the event.

Cheers.


gerjanvangeest (approved):

tlouisse:
Nice! Thanks

PR #1776, opened 2022-09-09
fix(form-core) sync autofocus to focusable node (fixes #1775)

### LINKED ISSUE #1775: [form-core] Lion form field components do not receive autofocus in dialog
## Expected behavior

When adding a lion-input with autofocus to a lion-dialog, on opening the dialog it should put initial focus on the lion-input.

See the following [code-example](https://studio.webcomponents.dev/edit/oCJHLBB0uzh4K3uas5Au/README.md?p=README.md) where the fourth button highlight the autofocus bug.

## Actual Behavior

The initial focus goes to the dialog element.

## Additional context

Package form-core version 0.18.3.
This bug was found while trying to find a workaround for a [bug in safari with the dialog](https://github.com/ing-bank/lion/issues/1773).

### Cause of the issue
The issue is caused by the implementation of intial focus in [contain-focus.js](https://github.com/ing-bank/lion/blob/650657231a4ec83592d45cd69836d28436635340/packages/overlays/src/utils/contain-focus.js#L74) where intial focus is set to the first focuasable element with the autofocus attribute, or the dialog element. While this is a fine implementation for native html elements, it fails when using lion form field components as from the outside it is hard to set autofocus on the input element. Setting autofocus on the component does not work as the component itself is not focusable, so is not accounted for when determining inital focus.

### Use case
While inappropriate use of the autofocus attribute can lead to bad user experience, especially for screen reader users. There are use cases for the dialog, like a confirmation dialog, where it can improve the user experience by immediately focussing the primary button, or a dialog containing a form where the first form field is focused. As long as the text content above the button/form field is linked to the dialog via aria-describedby this as a valid and accessible way to setup a dialog.

Sciurus7:
## What I did

1. In FocusMixin synced the autofocus attribute from the host to the focusable node.
2. Added tests to ensure the sync happens if autofocus was set on the host during render or when autofocus is added or removed from the host.

### DIFF
--- a/packages/form-core/src/FocusMixin.js
+++ b/packages/form-core/src/FocusMixin.js
@@ -27,6 +27,7 @@ const FocusMixinImplementation = superclass =>
       return {
         focused: { type: Boolean, reflect: true },
         focusedVisible: { type: Boolean, reflect: true, attribute: 'focused-visible' },
+        autofocus: { type: Boolean, reflect: true }, // Required in Lit to observe autofocus
       };
     }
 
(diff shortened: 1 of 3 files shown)


gerjanvangeest:
Hi @Sciurus7, thanks for the contribution! Somehow i missed this one completely, my excuses.
Could you resolve the conflicts and add a changeset (`npm run changeset`)?

fix: https://github.com/ing-bank/lion/issues/1775

Sciurus7:
Hi Gerjan, I finally managed to get my personal git account in a working state again. While trying to bring this pull request up to date I messed up a merge, so instead of trying to fix that particular mess, I branched of master again added the changes to an up to date branch and created a [new pull request](https://github.com/ing-bank/lion/pull/1963). This one should be up to date and has a change set. I will close this pull request accordingly.

PR #1777, opened 2022-09-14
Feat/providence update

tlouisse:
## What I did

- cleanup
- export map support
- 'custom project' support
- getSourceCodeFragmentOfDeclaration
- expose more via exports

### DIFF
--- a/packages-node/providence-analytics/src/program/analyzers/helpers/transform-into-iterable-find-exports-output.js
+++ b/packages-node/providence-analytics/src/program/analyzers/helpers/transform-into-iterable-find-exports-output.js
@@ -1,0 +1,67 @@
+/**
+ * @typedef {import('../../types/analyzers').FindExportsAnalyzerResult} FindExportsAnalyzerResult
+ */
+
+/**
+ * Convert to more easily iterable object
+ *
+ * From:
+ * ```js
+ * [
+ *  "file": "./file-1.js",
+ *  "result": [{
+ *    "exportSpecifiers": [ "a", "b"],
+ *    "localMap": [{...},{...}],
+ *    "source": null,
+ *    "rootFileMap": [{"currentFileSpecifier": "a", "rootFile": { "file": "[current]", "specifier": "a" }}]
+ *  }, ...],
+ * ```
+ * To:
+ * ```js
+ * [{
+ *   "file": ""./file-1.js",
+ *   "exportSpecifier": "a",
+ *   "localMap": {...},
+ *   "source": null,
+ *   "rootFileMap": {...}
+ * },
+ * {{
+ *   "file": ""./file-1.js",
+ *   "exportSpecifier": "b",
+ *   "localMap": {...},
+ *   "source": null,
+ *   "rootFileMap": {...}
+ * }}],
+ *
+ * @param {FindExportsAnalyzerResult} exportsAnalyzerResult
+ */
+function transformIntoIterableFindExportsOutput(exportsAnalyzerResult) {
+  /** @type {IterableFindExportsAnalyzerEntry[]} */
+  const iterableEntries = [];
+
+  for (const { file, result } of exportsAnalyzerResult.queryOutput) {
+    for (const { exportSpecifiers, source, rootFileMap, localMap, meta } of result) {
+      if (!exportSpecifiers) {
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+      for (const exportSpecifier of exportSpecifiers) {
+        const i = exportSpecifiers.indexOf(exportSpecifier);
+        /** @type {IterableFindExportsAnalyzerEntry} */
+        const resultEntry = {
+          file,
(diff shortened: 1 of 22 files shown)


packages-node/providence-analytics/src/program/analyzers/helpers/transform-into-iterable-find-exports-output.js
```diff
@@ -0,0 +1,67 @@
...
+
+  for (const { file, result } of exportsAnalyzerResult.queryOutput) {
+    for (const { exportSpecifiers, source, rootFileMap, localMap, meta } of result) {
+      if (!exportSpecifiers) {
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+      for (const exportSpecifier of exportSpecifiers) {
+        const i = exportSpecifiers.indexOf(exportSpecifier);
+        /** @type {IterableFindExportsAnalyzerEntry} */
```
narzac, line 50: i could be -1

packages-node/providence-analytics/src/program/services/InputDataService.js
```diff
@@ -205,6 +205,33 @@ function multiGlobSync(patterns, { keepDirs = false, root } = {}) {
...
+  }
+  return localPathWithoutDotSlash;
+}
+
+/**
+ * @param {{val:object|string;nodeResolveMode:string}} opts
+ * @returns {string}
+ */
+function getStringOrObjectValOfExportMapEntry({ val, nodeResolveMode, packageRootPath }) {
+  if (typeof val !== 'object') {
```
narzac, line 224: typeof null === 'object'

packages-node/providence-analytics/src/program/services/InputDataService.js
```diff
@@ -215,22 +242,30 @@ function multiGlobSync(patterns, { keepDirs = false, root } = {}) {
...
-        name: pathLib.basename(projectPath),
-        path: projectPath,
-      }),
-      entries: this.gatherFilesFromDir(projectPath, {
-        ...this.defaultGatherFilesConfig,
-        ...gatherFilesConfig,
-      }),
-    }));
+    const inputData = projectPaths.map(projectPathOrObj => {
+      if (typeof projectPathOrObj === 'object') {
```
narzac, line 252: typeof null === 'object'

packages-node/providence-analytics/src/program/utils/get-source-code-fragment-of-declaration.js
```diff
@@ -0,0 +1,98 @@
...
+ *    - is it a non ref declaration? Return the path of the node
+ * @param {{ referencedIdentifierName:string, globalScopeBindings:BabelBinding; }} opts
+ * @returns {BabelNodePath}
+ */
+function getReferencedDeclaration({ referencedIdentifierName, globalScopeBindings }) {
+  const [, refDeclaratorBinding] = Object.entries(globalScopeBindings).find(
+    ([key]) => key === referencedIdentifierName,
+  );
+
+  if (refDeclaratorBinding.path.node.init.type === 'Identifier') {
```
narzac, line 25: Maybe do below instead:
```js
const referencedIdentifierName = refDeclaratorBinding.path.node.init.name;
if (referencedIdentifierName) {
  return getReferencedDeclaration({ referencedIdentifierName, globalScopeBindings });
}
```
And if it is not guaranteed to deference `refDeclaratorBinding.path.node.init.name`, better to use `?.` to access properties

packages-node/providence-analytics/src/program/services/InputDataService.js
tlouisse: yeah indeed, we assume people to not have smth like `"exports": { "./": null }` in their export map. We could consider check general validity of exports obj before this method?

packages-node/providence-analytics/src/program/services/InputDataService.js
tlouisse: Yeah, we could validate the input again then. It would happen if someone does not comply with the supported type (string or {})

packages-node/providence-analytics/src/program/utils/get-source-code-fragment-of-declaration.js
tlouisse: At this point we can trust that `refDeclaratorBinding.path.node.init` exists (it has a name in case it's an Identifier)

packages-node/providence-analytics/src/program/services/InputDataService.js
tlouisse: Will add an input validation later 👍

packages-node/providence-analytics/src/program/services/InputDataService.js
tlouisse: Will add an input validation later 👍

narzac (approved):

PR #1779, opened 2022-09-16
Fix: calendar translation de.js strings corrected

vssn:
## What I did

1. Updated the translation strings for calendar. It should use the correct labelling of the year navigation buttons

Translation reference:
https://translate.google.de/?sl=en&tl=de&text=previous%20year%3B%0Anext%20year&op=translate


tlouisse (approved):

tlouisse:
Thanks!

PR #1788, opened 2022-10-07
chore: update dependency to fix "node16" module resolution

samuelstroschein:
Importing lion components when the module resolution of a project to "node16" (ESM) crashes node. 

Node can't find the exports of @open-wc/scoped-elements. The issue has been fixed in @open-wc/scoped-elements 2.0.2, see https://github.com/open-wc/open-wc/pull/2454

## What I did

1. Update the @open-wc/scoped-elements dependency in core to 2.0.3.


gerjanvangeest:
This PR is outdated, so i'll close it.

Issue #1791, opened 2022-10-11
[input-amount] removing digits on blur, when maximumFractionDigits is set to 0


gerjanvangeest:
I need the <ing-input-amount> tag, but without decimals and with a 12 digits max. In order to achieve this, I encountered an issue: while add some digits in input, delete one or two and, after that, adding one or two digits again, the input erase a group of three, four or event six digits once, after blur. 
 
Code here:

```html
 <lion-input-amount
   label="label"
   .validators="${[new Required(), new MaxNumber(999999999999)]}"
   currency='EUR'
   .formatOptions=${{
      minimumFractionDigits: 0,
      maximumFractionDigits: 0,
      locale: 'nl-BE'
}}
</lion-input-amount>
```

PR #1801, opened 2022-10-24
fix(localize): fix and simplify parseMode check

gerjanvangeest:
## What I did

1. fix: https://github.com/ing-bank/lion/issues/1791
2. simplify code of getParseMode()
3. make jsdoc and tests aligned


packages/localize/src/number/parseNumber.js
```diff
@@ -1,56 +1,27 @@
...
- * - else parse mode depends mostly on the last 4 chars
- * - 1234 => xxx1234 (heuristic)
- * - [space]123 => xxx123 (heuristic)
- * - ,123 => unclear
- *   - if 1.000,123 (we find a different separator) => 1000.123 (heuristic)
- *   - if 1,000,123 (we find only same separators) => 1000123 (unparseable)
- *   - if 100,123 (we find no more separators) => unclear
- *     - if en => 100123 (withLocale)
- *     - if nl => 100.123 (withLocale)
- *
```
tlouisse, line 25: Would it be possible to update this to the new outcomes? It's quite important for understanding all this

tlouisse: Also this needs to be updated? https://github.com/ing-bank/lion/blob/42f7026b5883e276859060008c4af15be79e20d8/packages/localize/src/number/parseNumber.js#L114

gerjanvangeest: This list was very unreadable, while the description of the parseNumber was much easier to understand.
Both descriptions were not inline with the old code, but are inline with the updated code.

packages/localize/src/number/parseNumber.js
```diff
@@ -1,56 +1,27 @@
...
  * See also {@link parseAmount}
```
gerjanvangeest, line 5:
```suggestion
* It has 3 "methods" of returning numbers
 * - 'unparseable': becomes just numbers
 * - 'withLocale': result depends on given or global locale
 * - 'heuristic': result depends on considering separators
 *
 * @example
 * parseNumber('1.234.567'); // method: unparseable => 1234567
 * parseNumber('1.234'); // method: withLocale => depending on locale 1234 or 1.234
 * parseNumber('1.234,56'); // method: heuristic => 1234.56
 * parseNumber('1 234.56'); // method: heuristic => 1234.56
 * parseNumber('1,234.56'); // method: heuristic => 1234.56
```

tlouisse: Looks good. Will approve the PR once it's rebased with the 'lion ui structure'.

.changeset/plenty-apples-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/localize': patch
```
gerjanvangeest, line 2: Does this need to change?

tlouisse: Yeah, should be `@lion/ui` indeed, with maybe `[localize]` in front of the description

.changeset/plenty-apples-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
```
tlouisse, line 2:
```suggestion
'@lion/ui': minor
```

tlouisse (approved):

PR #1802, opened 2022-10-24
feat: move all web pkgs into a single npm pkg @lion/ui

daKmoR:
## What I did

1. With export maps we can get the same runtime cost by using multiple entrypoints as with separate packages - It however simplifies usage and maintenance a LOT. 
2. This introduces a single `@lion/ui` package containing all web components.
3. `@lion/ajax` and `singleton-manager` will remain separate packages

PS: this also requires changes to our TS setup which will follow next...

PR #1804, opened 2022-10-25
safari mac fix for scrolling issues

iceicerickert:
## What I did

1. added the fix for scrolling issues on safari for mac too, the problems also occur on macs when only using your touchpad.

had contact with:
@tlouisse

### DIFF
--- a/.changeset/tiny-lamps-melt.md
+++ b/.changeset/tiny-lamps-melt.md
@@ -1,5 +1,5 @@
 ---
-"@lion/calendar": patch
+'@lion/calendar': patch
 ---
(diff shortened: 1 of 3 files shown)


tlouisse:
Nice. Thanks a lot. After this PR is merged to master, we will rebase it and add a changeset: https://github.com/ing-bank/lion/pull/1802

tlouisse:
Thanks again for committing this. Since we did an update of our component structure, I rebased it here (same contents still, with a test on top): https://github.com/ing-bank/lion/pull/1830

PR #1808, opened 2022-11-04
feat: introduce wireit and ts use node16 resolution

daKmoR:
## What I did

1. Add [wireit](https://github.com/google/wireit)
2. Update Typescript to the latest version and use `moduleResolution: "NodeNext"`
3. Types are now defined as `*.ts` files
4. Each package builds their own types via `npm run types`
5. Types are built into `dist-types` which is added to the exports via 
   ```
    "./*": {
      "types": "./dist-types/exports/*",
      "default": "./exports/*"
    },
    ```
6. Needed to patch [awesome-phonenumber](https://github.com/grantila/awesome-phonenumber/issues/86) to add a types export
7. Needed to patch `@open-wc/scoped-elements` to add types export
8. Types for `@lion/ajax` needed to be adjusted


.changeset/red-pugs-promise.md
```diff
@@ -0,0 +1,32 @@
+---
+'@lion/ui': patch
+---
+
+This introduces a new package `@lion/ui` which is a collection of UI components that can be used in your application. It contains all the components/system that used to be distributed via separate `@lion/*` packages.
```
tlouisse, line 5:
```suggestion
This introduces a new package `@lion/ui` which is a collection of UI components that can be used in your application. It contains all the components/systems that used to be distributed via separate `@lion/*` packages.
```

tsconfig.json
```diff
@@ -1,35 +1,40 @@
...
-    "packages-node/providence-analytics/**/*.js",
-    "packages-node/remark-extend/**/*.js"
-  ]
+    "skipLibCheck": true,
+    "incremental": true,
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true
+  }
+  // "include": ["packages/**/*.js", "packages-node/**/*.js"],
```
tlouisse, line 23: should we delete this (and add on package level in future)?

tlouisse (approved):

.changeset/red-pugs-promise.md
daKmoR: this will start a new changelog for a new package... I don't think it makes sense to start a new release with "breaking"...

tsconfig.json
daKmoR: deleted

PR #1809, opened 2022-11-07
fix: relax check for content types when setting cache

thepassle:
## What I did

1. Relaxed the check for content type when checking whether or not a request should be cached

cc @LarsDenBakker


daKmoR:
hey, we just merged a change to improve our way of generating types... https://github.com/ing-bank/lion/pull/1808 which sadly introduced a merge conflict...

could you do a rebase? if not I can take a look later

PR #1827, opened 2022-11-18
fix(overlay-manager): disable transition when reduced motion enabled

jcperea:
## What I did

1. Disables (actually reduces to 1ms) the 300ms opacity transition in the overlay manager when the customer has the reduced motion capability enabled on its browser.

The animation is not removed, but reduced to 1ms, so everything is working as expected, but the customer won't see the transition. Also, this helps the use of visual testing (screenshot testing) when opening an element from the overlay manager suite.

### DIFF
--- a/.changeset/five-dolls-begin.md
+++ b/.changeset/five-dolls-begin.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


tlouisse (approved):

tlouisse:
Nice addition! Thanks

PR #1829, opened 2022-11-18
fix(lion-input-tel): polish/cleanup input-tel(-dropdown)

tlouisse:
## What I did

Polish SlotMixin docs and LionInputTel(Dropdown)

### DIFF
--- a/docs/fundamentals/systems/core/SlotMixin.md
+++ b/docs/fundamentals/systems/core/SlotMixin.md
@@ -4,7 +4,7 @@ The SlotMixin is made for solving accessibility challenges that inherently come 
 Until [AOM](https://wicg.github.io/aom/explainer.html) is in place, it is not possible to create relations between different shadow doms.
 The need for this can occur in the following situations:
 
-1. A user defined slot. For instance:
+- A user-defined slot. For instance:
 
 ```html
 <my-input>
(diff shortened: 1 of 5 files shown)


docs/fundamentals/systems/core/SlotMixin.md
```diff
@@ -4,7 +4,7 @@ The SlotMixin is made for solving accessibility challenges that inherently come
 Until [AOM](https://wicg.github.io/aom/explainer.html) is in place, it is not possible to create relations between different shadow doms.
 The need for this can occur in the following situations:
 
-1. A user defined slot. For instance:
+- A user defined slot. For instance:
```
Knorgias, line 7: Jut a detail:
"user-defined" could be a better way to write it, eg https://en.wikipedia.org/wiki/User-defined_function

docs/fundamentals/systems/core/SlotMixin.md
```diff
@@ -14,7 +14,7 @@ The need for this can occur in the following situations:
 
 The help text here needs to be connected to the input element that may live in shadow dom. The input needs to have `aria-describedby="help-text-id".`
 
-2. An interplay of multiple nested web components. For instance:
+- An interplay of multiple nested web components. For instance:
 
 ```html
 <my-fieldset>
```
Knorgias, line 20: Shouldn;t <div> here have a slot attribute, eg <div slot="msg">Group error message</div> ?

docs/fundamentals/systems/core/SlotMixin.md
```diff
@@ -26,13 +26,16 @@ The help text here needs to be connected to the input element that may live in s
...
-- easily render lit templates
+Edge cases that SlotMixin solves:
+
+- **mediates** between light dom provided by the user ('public slots') and light dom provided by the component author ('private slots').
+- allows to hook into the reactive update loop of LitElement (**rerendering** on property changes)
+- respects the **scoped registry** belonging to the shadow root.
+
+## The api
 
 So, what does the api look like? SlotMixin can be used like this:
```
Knorgias, line 40: An example where `AccessibleControl` is used "in action" ?

tlouisse: It's a fictional control that should indicate that SlotMixin contributes to a11y. We could emphasize this by calling it `MyAccessibleControl` if you want

docs/fundamentals/systems/core/SlotMixin.md
tlouisse: It should be clear that this element lives in light dom. Whether this will have a slot or not is dependent on the implementation detail of my-fieldset (does it have a shadow root? If so, are we dealing with a default or named slot?), which is unknown.

In this example I tried to focus on the part about connecting light dom of a child (an input with aria-describedby="msg") with an element of its parent.
I guess we could make this example clearer like this:

```html
<my-fieldset>
  <my-input><input aria-describedby="parent-msg"></my-input>
  <my-input><input aria-describedby="parent-msg"></my-input>
  <div id="parent-msg">Group errror message</div>
</my-fieldset>
```

docs/fundamentals/systems/core/SlotMixin.md
tlouisse: True, will update it 👍

docs/fundamentals/systems/core/SlotMixin.md
```diff
@@ -120,16 +136,16 @@ For instance, a simple text label is provided like this:
...
   <label slot="label"><my-icon aria-hidden="true"></my-icon>My label</label>
 </my-input>
 ```
 
-- In the property/attribute case, SlotMixin adds the `<label slot="label">` under the hood, **unless** the developer already provided the slot.
-  This will make sure that the slot provided by the user always takes precedence and only one slot instance will be available in light dom per slot.
+In the property/attribute case, SlotMixin adds the `<label slot="label">` under the hood, **unless** the developer already provided the slot.
+This will make sure that the slot provided by the user always takes precedence and only one slot instance will be available in light dom per slot.
 
 ### Default slot
```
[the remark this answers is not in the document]
tlouisse: Indeed, thanks :) Fixed it

Knorgias (approved):

gerjanvangeest (approved):

PR #1857, opened 2022-12-13
Fix/accordion selectors

DannyMoerkerke:
## What I did

1. narrowed the scope of the selectors that query [slot=invoker] and [slot=content] to prevent that any nested elements with [slot=invoker] and [slot=content] are moved to slot=_accordion as well

### DIFF
--- a/.changeset/strange-dots-draw.md
+++ b/.changeset/strange-dots-draw.md
@@ -2,4 +2,4 @@
 '@lion/ui': patch
 ---
 
-narrowed the scope of the selectors that query [slot=invoker] and [slot=content] to prevent that any nested elements with [slot=invoker] and [slot=content] are moved to slot=\_accordion as well
+`accordion`: narrowed the scope of the selectors that query [slot=invoker] and [slot=content] to prevent that any nested elements with [slot=invoker] and [slot=content] are moved to slot=\_accordion as well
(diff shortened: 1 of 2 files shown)


packages/ui/components/accordion/src/LionAccordion.js
```diff
@@ -210,10 +210,10 @@ export class LionAccordion extends LitElement {
    */
   __rearrangeInvokersAndContent() {
     const invokers = /** @type {HTMLElement[]} */ (
-      Array.from(this.querySelectorAll('[slot="invoker"]'))
+      Array.from(this.querySelectorAll(':scope > [slot="invoker"]'))
```
daKmoR, line 213: just as a consideration - in tabs we filter on children... potentially faster
https://github.com/ing-bank/lion/blob/master/packages/ui/components/tabs/src/LionTabs.js#L170

.changeset/strange-dots-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+narrowed the scope of the selectors that query [slot=invoker] and [slot=content] to prevent that any nested elements with [slot=invoker] and [slot=content] are moved to slot=\_accordion as well
```
daKmoR, line 5: Should start with the element e.g.

```suggestion
`accordion`: narrowed the scope of the selectors that query [slot=invoker] and [slot=content] to prevent that any nested elements with [slot=invoker] and [slot=content] are moved to slot=\_accordion as well
```

daKmoR (approved):

Issue #1478, opened 2021-08-12
parseDate() returns undefined with valid dates, depending on time & timezone


nathangray:
## Expected behavior

Using locale format "year-month-day", if I call parseDate("2021/09/01"), it should return a Date, that is more or less equal to "2021-09-01"

import { formatDate, parseDate } from '@lion/localize';

console.log(parseDate("2021/09/01"));

`Tue Aug 31 2021 18:00:00 GMT-0600 (Mountain Daylight Time)`

## Actual Behavior

I get undefined

## Additional context

I'm trying this in the afternoon, in Mountain Daylight Time (UTC -6) so when the test is done, UTC time is a different day from local time. 

https://github.com/ing-bank/lion/blob/master/packages/localize/src/date/parseDate.js#L66
I think what is happening is due to my timezone, when the date is checked in line 66 there is a difference in date between local and UTC.  
Unfortunately changing that check to parsedDate.getUTCDate() and parsedDate.getUTCMonth() causes the tests to fail, and I'm not sure how to deal with that or if this is the correct solution.

mauricioaraldi:
I also am facing this issue. Any timezone that has a negative timezone will actually break, based on the [line](https://github.com/ing-bank/lion/blob/5d30258dd6d16bf8c777841509037aed284bc8c5/packages/ui/components/localize/src/date/parseDate.js#L66) that you sent , because the parsed date will be one day less than the original string date.

PR #1866, opened 2022-12-29
fix(localize): don't set modelValue to unparseable if date has a negative timezone

gerjanvangeest:
## What I did

fix: https://github.com/ing-bank/lion/issues/1478

I tested the suggested solution inside the input-datepicker with MinMaxDate validator. But then a day before the min date was approved. With the switch to locale DateString the min date worked as expected.
Only i had a hard time to get a working test for this. I tried `.formatOptions="${{ timeZone: "America/New_York" }}"`, but this changed the modelValue date (and others) to the day before.

### DIFF
--- a/packages/ui/components/localize/src/date/parseDate.js
+++ b/packages/ui/components/localize/src/date/parseDate.js
@@ -58,7 +58,7 @@ export function parseDate(dateString) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
-  const parsedDate = new Date(Date.UTC(year, month - 1, day));
+  const parsedDate = new Date(new Date(year, month - 1, day));
 
   // Check if parsedDate is not `Invalid Date` or that the date has changed (e.g. the not existing 31.02.2020)
   if (
--- a/packages/ui/components/localize/test/date/parseDate.test.js
+++ b/packages/ui/components/localize/test/date/parseDate.test.js
@@ -5,13 +5,13 @@ import { localizeTearDown } from '@lion/ui/localize-test-helpers.js';
 /**
  *
  * @param {Date | undefined} value
- * @param {Date} date
+ * @param {Date | undefined} date
  */
 function equalsDate(value, date) {
   return (
     Object.prototype.toString.call(value) === '[object Date]' && // is Date Object
     value &&
-    value.getDate() === date.getDate() && // day
+    value.getDate() === date?.getDate() && // day
     value.getMonth() === date.getMonth() && // month
     value.getFullYear() === date.getFullYear() // year
   );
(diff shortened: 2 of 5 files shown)


packages/ui/components/localize/src/date/parseDate.js
```diff
@@ -56,7 +56,7 @@ export function parseDate(dateString) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
-  const parsedDate = new Date(Date.UTC(year, month - 1, day));
+  const parsedDate = new Date(new Date(year, month - 1, day).toLocaleDateString());
```
tlouisse, line 59: Should we align with the locale defined locally in `localize` here, instead of browser defaults?
I most cases, they are the same, but they could be different on purpose. And it would probably be more consistent with the rest of the code in localize.

So that we have: `new Date(new Date(year, month - 1, day).toLocaleDateString(localize.locale))` ?

packages/ui/components/localize/test/date/parseDate.test.js
```diff
@@ -51,6 +51,57 @@ describe('parseDate()', () => {
...
+    function convertTZ(date, tzString) {
+      if (date === undefined) {
+        return undefined;
+      }
+      return new Date(date.toLocaleString('en-US', { timeZone: tzString }));
+    }
+
+    expect(
+      equalsDate(
+        convertTZ(parseDate('01-01-1979'), 'Asia/Manila'),
```
tlouisse, line 69: If parseDate does `new Date(...)` under the hood, these tests seem to always pass I think, since the timezone info is only added in the test atm.

tlouisse: Imho this test still gives a false sense of reliability. Unfortunately it's just not possible to mock the timezone of the browser. Now we have a situation where `equalsDate(parseDate('01-01-1979'), new Date('1979/01/01'))` passes and then we apply the exact same timezone transformation to both dates and test for equality... (which will also pass, but what are we testing then exactly ? 🤔 )

We probably want to add `parseDate(x).getTimezoneOffset === Intl.DateTimeFormat().resolvedOptions().timeZone` as a test instead.
(So we **always use the timezone from the browser, regardless of `localize` config**)

packages/ui/components/localize/src/date/parseDate.js
tlouisse: This has become `const parsedDate = new Date(new Date(year, month - 1, day));`

So we always use the browser locale, for two reasons:
- it's just the most logical for the end user. If it's 9pm, he/she doesn't want to see the date of tomorrow, because the locale is configured in a certain way 
- there's no 1-1 mapping from locale to timezone (a locale/geographical area could have multiple timezones)

DannyMoerkerke (approved):

tlouisse (approved):

PR #1879, opened 2023-01-10
fix(select-rich): set focusableNode correctly so focused and focused-…

gerjanvangeest:
…visible attributes are set when invoker gets focus

## What I did

1.

### DIFF
--- a/packages/ui/components/select-rich/src/LionSelectRich.js
+++ b/packages/ui/components/select-rich/src/LionSelectRich.js
@@ -86,6 +86,15 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
   }
 
   /**
+   * @protected
+   * @configure FocusMixin
+   */
+  // @ts-ignore
+  get _focusableNode() {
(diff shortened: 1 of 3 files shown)


packages/ui/components/select-rich/src/LionSelectRich.js
```diff
@@ -85,6 +85,15 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     );
   }
 
+  /**
+   * @protected
+   * @configure FocusMixin
+   */
+  // @ts-ignore
+  get _focusableNode() {
```
DannyMoerkerke, line 93: Why `get _focusableNode` when the element is named `_invokerNode`?

gerjanvangeest: The `_focusNode` is used inside the `FocusMixin`. For all inputs this corresponds with the `_inputNode`, only in case of the SelectRich it has to be the `_invokerNode`, since that is the element that will get focus.

daKmoR (approved):

PR #1881, opened 2023-01-11
Feat/input file

gerjanvangeest:
## What I did

1.


packages/ui/components/input-file/translations/bg.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  ALLOWED_FILE_SIZE: 'Моля, качете файл с размер, не по-голям от {params}.',
```
tlouisse, line 2: Is there a reason we have a different naming convention (instead of camelCased) here?

packages/ui/components/form-core/types/validate/ValidateMixinTypes.ts
```diff
@@ -9,9 +9,9 @@ import { FormControlHost } from '../FormControlMixinTypes.js';
 import { SyncUpdatableHost } from '../utils/SyncUpdatableMixinTypes.js';
 import { Validator } from '../../src/validate/Validator.js';
 
-type FeedbackMessage = {
+export type FeedbackMessage = {
   message: string | Node;
-  type: string;
+  type: ValidationType;
```
tlouisse, line 14: Someone extending lion should still be able to specify a type beyond error/warning/info/success, so I'm not 100% sure about this...

packages/ui/components/input-file/translations/bg.js
gerjanvangeest: Sole reason that the codebase i used had this already. I will turn it to camelCase :)

packages/ui/components/input-file/src/FileHandle.js
```diff
@@ -0,0 +1,84 @@
+import { FileValidation } from './validators.js';
+
+/**
+ * @typedef {import('../types/input-file.js').SystemFile} SystemFile
+ */
+
+export const MAX_FILE_SIZE = 117654604;
```
tlouisse, line 7: Is there a particular reason for this limit?

packages/ui/components/input-file/src/FileHandle.js
```diff
@@ -0,0 +1,84 @@
+import { FileValidation } from './validators.js';
```
tlouisse, line 1: we don't use the convention of capitalizing constant names elsewhere in repo, so it would be nicer if we used the same convention (just camelCase) here

packages/ui/components/input-file/src/FileHandle.js
```diff
@@ -0,0 +1,84 @@
...
+    this.systemFile = systemFile;
+    this.fileLimit = fileLimit;
+    this.uploadFileStatus();
+    if (this.failedProp.length === 0) {
+      this.createDownloadUrl(systemFile);
+    }
+  }
+
+  /**
+   * Syntactic sugar for getting the file extension
```
tlouisse, line 43: Think we can drop this description. Is it really syntactic sugar? :)

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+    this.uploadResponse = [];
+    this.__initialUploadResponse = this.uploadResponse;
+    this.uploadOnFormSubmit = true;
+    this.multiple = false;
+    this.enableDragAndDrop = false;
+    this.isDragging = false;
+    this.maxFileSize = MAX_FILE_SIZE;
+    this.allowedFileTypes = '';
+    this.allowedFileExtensions = '';
+    this._isUserProvidedLabel = '';
```
tlouisse, line 193: sounds more lik a boolean? Maybe `_hasUserProvidedFileUploadButtonLabel`? Or better: I think we should leave `fileUploadButtonLabel` to the Subclasser (who can provide it in a template, or, we provide a TemplateData object like in InputTelephone) like we do in all other components. 

The default text for Uploading a file ("Choose a file" by default in the platform) should be consistent across all forms within a Design System; an application developer should not change the button text for every instance. For a regular label/help-text, the latter does make sense

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+      },
+      /**
+       * style property
+       */
+      isDragging: { type: Boolean, reflect: true, attribute: 'is-dragging' },
+      maxFileSize: {
+        type: Number,
+        attribute: 'max-file-size',
+      },
+      fileUploadButtonLabel: { type: String, attribute: 'file-upload-button-label' },
```
tlouisse, line 67: Can we drop this prop? See explanation below...

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+  }
+
+  /**
+   * @configure SlotMixin
+   */
+  // @ts-ignore
+  get slots() {
+    return {
+      ...super.slots,
+      input: () => {
```
tlouisse, line 96:
```suggestion
      input: () => html\`<input value="${ifDefined(this.getAttribute('value'))}">\`
```

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+    };
+  }
+
+  static get properties() {
+    return {
+      ...super.properties,
+      _uploadedFilesMetaData: {
+        type: Array,
+      },
+      allowedFileExtensions: {
```
tlouisse, line 47: Maybe we should align this as much as possible with the platform: one`accept` attr that combines extensions and types? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+        attribute: 'allowed-file-types',
+      },
+      enableDragAndDrop: {
+        type: Boolean,
+        attribute: 'enable-drag-and-drop',
+      },
+      /**
+       * style property
+       */
+      isDragging: { type: Boolean, reflect: true, attribute: 'is-dragging' },
```
tlouisse, line 62: would be nice if we could make this read only by disabling the setter (see LionInputTelephone for examples)

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+      super.styles,
+      css`
+        .input-group__container {
+          position: relative;
+          display: flex;
+          flex-direction: column;
+          width: fit-content;
+        }
+
+        :host([drag-and-drop]) .input-group__container {
```
tlouisse, line 822: This should be aligned with the atrr name, I think

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+      },
+      allowedFileExtensions: {
+        type: String,
+        attribute: 'allowed-file-extensions',
+      },
+      allowedFileTypes: {
+        type: String,
+        attribute: 'allowed-file-types',
+      },
+      enableDragAndDrop: {
```
tlouisse, line 55: So basically (since we use native file input under the hood) drag and drop is enabled, but this flag shows a bigger "drop zone"?
So maybe call it "enable-drop-zone"?

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,848 @@
...
+
+    if (_newFiles.length > 0) {
+      this.dispatchFileListChangeEvent(_newFiles);
+    }
+  }
+
+  /**
+   * @param {InputFile[]} _newFiles
+   */
+  dispatchFileListChangeEvent(_newFiles) {
```
tlouisse, line 603:
```suggestion
   * @protected
   */
  _dispatchFileListChangeEvent(_newFiles) {
```

tkgroot (changes requested):
have reviewed and found some minor things to discuss

packages/ui/components/input-file/src/validators.js
```diff
@@ -0,0 +1,110 @@
...
+    return 'FileValidation';
+  }
+
+  /**
+   * Syntactic sugar for getting the file extension
+   * @param {typeof File.name} fileName
+   * @return {string}
+   */
+  // eslint-disable-next-line class-methods-use-this
+  _getExtension = fileName => fileName?.slice(fileName.lastIndexOf('.'));
```
DannyMoerkerke, line 23: This should be `_getExtension = fileName => fileName?.slice(fileName.lastIndexOf('.') + 1)` since it now compares for example `.txt` to `txt`.

docs/components/input-file/use-cases.md
```diff
@@ -0,0 +1,424 @@
...
+
+```js preview-story
+const myFormReset = ev => {
+  const input = ev.target.querySelector('lion-input-file');
+  input.reset();
+  console.log(input.modelValue);
+};
+
+const myFormSubmit = ev => {
+  ev.preventDefault;
```
DannyMoerkerke, line 245: Should be `ev.preventDefault()`

docs/components/input-file/use-cases.md
```diff
@@ -0,0 +1,424 @@
...
+  const input = ev.target.querySelector('lion-input-file');
+  console.log(input.hasFeedbackFor);
+  console.log(input.serializedValue);
+};
```
DannyMoerkerke, line 249: Should also have `return false;` to prevent submitting of the form.

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,879 @@
...
+      }),
+      new DuplicateFileNames({ show: false }),
+    ];
+  }
+
+  /**
+   * @param {InputFile[]} uploadedFiles
+   * @protected
+   */
+  _uploadFiles(uploadedFiles) {
```
DannyMoerkerke, line 568: This method does not actually upload the files to a server and is therefore a bit misleading. Should we rename it to something like `processFiles` or `parseFiles`?

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,879 @@
...
+        this.validate();
+      }
+    }
+    return computedFileList;
+  }
+
+  /**
+   * @param {DragEvent} ev
+   */
+  uploadDroppedFiles(ev) {
```
DannyMoerkerke, line 466: This method does not actually upload the files to a server and is therefore a bit misleading. Should we rename it to something like `processDroppedFiles` or `parseDroppedFiles`?

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -0,0 +1,879 @@
...
+  // @ts-ignore
+  get slots() {
+    return {
+      ...super.slots,
+      input: () => html`<input .value="${ifDefined(this.getAttribute('value'))}" />`,
+      'file-upload-button': () => {
+        const button = document.createElement('button');
+        button.setAttribute('id', `upload-button-${this._inputId}`);
+        button.setAttribute('type', 'button');
+        button.setAttribute('click', `${() => this._inputNode.click()}`);
```
DannyMoerkerke, line 100: This doesn't work, we should just use `addEventListener` here.

docs/components/input-file/use-cases.md
```diff
@@ -0,0 +1,424 @@
...
+
+```js preview-story
+export const prefilledState = () => {
+  return html`
+    <lion-input-file
+      label="Upload"
+      name="myFiles"
+      .multiple="${true}"
+      .validators="${[new Required()]}"
+      .uploadResponse="${[
```
DannyMoerkerke, line 213: Can we not just let the consumer only specify the filenames and we fill in the rest to make it easier for consumers?
What exactly is the purpose of prefilling? Are the specified files also uploaded somehow?

packages/ui/components/input-file/src/LionInputFile.js
DannyMoerkerke: fixed

docs/components/input-file/use-cases.md
DannyMoerkerke: fixed

packages/ui/components/input-file/src/validators.js
DannyMoerkerke: fixed

tkgroot (approved):

packages/ui/exports/define/lion-uploaded-file-list.js
```diff
@@ -0,0 +1,3 @@
+import { LionSelectedFileList } from '../input-file.js';
+
+customElements.define('lion-uploaded-file-list', LionSelectedFileList);
```
gerjanvangeest, line 3:
```suggestion
customElements.define('lion-selected-file-list', LionSelectedFileList);
```

And we should also update this file name.

packages/ui/components/input-file/src/LionInputFile.js
DannyMoerkerke: renamed

DannyMoerkerke (approved):

PR #1884, opened 2023-01-16
fix(icon): non-fatal console error for missing icons

jorenbroekema:
## What I did

1. Wrap icon manager resolve icon in try-catch to catch fatal error, then console.error it (non-fatal)

## Rationale

In short, missing assets like images, svgs, etc. should not cause fatal errors and cause your entire program (app) to break. 
With lion, icon assets are JS files (SVG-in-JS if you will), so we should take this extra precaution in my opinion.

Another argument can be that apps might choose to create a robots.txt file where they exclude icon files. This is necessary so that crawlers (like Googlebot for Google SEO), do not hit their [crawl request limits](https://support.google.com/webmasters/thread/173793044?hl=en&msgid=173822151) due to many icons. Icons, in the end, are not necessary to render the app for indexing it. When not bundling, e.g. loading Lion (or a Lion extension) from a CDN, this also becomes important as you get many requests due to not doing any bundling.

### DIFF
--- a/packages/ui/components/icon/src/LionIcon.js
+++ b/packages/ui/components/icon/src/LionIcon.js
@@ -194,11 +194,17 @@ export class LionIcon extends LitElement {
       }
     } else {
       const iconIdBeforeResolve = this.iconId;
-      const svg = await this._iconManager.resolveIconForId(iconIdBeforeResolve);
 
-      // update SVG if it did not change in the meantime to avoid race conditions
-      if (this.iconId === iconIdBeforeResolve) {
-        this.svg = svg;
+      // Wrap in try-catch so error is non-fatal.
+      // Failure to load an icon (asset) should not crash the entire app.
+      try {
+        const svg = await this._iconManager.resolveIconForId(iconIdBeforeResolve);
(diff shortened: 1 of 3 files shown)


packages/ui/components/icon/src/LionIcon.js
```diff
@@ -194,11 +194,17 @@ export class LionIcon extends LitElement {
...
+      // Wrap in try-catch so error is non-fatal.
+      // Failure to load an icon (asset) should not crash the entire app.
+      try {
+        const svg = await this._iconManager.resolveIconForId(iconIdBeforeResolve);
+        // update SVG if it did not change in the meantime to avoid race conditions
+        if (this.iconId === iconIdBeforeResolve) {
+          this.svg = svg;
+        }
+      } catch (e) {
+        console.error(e);
```
tlouisse, line 207:
```suggestion
        // eslint-disable-next-line no-console
        console.error(e);
```

tlouisse (approved):

jorenbroekema:
Thanks!

PR #1907, opened 2023-02-09
fix:  "autofocus" attribute added to _listboxNode

DannyMoerkerke:
## What I did

1. When the overlay is shown, the "autofocus" attribute is added to _listboxNode (_inputNode) to make sure that keyboard navigation continues to work when the element is inside a bottomsheet. When the overlay is closed the attribute is removed.
2. Added test
3. Added changes

### DIFF
--- a/packages/ui/components/select-rich/src/LionSelectRich.js
+++ b/packages/ui/components/select-rich/src/LionSelectRich.js
@@ -369,6 +369,8 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     if (this.hasNoDefaultSelected) {
       this._noDefaultSelectedInheritsWidth();
     }
+
+    this._listboxNode.setAttribute('autofocus', '');
   }
 
   /** @private */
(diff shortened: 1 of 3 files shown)


packages/ui/components/select-rich/src/LionSelectRich.js
```diff
@@ -369,6 +369,8 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
     if (this.hasNoDefaultSelected) {
       this._noDefaultSelectedInheritsWidth();
     }
+
+    this._listboxNode.setAttribute('autofocus', '');
```
tlouisse, line 373:
```suggestion
    // When `trapsKeyboardFocus:true` is configured in our overlay, we need to make sure
    // that our element with [role=listbox] gets focus. The `containFocus` util will look 
    // for an element with [autofocus] (otherwise it sets focus to `contentNode` (the 'root' of the overlay))
    this._listboxNode.setAttribute('autofocus', '');
```

tlouisse (approved):

PR #1906, opened 2023-02-09
fix(@lion/ui): export types necessary for type inference of mixins (fixes #1903)

riovir:
## What I did

Export types necessary to generate type definitions of components that incorporate Lion mixins.

1. Walked through the exported mixins and ensured that the [Mixin]Host types are exported
2. Noticed that each export is a line on its own so I stuck with that pattern
3. Additionally, exported the types used by the [Mixin]Host types
4. Exported the following utility types to prevent extenders from copying them out of the Lion source. These are needed when TypeScript asks the extenders to clarify the types:
- SlotsMap
- SlotFunctionResult
- SlotRerenderObject
- HTMLElementWithValue
- FormControl
- FormatOptions
- InteractionStates
- ElementWithParentFormGroup
- ValidationType
- StringToFunctionMap
- ViewportPlacement

I did my best not to over-export. The [Mixin]Host types are required by the @rollup/typescript-plugin, when it's asked to generate type declarations. Naturally, it needs to be able to import the types in order to explain them to users of the components. The utilities were mostly needed as they were already in use by existing codebases.

PR #1909, opened 2023-02-10
fix(@lion/ui): export types necessary for type inference of core.js

### LINKED ISSUE #1903: [lion/ui] Mixins create classes that can't be typed
## Expected behavior

Using a Lion mixin to create a component allows the (Rollup) TypeScript plugin to generate declaration files.

## Actual Behavior

```
git clone https://gitlab.com/riovir/lion-typescript-rollup-reprod.git
cd lion-typescript-rollup-reprod
npm ci
npm test
```

This silly example of adding a SlotMixin to a LionButton demonstrates the error:

```
(!) Plugin typescript: @rollup/plugin-typescript TS2742: The inferred type of 'MyButton' cannot be named
without a reference to '../node_modules/@lion/ui/dist-types/components/core/types/SlotMixinTypes.js'.
This is likely not portable. A type annotation is necessary.
```

This error consistently pops up on all mixins I've seen so far, particularly the ones in the `form-core` and `core`.

## Additional context

The error happens when the @rollup/plugin-typescript to set to generate declarations. I've also added the old form of lion commented out to show that version works. If this is in fact not a problem with Lion, and someone could educate me on how to properly configure my Rollup + TypeScript combo, that would also be much appreciated.

riovir:
Regretfully, I overlooked two Host types in the core in the https://github.com/ing-bank/lion/pull/1906 . Mea culpa. Fixes #1903 some more, this time for extenders who create a component that is tied to the `DisabledMixin`.

## What I did

1. Added the overlooked `DisabledHost` and `DisabledWithTabIndexHost` types as well.

### DIFF
--- a/.changeset/ten-plums-notice.md
+++ b/.changeset/ten-plums-notice.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(@lion/ui): export types necessary for type inference of core.js
--- a/packages/ui/exports/types/core.ts
+++ b/packages/ui/exports/types/core.ts
@@ -1,3 +1,5 @@
+export { DisabledHost } from '../../components/core/types/DisabledMixinTypes.js';
+export { DisabledWithTabIndexHost } from '../../components/core/types/DisabledWithTabIndexMixinTypes.js';
 export { SlotHost } from '../../components/core/types/SlotMixinTypes.js';
 export { SlotsMap } from '../../components/core/types/SlotMixinTypes.js';
 export { SlotFunctionResult } from '../../components/core/types/SlotMixinTypes.js';


riovir:
@tlouisse I've added the missing types. The pipeline seems to be hitting an infrastructure issue for the windows, as I don't think this change caused it to fail.

tlouisse:
> @tlouisse I've added the missing types. The pipeline seems to be hitting an infrastructure issue for the windows, as I don't think this change caused it to fail.

Np, thanks again! 
Maybe you can check the types generated via `npm run types`? 
If they solve your build, we could do another release (or you could temporarily solve your build with patch-package) :)

tlouisse (approved):

riovir:
@tlouisse I've checked out the change and our build would now indeed pass with the type definitions generated.

tlouisse:
> @tlouisse I've checked out the change and our build would now indeed pass with the type definitions generated.

Ok. A new patch release is coming 👍

Issue #1910, opened 2023-02-10
enhance dirty check of input-date / input-datepicker


tlouisse:
Prevent infinite loops by adding hasChanged...
See https://github.com/ing-bank/lion/issues/1908

PR #1914, opened 2023-02-13
[draft] incorporate hasChanged in sync updates

tlouisse:
- considers hasChanged function of `static get properties` for all sync updates in FormatMixin
- fixes dirty check of dates, so that `new Date('2020/10/10')` equals `new Date('2020/10/10')`
- fixes some behaviors with fomatting erroneous values wrt initial user-input-changed handling
 
TODO:
- make all tests run again

closes https://github.com/ing-bank/lion/issues/1910

### DIFF
--- a/packages/ui/components/input-date/src/LionInputDate.js
+++ b/packages/ui/components/input-date/src/LionInputDate.js
@@ -3,6 +3,32 @@ import { LionInput } from '@lion/ui/input.js';
 import { formatDate, LocalizeMixin, parseDate } from '@lion/ui/localize-no-side-effects.js';
 
 /**
+ *
+ * @param {Date | undefined} newV
+ * @param {Date | undefined} oldV
+ */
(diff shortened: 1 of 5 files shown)


gerjanvangeest:

packages/ui/components/input-date/src/LionInputDate.js
```diff
@@ -2,6 +2,32 @@ import { IsDate } from '@lion/ui/form-core.js';
...
+      newV.getFullYear() === oldV.getFullYear?.()
+    ) // year
+  );
+}
+
+/**
+ * Change function that returns true if `value` is different from `oldValue`.
+ * This method is used as the default for a property's `hasChanged` function.
+ */
+export const notEqual = (value, old) => {
```
tlouisse, line 26: TODO: in latest lit, we can import this function: https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L376

PR #1916, opened 2023-02-13
fix(@lion/ui): declare sideEffects in package.json (fixes #1915)

### LINKED ISSUE #1915: [@lion/ui]: Missing sideEffects declaration causes bloated bundles
## Expected behavior

```
import @lion/core; // Tree-shaken to blank bundle
import @lion/[other-exports]; // Tree-shaken to blank bundle
````
The `sideEffects` declaration is filled out in the `package.json`.

## Actual Behavior

```
import @lion/ui/core.js; // Includes about 20KB 
import @lion/ui/[other-exports].js; // Includes varying sizes of imports
````

The `sideEffects` declaration is missing from the `package.json`.

## Additional context

In our case after migrating our design system to the `@lion/ui` package we observed all the bundle size tests failing. Zero-import cases were resulting builds around 190KB, while importing just one component was somewhere around 220-260KB.

### PR DESCRIPTION
**Pull request by riovir**
## What I did

1. Added the following side effects (between scripts and dependencies, where I found them in older Lion packages):
```
  "sideEffects": [
    "exports/define/**/*.js",
    "exports/define-helpers/**/*.js",
    "exports/icon.js",
    "exports/localize.js",
    "exports/overlays.js"
  ],
```
2. Did the same change in the node_modules of our design system, and checked that indeed they fix the broken tree-shaking. All of our components now report bundle sizes similar to the pre-@lion/ui migration, including zero-import, single component imports and import everything tests.
3. Looked for the hint "singleton" in the `@lion/ui` source js files to see if any `singleton-manager` usages get caught that were not mentioned in the issue.

All the `define` and `define-helpers` were my take, not directly suggested by @tlouisse. It seems to make sense to mark the self-defining components as sideEffects.

### DIFF
--- a/.changeset/witty-carpets-clap.md
+++ b/.changeset/witty-carpets-clap.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(@lion/ui): declare sideEffects in package.json
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -45,6 +45,13 @@
     "types-check-only": "tsc --project tsconfig-check-only.json",
     "types-correct-after-build": "node ./scripts/types-correct-after-build.js"
   },
+  "sideEffects": [
+    "exports/define/**/*.js",
+    "exports/define-helpers/**/*.js",
+    "exports/icon.js",
+    "exports/localize.js",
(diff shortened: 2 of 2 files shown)


tlouisse:
Awesome 👍

PR #1918, opened 2023-02-14
feat(combobox): allow complex objects in choiceValue

gerjanvangeest:
## What I did

1. support complex objects in choiceValue
2. add an option to filter only on specific keys of the choiceValue
3. Make the first occurrence of a string highlighted, instead of the last


DannyMoerkerke (approved):

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -642,14 +657,27 @@ export class LionCombobox extends OverlayMixin(LionListbox) {
    * @param {string} matchingString
    * @protected
    */
-  // eslint-disable-next-line class-methods-use-this
   _onFilterMatch(option, matchingString) {
-    makeMatchingTextBold(option, matchingString, option);
+    this._highlightMatchedOptions(option, matchingString);
```
tlouisse, line 661: It would be nice if these would be tested as well in Subclasser describe and its use case would be clear from it.

Since these methods only handle one option a time, maybe we could consider renaming to:
- `_highlightMatchedOption`
- `_unhighlightMatchedOption`

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -1999,6 +2052,98 @@ describe('lion-combobox', () => {
           await performChecks();
         });
       });
+
+      it('allows to override "_onFilterMatch" and "_onFilterUmatch"', async () => {
```
tlouisse, line 2056:
```suggestion
      it('allows to override "_onFilterMatch" and "_onFilterUnmatch"', async () => {
```

tlouisse (approved):

PR #1928, opened 2023-02-27
fix(combobox): add translations

gerjanvangeest:
## What I did

1.


packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -880,8 +880,26 @@ describe('lion-combobox', () => {
...
         expect(el.checkedIndex).to.equal(-1);
         await el.feedbackComplete;
-        expect(el.hasFeedbackFor).to.include('error');
-        expect(el.showsFeedbackFor).to.include('error');
+        expect(el.hasFeedbackFor).to.include('error', 'hasFeedbackFor');
+        await el.feedbackComplete;
+        expect(el.showsFeedbackFor).to.include('error', 'showsFeedbackFor');
+      });
+
+      it('dropdown has a label', async () => {
```
Knorgias, line 888: How about testing the label in a different locale / language?

tlouisse: We usually trust here that localize does a fine job here when we provide multiple locales (so that's already covered in localize tests).

tlouisse (approved):

Issue #1664, opened 2022-04-14
[Feature request][combo box]: always let the modelValue reflect the value of the text box


M49A2:
Currently when you just type in some free text (without selecting anything in the list with suggestions), the modelValue will always be empty. Also no model-value-changed event will be thrown when typing. The only way to get the modelValue updated is by selecting a suggestion.

I would like to have the current text box value available as modelValue and also want/need to listen for model-value-changed events when the user starts typing.

tlouisse:
Thanks for the request. This definitely makes sense for a search input, where the text box is leading.
Currently, in the combobiox the listbox is leading. This however also has its usecases, so I think the 'fallback' to modelValue *(which basically makes the text box leading) should be a configurable option on top 👍 
It's been a while since I worked on the code, but I hope it would be conceptually smth like:

```js
get modelValue() {
  const listboxVal = super.modelValue;
  // TODO: come up with flag name less horrible than allowNonPredeterminedValues
  if (!listboxVal && this.allowNonPredeterminedValues) {
    return this._inputNode.value;
  }
  return listboxVal;
}

```

Question would be how to do this when `multiple` flag is enabled as well

gerjanvangeest:
@M49A2 This is fixed in mentioned PR.

PR #1930, opened 2023-03-01
feat(combobox): add `requireOptionMatch` flag and MatchesOption validator

gerjanvangeest:
## What I did

1. Allow unmatching values to be typed, before this returned an empty string.
2. Added an IsMatchingAnOption validator to check if the value is matching an option.

### DIFF
--- a/.changeset/soft-eagles-reply.md
+++ b/.changeset/soft-eagles-reply.md
@@ -5,5 +5,5 @@
 [combobox] Multiple improvements:
 
 - Allow textbox values to be entered that do not match a listbox option, via `requireOptionMatch` flag.
-- Added an `IsMatchingAnOption` validator to check if the value is matching an option.
+- Added an `MatchesOption` validator to check if the value is matching an option.
 - Exports combobox test helpers
--- a/.changeset/soft-eagles-reply.md
+++ b/.changeset/soft-eagles-reply.md
@@ -4,6 +4,6 @@
 
 [combobox] Multiple improvements:
 
-- Allow unmatching values to be typed, with option `requireOptionMatch`.
+- Allow textbox values to be entered that do not match a listbox option, via `requireOptionMatch` flag.
 - Added an `IsMatchingAnOption` validator to check if the value is matching an option.
 - Exports combobox test helpers
--- a/.changeset/soft-eagles-reply.md
+++ b/.changeset/soft-eagles-reply.md
@@ -1,0 +1,9 @@
+---
+'@lion/ui': patch
+---
+
+[combobox] Multiple improvements:
+
+- Allow unmatching values to be typed, with option `requireOptionMatch`.
+- Added an `IsMatchingAnOption` validator to check if the value is matching an option.
+- Exports combobox test helpers
--- a/docs/components/combobox/overview.md
+++ b/docs/components/combobox/overview.md
@@ -41,7 +41,7 @@ The combobox has many configurable properties to fine-tune its behaviour:
 
 - **Multiple choice** - Allows multiselection of options.
 - **requireOptionMatch**
-  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `MatchesOption` set as a default validator.
   - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. a search input.
 - **Autocomplete** - When the autocompletion will happen: `none`, `list`, `inline` and `both`.
 - **Matchmode** - Which part of the value should match: `begin` and `all`.
--- a/docs/components/combobox/overview.md
+++ b/docs/components/combobox/overview.md
@@ -37,7 +37,17 @@ export const main = () => html`
 
 ## Features
 
-> tbd
+The combobox has many configuration options to fine tune its behaviour:
+
+- **Multiple choice** - Allows multiselection of options.
+- **requireOptionMatch**
+  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+  - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input.
+- **Autocomplete** - When the autocompletion will happen: `none`, `list`, `inline` and `both`.
+- **Matchmode** - Which part of the value should match: `begin` and `all`.
+- **Show all on empty** - Shows the options list on empty.
+- **Selection follows focus** - When false the active/focused and checked/selected values will be kept track of independently.
+- **Rotate keyboard Navigation** - When false it won't rotate the navigation.
 
 ## Installation
 
--- a/docs/components/combobox/overview.md
+++ b/docs/components/combobox/overview.md
@@ -37,12 +37,12 @@ export const main = () => html`
 
 ## Features
 
-The combobox has many configuration options to fine tune its behaviour:
+The combobox has many configurable properties to fine-tune its behaviour:
 
 - **Multiple choice** - Allows multiselection of options.
 - **requireOptionMatch**
   - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
-  - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input.
+  - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. a search input.
 - **Autocomplete** - When the autocompletion will happen: `none`, `list`, `inline` and `both`.
 - **Matchmode** - Which part of the value should match: `begin` and `all`.
 - **Show all on empty** - Shows the options list on empty.
--- a/docs/components/combobox/use-cases.md
+++ b/docs/components/combobox/use-cases.md
@@ -16,11 +16,30 @@ availability of the popup.
 import { LitElement, html, repeat } from '@mdjs/mdjs-preview';
 import { listboxData, listboxComplexData } from '../listbox/src/listboxData.js';
 import { LionCombobox } from '@lion/ui/combobox.js';
+import { Required } from '@lion/ui/form-core.js';
 import '@lion/ui/define/lion-combobox.js';
 import '@lion/ui/define/lion-option.js';
 import './src/demo-selection-display.js';
 import { lazyRender } from './src/lazyRender.js';
 import levenshtein from './src/levenshtein.js';
+import { loadDefaultFeedbackMessages } from '@lion/ui/validate-messages.js';
+loadDefaultFeedbackMessages();
+```
+
+## Require option match
+
+By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+
+When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input. This means that all input values are allowed.
+
+```js preview-story
+export const optionMatch = () => html`
+  <lion-combobox name="search" label="Search" .requireOptionMatch=${false}>
+    ${lazyRender(
+      listboxData.map(entry => html` <lion-option .choiceValue="${entry}">${entry}</lion-option> `),
+    )}
+  </lion-combobox>
+`;
 ```
 
 ## Autocomplete
@@ -225,6 +244,22 @@ export const multipleChoice = () => html`
         (entry, i) =>
           html` <lion-option .choiceValue="${entry}" ?checked=${i === 0}>${entry}</lion-option> `,
       ),
+    )}
+  </lion-combobox>
+`;
+```
+
+## Validation
+
+The combobox works with a `Required` validator to check if it is empty.
+
+By default the a check is made which makes sure the value matches an option. This only works if `requireOptionMatch` is set to true.
+
+```js preview-story
+export const validation = () => html`
+  <lion-combobox name="combo" label="Validation" .validators=${[new Required()]}>
+    ${lazyRender(
+      listboxData.map(entry => html` <lion-option .choiceValue="${entry}">${entry}</lion-option> `),
     )}
   </lion-combobox>
 `;
--- a/docs/components/combobox/use-cases.md
+++ b/docs/components/combobox/use-cases.md
@@ -30,7 +30,7 @@ loadDefaultFeedbackMessages();
 
 By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
 
-When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input. This means that all input values are allowed.
+When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. a search input. This means that all input values are allowed.
 
 ```js preview-story
 export const optionMatch = () => html`
--- a/docs/components/combobox/use-cases.md
+++ b/docs/components/combobox/use-cases.md
@@ -28,7 +28,7 @@ loadDefaultFeedbackMessages();
 
 ## Require option match
 
-By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `MatchesOption` set as a default validator.
 
 When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. a search input. This means that all input values are allowed.
 
--- a/packages/ui/components/combobox/src/LionCombobox.js
+++ b/packages/ui/components/combobox/src/LionCombobox.js
@@ -171,6 +171,11 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
   }
 
   /**
+   * We define the value getter/setter below as also defined in LionField (via FormatMixin). 
+   * Since FormatMixin is meant for Formgroups/ChoiceGroup it's not applied on Combobox; 
+   * Combobox is somewhat of a hybrid between a ChoiceGroup and LionField, therefore we copy over 
+   * some of the LionField members to align with its interface. 
+   * 
    * The view value. Will be delegated to `._inputNode.value`
    */
   get value() {
@@ -382,7 +387,8 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
      */
     this.showAllOnEmpty = false;
     /**
-     * If set to false, the value is allowed to not match any of the options
+     * If set to false, the value is allowed to not match any of the options.
+     * We set the default to true for backwards compatibility
      */
     this.requireOptionMatch = true;
     /**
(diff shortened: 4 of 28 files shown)


tlouisse:
closes https://github.com/ing-bank/lion/issues/1664

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +141,54 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
     ];
   }
 
+  /**
+   * @override ChoiceGroupMixin
+   */
+  get modelValue() {
```
tlouisse, line 147: Could we do smth like:
```js
get modelValue() {
  if (!this.requireOptionMatch) {
    if (!this.multipleChoice) {
      return this._inputNode.value;
    } else {
      // This must be bound to a chip system or smth that would be maintained seperately...
      // It might be smth we want to implement later and for now we could throw an error on "!this.requireOptionMatch && this.multipleChoice"
    }
  }
  return super.modelValue;
}
```

packages/ui/components/combobox/src/validators.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  /**
+   * @param {unknown} [value]
+   * @param {string | undefined} [options]
+   * @param {{ node: any }} [config]
+   */
+  // eslint-disable-next-line class-methods-use-this
+  execute(value, options, config) {
+    const hasMatchingOptions =
```
tlouisse, line 16: What does this return? `aria-setsize` seems to be available on every option: https://github.com/ing-bank/lion/blob/fd09f652d450580d7ad43d2c4466cb4ed42d30f0/packages/ui/components/listbox/src/ListboxMixin.js#L455

`hasMatchingOptions` will then always be false (because .length > 0): 
- the boolean returned (indicating whether feedback should be shown) will thus be `false`
- this indicates that there is never an error, even if the textbox value is not found...

Maybe I'm wrong btw, but it would be good to (dis)prove it with a test :)

tlouisse: I think `hasMatchingOption(s)` can be determined with `this.checkedIndex !== -1`.
And then it might be good to think about the multipleChoice case as well

packages/ui/components/combobox/src/LionCombobox.js
gerjanvangeest: You want the `requireOptionMatch` to be the default, so non breaking?

`super.modelValue` does not exists.

packages/ui/components/combobox/src/validators.js
gerjanvangeest: Actually it worked, i got confused by my own naming :)

packages/ui/components/combobox/src/LionCombobox.js
tlouisse: Seems this is inherited indirectly from ChoiceGroupMixin (via ListboxMixin)?

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +141,64 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+      this._setCheckedElements(value, checkCondition);
+      this.requestUpdate('modelValue', this._oldModelValue);
+    }
+    this._oldModelValue = this.modelValue;
+  }
+
+  reset() {
+    super.reset();
+    // @ts-ignore
+    this._inputNode.value = this._initialModelValue;
```
tlouisse, line 199: Here we also should account for multiple choice?

packages/ui/components/combobox/src/validators.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  /**
+   * @param {unknown} [value]
+   * @param {string | undefined} [options]
+   * @param {{ node: any }} [config]
+   */
+  // eslint-disable-next-line class-methods-use-this
+  execute(value, options, config) {
+    const hasNoMatchingOptions = config?.node.checkedIndex === -1;
```
tlouisse, line 16: Nice

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -428,6 +312,27 @@ describe('lion-combobox', () => {
       expect(el.formElements[0].checked).to.be.false;
     });
 
+    it('sets modelValue to _inputNode.value if no option is selected when requireOptionMatch is false', async () => {
+      const el = /** @type {LionCombobox} */ (
+        await fixture(html`
+          <lion-combobox name="foo" .modelValue="${'Artichoke'}">
```
tlouisse, line 318:
```suggestion
          <lion-combobox name="foo" .modelValue="${'Artichoke'}" .requireOptionMatch="${false}">
```

tlouisse:
Looks good (w/o checking locally). 

Think all multipleChoice edge cases still need to be tackled, though. 
Later I will also checkout locally to have a deeper look

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +141,64 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+    if (!this.requireOptionMatch && this._inputNode) {
+      return this._inputNode.value;
+    }
+    return '';
+  }
+
+  /**
+   * @override ChoiceGroupMixin
+   */
+  set modelValue(value) {
```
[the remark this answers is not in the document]
gerjanvangeest: Yes it was. I was only not aware that i could do: `super.modelvalue = value`, now i am :)

packages/ui/components/combobox/src/LionCombobox.js
gerjanvangeest: Added a test, i don't think anything special is needed here.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -557,7 +600,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
    */
   // eslint-disable-next-line class-methods-use-this
   _getTextboxValueFromOption(option) {
-    return option.choiceValue;
+    if (option && option.choiceValue) {
+      return option.choiceValue;
+    }
+    return this.modelValue;
```
tlouisse, line 606:
```suggestion
    return option?.choiceValue || this.modelValue;
```

packages/ui/components/combobox/src/LionCombobox.js
tlouisse: Yeah, you're right (only in the unlikely edge case where one switches dynamically from .multipleChoice true to false or vice versa).

This is maybe not the best example where multipleChoice should be taken into account, but I would still have a check for .modelValue and IsMatchingOption Validator

gerjanvangeest:
Me and @tlouisse have concluded it should work as follows:

- **requireOptionMatch: true** - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching values should become Unparseable, with the `IsMatchingAnOption` set as a default validator.
- **requireOptionMatch: false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -480,6 +549,17 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
     }
+
+    if (changedProperties.has('requireOptionMatch') || changedProperties.has('multipleChoice')) {
+      if (!this.requireOptionMatch && this.multipleChoice) {
+        // TODO implement !requireOptionMatch and multipleChoice flow
+
+        // eslint-disable-next-line no-console
+        console.warn(
+          "multipleChoice and requireOptionMatch=false can't be used at the same time (yet).",
+        );
```
tlouisse, line 560:
```suggestion
        throw new Error(
          "multipleChoice and requireOptionMatch=false can't be used at the same time (yet).",
        );
```

I think throwing an error makes more sense, since you enter an illegal (for now) mode and a consuming dev might not check the console for this.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1090,6 +1178,6 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
 
   clear() {
     super.clear();
-    this._inputNode.value = '';
+    this.value = '';
```
tlouisse, line 1181: When FormatMixin is applied, it would delegate this to _inputNode. Does it do in this case as well?

gerjanvangeest: Yes, this is delegated as can be seen in the tests.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +146,55 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+   */
+  set modelValue(value) {
+    super.modelValue = value;
+  }
+
+  /**
+   * The view value. Will be delegated to `._inputNode.value`
+   */
+  get value() {
+    return (this._inputNode && this._inputNode.value) || this.__value || '';
```
tlouisse, line 177:
```suggestion
    return this._inputNode?.value || this.__value || '';
```

packages/ui/components/combobox/src/LionCombobox.js
tlouisse: Nvm, see the getter now 👍

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +146,62 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+   */
+  // @ts-ignore
+  get modelValue() {
+    const elems = this._getCheckedElements();
+    if (this.multipleChoice) {
+      return elems.map(el => el.choiceValue);
+    }
+    if (elems[0]) {
+      return elems[0].choiceValue;
+    }
```
tlouisse, line 160:
```suggestion
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -480,6 +556,17 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
     if (typeof this._selectionDisplayNode?.onComboboxElementUpdated === 'function') {
       this._selectionDisplayNode.onComboboxElementUpdated(changedProperties);
     }
+
+    if (changedProperties.has('requireOptionMatch') || changedProperties.has('multipleChoice')) {
+      if (!this.requireOptionMatch && this.multipleChoice) {
+        // TODO implement !requireOptionMatch and multipleChoice flow
+
+        // eslint-disable-next-line no-console
```
tlouisse, line 564:
```suggestion
```

docs/components/combobox/overview.md
```diff
@@ -37,7 +37,17 @@ export const main = () => html`
 
 ## Features
 
-> tbd
+The combobox has many configuration options to fine tune its behaviour:
+
+- **Multiple choice** - Allows multiselection of options.
+- **requireOptionMatch**
+  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+  - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input.
```
tlouisse, line 45:
```suggestion
  - **false** - The textbox is leading, with the listbox as an aid to supply suggestions, e.g. asearch input.
```

.changeset/soft-eagles-reply.md
```diff
@@ -0,0 +1,9 @@
+---
+'@lion/ui': patch
+---
+
+[combobox] Multiple improvements:
+
+- Allow unmatching values to be typed, with option `requireOptionMatch`.
```
tlouisse, line 7:
```suggestion
- Allow textbox values to be entered that do not match a listbox option, via `requireOptionMatch` flag.
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -141,6 +146,55 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+  // Duplicating from ChoiceGroupMixin, because you cannot independently inherit/override getter + setter.
+  // If you override one, gotta override the other, they go in pairs.
+  /**
+   * @override ChoiceGroupMixin
+   */
+  set modelValue(value) {
+    super.modelValue = value;
+  }
+
+  /**
```
tlouisse, line 173:
```suggestion
  /**
   * We define the value getter/setter below as also defined in FormatMixin (since FormatMixin is meant for Formgroups/ChoiceGroup and Combobox is somewhat of an hybrid between a ChoiceGroup and Field, we copy them over to align with the LionField interface). 
   * 
```

docs/components/combobox/use-cases.md
```diff
@@ -16,11 +16,30 @@ availability of the popup.
...
 import levenshtein from './src/levenshtein.js';
+import { loadDefaultFeedbackMessages } from '@lion/ui/validate-messages.js';
+loadDefaultFeedbackMessages();
+```
+
+## Require option match
+
+By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
+
+When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. an search input. This means that all input values are allowed.
```
tlouisse, line 33:
```suggestion
When `requireOptionMatch` is set to false the textbox is leading, with the listbox as an aid to supply suggestions, e.g. a search input. This means that all input values are allowed.
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -327,7 +381,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
      * By default, the listbox closes on empty, similar to wai-aria example and <datalist>
      */
     this.showAllOnEmpty = false;
-
+    /**
+     * If set to false, the value is allowed to not match any of the options
```
tlouisse, line 385:
```suggestion
     * If set to false, the value is allowed to not match any of the options.
     * We set the default to `true` for backwards compatibility
```

.changeset/soft-eagles-reply.md
```diff
@@ -0,0 +1,9 @@
+---
+'@lion/ui': patch
+---
+
+[combobox] Multiple improvements:
+
+- Allow textbox values to be entered that do not match a listbox option, via `requireOptionMatch` flag.
+- Added an `IsMatchingAnOption` validator to check if the value is matching an option.
```
tlouisse, line 8:
```suggestion
- Added an `MatchesOption` validator to check if the value is matching an option.
```

docs/components/combobox/overview.md
```diff
@@ -37,7 +37,17 @@ export const main = () => html`
 
 ## Features
 
-> tbd
+The combobox has many configurable properties to fine-tune its behaviour:
+
+- **Multiple choice** - Allows multiselection of options.
+- **requireOptionMatch**
+  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
```
tlouisse, line 44:
```suggestion
  - **true** (default) - The listbox is leading, the textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `MatchesOption` set as a default validator.
```

docs/components/combobox/use-cases.md
```diff
@@ -16,11 +16,30 @@ availability of the popup.
...
 import './src/demo-selection-display.js';
 import { lazyRender } from './src/lazyRender.js';
 import levenshtein from './src/levenshtein.js';
+import { loadDefaultFeedbackMessages } from '@lion/ui/validate-messages.js';
+loadDefaultFeedbackMessages();
+```
+
+## Require option match
+
+By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `IsMatchingAnOption` set as a default validator.
```
tlouisse, line 31:
```suggestion
By default `requireOptionMatch` is set to true, which means that the listbox is leading. The textbox is a helping aid to quickly select an option/options. Unmatching input values become Unparseable, with the `MatchesOption` set as a default validator.
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1,9 +1,11 @@
 import { browserDetection } from '@lion/ui/core.js';
+import { Unparseable } from '@lion/ui/form-core.js';
 import { LionListbox } from '@lion/ui/listbox.js';
 import { LocalizeMixin } from '@lion/ui/localize-no-side-effects.js';
 import { OverlayMixin, withDropdownConfig } from '@lion/ui/overlays.js';
 import { css, html } from 'lit';
 import { makeMatchingTextBold, unmakeMatchingTextBold } from './utils/makeMatchingTextBold.js';
+import { IsMatchingAnOption } from './validators.js';
```
tlouisse, line 8:
```suggestion
import { MatchesOption } from './validators.js';
```

packages/ui/components/combobox/src/validators.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable max-classes-per-file */
+import { Unparseable, Validator } from '@lion/ui/form-core.js';
+
+export class IsMatchingAnOption extends Validator {
```
tlouisse, line 4:
```suggestion
export class MatchesOption extends Validator {
```

packages/ui/components/combobox/test/validators.test.js
```diff
@@ -0,0 +1,62 @@
+import { mimicUserTyping } from '@lion/ui/combobox-test-helpers.js';
+import { IsMatchingAnOption } from '@lion/ui/combobox.js';
```
tlouisse, line 2:
```suggestion
import { MatchesOption } from '@lion/ui/combobox.js';
```

packages/ui/components/combobox/test/validators.test.js
```diff
@@ -0,0 +1,62 @@
...
+import '@lion/ui/define/lion-combobox.js';
+import '@lion/ui/define/lion-option.js';
+import { expect, fixture, html } from '@open-wc/testing';
+
+/**
+ * @typedef {import('@lion/ui/combobox.js').LionCombobox} LionCombobox
+ */
+
+describe('IsMatchingAnOption validation', () => {
```
tlouisse, line 11:
```suggestion
describe('MatchesOption validation', () => {
```

packages/ui/components/combobox/test/validators.test.js
```diff
@@ -0,0 +1,62 @@
...
+          <lion-option .choiceValue="${'Artichoke'}">Artichoke</lion-option>
+          <lion-option .choiceValue="${'Chard'}">Chard</lion-option>
+          <lion-option .choiceValue="${'Chicory'}">Chicory</lion-option>
+          <lion-option .choiceValue="${'Victoria Plum'}">Victoria Plum</lion-option>
+        </lion-combobox>
+      `)
+    );
+    const config = {};
+    config.node = el;
+    const validator = new IsMatchingAnOption();
```
tlouisse, line 54:
```suggestion
    const validator = new MatchesOption();
```

packages/ui/components/validate-messages/src/loadDefaultFeedbackMessagesNoSideEffects.js
```diff
@@ -1,22 +1,23 @@
 /* eslint-disable import/no-extraneous-dependencies */
+import { IsMatchingAnOption } from '@lion/ui/combobox.js';
```
tlouisse, line 2:
```suggestion
import { MatchesOption } from '@lion/ui/combobox.js';
```

packages/ui/components/validate-messages/src/loadDefaultFeedbackMessagesNoSideEffects.js
```diff
@@ -162,6 +163,8 @@ export function loadDefaultFeedbackMessagesNoSideEffects({ localize }) {
   MinMaxDate.getMessage = async data => getLocalizedMessage(data);
   /** @param {FeedbackMessageData} data */
   IsDateDisabled.getMessage = async data => getLocalizedMessage(data);
+  /** @param {FeedbackMessageData} data */
+  IsMatchingAnOption.getMessage = async data => getLocalizedMessage(data);
```
tlouisse, line 167:
```suggestion
  MatchesOption.getMessage = async data => getLocalizedMessage(data);
```

PR #1932, opened 2023-03-07
fix(listbox): allow use of arrow keys when focus is not on listbox

gerjanvangeest:
## What I did

1.

### DIFF
--- a/.changeset/eighty-dogs-travel.md
+++ b/.changeset/eighty-dogs-travel.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+[listbox] allow use of arrow keys when focus is not on listbox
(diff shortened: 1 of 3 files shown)


gerjanvangeest:
Error originates from:
https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/src/LionCombobox.js#L975

Which calls a super._listboxOnKeyDown:
https://github.com/ing-bank/lion/blob/master/packages/ui/components/listbox/src/ListboxMixin.js#L638

Which blocks the `ev. preventDefault` for ArrowLeft and ArrowRight.

tlouisse (approved):

gerjanvangeest:
The `_listboxOnKeyDown` is set on the `_inputNode`: https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/src/LionCombobox.js#L1065

I assume the ArrowLeft and Right were added in a later iteration?

PR #1934, opened 2023-03-09
fix: changed selectors for invokers and content to only select slotte…

DannyMoerkerke:
…d elements that are direct descendants. This is to prevent that slotted elements in accordion content and invokers are also selected and the amount of invokers and content is incorrect

## What I did

1. changed selectors for invokers and content to only select slotted elements that are direct descendants. This is to prevent that slotted elements in accordion content and invokers are also selected and the amount of invokers and content is incorrect

### DIFF
--- a/.changeset/silly-beds-teach.md
+++ b/.changeset/silly-beds-teach.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


gerjanvangeest (commented):
Can we have tests for this as well?

DannyMoerkerke:
> 

Good point, added a test.

PR #1938, opened 2023-03-12
Fixed grammar

passchn:
IBAN is feminin in German: <https://www.duden.de/rechtschreibung/IBAN>

## What I did

1. Changed spelling/grammar of the german error message for an invalid IBAN.

### DIFF
--- a/.changeset/nervous-numbers-smash.md
+++ b/.changeset/nervous-numbers-smash.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': minor
+---
+
+Changed spelling/grammar of the german error message for an invalid IBAN.
--- a/packages/ui/components/input-iban/translations/de.js
+++ b/packages/ui/components/input-iban/translations/de.js
@@ -1,6 +1,6 @@
 export default {
   error: {
-    IsIBAN: 'Geben Sie ein gültiges {fieldName} ein.',
+    IsIBAN: 'Geben Sie eine gültige {fieldName} ein.',
     IsCountryIBAN:
       'Geben Sie eine gültige {params, select,\n' +
       'AT {Österreichisch}\n' +


gerjanvangeest (commented):
thanks for the fix 💪 

Can you add a changeset (`npm run changeset`) so it will end up in our changelog?

tlouisse:
> thanks for the fix 💪
> 
> Can you add a changeset (`npm run changeset`) so it will end up in our changelog?

Good that you point this out indeed. However, we have to deal with some edge cases.

Maybe it can be solved with the same approach as in the Dutch language? https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-iban/translations/nl.js#L3
So that would be `gültige(s)` in German?

### Why that's needed
Since `{fieldName}` (corresponding to the `.fieldName` prop of the FormControl) can be:
- the `.fieldName` (when provided by the user)
- the `.label` (when no `.fieldName` provided)

Since we don't know whether the **noun** (IBAN or anything else) is going to be feminine or not, we don't know what our **adjective** (gültiges/gültige) should be.


### 💡 Possible ways of making this better...
Although "gültige(s)" would provide a valid outcome in all scenarios, we might look into a way to also configure the adjective of a Validator message for a certain adjective+fieldName combo

passchn:
I can't really think of a probable scenario where the noun is neutrum in case of the IBAN field's label. 

Useful labels are "IBAN", "Nummer", maybe "Kontonummer" which are all feminin. You really have to be creative to come up with a neutrum word. 

Anyway, it would be great to be able to configure the messages, I think! Maybe they could also be passed as attributes to the WebComponent.

tlouisse:
> I can't really think of a probable scenario where the noun is neutrum in case of the IBAN field's label.
> 
> Useful labels are "IBAN", "Nummer", maybe "Kontonummer" which are all feminin. You really have to be creative to come up with a neutrum word.
> 
> Anyway, it would be great to be able to configure the messages, I think! Maybe they could also be passed as attributes to the WebComponent.

It's definitely a more sensible default than wat it was before, but in case the label would be "Iban Text", the fieldName would be masculine for instance...

Having a closer look at it: in German it can be `gültiger`|`gültige`| `gültiges` etc.: https://www.verbformen.com/declension/adjectives/gu3ltig.htm

So a one-size-fits-all solution like we have in Dutch is not possible for the German language.


The best solution for now will then be:
- use the translation you proposed
- give lion-input-iban a default fieldName of 'iban' (then it takes precedence over the label text and it will always be correct, unless the developer overrides .fieldName)

In case the user overrides .fieldName, it will always be possible to override the complete message.

tlouisse (approved):

PR #1939, opened 2023-03-14
fix: checked-changed event is no longer fired on element initializati…

DannyMoerkerke:
…on when checked is set through attribute

## What I did

1. checked-changed event is no longer fired on element initialization when checked is set through attribute
2. added test
3. added changes

### DIFF
--- a/packages/ui/components/switch/src/LionSwitchButton.js
+++ b/packages/ui/components/switch/src/LionSwitchButton.js
@@ -12,10 +12,6 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
         type: Boolean,
         reflect: true,
       },
-      __initialized: {
-        type: Boolean,
-        state: true,
-      },
     };
   }
 
--- a/packages/ui/components/switch/src/LionSwitchButton.js
+++ b/packages/ui/components/switch/src/LionSwitchButton.js
@@ -14,6 +14,7 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
       },
       __initialized: {
         type: Boolean,
+        state: true,
       },
     };
   }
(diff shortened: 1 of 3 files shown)


packages/ui/components/switch/src/LionSwitchButton.js
```diff
@@ -12,6 +12,9 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
         type: Boolean,
         reflect: true,
       },
+      __initialized: {
```
tlouisse, line 15: I think as long as this prop is not reactive, we can omit it here.

packages/ui/components/switch/src/LionSwitchButton.js
```diff
@@ -12,6 +12,9 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
...
+        type: Boolean,
+      },
```
tlouisse:
```suggestion
```

packages/ui/components/switch/src/LionSwitchButton.js
```diff
@@ -158,8 +162,18 @@ export class LionSwitchButton extends DisabledWithTabIndexMixin(LitElement) {
...
+      this.isConnected &&
+      name === 'checked' &&
+      this.checked !== oldValue &&
+      !this.disabled
+    ) {
       this.__checkedStateChange();
     }
   }
+
+  firstUpdated() {
```
tlouisse, line 173:
```suggestion
  /** @param {import('lit').PropertyValues } changedProperties */
  firstUpdated(changedProperties) {
    super.firstUpdated(changedProperties);
```

tlouisse: Even though ]lit says it's not necessary to call `super`](https://lit.dev/docs/components/lifecycle/#firstupdated ), inside lion (complex inheritance hierarchy of mixins + superclasses) it is... 

It would be nice to have a linter for this

PR #1942, opened 2023-03-15
fix(select-rich): update WAI ARIA pattern to combobox-select-only

gerjanvangeest:
## What I did

1. [select-rich] update WAI ARIA pattern to combobox-select-only
2. Add docs about why we use a native dialog inside all our overlays


gerjanvangeest:
This change works good in Chrome + Voiceover, Firefox + Voiceover, Edge + NVDA, Firefox + NVDA. But in Safari + Voiceover i get the following:

Old:
image


New:
image

Which isn't an improvement.
So I would say. Wait half a year and check again before merging this PR.

PR #1952, opened 2023-03-29
chore: make default public npm registry explicit

Knorgias:
## What I did

1. Make registry=https://registry.npmjs.org/ the default by adding it in .npmrc on project level
2. Update package-lock.json

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.org


package-lock.json
```diff
@@ -22334,7 +22334,7 @@
     },
     "packages/ui": {
       "name": "@lion/ui",
-      "version": "0.1.5",
```
tkgroot, line 22337: this change shouldn't be part of this PR, I assume.

tkgroot (changes requested):

package-lock.json
Knorgias: Ran an `npm i` and quickly also updated the package-lock.
I also added it as (2) on the description above, so we can include this in this PR as well, if you agree.

tkgroot: sure, i dont know how strict you are with these things. Fine by me

tkgroot (approved):

tlouisse (approved):

PR #1957, opened 2023-04-11
Fix/dialog scroll to top ios

DannyMoerkerke:
## What I did

1. OverlayController: fixed check to determine if native dialog is supported, fixed check to determine if user has moved focus while dialog is open.
2. lion-dialog: added test to assert if element specified in dialog config key elementToFocusAfterHide is in viewport when dialog is closed
3. lion-dialog: added test to assert that element specified in dialog config key elementToFocusAfterHide is not focused when the dialog is closed if the user deliberately moved focus to another element while the dialog was open

### DIFF
--- a/docs/components/dialog/use-cases.md
+++ b/docs/components/dialog/use-cases.md
@@ -24,6 +24,8 @@ import './src/slots-dialog-content.js';
 
 ## Placement overrides
 
+<button type="button" id="focus-test">Focus test</button>
+
 ```js preview-story
 export const placementOverrides = () => {
   const dialog = placement => {
--- a/docs/components/dialog/use-cases.md
+++ b/docs/components/dialog/use-cases.md
@@ -24,8 +24,6 @@ import './src/slots-dialog-content.js';
 
 ## Placement overrides
 
-<button type="button" id="focus-test">Focus test</button>
-
 ```js preview-story
 export const placementOverrides = () => {
   const dialog = placement => {
--- a/packages/ui/components/dialog/test/lion-dialog.test.js
+++ b/packages/ui/components/dialog/test/lion-dialog.test.js
@@ -72,4 +72,36 @@ describe('lion-dialog', () => {
       expect(nestedDialogEl.opened).to.be.true;
     });
   });
+
+  describe('focus handling', () => {
+    it('should focus the element specified in the "elementToFocusAfterHide" config key', async () => {
+      const cfg = {
+        trapsKeyboardFocus: false,
+        elementToFocusAfterHide: document.querySelector('#focus-button'),
+      };
+
+      const el = await fixture(html`
+        <div>
+          <button type="button" id="focus-button">Focus</button>
+          <lion-dialog .config="${cfg}">
+            <div slot="content" class="dialog">Hey there</div>
+            <button slot="invoker">Popup button</button>
+          </lion-dialog>
+        </div>
+      `);
+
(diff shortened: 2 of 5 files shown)


docs/components/dialog/use-cases.md
```diff
@@ -24,6 +24,8 @@ import './src/slots-dialog-content.js';
 
 ## Placement overrides
 
+<button type="button" id="focus-test">Focus test</button>
```
gerjanvangeest, line 27: Should this button stay?

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -828,6 +831,12 @@ export class OverlayController extends EventTarget {
       this._hideResolve = resolve;
     });
 
+    // save the current activeElement so we know if the user set focus to another element than the invoker of the dialog
+    // while the dialog was open.
+    // We need this in the _restoreFocus method to determine if we should focus this.elementToFocusAfterHide when the
+    // dialog is closed or keep focus on the element that the user deliberately gave focus
+    this.__activeElement = document.activeElement;
```
tlouisse, line 838:
```suggestion
    this.__activeElement = /** @type {ShadowRoot} */ (this.contentNode.getRootNode()).activeElement;
```
Because document.activeElement only gets the "outer shadow root".

For instance in example below:
```js
<el-in-body>
  <template shadowrootmode="open">
    <lion-dialog>
.     <button id="activeEl"></button>
    </lion-dialog>
  </template>
</el-in-body>
```

document.activeElement will be el-body, but you want to register button#activeEl (else the  "inside dialog check" won't work)

tlouisse: Can we maybe call it `__activeElementRightBeforeHide`?

tlouisse: Can we maybe call it `__activeElementRightBeforeHide`?

packages/ui/components/dialog/test/lion-dialog.test.js
```diff
@@ -72,4 +85,117 @@ describe('lion-dialog', () => {
       expect(nestedDialogEl.opened).to.be.true;
     });
   });
+
+  describe('focus handling', () => {
```
tlouisse, line 89: Since lion-dialog is only a "convenience component" with a preconfigured OverlayController, can we move these tests to OverlayController?

docs/components/dialog/use-cases.md
DannyMoerkerke: fixed

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -828,6 +831,14 @@ export class OverlayController extends EventTarget {
       this._hideResolve = resolve;
     });
 
+    // save the current activeElement so we know if the user set focus to another element than the invoker of the dialog
+    // while the dialog was open.
+    // We need this in the _restoreFocus method to determine if we should focus this.elementToFocusAfterHide when the
+    // dialog is closed or keep focus on the element that the user deliberately gave focus
```
tlouisse, line 837: What we mainly solved here was that the native `<dialog>` automatically does focus delegation on close if I'm correct?
By storing this right before hide, we know for sure that we get the activeElement that reflects the current app state (not the one that was set by the native dialog after hide)

tlouisse (approved):

PR #1963, opened 2023-04-14
fix(form-core) sync autofocus to focusable node (fixes #1775)

### LINKED ISSUE #1775: [form-core] Lion form field components do not receive autofocus in dialog

Sciurus7:
## What I did

1. In FocusMixin synced the autofocus attribute from the host to the focusable node.
2. Added tests to ensure the sync happens if autofocus was set on the host during render or when autofocus is added or removed from the host.

### DIFF
--- a/packages/ui/components/form-core/src/FocusMixin.js
+++ b/packages/ui/components/form-core/src/FocusMixin.js
@@ -27,6 +27,7 @@ const FocusMixinImplementation = superclass =>
       return {
         focused: { type: Boolean, reflect: true },
         focusedVisible: { type: Boolean, reflect: true, attribute: 'focused-visible' },
+        autofocus: { type: Boolean, reflect: true }, // Required in Lit to observe autofocus
       };
     }
 
@@ -52,11 +53,41 @@ const FocusMixinImplementation = superclass =>
     connectedCallback() {
       super.connectedCallback();
       this.__registerEventsForFocusMixin();
+      this.__syncAutofocusToFocusableElement();
     }
 
     disconnectedCallback() {
       super.disconnectedCallback();
       this.__teardownEventsForFocusMixin();
+    }
+
+    /**
+     * Gets called when an attribute is changed.
+     * @param {String} name
+     * @param {String | null} _old
+     * @param {String | null} value
+     * @protected
+     */
+    attributeChangedCallback(name, _old, value) {
+      super.attributeChangedCallback(name, _old, value);
+      if (name === 'autofocus') {
+        this.__syncAutofocusToFocusableElement();
+      }
+    }
+
+    /**
+     * @private
+     */
+    __syncAutofocusToFocusableElement() {
+      if (!this._focusableNode) {
+        return;
+      }
+
+      if (this.hasAttribute('autofocus')) {
+        this._focusableNode.setAttribute('autofocus', '');
+      } else {
+        this._focusableNode.removeAttribute('autofocus');
+      }
     }
 
     /**
--- a/packages/ui/components/form-core/src/FocusMixin.js
+++ b/packages/ui/components/form-core/src/FocusMixin.js
@@ -48,6 +48,7 @@ const FocusMixinImplementation = superclass =>
        * @type {boolean}
        */
       this.focusedVisible = false;
+      this.autofocus = false;
     }
 
     connectedCallback() {
@@ -62,15 +63,11 @@ const FocusMixinImplementation = superclass =>
     }
 
     /**
-     * Gets called when an attribute is changed.
-     * @param {String} name
-     * @param {String | null} _old
-     * @param {String | null} value
-     * @protected
+     * @param {import('lit').PropertyValues } changedProperties
      */
-    attributeChangedCallback(name, _old, value) {
-      super.attributeChangedCallback(name, _old, value);
-      if (name === 'autofocus') {
+    updated(changedProperties) {
+      super.updated(changedProperties);
+      if (changedProperties.has('autofocus')) {
         this.__syncAutofocusToFocusableElement();
       }
     }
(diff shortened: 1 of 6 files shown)


packages/ui/components/form-core/src/FocusMixin.js
```diff
@@ -52,13 +53,43 @@ const FocusMixinImplementation = superclass =>
...
+    /**
+     * @private
+     */
+    __syncAutofocusToFocusableElement() {
+      if (!this._focusableNode) {
+        return;
+      }
+
+      if (this.hasAttribute('autofocus')) {
+        this._focusableNode.setAttribute('autofocus', '');
```
tlouisse, line 87: Hey, nice PR :)

Could this be tested in an integration test with our dialog system as well?

The spec says:
>No more than one element in the document or dialog may have the autofocus attribute. If  applied to multiple elements the first one will receive focus.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

Since recently we [use a native `<dialog>`](https://github.com/ing-bank/lion/pull/1712) under the hood.
I suspect that if both the host and _focusableNode have focus, the platform might attempt to put focus on the parent element (in case we don't configure the OverlayController with `trapsKeyboardFocus: true`, in which case [it calls contain-focus, like mentioned here](https://github.com/ing-bank/lion/pull/1776#event-9007140697))

Could we rule this out via an integration test in the overlay system?

packages/ui/components/form-core/src/FocusMixin.js
```diff
@@ -52,13 +53,43 @@ const FocusMixinImplementation = superclass =>
...
     }
 
+    /**
+     * Gets called when an attribute is changed.
+     * @param {String} name
+     * @param {String | null} _old
+     * @param {String | null} value
+     * @protected
+     */
+    attributeChangedCallback(name, _old, value) {
```
tlouisse, line 71: Is there a particular reason to use attributeChangedCallback? 
Not necessarily wrong ofc, but we usually rely on the `updated` callback from lit for perf/consistency

packages/ui/components/form-core/types/FocusMixinTypes.ts
```diff
@@ -25,6 +25,12 @@ export declare class FocusHost {
    */
   blur(): void;
 
+  /**
+  * Synchronizes property values when attributes change.
+  * @category attributes
+  */
+    attributeChangedCallback(name: string, _old: string | null, value: string | null): void;
```
tlouisse, line 32: Was TS complaining about it? I expect this to be already available on HTMLElement, which LItElement inherits from

packages/ui/components/form-core/src/FocusMixin.js
Sciurus7: Hi @tlouisse, honestly it's been a while since I actually created the original pull request, so I'm not quite sure why I decided to go with attributeChangedCallback. My suspicion would be, that, as we're dealing with an attribute this felt like the more appropriate place to listen to the change. That being said I've tested the change with updated and it works just as well, so I've added a commit using the updated, this also means the change to FocusMixinTypes is not required anymore.

packages/ui/components/form-core/src/FocusMixin.js
Sciurus7: Hi Thijs, good point. I've tried to write a test for what happens if "trapsKeyboardFocus: false" is set. 
The good news is that it won't focus the host. The potentially bad news is that it will not focus the _focusableNode either. 

I've done some manual testing on the desktop versions of Firefox, Chromium and Safari and found the following. In all cases I tested with autofocus synced: 

1. Test case: A LionDialog with trapsKeyboardFocus set to true and an LionCheckbox with autofocus. Result: The focusable node was focused.
2. Test

If trapsKeyboardFocus is set to true autofocus works as intended. If trapsKeyboardFocus is set to false autofocus will not work, pressing tab once will set the focus to the first focusable element in the dialog. 
This holds true both for a LionField with autofocus or a native html input with autofocus. Using a LionField with autofocus in a native dialog will focus the focusable_node, not the host

gerjanvangeest: Yes indeed for nr 2. Its not a new page, so the focus stays where it is.

I've added some focus tests to the dialog and form-integration.

tlouisse: Hey @Sciurus7 , thanks for all the great help and testing this. Sorry that I postponed my repsonse (wanted to test some things in the browser but didn't get to it...)
Luckily Gerjan took over so I'm happy to merge this :)

Knorgias (approved):

Issue #1964, opened 2023-04-14
[LionSelectRich] form event not dispatched


MatthieuLebigre:
## Expected behavior

When I select an option from a LionSelectRich component, I expect it to dispatch the `select` event, and/or the lion `user-input-changed` event 

## Actual Behavior

When I select an option from a LionSelectRich component, no event is dispatched

See https://studio.webcomponents.dev/edit/RWQMWWVDU73unFNft6NP/src/index.js?p=README.md

## Additional context

@lion/ui: "0.2.0",

DannyMoerkerke:
The only event that is currently dispatched is `model-value-changed`.

tlouisse:
Would it also be possible to do for now:

```js
<lion-select-rich @model-value-changed="${({detail}) => {
  if(detail.isTriggeredByUser) {
    // do my thing
  }
}}">
...
<lion-select-rich>
```
See: https://lion-web.netlify.app/fundamentals/systems/form/model-value/#event-meta


## Why?
`input-node-changed` was meant to **internally normalize** [interaction element / `._inputNode`](https://lion-web.netlify.app/guides/how-to/create-a-custom-field/#b-your-lionfield-extension) events inside FormatMixin...[^1]
To explain the need for normalization a little better:
- input/textarea fire `input` event on keystroke
- select fire  `change` event on keystroke[^2]
- my-custom-interaction-element fires `my-custom-change` event

[`FormatMixin` should only be bothered with one (protected) event](https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/src/FormatMixin.js#L522), w/o knowing implementation details of _inputNode.
It's up to a Subclasser creating a [LionField extension](https://lion-web.netlify.app/guides/how-to/create-a-custom-field/) to normalize an event.

Hindsight wisdom: `_normalized-input-node-changed` would have caused less confusion as a name 😅 

## For future discussion...

I would be really happy to see if we can improve our apis. 
The outcome might be that we agree to use `user-input-changed` everywhere as a public event :)
Then we would need to apply this in a consistent way throughout the whole codebase.
But maybe we want to align more with the platform as we integrate [Form Participation/ElementInternals](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) in the future and we go for just a `change` or `input` event. Also see: https://github.com/ing-bank/lion/discussions/24 
Really happy to hear your suggestions here as well! 🙂 


[...]

PR #1971, opened 2023-04-26
Fix/user input changed listbox

gerjanvangeest:
## What I did

fix: https://github.com/ing-bank/lion/issues/1964


tlouisse:
Also see https://github.com/ing-bank/lion/issues/1964#issuecomment-1529928784 
If we use `user-input-changed` as a public event, we would need to adjust our comments and docs as well, but we probably need to internally discuss a consistent future plan first.

The less apis that do the same thing, the better.  But as always, we should really work on docs and api tables now.

PR #1982, opened 2023-05-19
Fix/lion suggested changes

an8h:
## What I did

1. Unified the process for translations. To keep code consistent, readable, and maintainable I create a map object that contains all the locales and based on this mapping we can easily use the proper translation file. 
2. Added sanitization for string validators. First, it was only checking if the passed value is a string but to prevent XSS attacks a sanitization should happen after the existing check.


gerjanvangeest:
Hi @an8h thanks for taking the time for this contribution.

1. You should split 2 distinct works (translation files work + XSS work) into 2 separate PRs. Chances are we can merge work much quicker then. Currently the browser tests are failing. It might be that "translations file work" part of the PR is causing it, which blocks the other part.
2. From security point of view, the code for escaping HTML is very basic, and can be exploited, and not the best practice. The best practice is to use either [Sanitizer API (once fully supported)](https://caniuse.com/mdn-api_sanitizer) or a proven library to sanitize like [js-xss](https://github.com/leizongmin/js-xss) or [DOMPurify](https://github.com/cure53/DOMPurify). These are bigger libraries though, we should discuss this within the team.
3. Having the locale as a `var` in the string is something we also had in the beginning. The reason we dropped it, is bc Webpack couldn't resolve dynamic imports with variables. Probably it is solved by now, but we like to support older versions of Webpack.

So that would currently be a no for either fixes. But if you are eager to contribute we would love you to help us with:

- **cleaner/more consistent**
  It would be nice if we can remove the bloat from all files by moving it in a separate file next to the main component file. And we should probably add a comment everywhere that we need the bloat to support optimal bundling with webpack.
- **better documented with examples** 
  The sanitize example could be added as an example. Instead of a validator, it could be written as a live formatter (so that it actually sanitizes the input value). As text in the example we can mention the sanitization libs as mentioned above.

gerjanvangeest:

PR #1983, opened 2023-05-20
fix: lion-checkbox-indeterminate initial state closes #1967

### LINKED ISSUE #1967: [LionCheckboxIndeterminate]
## Expected behavior

`LionCheckboxIndeterminate` should have the appropriate state (checked, unchecked, or indeterminate) based on the state of its child checkbox elements.

## Actual Behavior

The last child checkbox element state is never processed and hence the indeterminate checkbox state is incorrect in some cases.

https://studio.webcomponents.dev/edit/lqw33LqYQlzL0LC3rwzI

## Additional context

`lion-checkbox-group 0.21.1`

The problem seems to be in LionCheckboxIndeterminate on line 209:
```javascript
_onRequestToAddFormElement() {
    this._setOwnCheckedState();
}
```

According to the comments for `FormRegistrartMixin`, `_onRequestToAddFormElement` is called _before_ the element is added, so the current checkbox element being added will never be visible when `_setOwnCheckedState` is called.

MiB-1:
…rminate closes #1967

## What I did

1. When we load the `lion-checkbox-indeterminate` with some initial checked child checkbox, it is not showing the expected state initially by adding a async function solves this issue.

### DIFF
--- a/.changeset/small-apricots-allow.md
+++ b/.changeset/small-apricots-allow.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix: initial state is not working as expected in lion-checkbox-indeterminate
--- a/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -207,7 +207,7 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
    * @protected
    */
   _onRequestToAddFormElement() {
-    this._setOwnCheckedState();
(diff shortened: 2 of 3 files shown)


MiB-1:
thanks. now I'll push with username with the new test. but can't understand what exactly needs to be tested. could you please explain a bit here?

gerjanvangeest:
"Having the `lion-checkbox-indeterminate` with an initial checked child checkbox showing the expected state."

MiB-1:
Thanks @gerjanvangeest. I found the issue I'm not doing `clock.restore()`. I'll add the test case and I'll push.

PR #1988, opened 2023-05-25
feat(@lion/ui): add _currencyDisplayNodeSlot to LionInputAmount

gyulus3:
## What I did

Added a _currencyDisplayNodeSlot variable to the LionInputAmount component to be able to customize the slot to put the currency label.

### DIFF
--- a/.changeset/orange-eels-pay.md
+++ b/.changeset/orange-eels-pay.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+Allow customization of currency label's slot in Input Amount
--- a/docs/components/input-amount/use-cases.md
+++ b/docs/components/input-amount/use-cases.md
@@ -20,15 +20,17 @@ export const negativeNumber = () => html`
 `;
 ```
 
-## Set currency suffix
+## Set currency label
 
-You can optionally set a currency suffix with the `currency` attribute.
+You can optionally set a currency label with the `currency` attribute, which is visible in the after slot by default.
 
 ```js preview-story
 export const currencySuffix = () => html`
   <lion-input-amount label="Price" currency="USD" .modelValue=${123456.78}></lion-input-amount>
 `;
 ```
+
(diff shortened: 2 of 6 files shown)


gerjanvangeest:
Hi @gyulus3 thanks for the contribution.

Could you add tests and some documentation for this feature?

gyulus3:
Hey @gerjanvangeest!

Of course, I can do that! Investigating the component I found something strange, what I believe is a leftover from a previous state of the code.

The currency label can be visible only in the _after_ slot as it is said in the code [here](https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-amount/src/LionInputAmount.js#L182). But the [documentation of the LionInputAmount](https://github.com/ing-bank/lion/blob/master/docs/components/input-amount/use-cases.md) mentions that setting the currency attribute will set the currency _suffix_, what is a bit misleading.

Besides the tests of the component I found a [test case](https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-amount/test/lion-input-amount.test.js#L197) for this situation and it says that it should ignore the display of the currency if the _suffix_ slot is already present. Indeed this test case succeeds because the currency is set in the _after_ slot and never overwrites the _suffix_ slot. Later I tested it with a little modification, where I put the currency in the _suffix_ slot, and when this slot was already set then the currency was still visible, which is the opposite behavior that the case wanted to test.

If you agree to this, I can fix the documentation with a little effort and remove the test cases that are not appropriate anymore.

gerjanvangeest:
@gyulus3 That sounds like a great fix!

gyulus3:
@gerjanvangeest I removed the inappropriate test case and created a new one for the requested feature. The docs have been updated, so the use-case contains now the customizable slot for the currency label.

gerjanvangeest (approved):

gyulus3:
@gerjanvangeest when will be the next release? I would be glad if it would be available soon.

gerjanvangeest:
Let me check with my team if we have anything else pending. Otherwise i'll make a release today/tomorrow.

PR #1989, opened 2023-05-26
Feat/workflow improvement

Farzannajipour:
## What I did

1. Updated the versions of actions/setup-node and actions/checkout in the release.yml file. 
2. Updated the `verify.yml` file to include the setup-common-steps job. The setup-common-steps job is responsible for setting up the common steps required for the verify workflow. By adding this job, we ensure that the necessary setup is performed before running the verify job.

The update includes the following changes:

Added the setup-common-steps job to the jobs section.
Added the needs keyword to the verify, browser-tests, node-tests, and node-tests-windows jobs, specifying that they depend on the setup-common-steps job.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -3,18 +3,23 @@ name: Verify changes
 on: pull_request
 
 jobs:
+  setup-common-steps:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+      - uses: google/wireit@setup-github-actions-caching/v1
+
   verify:
     name: Verify changes
     runs-on: ubuntu-latest
+    needs: setup-common-steps
     steps:
-      - uses: google/wireit@setup-github-actions-caching/v1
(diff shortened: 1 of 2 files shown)


.github/workflows/verify.yml
```diff
@@ -3,18 +3,23 @@ name: Verify changes
 on: pull_request
 
 jobs:
-  verify:
-    name: Verify changes
+  setup-common-steps:
```
gerjanvangeest, line 6:
```suggestion
  init:
```

gerjanvangeest (changes requested):
Hi @Farzannajipour,

Thanks for updating these outdated files. But why not go with the latest [actions/checkout](https://github.com/actions/checkout/blob/main/package.json) and [action/setup-node](https://github.com/actions/setup-node/blob/main/package.json) (v3)?

gerjanvangeest (approved):

tkgroot (approved):

tlouisse (approved):

PR #2018, opened 2023-06-20
Fix/remove file

DannyMoerkerke:
## What I did

1. lion-selected-filelist: restored composed: true and bubbles: true for 'file-remove-requested' event, otherwise removing files for subclassers won't work

### DIFF
--- a/packages/ui/components/input-file/src/LionInputFile.js
+++ b/packages/ui/components/input-file/src/LionInputFile.js
@@ -349,7 +349,6 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
       this.setAttribute('drop-zone', '');
     }
 
-    console.log(this._fileListNode);
 
     /** @type {LionSelectedFileList} */ (this._fileListNode).addEventListener(
       'file-remove-requested',
--- a/packages/ui/components/input-file/src/LionInputFile.js
+++ b/packages/ui/components/input-file/src/LionInputFile.js
@@ -176,16 +176,17 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
     this.__duplicateFileNamesValidator = new DuplicateFileNames({ show: false });
   }
 
+  get _fileListNode() {
+    return Array.from(this.children).find(child => child.slot === 'selected-file-list')
+      ?.children[0];
+  }
+
   connectedCallback() {
     super.connectedCallback();
     this.__initialUploadResponse = this.uploadResponse;
 
     this._inputNode.addEventListener('change', this._onChange);
     this._inputNode.addEventListener('click', this._onClick);
-    this.addEventListener(
-      'file-remove-requested',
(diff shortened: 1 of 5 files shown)


tlouisse:
Yeah, noticed this before as well. Would be nice if there was a faling test for this.

I don't think it's a good idea to leak private events, though.

We could replace this:
This can be replaced: https://github.com/ing-bank/lion/blob/a19e02c443e21d4e6c62bb08bbe6db64f7446010/packages/ui/components/input-file/src/LionInputFile.js#L185



```js
get _fileListNode() {
  return Array.from(this.children).filter(child => child.slot === 'selected-file-list');
}

connectedCallback() {
  super.connectedCallback();
  
  // ...
  
  this._fileListNode.addEventListener(
    'file-remove-requested',
     /** @type {EventListener} */ (this._onRemoveFile),
  );
}

// and remove on disconnected
```

The event then only needs to communicate with its parent without "polluting" the rest of the dom tree.

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -347,6 +348,12 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
       this.__setupDragDropEventListeners();
       this.setAttribute('drop-zone', '');
     }
+
```
tlouisse, line 351:
```suggestion
```

tlouisse (approved):

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -176,16 +176,17 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
...
+    return Array.from(this.children).find(child => child.slot === 'selected-file-list')
+      ?.children[0];
+  }
+
   connectedCallback() {
     super.connectedCallback();
     this.__initialUploadResponse = this.uploadResponse;
 
     this._inputNode.addEventListener('change', this._onChange);
     this._inputNode.addEventListener('click', this._onClick);
```
tlouisse, line 189:
```suggestion
    this._inputNode.addEventListener('click', this._onClick);
    
    this._fileListNode.addEventListener(
      'file-remove-requested',
      /** @type {EventListener} */ (this._onRemoveFile),
    );
```

packages/ui/components/input-file/test/lion-input-file.test.js
```diff
@@ -852,7 +852,8 @@ describe('lion-input-file', () => {
       // @ts-expect-error [allow-protected-in-test]
       expect(el._selectedFilesMetaData[1].systemFile.name).to.equal('file2.txt');
 
-      el.dispatchEvent(
+      // @ts-expect-error [allow-protected-in-test]
+      el._fileListNode.dispatchEvent(
         new CustomEvent('file-remove-requested', {
           composed: true,
           bubbles: true,
```
tlouisse, line 859:
```suggestion
```

PR #2033, opened 2023-07-06
fix: fixed infinite loop on progress indicator on max < min

NeelSinha:
## What I did

1. Added a check for max < min on updated lifecycle of progress indicator, throws an error

### DIFF
--- a/packages/ui/components/progress-indicator/src/LionProgressIndicator.js
+++ b/packages/ui/components/progress-indicator/src/LionProgressIndicator.js
@@ -148,7 +148,7 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
         this._resetAriaValueAttributes();
         this._setDefaultLabel();
       }
-    } else {
+    } else if (this.min < this.max) {
       if (changedProperties.has('value')) {
         if (!this.value || typeof this.value !== 'number') {
           this.removeAttribute('value');
@@ -177,6 +177,8 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
           this.value = this.max;
         }
       }
+    } else {
+      throw new Error('min greater than max');
     }
   }
 


packages/ui/components/progress-indicator/src/LionProgressIndicator.js
```diff
@@ -177,6 +177,8 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
           this.value = this.max;
         }
       }
+    } else {
+      throw new Error('min greater than max');
```
jorenbroekema, line 181:
```suggestion
      throw new Error(`progress-indicator min value (${this.min}) should always be smaller than max value (${this.max})`);
```

jorenbroekema (changes requested):
Suggestion for more descriptive error message.

Can you also add a test showing the scenario and the fact that it will throw in this case?

packages/ui/components/progress-indicator/src/LionProgressIndicator.js
NeelSinha: it("throws an error when min is greater than max", async () => {

      let errorCaught = false;
     try{ 
      const el = await fixture(
        html`<lion-progress-indicator min="100" max="90" aria-label="foo"></lion-progress-indicator> `,
      );
      // await el.updateComplete;
    }catch(error){ 
        errorCaught = true
        expect(error).to.be.an.instanceOf(Error);
        expect(error.message).to.equal("progress-indicator min value (100) should always be smaller than max value (90)")
      }

      console.log("error", errorCaught)
      expect(errorCaught).to.be.true
    })
    
    
    I added this test but this is not able to catch the error, seems like updated does not get called, this works fine on my setup which I have done with lit elements and I'm using lion components as base.
    
    Can you tell me what I am doing wrong

NeelSinha: @jorenbroekema I have committed the changes, was not able to add test cases to catch this as an error, please have a look :)

jorenbroekema: Ah thanks for the ping, I'll take a look as soon as I can :)

gerjanvangeest:

PR #2039, opened 2023-07-17
feat: parsing body for failed ajax responses

jorenbroekema:
## What I did

1. Add automatic response body parsing for fetchJson failed responses
2. Add test for manually parsing body for fetch failed responses

### DIFF
--- a/.changeset/quiet-insects-shake.md
+++ b/.changeset/quiet-insects-shake.md
@@ -1,0 +1,5 @@
+---
+'@lion/ajax': minor
+---
+
+Parses response body automatically for fetchJson failed responses. Add test for reading out the response body in cases of a failed response in regular fetch call.
--- a/package-lock.json
+++ b/package-lock.json
@@ -22384,7 +22384,7 @@
     },
     "packages/ajax": {
       "name": "@lion/ajax",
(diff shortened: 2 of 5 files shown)


gerjanvangeest (approved):

thepassle:
This contains a bug when using `ajaxApi.fetchJson()`

`fechJson` internally calls `this.fetch`, where `parseErrorResponse` is set to `true`:
```js
const response = await this.fetch(info, jsonInit, true);
```

However, it doesnt first call `fetch` on the `Ajax` class, but on the `AjaxApi` class, which signature looks like this:
```js
fetch(apiPath, init) {} 
```

The `AjaxApi` class's `fetch` method then calls: 

```js
super.fetch(getApiUrl(apiPath), {
  credentials: "include",
  ...init,
});
```

reaching `Ajax`'s `fetch` method, where `parseErrorResponse` default to false:
```js
async fetch(info, init, parseErrorResponse = false) {}
```

Meaning that it doesn't parse the body for failed requests when using `AjaxApi`. Can you take a look at this? :)

thepassle:
Ah yes indeed, this is in `AjaxApi` which is in ING. My bad. Still need to fix it there too though :)

PR #2049, opened 2023-08-01
fix: clone response when reading the response for error

jorenbroekema:
## What I did

1. Fix a regression that I introduced by reading out the response. People were doing error.response.text() themselves, and now they get an error that the body stream has already been read, because in fetchJson I added the feature that it reads out the body for you and puts it on error.body already.

### DIFF
--- a/packages/ajax/test/Ajax.test.js
+++ b/packages/ajax/test/Ajax.test.js
@@ -215,6 +215,8 @@ describe('Ajax', () => {
         expect(_e.request).to.be.an.instanceOf(Request);
         expect(_e.response).to.be.an.instanceOf(Response);
         expect(_e.body).to.equal('my response');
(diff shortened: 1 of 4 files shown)


packages/ajax/test/Ajax.test.js
```diff
@@ -215,6 +215,8 @@ describe('Ajax', () => {
         expect(_e.request).to.be.an.instanceOf(Request);
         expect(_e.response).to.be.an.instanceOf(Response);
         expect(_e.body).to.equal('my response');
+        const bodyFromResponse = await _e.response.text();
```
jorenbroekema, line 218: Before this change, this would throw.. which is a regression of 1.2.0

Knorgias (approved):

PR #2053, opened 2023-08-02
feat: init nodejs-helpers package, add bypass-import-map, bypass-impo…

narzac:
…rt-map tasks, and run part of lion prepublish

## What I did

1. init nodejs-helpers package
2. add bypass-import-map
3. add bypass-export-map
4. run part of lion prepublish


packages-node/nodejs-helpers/test-node/index.test.js
```diff
@@ -0,0 +1,33 @@
...
+  asyncSerialForEach,
+  asyncConcurrentForEach,
+  // Tasks
+  bypassImportMap,
+  bypassExportMap,
+} from '../src/index.js';
+
+describe('Public API', () => {
+  it('should expose the agreed public API', () => {
+    expect(parseCode).to.exist;
```
tlouisse, line 20: maybe we can test whether they are functions, since otherwise the import statement above would already have thrown an error?

narzac: Done

packages-node/nodejs-helpers/test-node/tasks/bypass-export-map.test.js
```diff
@@ -0,0 +1,328 @@
...
+      const resolvedPath = require.resolve(initialSource, { paths });
+      resolvedPaths.push(resolvedPath);
+    },
+  };
+  transformCode(code, visitor);
+  return resolvedPaths;
+};
+
+describe('bypassExportMap simple export map case', async () => {
+  const packageDir = 'test-node/fixtures/simple-export-map';
```
tlouisse, line 104: Did you consider using mock-fs (for readability and perf)?

Then we can do:

```js
mockFs({
  '/simple-export-map/components/icon/xyz.js': "export x = () => 'bla';"
})

```

packages-node/nodejs-helpers/src/tasks/bypass-import-map.js
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Bypasses import map for a package given by `packageDir`
+ *
+ * @param {string} packageDir
+ * @param {{ignoredDirs?: string[]}} [options]
+ * @throws ERROR_CAN_NOT_ACCESS_PACKAGE_DIR
+ * @returns {Promise<any>}
+ */
+export const bypassImportMap = async (packageDir, options = {}) => {
+  const ignoredDirs = options.ignoredDirs || ['node_modules'];
```
tlouisse, line 131: For a future improvement: maybe we can leverage globby's gitignore option?

packages-node/nodejs-helpers/src/tasks/bypass-export-map.js
```diff
@@ -0,0 +1,193 @@
...
+ * @param {string} packageDir
+ * @param {string} exportMapKey
+ * @param {string} exportMapValue
+ * @returns {Promise<any>}
+ */
+const bypassExportMapItem = async (packageDir, exportMapKey, exportMapValue) => {
+  const exportMapValuePath = path.join(packageDir, exportMapValue);
+  const exportMapKeyPath = path.join(packageDir, exportMapKey);
+  const searchPattern = isDirectorySync(exportMapValuePath)
+    ? path.join(exportMapValuePath, '**', '*.js')
```
tlouisse, line 131: At some point, we might also support css import assertions. At that point, we can probably extract the generic export map resolver from providence and use it here as well.

tlouisse (approved):

packages-node/nodejs-helpers/test-node/tasks/bypass-export-map.test.js
narzac: Yes I did consider mocking filesystem, however it is another layer of abstraction on top of the reality, which will make it less readable imo, and more of a complex solution. Of course, as you suggested we could still consider it, say for performance, and that would certainly be the case, if we were mocking timers and so on, yet performance wise the whole tests that access the filesystem are taking ~0.247 seconds to be precise.

To get the ~0.247, we can

First run the tests without the tests which uses the filesystem

`rm -rf packages-node/nodejs-helpers/test-node/tasks/ && time npm run test -w packages-node/nodejs-helpers/`
 
real	0m1.323s

And then, run it with them: 
`g checkout packages-node/nodejs-helpers/test-node/tasks/ && time npm run  test -w packages-node/nodejs-helpers/ `

real	0m1.570s

Then the difference is ~0.247,, which is insignificant.

PR #2064, opened 2023-08-24
Ajax fetch error fixes

koddsson:
## What I did

1. I set the `name` property of the error so that the name is preserved. Without setting the `name` property it's overwritten when calling `super`.

### Input
```js
class MyError extends Error {
    constructor() {
        super('My error!')
    }
}
try { throw new MyError('foobar') } catch(err) { console.log(err.name) }
```

### Output
```
Error
```

Compare that with the same code where we actually set the `name` property:

### Input
```js
class MyError extends Error {
    constructor() {
        super('My error!')
        this.name = 'MyError'
    }
}
try { throw new MyError('foobar') } catch(err) { console.log(err.name) }
```

### Output
```
MyError
```

2. Added the response status code and text to the message.

### DIFF
--- a/.changeset/hot-melons-grab.md
+++ b/.changeset/hot-melons-grab.md
@@ -1,0 +1,5 @@
+---
+'@lion/ajax': patch
+---
+
+Preserve error name and enrich it with the response status
--- a/package-lock.json
+++ b/package-lock.json
@@ -2893,10 +2893,6 @@
         "node": ">= 8.0.0"
       }
     },
-    "node_modules/@lion/nodejs-helpers": {
-      "resolved": "packages-node/nodejs-helpers",
-      "link": true
-    },
     "node_modules/@lion/accordion": {
       "version": "0.9.0",
       "resolved": "https://registry.npmjs.org/@lion/accordion/-/accordion-0.9.0.tgz",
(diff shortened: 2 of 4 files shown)


koddsson:
And what do you think about linting for this in the future?

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6d15a02d48de7ecfc38d0683a8487b2f937d83a0/docs/rules/custom-error-definition.md

tlouisse:
> And what do you think about linting for this in the future?
> 
> https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6d15a02d48de7ecfc38d0683a8487b2f937d83a0/docs/rules/custom-error-definition.md

Seems like something that you would easily forget, so good idea to add the rule. Would you like to add a PR for it?

koddsson:
@tlouisse; I think this latest change should be good but CI isn't running. You probably need to push the button again :)

tlouisse:
> @tlouisse; I think this latest change should be good but CI isn't running. You probably need to push the button again :)

Awesome. We will merge it soon to master (master is currently in a 'prerelease mode', it will be merged when it's back in 'normal mode') and let you know when it will be released on npm.

PR #2068, opened 2023-09-01
fix(progress-indicator): accept 0 as a valid value

gafortuno:
## What I did

1. Updated the condition in removing the value attribute.
2. When the value is set to 0, currently, without this fix, the value is being removed because the condition is **!this.value || typeof this.value !== 'number'** => !0 is equal to true that's why the progress bar value is being removed (**this.removeAttribute('value')**)
3. I updated the condition to check if the value is 0, it will be accepted and the value attribute will not be removed.
4. I also added and updated the unit test. The current unit test on setting the value of the value attribute to an empty string will fail because according to this [documentation](https://lit.dev/docs/components/properties/), on the convertion for Number (attribute to property) **_If the element has the corresponding attribute, set the property to Number(attributeValue)_**. So, if the value is set to '', it will be Number('') which is equal to 0.

### DIFF
--- a/.changeset/early-bats-dance.md
+++ b/.changeset/early-bats-dance.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(progress-indicator): accept 0 as a valid value
--- a/packages/ui/components/progress-indicator/src/LionProgressIndicator.js
+++ b/packages/ui/components/progress-indicator/src/LionProgressIndicator.js
@@ -150,7 +150,7 @@ export class LionProgressIndicator extends LocalizeMixin(LitElement) {
       }
     } else {
       if (changedProperties.has('value')) {
-        if (!this.value || typeof this.value !== 'number') {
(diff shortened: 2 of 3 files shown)


gafortuno:
Before the fix:

**If value is set to `0`, it's taking the whole progress:**
Screenshot 2023-09-01 at 6 58 10 PM

**width is not being set properly because the value attribute is removed and it is considered `indeterminate`**
Screenshot 2023-09-01 at 6 58 25 PM

**value becomes `null`**
Screenshot 2023-09-01 at 6 58 48 PM

After the fix:
Screenshot 2023-09-01 at 12 14 48 PM
Screenshot 2023-09-01 at 12 14 58 PM

gerjanvangeest:
Looks good to me, can you add a changeset `npm run changeset`, to add the change to the changelog?

gafortuno:
Added changeset thanks @gerjanvangeest

gerjanvangeest (approved):

PR #2070, opened 2023-09-06
Fix undefined `document.body`

fabiofdsantos:
If we e.g. `import "@lion/ui/define/lion-dialog.js";`, `document.body` can be null.

`document.body` is a read-only property that gets set by the browser when it parses the HTML document.

As it's used only inside `__keepBodySize`, no need to declare `supportsCSSTypedObject` before.

### DIFF
--- a/.changeset/dull-tools-protect.md
+++ b/.changeset/dull-tools-protect.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix: undefined document.body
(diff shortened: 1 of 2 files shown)


gerjanvangeest:
Looks fine to me, can you add a changeset (`npm run changeset)` to get the change added to the changelog?

fabiofdsantos:
> Looks fine to me, can you add a changeset (`npm run changeset)` to get the change added to the changelog?

@gerjanvangeest Done! Thanks! 👍🏻

gerjanvangeest (approved):

Issue #2075, opened 2023-09-12
[input-datepicker] Validators should not disable dates when another type of validator then error is used


gerjanvangeest:
## Expected behavior

When a validator other then "error" is used, the dates applied to MinDate or MaxDate should stay enabled.

## Actual Behavior

Dates are disabled, no matter the type of validator.

```html
<lion-input-datepicker
    name="start-date"
    label="Start date"
    .modelValue="${new Date()}"
    .validators="${[new MinDate(new Date(), { type: 'warning' })]}"
></lion-input-datepicker>
```

PR #2076, opened 2023-09-12
fix(input-datepicker): only disable dates if validator is of type error

gerjanvangeest:
## What I did

1. fix: https://github.com/ing-bank/lion/issues/2075

### DIFF
--- a/packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
+++ b/packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
@@ -323,6 +323,29 @@ describe('<lion-input-datepicker>', () => {
         expect(elObj.calendarEl.maxDate).to.equal(myMaxDate);
       });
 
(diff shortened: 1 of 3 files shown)


packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -323,6 +323,29 @@ describe('<lion-input-datepicker>', () => {
         expect(elObj.calendarEl.maxDate).to.equal(myMaxDate);
       });
 
+      it('does not converts MaxDate validator to "maxDate" property if validator type other then "error"', async () => {
```
Knorgias, line 326: Would be a good practice to test for all other validator types other than "error", besides "warning", wdyt?

gerjanvangeest: On lion we only have `error`.

tkgroot (approved):

Knorgias (approved):

Issue #2084, opened 2023-10-05
[LionCombobox]: Only searching for part of the input string


gyulus3:
## Reproduction recipe
 - Type 'Appl' in the Combobox input.
 - Remove the second 'p' letter from the word using the keyboard.

## Expected behavior

The combobox searches for values containing the word 'Apl'.

## Actual Behavior

The combobox searches for values containing the word 'Ap'.

PR #2085, opened 2023-10-05
fix(combobox): Fix search issues when modifying the middle of the input word

gyulus3:
## What I did
The _handleAutocompletion recognized a selection in the input even if the selection was empty.

Corresponding issue and the problem explanation: fixes https://github.com/ing-bank/lion/issues/2084


gerjanvangeest (approved):
@gyulus3 Thanks for the fix 💪!

PR #2087, opened 2023-10-06
fix(overlay): avoid calling push in iterable object to avoid error lo…

### LINKED ISSUE #2086: Error with overlays in lion/ui 0.4.X with old chromium versions
## Components with overlays should work correctly
Whe  you load a component that uses overlay it should work correctly

## Components that use overlay doesn't work in old chromium versions

We are using lion/ui 0.4.0 o 0.4.1. Component that use that version can't load overlays in old chromium verisons. If you try to load a component with overlay in Chromium Versión 87.0.4280.0 (Build for developers) (x86_64) you will receive in console the error I attach to the issue report. The error comes from "packages\ui\components\overlays\src\utils\adopt-styles.js". It looks like adoptedStyleSheets is an iterable but it is not an Array and it doesn't have the push function. 

## How to fix it
Use [...adoptedStyleSheets , style] instead adoptedStyleSheets.push

Console error example:
aa79cc27-f66a-45ef-a1db-6dbf290f1229

alejandrofernandezcanton:
…ading overlays in old chromium versions

## What I did

1. Change .push in iterator to avoid error loading overlays. This only happens in old chromium version but we support old versions. 

Resolves #2086

### DIFF
--- a/.changeset/forty-cats-unite.md
+++ b/.changeset/forty-cats-unite.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+avoid calling push in iterable object to avoid error loading overlays in old chromium versions
--- a/package-lock.json
+++ b/package-lock.json
@@ -22698,7 +22698,7 @@
     },
     "packages/ajax": {
       "name": "@lion/ajax",
-      "version": "1.2.1",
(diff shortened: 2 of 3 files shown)


alejandrofernandezcanton:
problem with cla

PR #2088, opened 2023-10-09
fix(overlay): avoid calling push in iterable object to avoid error lo…

alejandrofernandezcanton:
…ading overlays

## What I did

Change .push in iterator to avoid error loading overlays. This only happens in old chromium version but we support old versions.
Resolves https://github.com/ing-bank/lion/issues/2086

### DIFF
--- a/.changeset/forty-cats-unite.md
+++ b/.changeset/forty-cats-unite.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+avoid calling push in iterable object to avoid error loading overlays in old chromium versions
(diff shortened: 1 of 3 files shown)


gerjanvangeest (approved):
@alejandrofernandezcanton Thanks for the fix 💪!

alejandrofernandezcanton:
@gerjanvangeest thanks for merging the fix. Could you tell me when the next release will be?

gerjanvangeest:
We normally do it on request, so i just created the release :)

alejandrofernandezcanton:
Thanks you @gerjanvangeest

Issue #2089, opened 2023-10-11
[LionInputAmount]: Doesn't link currency to label when currency is set after first render


gyulus3:
## Reproduction recipe
- Open an Input Amount component with no currency set and label set to 'Amount'.
- Set the currency using the browser console to 'USD'.

## Expected behavior

The currency name is part of the accessible name of the input, the value is 'Amount US dollars'.

## Actual Behavior

The currency name is not part of the accessible name of the input, the value is 'Amount' only.

PR #2091, opened 2023-10-11
fix(input-amount): Link currency to label after setting currency from…

gyulus3:
Now it adds the currency label id to aria-labelledby of input when the currency was set from undefined to a valid currency.

Corresponding issue and the problem explanation: fixes https://github.com/ing-bank/lion/issues/2089


gerjanvangeest (approved):
@gyulus3 thanks for the fix 💪!

PR #2096, opened 2023-10-16
added file last modified validator

labibismaiel:
## File last modified date validator
This validator can be used to ensure that the uploaded file was last modified within a specific date range. This can be useful in scenarios where you want to ensure that users are uploading recent files, such as recent documents, photos, or reports.

To use this validator, you would provide the desired date range as parameters when adding it to a form element:
```
const fileInput = document.querySelector('lion-input-file');
const validator = new  FileLastModifiedDateValidator({ 
	startDate: new  Date('2022-01-01'),
	endDate: new  Date('2023-01-01') 
});
fileInput.addValidator(validator);
```
This would ensure that only files modified between January 1, 2022, and January 1, 2023, are considered valid.

### DIFF
--- a/packages/ui/components/form-core/src/validate/FileLastModifiedDateValidator.js
+++ b/packages/ui/components/form-core/src/validate/FileLastModifiedDateValidator.js
@@ -1,0 +1,33 @@
+import { Validator } from './Validator.js';
+
+export class FileLastModifiedDateValidator extends Validator {
+  static validatorName = 'FileLastModifiedDate';
+
+  /**
+   * Checks if the file's last modified date is within the specified range.
+   * @param {Date} fileLastModifiedDate - The file's last modified date.
+   * @param {Date} startDate - The start date of the allowed range.
+   * @param {Date} endDate - The end date of the allowed range.
+   * @returns {boolean}
+   */
+  static isDateWithinRange(fileLastModifiedDate, startDate, endDate) {
+    return fileLastModifiedDate >= startDate && fileLastModifiedDate <= endDate;
+  }
+
+  /**
+   * @param {Array<File>} modelValue - Array of file objects.
+   * @param {{ startDate: Date; endDate: Date; }} params - Validation parameters.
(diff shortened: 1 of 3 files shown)


gerjanvangeest:
Hi @labibismaiel, 

Thanks for the PR 💪 !!

- Since its an input-file specific valdator, could you move this code to the [input-file](https://github.com/ing-bank/lion/tree/master/packages/ui/components/input-file/src)? 
- Add a changeset `npm run changset`, so the change will end up in our changelog?
- Fix the failing tests
- Sign the CLA

packages/ui/components/form-core/src/validate/FileLastModifiedDateValidator.js
```diff
@@ -0,0 +1,33 @@
+import { Validator } from './Validator.js';
+
+export class FileLastModifiedDateValidator extends Validator {
+  static validatorName = 'FileLastModifiedDate';
+
+  /**
+   * Checks if the file's last modified date is within the specified range.
+   * @param {Date} fileLastModifiedDate - The file's last modified date.
+   * @param {Date} startDate - The start date of the allowed range.
```
gerjanvangeest, line 9:
```suggestion
   * @param {Date} minDate - The start date of the allowed range.
```

packages/ui/components/form-core/src/validate/FileLastModifiedDateValidator.js
```diff
@@ -0,0 +1,33 @@
...
+   * @param {Date} endDate - The end date of the allowed range.
```
gerjanvangeest, line 10:
```suggestion
   * @param {Date} maxDate - The end date of the allowed range.
```

packages/ui/components/form-core/src/validate/FileLastModifiedDateValidator.js
gerjanvangeest: Make this inline with other date validators

Issue #2100, opened 2023-10-17
[LionCombobox]


riovir:
## Expected behavior

The disptatched `model-value-changed` event on calling `clear()` reflects a blank value in `event.target.modelValue`.

## Actual Behavior

Get a stale, before clear() value instead. The following reproduction demonstrates that the `model-value-changed` event is dispatched well before the computed `modelValue` settles on an empty value.

https://studio.webcomponents.dev/edit/cmj4eJXT9nPgn1ydVTpY/src/index.js?p=README.md

## Additional context

This is particularly painful in a Vue app, where the component is wrapped as a Vue component. It listens to `model-value-changed` and emits an `update:modelValue` event, with the `event.target.modelValue`, which is an `Unparseable`. This in then gets fed back into the webcomponent by Vue causing the component to settle on an Unparseable value, rather than on a blank one. However, the issue also impacts anyone relying on the modelValue to be blank after calling clear() when the event dispatches.

gerjanvangeest:
This issue is resolved in the already merged PR above.

PR #2104, opened 2023-10-18
fix(lion-combobox): The event emitted during clear has stale value

GuiAmPm:
## What I did

1. Created a test to cover the issue reported at
https://github.com/ing-bank/lion/issues/2100
Note: I tried to use Spy instead of an arrow function, but I learned that sinon's `.calledWith` compares to the reference of the object used in the call, and since the value object is updated between call and the assertion it reports the "updated" value.

2. Now the value is first set to `''` before calling `super.clean()`, since no other event is called when changing this value, I assume this has no further impact.

### DIFF
--- a/.changeset/poor-trains-drive.md
+++ b/.changeset/poor-trains-drive.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': minor
+---
+
+Fix: Event emited when clearing a combobox sends stale value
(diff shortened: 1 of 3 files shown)


gerjanvangeest (approved):
@GuiAmPm thanks for the fix 💪!

Issue #1950, opened 2023-03-22
[combobox] Implement !requireOptionMatch and multipleChoice flow 


gerjanvangeest:
The combobox with multiple choice does not support the use case in which the user can freely create valid values (requireOptionMatch=false).

https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/src/LionCombobox.js#L561

gerjanvangeest:
Hi @GuiAmPm,

Great to hear that you are willing to deepdive into this!

The suggestions you make sounds good to me, we can always refactor it a bit once it becomes clear what the mixin actually will do. Although i have to say that writing a mixin might be a bit more complex, than adding it directly to the combobox itself.

I would like to suggest to start with some BDD tests, so we can already review that before you did a lot of work in writing the code itself. Feel free to reach out if your stuck.

Gerjan

PR #2108, opened 2023-10-21
Feature: Added support to multiselect and requireOptionMatch=false at the  same time for lion-combobox

GuiAmPm:
## What I did

1. Added support to multiselect combined with requireOptionCombobox=false via a mixin (CustomChoiceMixin) to avoid cluttering both combobox and the listbox/choicegroup mixins 
https://github.com/ing-bank/lion/issues/1950
2. Wrote new tests to cover the scenarios and updated 
3. Added a new example to the documentation


gerjanvangeest (commented):
@GuiAmPm looks great! I asked a colleague also to look at this PR, so he might add some extra feedback.

- I see this is already a bug for `single-option`, but you might want to fix this while at it. Currently if you type `art` it autocompletes to `artichoke`, if you hit `delete` it changes to `art`, since `choke` was already selected. But if you hit enter the `modelValue` will still be `artichoke`. 
- Can we add the `runCustomChoiceMixinSuite` to the `lion-combobox-integrations.test.js`?

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -364,6 +364,38 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
     return this._inputNode;
   }
 
+  /**
+   * @type {number | number[]}
+   */
+  // @ts-ignore
+  get checkedIndex() {
```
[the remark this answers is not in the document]
GuiAmPm: Nope, leftovers!

packages/ui/components/form-core/src/choice-group/CustomChoiceMixin.js
```diff
@@ -0,0 +1,182 @@
...
+      }
+
+      if (values === '') {
+        return this._customChoices.values().next().value || '';
+      }
+
+      return values;
+    }
+
+    /** @param {import('lit').PropertyValues} changedProperties */
```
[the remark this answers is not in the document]
GuiAmPm: I made a mistake with the name, actually it's supposed to dispatch the event only in subsequent events after the initialization.
Groups are not dispatching events if not caused by one of it's children changing, so we have to manually dispatch it when a value without a representing child component is selected.

packages/ui/components/form-core/src/choice-group/CustomChoiceMixin.js
```diff
@@ -0,0 +1,182 @@
...
+      return Array.from(this._customChoices).filter(
+        choice => !elems.some(elem => elem.choiceValue === choice),
+      );
+    }
+
+    constructor() {
+      super();
+
+      this.allowCustomChoice = false;
+
```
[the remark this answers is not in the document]
GuiAmPm: Can you describe a scenario? Not sure I follow.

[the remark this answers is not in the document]
gerjanvangeest: Something like:

```js
  it('ignores empty string modelValue inputs', async () => {
    const el = /** @type {LionCombobox} */ (
      await fixture(html`
        <lion-combobox
          name="foo"
          multiple-choice
          autocomplete="none"
        >
          <lion-option .choiceValue="${'Artichoke'}">Artichoke</lion-option>
          <lion-option checked .choiceValue="${'Chard'}">Chard</lion-option>
          <lion-option .choiceValue="${'Chicory'}">Chicory</lion-option>
          <lion-option .choiceValue="${'Victoria Plum'}">Victoria Plum</lion-option>
        </lion-combobox>
      `)
    );

    el.requireOptionMatch = false;
    await el.updateComplete;
    const { _inputNode } = getComboboxMembers(el);

    mimicKeyPress(_inputNode, 'Enter');
    await el.updateComplete;
    expect(el.modelValue).to.eql([]);

    mimicUserTyping(el, ' ');
    await el.updateComplete;
    mimicKeyPress(_inputNode, 'Enter');
    await el.updateComplete;
    expect(el.modelValue).to.eql([]);
  });
```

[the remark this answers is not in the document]
GuiAmPm: Thanks, I've added the checks and updated unit tests

docs/components/combobox/use-cases.md
```diff
@@ -231,6 +231,7 @@ This will:
 > Please note that the lion-combobox-selection-display below is not exposed and only serves
 > as an example. The selection part of a multiselect combobox is not yet accessible. Please keep
 > in mind that for now, as a Subclasser, you would have to take care of this part yourself.
+> Also keep in mind that the combobox organizes the selected list by it's original index in the option list
```
tlouisse, line 234:
```suggestion
> Also keep in mind that the combobox organizes the selected list by its original index in the option list
```

GuiAmPm: Nice catch, thanks!

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -43,6 +43,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
       requireOptionMatch: {
         type: Boolean,
       },
+      allowCustomChoice: {
```
tlouisse, line 46: I think this leaves us with two properties that have the same goal, but are each others inverse (allowCustomChoice === ! requireOptionMatch). 

We usually have these rule for a boolean flag: 
1. it should be `false` by default, (so the non-default can be enabled with a boolean attribute).
2. it should preferably not be a negation, since usually it's harder to reason about

Why do we have `this.requireOptionMatch = true` in the constructor of combobox, then?
Because in this case:
- we thought allowing a custom value should be the default behavior
- we didn't implement this default yet, bc it would be a breaking change. And we want to do this in the future (so then we would comply to rule 1 above again and in our constructor we would do `this.requireOptionMatch = false` again, and then we could introduce `require-option-match` attribute).

Hope this makes sense to you? :)

If so, could you maybe stick to `requireOptionMatch` everywhere? 
And then it would be good if we explained the above in our code, making clear we want to switch the default behavior of the combobox in. a future release

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * @type {CustomChoiceGroupMixin}
+ * @param {import('@open-wc/dedupe-mixin').Constructor<import('lit').LitElement>} superclass
+ */
+const CustomChoiceGroupMixinImplementation = superclass =>
+  // @ts-ignore https://github.com/microsoft/TypeScript/issues/36821#issuecomment-588375051
+  class CustomChoiceGroupMixin extends ChoiceGroupMixin(superclass) {
+    static get properties() {
+      return {
+        allowCustomChoice: {
```
tlouisse, line 28: Considering the fact that this is designed to cover for multiple scenarios (including radiogroups), allowCustomChoice might be preferable over requireOptionMatch.

Let me give it a second thought, I will come back to it later :)

tlouisse: Can you maybe give us an idea how this mixin would be used outside the combobox? With LionRadioGroup for instance?

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * @typedef {import('../../types/choice-group/CustomChoiceGroupMixinTypes.js').CustomChoiceGroupMixin} CustomChoiceGroupMixin
+ * @typedef {import('../../types/choice-group/CustomChoiceGroupMixinTypes.js').CustomChoiceGroupHost} CustomChoiceGroupHost
+ */
+
+/**
+ * @param {any|any[]} value
+ * @returns {any[]}
+ */
+function normalizeArray(value) {
```
tlouisse, line 13: Nitpicking: could this be called `ensureArray`?

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
GuiAmPm: Your previous explanation makes sense, but indeed that is what I was thinking about the `allowCustomChoice`, and honestly I was expecting someone to comment on the way I went around it, inverting the value via property, as I didn't know about the intention for `requireOptionMatch = true`, and I didn't want to rename it completely as not to break compatibility (though I considered deprecating it, but that's not me to decide it).

I don't have a strong opinion either way other than `allowCustomChoice` matches better with the mixin's name. Let me know once you have a decision and I will gladly update it.

Cheers.

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
GuiAmPm: Will do

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
```diff
@@ -0,0 +1,204 @@
...
+      /**
+       * @type {Set<unknown>}
+       * @protected
+       */
+      this._customChoices = new Set();
+
+      /**
+       * @private
+       */
+      this.__initialized = false;
```
tlouisse, line 129: Is it possible to reuse `hasUpdated` instead? https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L1489

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -615,6 +619,35 @@ describe('lion-combobox', () => {
...
+      el.requireOptionMatch = false;
+      const { _inputNode } = getComboboxMembers(el);
+
+      mimicUserTyping(el, 'Art');
+      await el.updateComplete;
+
+      await mimicUserTypingAdvanced(el, ['Delete']);
+      await el.updateComplete;
+      await el.updateComplete;
+
```
tlouisse, line 642: Could we also add a test here that the listbox closes here (if I read the correctly)?

Is this behavior on delete somewhere documented (like a wai-aria spec)?

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
```diff
@@ -0,0 +1,198 @@
...
+
+      if (values === '') {
+        return this._customChoices.values().next().value || '';
+      }
+
+      return values;
+    }
+
+    /** @param {import('lit').PropertyValues} changedProperties */
+    updated(changedProperties) {
```
tlouisse, line 149: Could you elaborate on what happens here exactly?

(If I delete the `updated` callback, the tests still run)

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
tlouisse: Great, makes sense. I must say that I like `allowCustomChoice` better (it's easier to read), so happy to go with that (especially now it makes more sense in a broader perspective) :)

It would maybe be nice to make a small example (can be in the test of the CustomChoiceGroupMixin or as a demo of the mixin) with a radiogroup/checkbox/listbox with a custom text field.

If it means CustomChoiceGroupMixin can be applied to all ChoiceGroups, we might as well consider merging the allowCustomChoice functionality into ChoiceGroupMixin?

GuiAmPm: > It would maybe be nice to make a small example (can be in the test of the CustomChoiceGroupMixin or as a demo of the mixin) with a radiogroup/checkbox/listbox with a custom text field.
> If it means CustomChoiceGroupMixin can be applied to all ChoiceGroups, we might as well consider merging the allowCustomChoice functionality into ChoiceGroupMixin?

The way it's implemented currently we need to add the mixin to those other controls so I can make those examples. Makes sense to merge that to ChoiceGroupMixin, that's just raise the question: Do we want all the users have to handle Custom Choices every time a new ChoiceGroupMixin is created?

Also, do want that done in this one PR?
Also can I assume that renaming the property is not included?  Or this will break current users?

packages/ui/components/form-core/src/choice-group/CustomChoiceGroupMixin.js
GuiAmPm: I've added to fix a issue I had where it was not triggering the events properly, but indeed it's not necessary now. Might have solved while cleaning up the code at some point and forgot to remove it.

packages/ui/components/combobox/test/lion-combobox.test.js
GuiAmPm: I was not looking for that behavior specifically.
Before (probably still present in master), if you pressed enter even when the popup is closed, you would select the previously selected value. Delete was already closing the listbox before.

packages/ui/components/combobox/src/LionCombobox.js
tlouisse: See below how `allowCustomChoice` makes more sense in the broader picture (it just means that "no custom choice" will stay the default of combobox then and all reqs are still met)

packages/ui/exports/form-core-test-suites.js
```diff
@@ -8,3 +8,4 @@ export { runInteractionStateMixinSuite } from '../components/form-core/test-suit
 export { runNativeTextFieldMixinSuite } from '../components/form-core/test-suites/NativeTextFieldMixin.suite.js';
 export { runValidateMixinSuite } from '../components/form-core/test-suites/ValidateMixin.suite.js';
 export { runValidateMixinFeedbackPart } from '../components/form-core/test-suites/ValidateMixinFeedbackPart.suite.js';
+export { runCustomChoiceGroupMixinSuite } from '../components/form-core/test-suites/choice-group/CustomChoiceGroupMixin.suite.js';
```
tlouisse, line 11:
```suggestion
```
For now, we don't want to expose the mixin yet (as it probably will be merged into ChoiceGroupMixin)

packages/ui/exports/form-core.js
```diff
@@ -49,5 +49,6 @@ export { LionValidationFeedback } from '../components/form-core/src/validate/Lio
 
 export { ChoiceGroupMixin } from '../components/form-core/src/choice-group/ChoiceGroupMixin.js';
 export { ChoiceInputMixin } from '../components/form-core/src/choice-group/ChoiceInputMixin.js';
+export { CustomChoiceGroupMixin } from '../components/form-core/src/choice-group/CustomChoiceGroupMixin.js';
```
tlouisse, line 52:
```suggestion
```
For now, we don't want to expose the mixin yet (as it probably will be merged into ChoiceGroupMixin)

packages/ui/exports/types/form-core.ts
```diff
@@ -1,5 +1,6 @@
 export { ChoiceGroupHost } from '../../components/form-core/types/choice-group/ChoiceGroupMixinTypes.js';
 export { ChoiceInputHost } from '../../components/form-core/types/choice-group/ChoiceInputMixinTypes.js';
+export { CustomChoiceGroupHost } from '../../components/form-core/types/choice-group/CustomChoiceGroupMixinTypes.js';
```
tlouisse, line 3:
```suggestion
```
For now, we don't want to expose the mixin yet (as it probably will be merged into ChoiceGroupMixin)

packages/ui/components/form-core/test/choice-group/CustomChoiceGroupMixin.test.js
```diff
@@ -0,0 +1,15 @@
+import { ChoiceInputMixin, CustomChoiceGroupMixin, FormGroupMixin } from '@lion/ui/form-core.js';
```
tlouisse, line 1:
```suggestion
import { ChoiceInputMixin, FormGroupMixin } from '@lion/ui/form-core.js';
import { CustomChoiceGroupMixin } from '../../src/choice-group/CustomChoiceGroupMixin.js';
```

packages/ui/components/form-core/test/choice-group/CustomChoiceGroupMixin.test.js
```diff
@@ -0,0 +1,15 @@
...
+import { LionInput } from '@lion/ui/input.js';
+import { runCustomChoiceGroupMixinSuite } from '@lion/ui/form-core-test-suites.js';
```
tlouisse, line 3:
```suggestion
import { runCustomChoiceGroupMixinSuite } from '../../test-suites/choice-group/CustomChoiceGroupMixin.suite.js';
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1,5 +1,5 @@
 import { browserDetection } from '@lion/ui/core.js';
-import { Unparseable } from '@lion/ui/form-core.js';
+import { Unparseable, CustomChoiceGroupMixin } from '@lion/ui/form-core.js';
```
tlouisse, line 2:
```suggestion
import { Unparseable } from '@lion/ui/form-core.js';
import { CustomChoiceGroupMixin } from '../../form-core/src/choice-group/CustomChoiceGroupMixin.js';

```

packages/ui/components/combobox/test/lion-combobox-integrations.test.js
```diff
@@ -1,4 +1,9 @@
 import { runListboxMixinSuite } from '@lion/ui/listbox-test-suites.js';
+import { runCustomChoiceGroupMixinSuite } from '@lion/ui/form-core-test-suites.js';
```
tlouisse, line 2:
```suggestion
import { runCustomChoiceGroupMixinSuite } from '../../form-core/test-suites/choice-group/CustomChoiceGroupMixin.suite.js';
```

packages/ui/components/form-core/test/choice-group/CustomChoiceGroupMixin.integrations.test.js
```diff
@@ -0,0 +1,54 @@
+import { runChoiceGroupMixinSuite } from '@lion/ui/form-core-test-suites.js';
+import { LitElement } from 'lit';
+import '@lion/ui/define/lion-fieldset.js';
+import '@lion/ui/define/lion-checkbox-group.js';
+import '@lion/ui/define/lion-checkbox.js';
+import { FormGroupMixin, CustomChoiceGroupMixin, ChoiceInputMixin } from '@lion/ui/form-core.js';
```
tlouisse, line 6:
```suggestion
import { FormGroupMixin, ChoiceInputMixin } from '@lion/ui/form-core.js';
import { CustomChoiceGroupMixin } from '../../src/choice-group/CustomChoiceGroupMixin.js';
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1065,25 +1083,53 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(LionListbox)) {
...
+
+          this.modelValue = this.parser([...this.modelValue, this._inputNode.value]);
+
+          this._inputNode.value = '';
+          this.opened = false;
+        } else {
+          super._listboxOnKeyDown(ev);
+          if (this.requireOptionMatch) {
+            this._inputNode.value = '';
+          }
```
tlouisse, line 1123:
```suggestion
          // TODO: should we clear the input value here when allowCustomChoice is false?
          // For now, we don't...
```

tlouisse (approved):

tlouisse:
Thanks for this quality contribution @GuiAmPm! 🎉 (see https://github.com/ing-bank/lion/pull/2108#discussion_r1386471999)

PR #2117, opened 2023-11-02
chore: allow ajax test to work with different dev server ports

tlouisse:
hardcoded port of 8000 leads to failing tests when multiple local servers are running


gerjanvangeest:
@tlouisse probably also needs a changeset

tlouisse:
> @tlouisse probably also needs a changeset

It's only for local development (no need to release to npm, I think)?

okadurin (approved):

PR #2140, opened 2023-11-16
feat(singleton-manager): add node/ssr support

tlouisse:
add node/ssr support

### DIFF
--- a/packages/singleton-manager/src/SingletonManagerClass.js
+++ b/packages/singleton-manager/src/SingletonManagerClass.js
@@ -1,9 +1,18 @@
 const sym = Symbol.for('lion::SingletonManagerClassStorage');
 
+/**
+ * Allow compatibility with node-js (for ssr).
+ * In the future, we can just use globalThis directly
+ * (for now, we're backwards compatible with browsers that still only use window, since we don't know all contexts singleton-manager is used in).
+ */
+// eslint-disable-next-line no-undef
+const globalThisOrWindow = globalThis || window;
 export class SingletonManagerClass {
   constructor() {
     /** @protected */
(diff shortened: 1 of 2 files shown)


packages/singleton-manager/src/SingletonManagerClass.js
```diff
@@ -1,9 +1,18 @@
...
+ * Allow compatibility with node-js (for ssr).
+ * In the future, we can just use globalThis directly
+ * (for now, we're backwards compatible with browsers that still only use window, since we don't know all contexts singleton-manager is used in).
+ */
+// eslint-disable-next-line no-undef
+const globalThisOrWindow = globalThis || window;
 export class SingletonManagerClass {
   constructor() {
     /** @protected */
-    this._map = window[sym] ? window[sym] : (window[sym] = new Map());
```
okadurin, line 6: Should we provide `globalThisOrWindow ` as a constructor parameter and then update unit tests so that they all run for both `window` and `globalThis`?

tlouisse: Was indeed thinking of how we can test this. The right thing to do would probably be to run all the current test in a node runtime (via mocha) to ensure we can use singletonManager with node.
I would rather focus on that (at a later point), then providing globalThis to the constructor.

At some point (when we add ssr support for lion) we need to test all our code in node as well. But since this change was so small and straightforward (`globalThis` with a fallback to `window`) I think we could get away w/o a node test for now.

tlouisse: Update: added a commit to run the same tests in node as well :)

okadurin (approved):

PR #2150, opened 2023-11-29
fix(tabs): set selectedIndex only if differ from previous selectedIndex

gyulus3:
## What I did

I modified the selectedIndex setter. Now it sets the selectedIndex only if the new value is different than the current selectedIndex.


tlouisse (approved):

tlouisse:
Thanks @gyulus3 !

PR #2155, opened 2023-12-11
fix(overlay-controller): do not scroll to invoker if it is visible

albelop:
When closing an overlay, `scrollIntoView` forced the invoker element to be placed in the middle of the screen. This fix prevents the scroll if the invoker is already visible, but still scrolls it into view if needed.

Related to #1957. Existing tests still apply.

## What I did

1. Replace the `scrollIntoView` [block options](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#block) from `center` to `nearest`

### DIFF
--- a/.changeset/hot-otters-push.md
+++ b/.changeset/hot-otters-push.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+Fix scroll behavior when closing an overlay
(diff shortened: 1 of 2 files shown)


gerjanvangeest (approved):

gerjanvangeest:
@albelop Thank for the fix :)

PR #2159, opened 2023-12-18
chore: update to lit3 and other dependencies

gerjanvangeest:
BREAKING CHANGE: update to latest major versions of lit, autosize, awesome-phonenumber & ibantools

## What I did

1.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - lit3
 
 jobs:
   release:
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -59,14 +59,14 @@
     "exports/overlays.js"
   ],
   "dependencies": {
-    "@bundled-es-modules/message-format": "^6.0.4",
-    "@open-wc/dedupe-mixin": "^1.3.1",
+    "@bundled-es-modules/message-format": "^6.2.4",
+    "@open-wc/dedupe-mixin": "^1.4.0",
     "@open-wc/scoped-elements": "^2.1.4",
-    "@popperjs/core": "^2.11.6",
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
```
tlouisse, line 6: Think we can only do `lit3-alpha` (as a temp situation.)
Once that becomes master, we remove `lit3` branch and create a `lit2` branch might we need it in the future (I suspect the need for it is way less then for our [`lit1`](https://github.com/ing-bank/lion/tree/lit1) branch)

packages/ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lion/ui",
-  "version": "0.5.2",
+  "version": "0.5.2-alpha.0",
```
tlouisse, line 3: Maybe 0.6.0-alpha.

Also, we need to do this to prevent npm from referring to it as latest: https://github.com/ing-bank/lion/commit/8702381b52cd9ded130318397e3e3e5b30b17a4d

(but then with lit3 as tag).

.github/workflows/release.yml
tlouisse: And once we create the lit2 branch, we should do this as well with a lit2 tag: https://github.com/ing-bank/lion/commit/8702381b52cd9ded130318397e3e3e5b30b17a4d

tlouisse (changes requested):

.github/workflows/release.yml
tlouisse: Maybe it's also better to call the branch lit3-alpha instead of lit3 (for clarity)

PR #2173, opened 2024-01-23
Fix/input stepper a11y

gerjanvangeest:
## What I did

1. a11y enhancement & added translations

### DIFF
--- a/.changeset/pink-islands-return.md
+++ b/.changeset/pink-islands-return.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': minor
+---
+
+[input-stepper] a11y enhancement & added translations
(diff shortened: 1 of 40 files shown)


.changeset/pink-islands-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+[BREAKING-CHANGE] [input-stepper] a11y enhancement & added translations
```
tlouisse, line 5: How exactly is it breaking?

gerjanvangeest: Changed a function name, but that was private, so can be changed to none breaking.

Will then directly add it to master :)

docs/components/input-stepper/overview.md
```diff
@@ -4,16 +4,16 @@ A web component that enables the user to increase and decrease a numeric value b
 
 ```js script
 import { html } from '@mdjs/mdjs-preview';
+import { loadDefaultFeedbackMessages } from '@lion/ui/validate-messages.js';
```
tlouisse, line 7: Maybe we can move these to some global demo setups in our mdjs-viewer in the future

tlouisse (approved):

PR #2174, opened 2024-01-24
feat(form-core): add select and upload specific required messages

gerjanvangeest:
## What I did

1. add inputInteractionMode to validators, creates specific select and upload required message


packages/ui/components/validate-messages/translations/en.js
```diff
@@ -1,6 +1,8 @@
 export default {
   error: {
-    Required: 'Please enter a(n) {fieldName}.',
+    Required: 'Please {inputMode} a(n) {fieldName}.',
```
tlouisse, line 3:
```suggestion
    Required: 'Please enter a(n) {fieldName}.',
```

packages/ui/components/form-core/src/validate/Validator.js
```diff
@@ -27,12 +28,20 @@ export class Validator extends EventTarget {
...
+    /**
+     * Default type supported by ValidateMixin
+     * @type {ValidationType}
+     */
+    this.type = config?.type || 'error';
+    /**
+     * Way of interacting with the form-field: "enter" | "select" | "upload"
+     * @type {InputMode}
+     */
+    this.inputMode = config?.inputMode || 'enter';
```
tlouisse, line 40: Can we maybe rename this to interactionMode? Since inputMode is already occuppied: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

packages/ui/components/validate-messages/src/loadDefaultFeedbackMessagesNoSideEffects.js
```diff
@@ -124,9 +129,10 @@ export function loadDefaultFeedbackMessagesNoSideEffects({ localize }) {
    * @returns {Promise<string|Element>}
    */
   const getLocalizedMessage = async data => {
+    const inputMode = data.inputMode !== 'enter' ? capitalize(data.inputMode) : '';
     await forMessagesToBeReady();
     if (data) {
-      return localize.msg(`lion-validate:${data.type}.${data.name}`, data);
+      return localize.msg(`lion-validate:${data.type}.${data.name}${inputMode}`, data);
```
tlouisse, line 135: Should we consider a `_` inbetween here, to indicate it's a message variation for the same validator type?

tlouisse:
> ## What I did
> 1. add inputMode to validators, creates specific select and upload required message

Looks great. Added a few remarks that could make it more optimized

packages/ui/components/form-core/test-suites/ValidateMixinFeedbackPart.suite.js
```diff
@@ -622,6 +626,49 @@ export function runValidateMixinFeedbackPart() {
...
+      );
+      await el.updateComplete;
+      await el.feedbackComplete;
+
+      // ignore fieldName Promise as it will always be unique
+      const compare = spy.args[0][0];
+      delete compare?.config?.fieldName;
+      expect(compare).to.eql({
+        config: {
+          inputMode: 'select',
```
tlouisse, line 659: Not really sure if developers should be able to configure this.
It's up to the developer of a FormControl to make it accessible. 

In case a developer wants to change the message, he should provide the getMessage method

docs/fundamentals/systems/form/validate.md
```diff
@@ -229,10 +228,18 @@ The required validator can be put onto every form field element and will make su
 
 ```js preview-story
 export const requiredValidator = () => html`
-  <lion-input .validators=${[new Required()]} label="Required"></lion-input>
+  <lion-input .validators=${[new Required()]} label="Required" .fieldName="value"></lion-input>
 `;
 ```
 
+The default `Required` validation message is "Please, enter a(n) {fieldName}". In which the "fieldName" can be provided separately via the `fieldName` attribute, with a fallback to the label. See [override fieldname](#override-fieldname) for more information.
```
tlouisse, line 235: Nice 👍

packages/ui/components/form-core/src/validate/Validator.js
gerjanvangeest: InteractionMode is also already occupied in our combobox.

gerjanvangeest: Currently went for `inputInteractionMode`

docs/components/listbox/use-cases.md
```diff
@@ -87,61 +84,83 @@ This behavior can usually be seen in `<select>` on the Windows platform.
...
+  <lion-option .choiceValue=${'Beets'}>Beets</lion-option>
+  <lion-option .choiceValue=${'Bell pepper'}>Bell pepper</lion-option>
+  <lion-option .choiceValue=${'Broccoli'}>Broccoli</lion-option>
+  <lion-option .choiceValue=${'Brussels sprout'} disabled>Brussels sprout</lion-option>
+  <lion-option .choiceValue=${'Cabbage'}>Cabbage</lion-option>
+  <lion-option .choiceValue=${'Carrot'}>Carrot</lion-option>
+</lion-listbox>
+```
+
+## Validation
```
tlouisse, line 140: Maybe we should mention here that it inherits from LionField/FormControl and validation works the same for all FormControls (the less code examples we need to maintain, the better)

For Combobox the example made sense since it has a specific Validator

docs/components/select/use-cases.md
```diff
@@ -2,54 +2,75 @@
...
+  <select slot="input">
+    <option selected hidden value>Please select</option>
+    <option value="red">Red</option>
+    <option value="hotpink">Hotpink</option>
+    <option value="teal">Teal</option>
+  </select>
+</lion-select>
+```
+
+## Validation
```
tlouisse, line 54: Maybe we should mention here that it inherits from LionField/FormControl and validation works the same for all FormControls.

For Combobox the example made sense since it has a specific Validator

packages/ui/components/form-core/src/validate/Validator.js
tlouisse: `operationMode` maybe?

docs/components/select/use-cases.md
gerjanvangeest: So no visual code block, and just a link to `./fundatementals/systems/form/validate.md` page.

tlouisse (approved):

PR #2181, opened 2024-02-02
Add a DisabledController to replace the DisabledMixin

Ramon92:
## What I did
A first step to move away from `ClassMixins` on components and use controllers 
This should improve debugging and tracking back what methods belongs to which controller/Mixin
Properties are then also set by the component making it clear in code what the properties of the component are and no hidden properties added by `ClassMixins`

1. Add a DisabledController to replace the DisabledMixin

- DisabledController has slightly adjusted api compared to the DisabledMixin
- host requires to have a `disabled` property
- Host needs to provide a callback to set the `disabled` property
- Same tests as the `DisabledMixin` to asses the same functionality is still working


gerjanvangeest:
Hi @Ramon92,

Thanks for this contribution. Can you also add it to the `ValidateMixin` and `FormGroupMixin` to make sure it works for the full form system?

Ramon92:
> Hi @Ramon92,
> 
> Thanks for this contribution. Can you also add it to the `ValidateMixin` and `FormGroupMixin` to make sure it works for the full form system?

Hi @gerjanvangeest 
Sorry for the late reply, yes sounds good.
I'll make the changes this month to make it work with the form system.

gerjanvangeest:
I'll close this PR due to inactivity.

Also on a side node, the `FormGroupMixin` needs to be able to access the childs `makeRequestToBeDisabled()` function. Which cannot be done in a direct way when using a controller. So not sure using a controller for this use case is the best solution.

Ramon92:
Apologies, did not find the time after all.
You are right, I stumbled upon that issue, making it pretty hard to detach the mixin to a controller. it is tightly coupled together.

PR #2182, opened 2024-02-05
chore: add data-label to doc examples with explanation

gerjanvangeest:
## What I did

1.

### DIFF
--- a/docs/fundamentals/systems/form/styling.md
+++ b/docs/fundamentals/systems/form/styling.md
@@ -17,7 +17,6 @@ export const label = () => html`
   <style>
     .u-sr-only {
       position: absolute;
-      top: 0;
       width: 1px;
       height: 1px;
       overflow: hidden;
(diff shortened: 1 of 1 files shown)


docs/fundamentals/systems/form/styling.md
```diff
@@ -37,52 +37,56 @@ export const label = () => html`
 `;
 ```
 
-## Prefix
+## Input additions
+
+When using an addition to the input make sure to add `data-label` to add the content connected to the input via `aria-labelledby` or add `data-description` to connect it via the `aria-describedby` when the content is relevant for the screen-reader user.
+
```
tlouisse, line 43: It might seem a bit now if [data-label] is always needed, but that's probably not the case

tlouisse: I mean a suffix can also be purely visual (in case of a search icon for a field already labelled "Search") and we don't need data-label or data-description

gerjanvangeest: I changed the text to `you could` instead of `make sure to`, but i could make it more explicit.

tlouisse (approved):

PR #2186, opened 2024-02-07
fix(checkbox-group): add role=list and role=listitem to checkbox-inde…

gerjanvangeest:
…terminate and its children

## What I did

1.

### DIFF
--- a/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -197,17 +197,32 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
   // eslint-disable-next-line class-methods-use-this
   _afterTemplate() {
     return html`
-      <div class="choice-field__nested-checkboxes">
+      <div class="choice-field__nested-checkboxes" role="list">
         <slot></slot>
       </div>
     `;
   }
 
   /**
+   * @param {Event} ev
    * @protected
    */
-  _onRequestToAddFormElement() {
+  _onRequestToAddFormElement(ev) {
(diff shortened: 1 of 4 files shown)


packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -197,16 +197,20 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
...
     `;
   }
 
   /**
+   * @param {Event} ev
    * @protected
    */
-  _onRequestToAddFormElement() {
+  _onRequestToAddFormElement(ev) {
+    if (!(/** @type {HTMLElement} */ (ev.target).hasAttribute('role'))) {
```
tlouisse, line 211: are the [role=list] (shadow) and [role=lisitem] (probably light) in the same dom and does this work well in SR?

gerjanvangeest: They are in different doms, but this works in VoiceOver and NVDA

tlouisse: Ok nice, as long as we don't have idrefs it's fine indeed :)

packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -197,16 +197,20 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
...
+      /** @type {HTMLElement} */ (ev.target)?.setAttribute('role', 'listitem');
```
tlouisse, line 212: For completeness, it would be nice to clean up the dom we manipulated in `_onRequestToRemoveFormElement`

tlouisse (approved):

PR #2189, opened 2024-02-09
feat: split up validate-messages per entrypoint for smarter bundling

tlouisse:
For optimized bundling, it's reccommended to load feedback messages per entrypoint. For instance, when you only use form-core in your app:

```js
import { LionInputTel } from '@lion/ui/input-tel.js';
import { getLocalizeManager } from '@lion/ui/localize-no-side-effects.js';
import { loadInputTelMessagesNoSideEffects } from '@lion/ui/validate-messages-no-side-effects.js';

export class MyInputTel extends LionInputTel {
  constructor() {
    super();
    loadInputTelMessagesNoSideEffects({ localize: getLocalizeManager() });
  }
}
```

This prevents you from loading unused entrypoints like input-tel (which loads a full phone validation library) etc.

### DIFF
--- a/.changeset/swift-deers-taste.md
+++ b/.changeset/swift-deers-taste.md
@@ -1,0 +1,22 @@
+---
+'@lion/ui': patch
+---
+
+feat: split validate-messages-no-side-effects methods, so they can be bundled along with entrypoints.
+
+For optimized bundling, it's reccommended to load feedback messages per entrypoint. For instance, when you only use form-core in your app:
+
+```js
+import { LionInputTel } from '@lion/ui/input-tel.js';
+import { getLocalizeManager } from '@lion/ui/localize-no-side-effects.js';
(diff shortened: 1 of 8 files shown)


gerjanvangeest:
A changeset is needed.

tlouisse:
> A changeset is needed.

True, fixed. Could you have another look?

gerjanvangeest (approved):

packages-node/providence-analytics/test-node/program/analyzers/e2e/match-imports_-_importing-target-project_0.0.2-target-mock_+_exporting-ref-project_1.0.0__1789378150.json
```diff
@@ -0,0 +1,162 @@
+{
```
gerjanvangeest, line 1: I don't think this file has anything to do with the rest of this change.

PR #2198, opened 2024-02-26
fix(overlays): only close the top overlay on [Esc]

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -121,7 +121,7 @@ export class OverlayController extends EventTarget {
     /** @private */
     this.__sharedConfig = config;
 
-    this.blockEscKeyHandler = false;
+    this._blocksEscKeyHandler = false;
 
     /** @private */
     this.__activeElementRightBeforeHide = null;
@@ -1128,9 +1128,12 @@ export class OverlayController extends EventTarget {
     }
   }
 
-  /** @private */
-  __escKeyHandler(/** @type {KeyboardEvent} */ ev) {
-    if (ev.key === 'Escape' && !this.blockEscKeyHandler) {
+  /**
+   * @param {KeyboardEvent} ev
+   * @private
+   */
+  __escKeyHandler(ev) {
+    if (ev.key === 'Escape' && !this._blocksEscKeyHandler) {
       this.hide();
     }
   }
@@ -1159,6 +1162,8 @@ export class OverlayController extends EventTarget {
    */
   _handleHidesOnOutsideEsc({ phase }) {
     if (phase === 'show') {
+      this.__escKeyHandler = (/** @type {KeyboardEvent} */ ev) =>
+        ev.key === 'Escape' && this.hide();
       document.addEventListener('keyup', this.__escKeyHandler);
     } else if (phase === 'hide') {
       document.removeEventListener('keyup', this.__escKeyHandler);
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -121,6 +121,8 @@ export class OverlayController extends EventTarget {
     /** @private */
     this.__sharedConfig = config;
 
+    this.blockEscKeyHandler = false;
+
     /** @private */
     this.__activeElementRightBeforeHide = null;
 
@@ -294,7 +296,7 @@ export class OverlayController extends EventTarget {
   }
 
   /**
-   * Hides other overlays when mutiple are opened (currently exclusive to globalOverlayController)
+   * Hides other overlays when multiple are opened (currently exclusive to globalOverlayController)
    * @type {boolean}
    */
   get isBlocking() {
@@ -302,7 +304,7 @@ export class OverlayController extends EventTarget {
   }
 
   /**
-   * Hides other overlays when mutiple are opened (currently exclusive to globalOverlayController)
+   * Hides other overlays when multiple are opened (currently exclusive to globalOverlayController)
    * @type {boolean}
    */
   get preventsScroll() {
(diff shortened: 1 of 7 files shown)


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -121,6 +121,8 @@ export class OverlayController extends EventTarget {
     /** @private */
     this.__sharedConfig = config;
 
+    this.blockEscKeyHandler = false;
```
tlouisse, line 124: If we want to make it part of OverlayController config:
- it should be added to OverlayConfig type (under User interaction): https://github.com/ing-bank/lion/blob/master/packages/ui/components/overlays/types/OverlayConfig.ts#L51
- in the past, we introduced a specific naming convention (3rd person instead of usual imperative). Would be nice to stay consistent => `blocksEscKeyHandler`
- to align with rest of config options
  - should be part of `_defaultConfig`
  - should have getter  reading from config (and no setter)

tlouisse: Also, having a closer look, shouldn't this be the default behavour of `hidesOnEsc`?

Only `hidesOnOutsideEsc` should close all ovelays

gerjanvangeest: The `blocksEscKeyHandler` is not a configurable option, it is set internally. So i don't think it should be part of the config options.

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -606,6 +606,31 @@ describe('OverlayController', () => {
...
+        await ctrl1.show();
+        await ctrl2.show();
+        expect(ctrl1.manager.shownList.length).to.equal(2);
+        expect(ctrl1._blocksEscKeyHandler).to.be.true;
+        expect(ctrl2._blocksEscKeyHandler).to.be.false;
+
+        ctrl2.contentNode.dispatchEvent(new KeyboardEvent('keyup', { key: 'Escape' }));
+        await aTimeout(0);
+        expect(ctrl2.isShown).to.be.false;
+        expect(ctrl1.isShown).to.be.true;
```
tlouisse, line 629: Does this test fail when `_blocksEscKeyHandler` is left out of the equation?

I would expect the keyup event to never reach ctrl1.contentNode, as ctrl2.contentNode and ctrl1.contentNode are siblings.

The test should probably create a setup where ctrl2.contentNode is nested inside ctrl1.contentNode

packages/ui/components/overlays/src/OverlaysManager.js
```diff
@@ -108,7 +116,14 @@ export class OverlaysManager {
     if (!this.list.find(ctrl => ctrlToHide === ctrl)) {
       throw new Error('could not find controller to hide');
     }
+    // eslint-disable-next-line no-param-reassign
+    ctrlToHide._blocksEscKeyHandler = false;
     this.__shownList = this.shownList.filter(ctrl => ctrl !== ctrlToHide);
+    setTimeout(() => {
```
tlouisse, line 122: Can we explain what this code does and why the timeout is needed?

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1126,9 +1128,14 @@ export class OverlayController extends EventTarget {
...
   }
 
-  /** @private */
-  __escKeyHandler(/** @type {KeyboardEvent} */ ev) {
-    return ev.key === 'Escape' && this.hide();
+  /**
+   * @param {KeyboardEvent} ev
+   * @private
+   */
+  __escKeyHandler(ev) {
```
tlouisse, line 1135: Maybe we can also check this.handleHidesOnOutsideEsc here (instead of overriding the `__escKeyHandler`in `_handleHidesOnOutsideEsc`...)

(Know you didn't write it, but while we're at it :))

packages/ui/components/overlays/src/OverlayController.js
tlouisse: Ah nice, you made it a protected variable now

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -121,6 +121,8 @@ export class OverlayController extends EventTarget {
     /** @private */
     this.__sharedConfig = config;
 
+    this._blocksEscKeyHandler = false;
```
tlouisse, line 124:
```suggestion
   /** @protected */
    this._blocksEscKeyHandler = false;
```

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -629,6 +654,30 @@ describe('OverlayController', () => {
         ctrl.contentNode.dispatchEvent(new KeyboardEvent('keyup', { key: 'Escape' }));
         expect(ctrl.isShown).to.be.true;
       });
+
+      it('hides all overlays when [escape] is pressed on outside element', async () => {
```
tlouisse, line 658: I wonder if the approach still works if we have two ctrls: ctrl1 with hidesOnOutsideEsc and ctrl2 with hidesOnEsc. And ctrl2 is nested in ctrl1 (its contentNode is nested).

Would ctrl1 still be autonomous and be closed? (I suspect not, as `_blocksEscKeyHandler` works as a global option (on OverlaysManager level) now).

tlouisse: Maybe we can solve this in a different way?
You could for instance check whether the esc event came from a child in the dom. If so, and we have `closeOnEsc` and not `closeOnOutsideEsc`, we should not close ourselves.

packages/ui/components/overlays/src/OverlaysManager.js
gerjanvangeest: It removes the `_blocksEscKeyHandler` from the next overlay.
If there is no timeout, the unblocking will be done before the `Esc` will trigger the hide overlay. If you unblock it one tick later, it all works as expected.

packages/ui/components/overlays/src/OverlayController.js
gerjanvangeest: Yes, but we don't want to have the `_blocksEscKeyHandler` in that function.
Then its better to have a separate `escKeyOutisideHandler`

PR #2200, opened 2024-02-29
@lion/ajax: BREAKING: Only add XSRF token on mutable requests and on same origin or whitelisted origins

hvdb:
Previously the XSRF token was added to any call to any origin.  
This is changed to only attach the token to requests that are POST/PUT/PATCH/DELETE.  
And another change is that it will validate if the request origin is the same as current origin or when the origin is in the xsrfTrustedOrigins.

This is a fix for a vulnerability found, as we inadvertently revealed the confidential XSRF-TOKEN stored in cookies by including it in the HTTP header X-XSRF-TOKEN for every request made to any host allowing attackers to view sensitive information.

### DIFF
--- a/packages/ajax/src/interceptors/xsrfHeader.js
+++ b/packages/ajax/src/interceptors/xsrfHeader.js
@@ -3,6 +3,61 @@
  */
 
 /**
+* Parse a URL to discover it's components
+*
+* @param {String | object} url The URL to be parsed
+* @returns {{protocol: string, host: string }}
+*/
+function resolveURL(url) {
+  if (typeof url !== 'string') {
+    return url;
+  }
+  let href = url;
+  let urlParsingNode = document.createElement('a');
+  urlParsingNode.setAttribute('href', href);
+
+  // urlParsingNode provides the UrlUtils interface - http://url.spec.whatwg.org/#urlutils
+  return {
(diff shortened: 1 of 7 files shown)


packages/ajax/src/interceptors/xsrfHeader.js
```diff
@@ -2,6 +2,59 @@
  * @typedef {import('../../types/types.js').RequestInterceptor} RequestInterceptor
  */
 
+/**
+* Parse a URL to discover it's components
```
tlouisse, line 6:
```suggestion
* Parse a URL to discover its components
```

packages/ajax/src/interceptors/xsrfHeader.js
```diff
@@ -2,6 +2,59 @@
...
+*
+* @param {String | object} url The URL to be parsed
+* @returns {{protocol: string, host: string }}
+*/
+function resolveURL(url) {
+  if (typeof url !== 'string') {
+    return url;
+  }
+  let href = url;
```
tlouisse, line 15:
```suggestion
  const href = url;
```

packages/ajax/src/interceptors/xsrfHeader.js
```diff
@@ -2,6 +2,59 @@
...
+  let urlParsingNode = document.createElement('a');
```
tlouisse, line 16:
```suggestion
  const urlParsingNode = document.createElement('a');
```

PR #2210, opened 2024-03-05
fix(textarea): fix box-sizing inside tests to make it work cross browser

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/textarea/test/lion-textarea.test.js
+++ b/packages/ui/components/textarea/test/lion-textarea.test.js
@@ -12,11 +12,6 @@ import '@lion/ui/define/lion-textarea.js';
 
 const fixture = /** @type {(arg: TemplateResult|string) => Promise<LionTextarea>} */ (_fixture);
 
-const isFirefox = (() => {
-  const ua = navigator.userAgent.toLowerCase();
-  return ua.indexOf('firefox') !== -1 && ua.indexOf('safari') === -1 && ua.indexOf('chrome') === -1;
-})();
-
 function hasBrowserResizeSupport() {
   const textarea = document.createElement('textarea');
   return textarea.style.resize !== undefined;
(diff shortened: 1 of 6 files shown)


packages/ui/components/textarea/test/lion-textarea.test.js
```diff
@@ -98,10 +93,13 @@ describe('<lion-textarea>', () => {
...
   });
 
-  // TODO: To be fixed in 4096171
-  it.skip(`starts growing when content is bigger than "rows"
+  it(`starts growing when content is bigger than "rows"
     'and stops growing after property "maxRows" is reached`, async () => {
-    const el = await fixture(`<lion-textarea></lion-textarea>`);
+    const el = await fixture(html`
+      <lion-textarea>
+        <textarea slot="input" style="box-sizing: content-box;"></textarea>
```
okadurin, line 100: Could we keep the old API so that `textarea slot="input"` does not have to be added by users as a mandatory thing to make this example work?

gerjanvangeest: I moved this to this part of styling to the `web-test-runner.config.mjs`

packages/ui/package.json
```diff
@@ -65,12 +65,15 @@
...
     "@popperjs/core": "^2.11.8",
-    "autosize": "^6.0.1",
+    "autosize": "^6.0.0",
     "awesome-phonenumber": "^6.4.0",
     "ibantools": "^4.3.9",
     "lit": "^3.1.2",
     "singleton-manager": "^1.7.0"
   },
+  "devDependencies": {
+    "@types/autosize": "^4.0.3"
```
gerjanvangeest, line 75: Should this be here, or with the other devDeps in the root `package.json`?

tlouisse: Maybe the root is better indeed

tlouisse (approved):

okadurin (approved):

PR #2212, opened 2024-03-06
fix(overlays): skip local-positioning tests for Firefox, and added a …

gerjanvangeest:
…todo to fix later

## What I did

1.

### DIFF
--- a/packages/ui/components/overlays/test/local-positioning.test.js
+++ b/packages/ui/components/overlays/test/local-positioning.test.js
@@ -90,7 +90,7 @@ describe('Local Positioning', () => {
       `);
       await ctrl.show();
 
-      // TODO: fails on Firefox => fix it
+      // TODO: test fails on Firefox, but looks fine in browser => try again in a later version and investigate when persists (or move to anchor positioning when available in all browsers)
       if (!isFirefox) {
         expect(normalizeTransformStyle(ctrl.contentWrapperNode.style.transform)).to.equal(
           'translate(70px, -508px)',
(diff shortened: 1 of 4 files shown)


packages/ui/components/overlays/test/local-positioning.test.js
```diff
@@ -225,18 +231,21 @@ describe('Local Positioning', () => {
 
       await ctrl.show();
 
-      // N.B. margin between invoker and content = 8px
-      expect(normalizeTransformStyle(ctrl.contentWrapperNode.style.transform)).to.equal(
-        'translate(110px, -308px)',
-        '110 = (100 + (100-80)/2); -308= 300 + 8',
-      );
+      // TODO: fails on Firefox => fix it
```
tlouisse, line 234:
```suggestion
      // TODO: test fails on Firefox, but looks fine in browser => try again in a later version and investigate when persists (or move to anchor positioning when available in all browsers)
```

okadurin (approved):

packages/ui/components/overlays/test/local-positioning.test.js
tlouisse: Maybe also return early here? https://dev.to/jpswade/return-early-12o5

(better readable and easier to remove 1 line later)

packages/ui/components/overlays/test/local-positioning.test.js
```diff
@@ -9,6 +9,11 @@ import { normalizeTransformStyle } from '../test-helpers/normalizeTransformStyle
  * @typedef {import('../types/OverlayConfig.js').ViewportPlacement} ViewportPlacement
  */
 
+const isFirefox = (() => {
```
tlouisse, line 12: This should also be available in core/browserDetection

tlouisse (approved):

PR #2222, opened 2024-03-15
Added a flag to composed property of form-element-register event to support children elements registration

maxevilmind:
## What I did

1. Added a flag to composed property of form-element-register event's parameter.

This fixes a scenario when a component wraps several other components and we want to have serialized value taken from all children, not only a first child.

On the previous version of lit we could bypass it with `createRenderRoot() {return this;}`, disabling shadow DOM.
However, with the latest updates to Lit this fix does not work anymore.

Name of the flag is, of course, negotiable. If you have more clear option, or the option that suits your standards better, please let me know, I will quickly update it.

PS:
I did not add any tests because I could not find any tests for this particular event present. If the unit test is necessary in this case, please suggest the right place to add it
As per docs, I don't see docs for this event either, please suggest the right place to document it.


gerjanvangeest (commented):
Thanks for the contribution :)

I would expect the event to have tests in the [FormRegistrationMixins.suite.js](https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js). If you could add some that would be great.

Can you add `this.formElementRegisterComponsed = false;` to the constructor, to make the linter happy. And maybe add a one liner JSdoc.

And can you add a changeset `npm run changeset` to add the change to our release and changelog?

packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
```diff
@@ -36,6 +36,7 @@ const FormRegisteringMixinImplementation = superclass =>
         new CustomEvent('form-element-register', {
           detail: { element: this },
           bubbles: true,
+          composed: this.formElementRegisterComponsed ? true : false,
```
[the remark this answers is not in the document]
maxevilmind: Always glad to contribute to lion :)
Thanks for the feedback, will pick it up today.
And yeah, will run the `npm run changeset` too.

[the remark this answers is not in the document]
maxevilmind: @gerjanvangeest do you think it's better to have tests in:
packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js (as you mentioned before)
Or this is a better place?
packages/ui/components/form-integrations/test/form-integrations.test.js
Cause it seems that form registration and serializedValue, formattedValue are tested there...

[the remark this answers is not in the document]
gerjanvangeest: Both fine with me :)

packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
```diff
@@ -36,6 +36,7 @@ const FormRegisteringMixinImplementation = superclass =>
         new CustomEvent('form-element-register', {
           detail: { element: this },
           bubbles: true,
+          composed: this.formElementRegisterComposed ? true : false,
```
tlouisse, line 39:
```suggestion
          composed: this.allowCrossRootRegistration,
```

tlouisse: Can you add this to constructor?

```js
/**
 * To encourage accessibility best practices, `form-element-register` events
 * do not pierce to shadow roots. This forces the developer to create form groups and fieldsets that automatically can create accessible relationships in the same dom tree.
 If you know what you're doing, it is possible to nest FormControls
 inside shadow dom. See https://lion-web.netlify.app/fundamentals/rationales/accessibility/#shadow-roots-and-accessibility
 */
this.allowCrossRootRegistration = false;
```

tlouisse: And the same field should be added to the mixin types: https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/types/registration/FormRegisteringMixinTypes.ts

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -256,6 +257,23 @@ export const runRegistrationSuite = customConfig => {
...
+        const el = /** @type {RegistrarClass} */ (
+          await fixture(html`
+            <${parentTag} .formElementRegisterComposed="${true}">
+              <${childTag}></${childTag}>
+            </${parentTag}>
+          `)
+        );
+        const eventSpy = sinon.spy();
+        el.addEventListener('form-element-register', eventSpy);
+        el.connectedCallback();
```
tlouisse, line 271: this should not be needed
```suggestion
```

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -256,6 +257,23 @@ export const runRegistrationSuite = customConfig => {
       ]);
     });
 
+    describe('FormRegisteringMixin', () => {
+      it('dispatches the form-element-register event with correct details upon connectedCallback', async () => {
+        const el = /** @type {RegistrarClass} */ (
+          await fixture(html`
```
tlouisse, line 263: It would be best to prove the use case here: add childTag as part of the shadow dom of parent (by creating a separate class).

And then it would also be good to test in a separate fixture that the spy will not be called without `allowCrossRootRegistration`

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -256,6 +257,23 @@ export const runRegistrationSuite = customConfig => {
...
+            <${parentTag} .formElementRegisterComposed="${true}">
```
tlouisse, line 264:
```suggestion
            <${parentTag} allow-cross-root-registration>
```

packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
tlouisse: And can you add this to properties?

```js
allowCrossRootRegistration: {type: Boolean, attribute: allow-cross-root-registration}
```

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
```diff
@@ -256,6 +257,23 @@ export const runRegistrationSuite = customConfig => {
...
+              <${childTag}></${childTag}>
```
tlouisse, line 265:
```suggestion
              <${childTag} allow-cross-root-registration></${childTag}>
```

tlouisse (changes requested):
See comments

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
maxevilmind: The test breaks without calling `el.connectedCallback()` explicitly

packages/ui/components/form-core/test-suites/FormRegistrationMixins.suite.js
maxevilmind: Added a case for a spy not called if the flag is not set

tlouisse (approved):

Issue #2224, opened 2024-03-19
[@lion/ui]: Depending on @open-wc/scoped-elements ^2.1.4 causes "Multiple versions of Lit loaded"


riovir:
## Expected behavior

A fresh install of Lion does not cause version conflicts of `@lit/reactive-element`.

## Actual Behavior

Installing `@lion/ui@0.5.5` causes a `@lit/reactive-element@2.0.4` and a `@lit/reactive-element@1.6.3` version conflict.

## Additional context

Lion UI is the main dependency of our Design System. We also use `lit` and `@open-wc/scoped-elements` pinned to Lion-compatible versions. Unfortunately, here is the result of `npm ls npm ls @lit/reactive-element` of a project freshly installing the package:

```
  `-- @example/design-system@0.4.19
    +-- @lion/ui@0.5.5
    | `-- @open-wc/scoped-elements@2.2.4
    |   `-- @lit/reactive-element@2.0.4
    +-- @open-wc/scoped-elements@2.2.0
    | `-- @lit/reactive-element@1.6.3
    `-- lit@2.8.0
      +-- @lit/reactive-element@1.6.3
      `-- lit-element@3.3.3
        `-- @lit/reactive-element@1.6.3
```

Looking at the `@lion/ui` [package.json](https://github.com/ing-bank/lion/blob/master/packages/ui/package.json#L66) we can see that Lion is okay to use use `@open-wc/scoped-elements` v2.1.4+ Unfortunately, this includes `v2.2.4` which did a major dependency bump on `@lit/reactive-element` to v2.

The simplest fix would be for Lion to pin the @open-wc/scoped-elements version to the last non-conflicting one: `2.2.0` until scoped-elements is ready to be upgraded to v3 (personally, not a fan of making the polyfill mandatory). Then later, once Lion has updated to Lit v3 the dependency can be unpinned.

riovir:
Submutted PR: https://github.com/ing-bank/lion/pull/2225

riovir:
Note that needed to further limit the version back to 2.2.0 as the mixin started breaking types.

PR #2225, opened 2024-03-19
fix(ui): Fix multiple Lit versions - pin @open-wc/scoped-elements to 2.2.0 (#2224)

riovir:
## What I did

1. Pinned the version of `@open-wc/scoped-elements` to `2.2.0`

This is the last non-build breaking version that doesn't need `@lit/reactive-element` v2 for which AFAIK Lion is not yet ready due to using Lit v2 ATM. The issue with the `2.2.3` was that it failed type checks of some mixins.

### DIFF
--- a/.changeset/cold-pets-retire.md
+++ b/.changeset/cold-pets-retire.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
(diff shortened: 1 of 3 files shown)


gerjanvangeest (approved):
Thanks @riovir 💪 I will release this today.

I think the lit3 will be released within a week or 2.

Issue #2227, opened 2024-03-26
[@lion/ui/localize.js] parseNumber('--42') is parsed as NaN


riovir:
## Expected behavior

Calls of `parseNumber('--blah42blah')` or `parseNumber('--42')` would only return with a `type === 'number'` that's a usable number.
Based on `parseNumber('++++blah42blah')` and `parseNumber('++++42')` returning `42`. I'd expect the same to work with multiple minus signs.

## Actual Behavior

The `parseNumber('--blah42blah')` or `parseNumber('--42')` calls return with a `NaN`.

## Additional context

Considering how `NaN !== NaN` this turned out to be the source of an infinite update loop that froze the browser tab if the customer accidentally tapped the '-' sign twice when entering a number. 

Note that the loop is not directly caused by Lion, but is prone to happen when the LionInputAmount is used together with a LionForm in some apps.

riovir:
I'm planning on posting a PR for this line: https://github.com/ing-bank/lion/blob/master/packages/ui/components/localize/src/number/parseNumber.js#L118

```
const cleanedInput = matchedInput.join(''); // Instead of this
const cleanedInput = dropExtraMinuses(matchedInput.join('')) // This where dropExtraMinuses() keeps the first '-' and wipes the rest of the occurrences from the string. So '---42' -> '-42'
```

tlouisse:
@riovir Thanks for proposing the fix!

Which component causes the infinite loop if I may ask? LionInputAmount?

tlouisse:
Thanks for the good explanation 👍

Maybe we are also able to catch this here:
https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/src/FormatMixin.js#L81
(probably using the [notEqual](https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts#L376) from lit here, would also fix it)

riovir:
One question though: when the value of LionInputAmount is resulting in a `NaN` wouldn't a `new Unparseable(this.value)` work better? In our apps we type the modelValue as `number | null | Unparseable`. What 95% of app teams don't do is to write code that defends against `NaN` after they checked that the `typeof modelValue === 'number'`.

One possible approach here was my change to the `parseNumber` which is not enough, as we also got a reported bug about the LionInputAmount not handling a value like "999999999999999999999999999999934.42" well also resulting in a `NaN`. (In fact this was the original value that triggered the infinite loop: a customer pasting a waaay too big number)

[...]

PR #2232, opened 2024-03-27
fix(localize): let parseNumber() deal with extra '-' chars (#2227)

riovir:
## What I did

1. In the `@lion/ui/localize` wrapped a utility that drops any additional '-' character from the original `cleanedInput ` so that parsing will result in a negative number instead of `NaN`

This behavior will also align better with how the parser would behave if '+' character would be used in a similar fashion.


riovir:
As discussed in #2227 and #2233, there are better approaches available for this problem, so closing this PR.

Issue #2208, opened 2024-03-05
select-rich when [read-only] it still works with keyboard interaction


gerjanvangeest:
## Expected behavior

- Keyboard interaction to be blocked
- also the active state button is still there

## Actual Behavior

Via the arrow keys another option can still be selected.
https://lion-web.netlify.app/components/select-rich/use-cases/#read-only-prefilled

wessamzaghloul:
active state behaves the same as in the 'Single Option', would this be then working as expected?

PR #2235, opened 2024-03-31
fix(select-rich): readonly works with keyboard interaction

wessamzaghloul:
## What I did

1. fixes [#2208](https://github.com/ing-bank/lion/issues/2208)

### DIFF
--- a/packages/ui/components/select-rich/src/LionSelectRich.js
+++ b/packages/ui/components/select-rich/src/LionSelectRich.js
@@ -477,7 +477,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
    */
   // TODO: rename to _onKeyUp in v1
   __onKeyUp(ev) {
-    if (this.disabled) {
+    if (this.disabled || this.readOnly) {
       return;
     }
 
--- a/packages/ui/components/select-rich/test/lion-select-rich.test.js
+++ b/packages/ui/components/select-rich/test/lion-select-rich.test.js
@@ -329,6 +329,21 @@ describe('lion-select-rich', () => {
       expect(elSingleoption.opened).to.be.false;
     });
 
+    it('stays closed with [ArrowUp] or [ArrowDown] key in readonly mode', async () => {
+      const elReadOnly = await fixture(html`
+        <lion-select-rich readonly>
+          <lion-option .choiceValue=${10}>Item 1</lion-option>
+          <lion-option .choiceValue=${20} checked>Item 2</lion-option>
+        </lion-select-rich>
+      `);
+
+      elReadOnly.dispatchEvent(new KeyboardEvent('keyup', { key: 'ArrowDown' }));
+      expect(elReadOnly.opened).to.be.false;
+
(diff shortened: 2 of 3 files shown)


packages/ui/components/select-rich/test/lion-select-rich.test.js
```diff
@@ -329,6 +329,21 @@ describe('lion-select-rich', () => {
...
+    it('stays closed with [ArrowUp] or [ArrowDown] key in readonly mode', async () => {
+      const elReadOnly = await fixture(html`
+        <lion-select-rich readonly>
+          <lion-option .choiceValue=${10}>Item 1</lion-option>
+          <lion-option .choiceValue=${20} checked>Item 2</lion-option>
+        </lion-select-rich>
+      `);
+
+      elReadOnly.dispatchEvent(new KeyboardEvent('keyup', { key: 'ArrowDown' }));
+      expect(elReadOnly.opened).to.be.false;
```
ryubro, line 341: Fix looks good to me!

I just want to share my finding that `elReadOnly.opened` will be `false` even without `readonly` attribute because `opened` is updated asynchronously.

Adding `await aTimeout(500)` between dispatching event and the `expect(...)` will do the job, although introducing timeout is not ideal. If you have a better idea, go for it!

The test for disabled has the same problem which was addressed in https://github.com/ing-bank/lion/pull/2237.

wessamzaghloul: Before adding my code changes, I ensured the test failed, proving its validity. It passed only after applying my changes. I also removed the `readonly` attribute to test further, as you suggested. The test failed in this scenario, indicating that `elReadOnly.opened` updates on time as expected.

ryubro: Sorry, indeed I cannot reproduce it. Sorry for the confusion!

packages/ui/components/select-rich/src/LionSelectRich.js
```diff
@@ -526,7 +526,7 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
   _listboxOnKeyDown(ev) {
     super._listboxOnKeyDown(ev);
 
-    if (this.disabled) {
+    if (this.disabled || this.readOnly) {
```
gerjanvangeest, line 529: This part of the code will only be triggered when the focus is on the listbox, so when already opened. It doesn't make sense to check if `readOnly` is on since the listbox should not get opened in the first place.

I'm not sure why we check for disabled here, i would expect that is also not needed.

gerjanvangeest (commented):
Can you add a changeset (`npm run changeset`), so the change will trigger the release PR and will end up in our changelog?

packages/ui/components/select-rich/src/LionSelectRich.js
wessamzaghloul: Yes, makes sense! reverted this part

gerjanvangeest (approved):
@wessamzaghloul thanks for the fix!

PR #2236, opened 2024-03-31
chore: add test watch

wessamzaghloul:
## What I did

1. Added test watch to ensure tests re-run while coding.


gerjanvangeest:
Hi @wessamzaghloul thanks for the PR.

We already have a script that is doing the same thing: "debug".

PR #2237, opened 2024-03-31
chore: misleading test disabled select rich

ryubro:
## What I did

1. Fixed a misleading test - more details on line comments
    - Found this during fixing https://github.com/ing-bank/lion/issues/2208 but noticed that there's already a PR for it: https://github.com/ing-bank/lion/pull/2235
2. Updated `.prettierignore` to include machine-generated files

### DIFF
--- a/packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
+++ b/packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
@@ -1,4 +1,4 @@
-import { expect, triggerBlurFor, triggerFocusFor, fixture } from '@open-wc/testing';
+import { expect, triggerBlurFor, triggerFocusFor, fixture, aTimeout } from '@open-wc/testing';
 import { Required } from '@lion/ui/form-core.js';
 import { html } from 'lit/static-html.js';
 import { browserDetection } from '@lion/ui/core.js';
@@ -201,6 +201,8 @@ describe('lion-select-rich interactions', () => {
       );
       const { invoker } = getNodes(el);
       invoker.click();
+      // Without waiting, el.opened will be always false, as toggling is async
+      await aTimeout(500);
       expect(el.opened).to.be.false;
     });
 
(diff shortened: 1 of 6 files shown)


packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
```diff
@@ -201,6 +201,8 @@ describe('lion-select-rich interactions', () => {
       );
       const { invoker } = getNodes(el);
       invoker.click();
+      // Without waiting, el.opened will be always false, as toggling is async
+      await aTimeout(500);
       expect(el.opened).to.be.false;
```
ryubro, line 206: Even if we change the line 197 to `<lion-select-rich>`, `el.opened` would be still `false` without waiting because it's not synchronous updating `el.opened` to `true` by calling `show()`.

Certainly, waiting for a specific time is not ideal, I tried `await el.updateComplete` but it seems not relevant.

Maybe we can add another prop like `el.toggleComplete` but not sure if it is a good idea to add a prop just for the sake of testing.

ryubro: https://github.com/ing-bank/lion/assets/9986886/fe921ab5-83b1-453e-92f9-9025a8416f29

gerjanvangeest: Can you add the `_overlayCtrl` to the `getNodes()` function.

```js
const { invoker, _overlayCtrl } = getNodes(el)
invoker.click();
await _overlayCtrl._showComplete;
```

I don't like random timeouts.

Better would be even to reuse the `getSelectRichMembers()` from `lion-select-rich.test.js` and replace the `getNodes()` function. That would mean we create a `test-helpers` folder as we have for overlays.

ryubro: Didn't know `_overlayCtrl._showComplete` exists. Thank you!

ryubro: Done!

PR #2193, opened 2024-02-15
fix(input-tel-dropdown): prevent jumping to input field on each arrow…

gerjanvangeest:
… key in windows/linux

## What I did

1.

PR #2245, opened 2024-04-05
fix: set focus correctly on Linux

okadurin:


packages/ui/components/input-tel-dropdown/src/LionInputTelDropdown.js
```diff
@@ -364,10 +364,12 @@ export class LionInputTelDropdown extends LionInputTel {
     // will set each item on arrow key up/down to activeElement
     // which causes the focus to jump every time to the inputNode
     const overlayController = dropdownElement._overlayCtrl;
-    // @ts-ignore interactionMode only exists on LionSelectRich not on HTMLSelectElement
-    if (overlayController?.isShown && dropdownElement.interactionMode !== 'windows/linux') {
```
gerjanvangeest, line 368: We need this line of code:
https://github.com/ing-bank/lion/pull/2193

But feel free to adjust the test to make it work properly on linux.

gerjanvangeest (approved):

PR #2247, opened 2024-04-09
feat(form-core): add a way to prevent registration of a form-element

gerjanvangeest:
## What I did

1.


tlouisse:
Ideally the developer using composition inside a FormControl should not be bothered with this. 
I think we can do it in such a way that we align this behavior with how modelValues are propagated (LionInputTelDropdown is already considered an "endpoint". (Fields are endpoints, FormGroups aren't))

Screenshot 2024-04-09 at 12 10 05

Fields have `this._repropagationRole = 'child'`. How this is currently handled for the modelValue repropagation: https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-tel-dropdown/src/LionInputTelDropdown.js#L445

We should solve both (modelValue repropagation and registration)in a more generic way.
Maybe we can do:

In FormControlMixin:
```js
/**
 * Whether 
 * @readonly
 */
get _isEndPoint() {
  return this.__isEndpoint || (this._repropagationRole === 'child');
}
  
/**
 * We should only repropagate modelValue events if we are a group (not if we are an endpoint without children)
 */
_repropagationCondition() {
  return !this._isEndpoint;
}
```

In FormRegistrarMixin:
```js
// prevent child registration if _isEndpoint
```

N.B. above is pseude code and there will be a few edge cases we need to tackle as well. 
(edge cases will mainly be the situations with options (child) registering to choice-group.)

PR #2251, opened 2024-04-14
Update various github actions and node versions

koddsson:
## What I did

1. I updated some old node versions in CI (node 16 is no longer supported)
2. I updated some github actions
3. I changed some npm commands from `npm install --ci` to `npm ci`


koddsson:
I.. did not expect this many errors.

koddsson:
An admin is going to need to update the required PR builds so that node 18 is required and not 16 if we are going to merge this.

PR #2255, opened 2024-04-15
fix: optimize the way we focus the input field for LionInputTelDropdown

okadurin:
Instead of using setTimeout to focus the input field, let's use the `elementToFocusAfterHide` property


gerjanvangeest (approved):
Nice solution :)

PR #2261, opened 2024-04-18
Fix select rich key interaction not setting correct isTriggeredByUser value

01Sharpshooter:
## What I did

1. I fixed [ArrowUp]/[ArrowDown] not setting `isTriggeredByUser` to `true` when done directly on the select-rich component, not in the listbox


01Sharpshooter:
Closing this because I pushed with wrong user info. Still learning this open-source thing...

PR #2266, opened 2024-04-18
Make getCacheIdentifier config property asynchronous

ing-athirlwall:
## What I did

1. Make the getCacheItentifier function asynchronous. This would make the Lion Ajax library compatible with using the profilesReady function which retrieves the profiles within the Enhanced WebView.


packages/ajax/src/interceptors/cacheInterceptors.js
```diff
@@ -59,14 +59,14 @@ const isResponseSizeSupported = (responseSize, maxResponseSize) => {
...
- * @param {function(): string} getCacheId used to invalidate cache if identifier is changed
+ * @param {function(): Promise<string>} getCacheId used to invalidate cache if identifier is changed
  * @param {CacheOptions} globalCacheOptions
  * @returns {RequestInterceptor}
  */
 const createCacheRequestInterceptor =
   (getCacheId, globalCacheOptions) => /** @param {CacheRequest} request */ async request => {
     validateCacheOptions(request.cacheOptions);
-    const cacheSessionId = getCacheId();
+    const cacheSessionId = await getCacheId();
```
tlouisse, line 69: Although `createCacheRequestInterceptor` is an async function, 
inside our code we don't use `await` anywhere (yet). This means it can be called as a sync function until now as well.

Inside out own code, we do: https://github.com/ing-bank/lion/blob/master/packages/ajax/src/interceptors/cacheInterceptors.js#L168

So inside our publicly exported `createCacheInterceptors` function we catch and return `cacheRequestInterceptor` with type `RequestInterceptor` atm. This would change it to `Promise<RequestInterceptor>`.

 So even though `createCacheRequestInterceptor` is defined as async atm, it's currently treated as a sync function in our own code (and potentially by our consumers).
 
 The best way to make it backwards compatible:
 - createCacheRequestInterceptor returns `{RequestInterceptor|Promise<RequestInterceptor>}`
 - getCacheId param is type `() => string|Promise<string>`
 
 ```suggestion
    const getCacheIdResult = getCacheId();
    const isPromise = 'then' in getCacheIdResult;
    const cacheSessionId = isPromise ? await getCacheIdResult : getCacheIdResult;
```

Ideally the full chain should be promises (so we should add await here: https://github.com/ing-bank/lion/blob/master/packages/ajax/src/interceptors/cacheInterceptors.js#L168)
. But it would be a breaking change

Comment (ing-athirlwall) [reply packages/ajax/src/interceptors/cacheInterceptors.js:69]: So we can do the same "dual type" trick in Ajax.js where we can define the getCacheIdentifier option as string|Promise<string>.

**Review by tlouisse — APPROVED** [review]

File: packages/ajax/test/interceptors/cacheInterceptors.test.js [hunk 2]
```diff
@@ -36,6 +38,8 @@ describe('cache interceptors', () => {
   /** @type {Response} */
   let mockResponse;
   const getCacheIdentifier = () => String(cacheId);
+  const getCacheIdentifierAsync = () => String(cacheId);
```
tlouisse, line 41:
```suggestion
  const getCacheIdentifierAsync = () => Promise.resolve(String(cacheId));
```

PR #2279, opened 2024-05-13
fix(form-core): Implement conversion of name attribute value to string type

donecarlo:
## What I did

1. Updated the behavior of the `name` attribute to convert values to `string` type

### DIFF
--- a/packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
+++ b/packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
@@ -55,7 +55,7 @@ const FormRegisteringMixinImplementation = superclass =>
 
     /**
      * Converts values provided for the `name` attribute to string type.
-     * Mimicks the native `input` behavior.
+     * Mimics the native `input` behavior.
      * @param {string} newName
      */
     set name(newName) {
--- a/packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
+++ b/packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
@@ -23,6 +23,14 @@ const FormRegisteringMixinImplementation = superclass =>
     constructor() {
       super();
       /**
+       * The name the element will be registered with to the .formElements collection
+       * of the parent. Also, it serves as the key of key/value pairs in
+       *  modelValue/serializedValue objects
+       * @type {string}
+       */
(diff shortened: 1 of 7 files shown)


packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
```diff
@@ -37,8 +45,28 @@ const FormRegisteringMixinImplementation = superclass =>
...
+     * Name attribute for the control.
+     * @type {string}
+     */
+    get name() {
+      return this.__name || '';
+    }
+
+    /**
+     * Converts values provided for the `name` attribute to string type.
+     * Mimicks the native `input` behavior.
```
tlouisse, line 58:
```suggestion
     * Mimics the native `input` behavior.
```

packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
```diff
@@ -37,8 +45,28 @@ const FormRegisteringMixinImplementation = superclass =>
...
+     */
+    set name(newName) {
+      const oldName = this.name;
+      this.__name = newName.toString();
+      this.requestUpdate('name', oldName);
+    }
+
     static get properties() {
       return {
+        name: { type: String, reflect: true },
```
tlouisse, line 69: Since this will be inherited by FormControlMixin: can we remove it there?

donecarlo: Updated

packages/ui/components/form-core/src/registration/FormRegisteringMixin.js
donecarlo: Done

packages/ui/components/form-core/types/FormControlMixinTypes.ts
```diff
@@ -75,7 +71,7 @@ export declare class FormControlHost {
 
   /**
    * The name the element will be registered with to the .formElements collection
-   * of the parent. Also, it serves as the key of key/value pairs in
+   * of the parent. Also, it serves as the keny of key/value pairs in
```
tlouisse, line 74:
```suggestion
   * of the parent. Also, it serves as the key of key/value pairs in
```

PR #2281, opened 2024-05-14
fix(tabs): make tab panels focusable

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/tabs/src/LionTabs.js
+++ b/packages/ui/components/tabs/src/LionTabs.js
@@ -19,6 +19,16 @@ function setupPanel({ el, uid }) {
   el.setAttribute('id', `panel-${uid}`);
   el.setAttribute('role', 'tabpanel');
   el.setAttribute('aria-labelledby', `button-${uid}`);
+  /**
+   * Facilitates navigation to panel content for assistive technology users.
+   *
+   * Focusable tab panel elements are recommended if any panels in a set contain
+   * content where the first element in the panel is not focusable.
+   * https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/
+   */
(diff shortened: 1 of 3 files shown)


packages/ui/components/tabs/src/LionTabs.js
```diff
@@ -19,6 +19,16 @@ function setupPanel({ el, uid }) {
...
   el.setAttribute('aria-labelledby', `button-${uid}`);
+  /**
+   * Facilitates navigation to panel content for assistive technology users.
+   *
+   * Focusable tab panel elements are recommended if any panels in a set contain
+   * content where the first element in the panel is not focusable.
+   * https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/
+   */
+  if (!el.hasAttribute('tabindex')) {
+    el.setAttribute('tabindex', '0');
```
tlouisse, line 30: maybe we can also clean up on cleanButton?
(where it's usually better to restore initial state (could be stored in a Weakmap) then a regardless removal)

tlouisse: I think it's called here on slot change:
https://github.com/ing-bank/lion/blob/d493f6c82d82462ccbb0580531a90d1e74ad550d/packages/ui/components/tabs/src/LionTabs.js#L196

Basically we say here: "whatever content is now available in the slot (it might be different from the contents that were previously inside the slot), make sure that we leave the elements we altered in a clean state (their initial state)"

tlouisse: Created an issue for doing this consistently throughout the codebase: https://github.com/ing-bank/lion/issues/2295

tlouisse (approved):

PR #2291, opened 2024-05-17
fix(slotmixin): Now throws shadowRoot not found in case of shadowRoot…

Dozom:
… equals to null

## What I did

1. I throw an error if shadowRoot was not found.

### DIFF
--- a/packages/ui/components/core/src/SlotMixin.js
+++ b/packages/ui/components/core/src/SlotMixin.js
@@ -134,6 +134,7 @@ const SlotMixinImplementation = superclass =>
         const hasShadowRoot = Boolean(this.shadowRoot);
         if (!hasShadowRoot) {
           // TODO: throw an error in a breaking release
+          // eslint-disable no-console
           console.error(`[SlotMixin] No shadowRoot was found`);
         }
         const registryRoot = supportsScopedRegistry ? this.shadowRoot : document;
--- a/packages/ui/components/core/src/SlotMixin.js
+++ b/packages/ui/components/core/src/SlotMixin.js
@@ -134,7 +134,7 @@ const SlotMixinImplementation = superclass =>
         const hasShadowRoot = Boolean(this.shadowRoot);
         if (!hasShadowRoot) {
           // TODO: throw an error in a breaking release
-          // eslint-disable no-console
+          // eslint-disable-next-line no-console
           console.error(`[SlotMixin] No shadowRoot was found`);
         }
         const registryRoot = supportsScopedRegistry ? this.shadowRoot : document;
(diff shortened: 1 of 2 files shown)


packages/ui/components/core/src/SlotMixin.js
```diff
@@ -131,7 +131,11 @@ const SlotMixinImplementation = superclass =>
       if (isFirstRender) {
         // @ts-expect-error wait for browser support
         const supportsScopedRegistry = !!ShadowRoot.prototype.createElement;
-        const registryRoot = supportsScopedRegistry ? this.shadowRoot || document : document;
+        const hasShadowRoot = Boolean(this.shadowRoot);
+        if (!hasShadowRoot) {
+          throw new Error(`No shadowRoot was found`);
```
tlouisse, line 136:
```suggestion
          // TODO: throw an error in a breaking release
          console.error(`[SlotMixin] No shadowRoot was found`);
```

Hey, thanks for you contribution :) 
Since I think there are consumers that use this mixin, throwing an error would be breaking. So I think we can best leave that for a breaking release

tlouisse: Could you also add a changeset? (via `npm run hcangeset`)

Dozom:
Hello @tlouisse! It was my first contribution to open source, 
1. When do I have to execute npm run changelog, after pulling the changes or before? 
2. How can I add the changelog?

Actual step (local fork):
image

It created a .changeset/odd-poets-grow.md
with the following content, 
---
'@lion/ui': major
---

fix(slotmixin): Now prints a console error when shadowRoot is not found

What is this used for? 

Thank's for the answer!

tlouisse:
Hi @Dozom 
> 1. When do I have to execute npm run changelog, after pulling the changes or before?
It's a good practice to attach a changeset entry to a commit(`npm run changeset` writes an md file to  the `.changesets` folder) to the commit. 
> 2. How can I add the changelog?
This is done automatically in a separate pipeline (@changeset/cli generates the changelog based on the changeset files: https://github.com/changesets/changesets?tab=readme-ov-file)

Dozom:
Hi @tlouisse
I added the changeset entry, is it correct? 
On the other hand, thank's for documentation I really appreciated. 😄

packages/ui/components/core/src/SlotMixin.js
```diff
@@ -131,7 +131,12 @@ const SlotMixinImplementation = superclass =>
       if (isFirstRender) {
         // @ts-expect-error wait for browser support
         const supportsScopedRegistry = !!ShadowRoot.prototype.createElement;
-        const registryRoot = supportsScopedRegistry ? this.shadowRoot || document : document;
+        const hasShadowRoot = Boolean(this.shadowRoot);
+        if (!hasShadowRoot) {
+          // TODO: throw an error in a breaking release
+          console.error(`[SlotMixin] No shadowRoot was found`);
```
tlouisse, line 137:
```suggestion
          // eslint-disable no-console
          console.error(`[SlotMixin] No shadowRoot was found`);
```

tlouisse (approved):

tlouisse:
> Hi @tlouisse I added the changeset entry, is it correct? On the other hand, thank's for documentation I really appreciated. 😄

Yes, looks good. Thanks for the contribution!

Issue #2300, opened 2024-06-06
OverlayMixin causes a memory leak


Sciurus7:
## Reproduction Recipe
- Go to a page containing a component using the OverlayMixin that can be removed. (See [the example I created](https://lion-overlay-mixin-memory-leak.netlify.app/) for a very basic component containing the OverlayMixin.)
- Check and uncheck the checkbox. That'll create and remove a simple component using the overlay. 
- In the Chrome devtools, go to the memory tab. 
- Do a GC (the broomstick icon) just to be sure. 
- Take a heap snapshot
- In the list, find the entry for "Detached HTMLElement". That'll show you the elements that are not part of the DOM any more, but cannot be garbage collected. 

## Expected behavior
There are no detached HTMLElements

## Actual Behavior
The component with the overlay-mixin can not be garbage collected.

## Additional context
This bug has also been tested and reproduced with the LionSelectRich.
Based on research this bug is partially caused by the OverlayManager, which adds every OverlayController to a list, but never removes it. 
The OverlayController should probably be removed from the list during the disconnectedCallback of the OverlayMixin. Unfortunately even when taking this step the memory leak persists so there must be another cause for the memory leak.

### Lion version
0.7.3

### Code base example
https://github.com/Sciurus7/lion-overlay-mixin-memory-leak

### Additional research notes
The memory leak definitely seems to be linked to the OverlaysManager. When setting this.manager to null in the OverlayController constructor the memory leak will disappear. Some data collected during the investigation.

While the OverlayConstructor will add itself to a list in the OverlaysManager during its initialization, it never removes itself of this list while it is being removed.  This can be remedied by removing the OverlayController from that list when it’s torndown. This should be done done in the OverlayMixin:

```
/** @protected */
_teardownOverlayCtrl() {
   this._teardownOpenCloseListeners();
   this.__teardownSyncFromOverlayController();
   /** @type {OverlayController} */
   (this._overlayCtrl).teardown();
   this._overlayCtrl.manager.remove(this._overlayCtrl); // To be added in a Lion's code
}
```

Even after the above change the memory leak persists. This happens even when the OverlayManager is basically an empty class with just empty methods and no props. 

During a simple lifecycle of a component with an OverlayMixin where it is just created and removed he following methods are called on the manager:
- add
- hide
- hide
- remove

riovir:
[...]

PR #2302, opened 2024-06-11
fix(overlays): fix memory leaks of adopt-styles and OverlayMixin (#2300)

riovir:
## What I did

1. Adjust the `adopt-style` styleCache to be a `WeakMap` preventing the `renderRoot`s from being garbage collected
2. Let the `OverlayMixin` unwire its `OverlayController` from the `OverlayManager` on teardown

### DIFF
--- a/.changeset/quiet-apricots-sneeze.md
+++ b/.changeset/quiet-apricots-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+[overlays] Fixed memory leak caused adopted style cache and the `OverlayMixin` not releasing the `OverlayController` on teardown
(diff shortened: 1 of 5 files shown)


riovir:
Will need to look closer into how to keep the OverlayController reference in the OverlayManager without never releasing it. Weak collections seem to be a no-go, as iteration is required. I'm hoping that carefully cutting and restoring the references in the connected / disconnected or init / teardown hooks will get the right results without introducing regression.

Sciurus7:
That is looking good! 
It would be much appreciated if this PR could be handled and released relatively quickly. My company is working on a framework where this memory leak will wreak havoc unless fixed. Unfortunately it is nigh impossible to implement some temporary workaround in our own components that would apply the same fixes as this PR, so we are really dependent on this being fixed in Lion.

tlouisse:
Sorry for the late response, this is a really valuable PR! 
Thanks a lot for the great debugging work and the pull request @riovir and @Sciurus7 👏 

FYI: I  added a test and moved the logic from OverlayMixin to OverlayController (as the controller can be used in isolation as well).

tlouisse (approved):

PR #2308, opened 2024-06-27
fix: add files more than once doesn't overwrite anymore

ryubro:
## What I did

1. Fixed a bug that when user adds files more than once, the model value is overwritten by the last addition but the file list grows, adding the last addition to the existing list
   - Example: User added two files, `a` and `b` in a file dialog, then  added another file `c` opening a file dialog once more. The file list would have `a`, `b`, `c` but the model value would be `c`. After the fix, the model value also would be `a`, `b`, `c`.
   - Tested on both dialog and drag & drop

----
- [x] Fixing an issue when we upload the same file, the model value gets empty.

### DIFF
--- a/packages/ui/components/input-file/src/LionInputFile.js
+++ b/packages/ui/components/input-file/src/LionInputFile.js
@@ -499,13 +499,13 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
 
     this._inputNode.files = ev.dataTransfer.files;
 
-    const computedFiles = this.__computeNewAddedFiles(
-      /** @type {InputFile[]} */ (Array.from(ev.dataTransfer.files)),
-    );
     if (this.multiple) {
+      const computedFiles = this.__computeNewAddedFiles(
+        /** @type {InputFile[]} */ (Array.from(ev.dataTransfer.files)),
+      );
       this.modelValue = [...(this.modelValue ?? []), ...computedFiles];
     } else {
-      this.modelValue = computedFiles;
+      this.modelValue = Array.from(ev.dataTransfer.files);
     }
 
     // @ts-ignore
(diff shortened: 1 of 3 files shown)


ryubro:
Fixing an issue when we upload the same file, the model value gets empty.

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -481,15 +498,17 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
...
+    if (this.multiple) {
+      const computedFiles = this.__computeNewAddedFiles(
+        /** @type {InputFile[]} */ (Array.from(ev.dataTransfer.files)),
+      );
+      this.modelValue = [...(this.modelValue ?? []), ...computedFiles];
+    } else {
+      this.modelValue = Array.from(ev.dataTransfer.files);
+    }
+
     // @ts-ignore
```
gerjanvangeest, line 511: I guess this line can also be removed

packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -609,6 +628,7 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
       .map(({ systemFile }) => /** @type {InputFile} */ (systemFile));
 
     if (_successFiles.length > 0) {
+      // this.modelValue = [...this.modelValue, _successFiles];
```
gerjanvangeest, line 631: Don't add unused lines

ryubro: Sorry, a left over from a tinkering. Removed!

packages/ui/components/input-file/src/LionInputFile.js
ryubro: This one was required due to typing, but now it's replaced with type casting comment.

gerjanvangeest (approved):

PR #2310, opened 2024-07-03
fix(overlays): prevent closing of a modal dialog on pressing Esc and …

gerjanvangeest:
…hidesOnEsc is set to false

## What I did

1. prevent the default when the [cancel event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/cancel_event) was fired, so that `hidesOnEsc: false` works.


okadurin (approved):

tlouisse:
Nice 👍

PR #2311, opened 2024-07-10
Use `<dialog>` element in `<lion-dialog>`

koddsson:
I think we can probably remove a lot of the overlay code by using native web primitives at this point. Here's my attempt at replacing the overlays mixin in `lion-dialog` by just using a `<dialog>` element.

Note that I removed a couple of tests that I thought didn't make sense. 

- "with trapsKeyboardFocus set to false the focus stays on the invoker"
  - I don't know when you'd actually want this so I wanted to remove that functionality as well as the test.
- "sets focus on contentSlot by default"
  - I'm not sure this is the correct behaviour. Without changing anything, the `<dialog>` element will autofocus a `<input>` element like in the test so I'm inclined to think it's the correct behaviour. Happy to be proven wrong on that one.

### DIFF
--- a/packages/ui/components/dialog/src/LionDialog.js
+++ b/packages/ui/components/dialog/src/LionDialog.js
@@ -43,7 +43,7 @@ export class LionDialog extends LitElement {
       <button part="invoker" @click=${this.#open}>
         <slot name="invoker"></slot>
       </button>
-      <dialog id="overlay-content-node-wrapper">
+      <dialog>
         <slot name="content"></slot>
       </dialog>
     `;
--- a/packages/ui/components/dialog/src/LionDialog.js
+++ b/packages/ui/components/dialog/src/LionDialog.js
@@ -15,8 +15,11 @@ export class LionDialog extends LitElement {
     this.removeEventListener('close', this.#close);
   }
 
+  /**
+   * @returns {HTMLDialogElement}
+   */
   get #dialog() {
-    return this.shadowRoot.querySelector('dialog');
+    return /** @type {HTMLDialogElement} */ (this.renderRoot.querySelector('dialog'));
   }
 
   #open() {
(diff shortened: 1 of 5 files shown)


packages/ui/components/dialog/src/LionDialog.js
```diff
@@ -1,23 +1,51 @@
...
+    }
+  `;
+
   render() {
     return html`
-      <slot name="invoker"></slot>
-      <div id="overlay-content-node-wrapper">
+      <button part="invoker" @click=${this.#open}>
+        <slot name="invoker"></slot>
+      </button>
```
koddsson, line 45: I thought it was wrong to let the consumer decide what kind of invoker they wanted to use. I think it should always be a button so that the semantics are always correct and the accessibility is more sound.

packages/ui/components/dialog/src/LionDialog.js
```diff
@@ -1,23 +1,51 @@
...
+      <dialog id="overlay-content-node-wrapper">
```
koddsson, line 46: I don't think we need this anymore.

```suggestion
      <dialog>
```

packages/ui/components/dialog/src/LionDialog.js
```diff
@@ -1,23 +1,51 @@
...
+    this.opened = false;
+  }
+
+  static styles = css`
+    :host {
+      --border: 1px solid black;
+    }
+    dialog {
+      border: var(--border);
+    }
```
koddsson, line 38: Maybe this should just be a part too?

koddsson:
Weird because these tests run green for me when I run them against Safari headful on my local machine.

koddsson:
> Weird because these tests run green for me when I run them against Safari headful on my local machine.

It does fail when I run the tests headless but when then it fails when I add a `input.focus()` in there 🤔

koddsson:
> > Weird because these tests run green for me when I run them against Safari headful on my local machine.
> 
> It does fail when I run the tests headless but when then it fails when I add a `input.focus()` in there 🤔

Ah no, I was actually having two different tests failing (but with very similar names) and getting confused between the two. Digging into them now!

koddsson:
Now it seems this is only failing in tests because of some autofocus heuristics in Safari (and is backed up by https://github.com/shoelace-style/shoelace/issues/688).

I'm hesitant to call `.focus()` to force the autofocus in tests so I think I might just delete the test and rely on that the browser should be doing the correct thing.

koddsson:
Let me know what y'all think :)

PR #2322, opened 2024-07-26
Fix/deep contains

okadurin:
Fix `deep-contains` function to allow it to work correctly with slots


packages/ui/components/overlays/src/utils/deep-contains.js
```diff
@@ -10,11 +10,51 @@ export function deepContains(el, targetEl) {
...
+
   /** @param {HTMLElement|ShadowRoot} elem */
   function checkChildren(elem) {
     for (let i = 0; i < elem.children.length; i += 1) {
       const child = /** @type {HTMLElement}  */ (elem.children[i]);
-      if (child.shadowRoot && deepContains(child.shadowRoot, targetEl)) {
+      const slotProjectionElement = getSlotProjection(child);
+      if (
+        (child.shadowRoot && deepContains(child.shadowRoot, targetEl)) ||
+        (slotProjectionElement && deepContains(slotProjectionElement, targetEl))
```
tlouisse, line 56: In the next level shadowRoot (recursive call of deepContains), we will now check the shadowRoot of the parent again...

Maybe we can omit this slot check for level > 1 (to improve perfomance)?

okadurin: I have updated the code. Not exactly what you are proposing but it could cover you concern if I understood you correctly

tlouisse: I basically mean that if you have:

```html
<my-el>
  <div data-level="0" slot="projected-from-lvl-1"></div>
  #shadowroot
    <div data-level="1" slot="projected-from-lvl-2"></div>
    <slot name="project-from-lvl-1"></slot>
    #shadowroot
       <div data-level="2"></div>
       <slot name="projected-from-lvl-2"></slot>
      #shadowroot
        <div data-level="3"></div>
        #shadowroot
          <div data-level="3"></div>
          #shadowroot
```

If you provide an element (`<div data-level="1" slot="projected-from-lvl-2"></div>`) in first shadowRoot (`deepContains(elInFirstShadowRoot, ...)`) as search root, you basically want to make sure that the slotted content 
`<div data-level="0" slot="projected-from-lvl-1"></div>` is also found. 

We will not find it under the children of provided search root, as its parent shadowRoot hasn't been searched yet (so for this level you want to check assignedNodes of slots) . However, for recursive calls (when we reach level 2) this check has already taken place as the traversal of the content of shadowroot lvl1 has already encountered this (`<div data-level="1" slot="projected-from-lvl-2"></div>`) was already found...

tlouisse: So the solution would be to provide a shadowRootLevel = 0 as third param and for internal recursive calls it gets a +1 every time when dealing with a new shadowRoot. The slot only gets checked for shadowRootLevel of 0

okadurin: Thank you for details explanation, Thijs! Unfortunately I am yet not completely following.
I have added another test where the [target element](https://github.com/ing-bank/lion/blob/78229f4fcc20fe3dac2523910ad6986f56d4d4a0/packages/ui/components/overlays/test/utils-tests/deep-contains.test.js#L172) is located inside the multiple levels of ShadowRoot's of the [LightDom element](https://github.com/ing-bank/lion/blob/78229f4fcc20fe3dac2523910ad6986f56d4d4a0/packages/ui/components/overlays/test/utils-tests/deep-contains.test.js#L167). This way it does not look that we need to limit the search to `shadowRootLevel = 0`. Before I go ahead with other questions regarding your comment, could you have a look at the new tests and the template and tell me if your statement is still valid? Just to make sure we understand the task the same way.

tlouisse: Hey,

Sorry for the bad explanation on my end.

The current code works perfect functionality wise already 👍 . My suggestion was only to improve performance and avoid double checks.

So maybe it can be explained a bit more simplified by adding a third param:

```js

// Right now, the util only checks its children (the children it owns)
// So it does not have access to elements above searchRootEl (projected in a shadow dom of a parent that it does not own (as it owns the slot elements, but not the nodes assigned to them)
deepContains(searchRootEl, targetEl);

// Now we can change this behavior, as it might be convenient for our own overlays
deepContains(searchRootEl, targetEl, { searchInAssignedSlotElements: true });
```
The latter allows us to also search in the parent shadowRoot of searchRootEl.
However, if we call deepContains recursively (into a next shadowRoot) there is no need to look for assignedElements anymore. Since the potential slot contents are already searched by then...

So for recursive calls , you would then do:

```js
function deepContains(searchRootEl, targetEl, { searchInAssignedSlotElements } = {}) {
   // ...
   deepContains(searchRootEl, targetEl, { searchInAssignedSlotElements: false });
   // ...
}
```

okadurin: Thijs, I added the code for caching the element after it's been processed completely (its children and shadowRoot if any). Then if the code hits the element again it gets the deepContains function result for the element from the cache instead of processing the same file twice.

tlouisse: > Thijs, I added the code for caching the element after it's been processed completely (its children and shadowRoot if any). Then if the code hits the element again it gets the deepContains function result for the element from the cache instead of processing the same element twice.

Ah I see what you mean now 💪  (there can be situations where multiple level traversal is needed indeed).
I also added an update, since the light dom might contain multiple instances of `[slot=content]`.

On top, we can use declarative shadow dom these days. That might make the tests a bit more readable: https://caniuse.com/declarative-shadow-dom

okadurin: Thanks! I have fixed the test and updated the code to improve readability of caching feature

tlouisse (approved):

Issue #2323, opened 2024-07-31
[LionOverlays]: Close Button ('x') causes growing margin-right on body element


ka9de:
## Expected behavior

I expect the dialog/overlay to not add margin on my body element when I open and close it, if using the x-button.

## Actual Behavior

When I use a dialog with x-button in the header, **hovering** this button causes a growing **margin-right** on the body element. So if you open and close the dialog repeatedly (using the x-button and waiting 3 seconds after opening), your body element will get smaller and smaller, causing the background (white) to appear on the right side.

This is caused by the x-buttons **tooltip**, which is calling the _keepBodySize method of OverlayController.js and interfering with the values that have been set by the OverlayManager before.

tlouisse:
Likely solution:

We should add on top of https://github.com/ing-bank/lion/blob/master/packages/ui/components/overlays/src/OverlayController.js#L790
 
```
if (!this.preventsScroll) return;
```

And then we need a test for this

ka9de:
Hi @tlouisse 
I created a PR https://github.com/ing-bank/lion/pull/2331 but I am really not familiar with Github and changesets, obviously something did now work correctly.

tlouisse:
closed by https://github.com/ing-bank/lion/pull/2331

PR #2331, opened 2024-08-12
Fix/overlay growing margin, issue 2323 

ka9de:
## What I did

1. fix the _keepBodySize to return instantly if controller uses local placement
2. wrote a test to confirm this

### DIFF
--- a/.changeset/avoid-growing-margin.md
+++ b/.changeset/avoid-growing-margin.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(2323): avoid growing margin when using overlays with tooltips
--- a/.changeset/avoid-growing-margin.md
+++ b/.changeset/avoid-growing-margin.md
@@ -2,4 +2,4 @@
 '@lion/ui': patch
 ---
 
-fix(2323): avoid growing margin when using overlays with tooltips
+fix: [overlays] avoid growing margin when using `preventsScroll: false`
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -787,6 +787,10 @@ export class OverlayController extends EventTarget {
    * @protected
    */
   _keepBodySize({ phase }) {
+    if (this.placementMode !== 'global') {
+      return;
+    }
+
     switch (phase) {
       case 'before-show':
         this.__bodyClientWidth = document.body.clientWidth;
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -787,7 +787,7 @@ export class OverlayController extends EventTarget {
    * @protected
    */
   _keepBodySize({ phase }) {
-    if (this.placementMode !== 'global') {
+    if (!this.preventsScroll) {
       return;
     }
 
--- a/packages/ui/components/overlays/test/OverlayController.test.js
+++ b/packages/ui/components/overlays/test/OverlayController.test.js
@@ -1764,4 +1764,20 @@ describe('OverlayController', () => {
       );
     });
   });
+
+  describe('run _keepBodySize only with global placement', () => {
+    /**
+     * @type {OverlayController}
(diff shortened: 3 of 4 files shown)


.changeset/avoid-growing-margin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(2323): avoid growing margin when using overlays with tooltips
```
tlouisse, line 5:
```suggestion
fix: [overlays] avoid growing margin when using overlays with tooltips
```

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -787,6 +787,10 @@ export class OverlayController extends EventTarget {
    * @protected
    */
   _keepBodySize({ phase }) {
+    if (this.placementMode !== 'global') {
```
tlouisse, line 790:
```suggestion
    if (!this.preventsScroll) {
```

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1764,4 +1764,20 @@ describe('OverlayController', () => {
       );
     });
   });
+
+  describe('run _keepBodySize only with global placement', () => {
+    /**
+     * @type {OverlayController}
+     */
+    const overlayControllerLocal = new OverlayController({
+      ...withLocalTestConfig(),
```
tlouisse, line 1773:
```suggestion
      ...withLocalTestConfig(),
      preventsScroll: true,
```

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1764,4 +1764,20 @@ describe('OverlayController', () => {
...
+    });
+
+    it('should not run with local placement', async () => {
+      await overlayControllerLocal.show();
+
+      expect(overlayControllerLocal.__bodyMarginRightInline).to.equal(undefined);
+      expect(overlayControllerLocal.__bodyMarginRight).to.equal(undefined);
+    });
```
tlouisse, line 1781: Looks great, maybe you can also test here for the `preventsScroll: true` case...

```
      overlayControllerLocal.updateConfig({ preventsScroll: true }); 

      expect(overlayControllerLocal.__bodyMarginRightInline).to.not.equal(undefined);
      expect(overlayControllerLocal.__bodyMarginRight).to.not.equal(undefined);

```

tlouisse (changes requested):
Thanks for the PR!

The changelog was correct indeed 👍 

Can you look at the suggestions? It would also be nice if you could squash all commits into one (but we can also do it while merging...)

When these latest changes are implemented, we will merge it 💪

tlouisse:
closes https://github.com/ing-bank/lion/issues/2323

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1764,4 +1764,20 @@ describe('OverlayController', () => {
...
+     */
+    const overlayControllerLocal = new OverlayController({
+      ...withLocalTestConfig(),
+    });
+
+    it('should not run with local placement', async () => {
+      await overlayControllerLocal.show();
+
+      expect(overlayControllerLocal.__bodyMarginRightInline).to.equal(undefined);
+      expect(overlayControllerLocal.__bodyMarginRight).to.equal(undefined);
```
tlouisse, line 1780:
```suggestion
      expect(overlayControllerLocal.__bodyMarginRight).to.equal(undefined);
      
      overlayControllerLocal.updateConfig({ preventsScroll: true }); 
      expect(overlayControllerLocal.__bodyMarginRightInline).to.not.equal(undefined);
      expect(overlayControllerLocal.__bodyMarginRight).to.not.equal(undefined);
```

ka9de (commented):
I fixed some linting problem and updated the tests with some text

tlouisse (approved):

tlouisse:
Thanks for the PR!

PR #2332, opened 2024-08-15
fix(combobox): when multiple choice reset all listbox options on select

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/combobox/src/LionCombobox.js
+++ b/packages/ui/components/combobox/src/LionCombobox.js
@@ -22,6 +22,7 @@ const matchA11ySpanReverseFns = new WeakMap();
  * @typedef {import('@lion/ui/types/form-core.js').ChoiceInputHost} ChoiceInputHost
  * @typedef {import('@lion/ui/types/form-core.js').FormControlHost} FormControlHost
  * @typedef {import('../types/SelectionDisplay.js').SelectionDisplay} SelectionDisplay
+ * @typedef {LionOption & { onFilterUnmatch?:function, onFilterMatch?:function }} OptionWithFilterFn
  */
 
 /**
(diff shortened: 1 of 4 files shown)


packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -3018,6 +3063,37 @@ describe('lion-combobox', () => {
       expect(el.opened).to.equal(true);
     });
 
+    it('resets the inputNode.value and listbox options on [Enter]', async () => {
```
tlouisse, line 3066:
```suggestion
    it('clears texbox after selection of a new item', async () => {
```

tlouisse: Right now, it does not work on click for instance...

tlouisse: And listbox options should not be cleared right? (they are not atm, but previous it statement seemed to suggest they should)

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -201,8 +202,24 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
...
-    // @ts-ignore _initialModelValue comes from ListboxMixin
-    this.value = this._initialModelValue;
+    if (!this.multipleChoice) {
+      // @ts-ignore _initialModelValue comes from ListboxMixin
+      this.value = this._initialModelValue;
+    }
+    this.resetListboxOptions();
+  }
+
+  resetListboxOptions() {
```
tlouisse, line 212: Can this be a private method?

gerjanvangeest: Protected is possible, we need to override it internally.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -201,8 +202,24 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
 
   reset() {
     super.reset();
-    // @ts-ignore _initialModelValue comes from ListboxMixin
-    this.value = this._initialModelValue;
+    if (!this.multipleChoice) {
+      // @ts-ignore _initialModelValue comes from ListboxMixin
+      this.value = this._initialModelValue;
```
tlouisse, line 207: If we set the modelValue, it triggers the "main loop". Then the modelValue is synced with the "presentation" (how the individual options look and their a11y attrs are handled)

gerjanvangeest: It now added the value to the inputNode.Value, which is not needed for multipleChoice.

packages/ui/components/combobox/test/lion-combobox.test.js
tlouisse: https://github.com/user-attachments/assets/2488d593-059e-48af-8dcc-b37a415d422d

gerjanvangeest: Also the listbox options should be cleanup, otherwise only a limited amount of options with highlighting are shown.

ryubro:
When I click the field for the first time, I don't see any list until I type something. But once I select something, clicking the field opens the whole list. Is this behavior difference intended? 

https://github.com/user-attachments/assets/71c6fb20-0cf0-48cc-8ec0-e55aed17e13f

okadurin (approved):

gerjanvangeest:
> When I click the field for the first time, I don't see any list until I type something. But once I select something, clicking the field opens the whole list. Is this behavior difference intended?

Adjusted the code, so that this only happens when `showAllOnEmpty` is set.

PR #2333, opened 2024-08-17
Update README.md: grammatical fix

aghArdeshir:
I really find it hard to read this part:

> This makes sense as the main use case is to extend those components and if you do you do not want to override existing stylings.

Especially the part that says: "and if you do you do not want to"

Even though, after these commas I inserted, it is still hard to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,7 @@ They provide an unopinionated, white-label layer that can be extended to your ow
 - **Modern Code:** Lion is distributes as pure es modules.
 - **Exposes functions/classes and Web Components:** Ships a functionality in it's most appropriate form.
 
-> Note: Our demos may look a little bland but that is on purpose. They only come with functional stylings.
-> This makes sense as the main use case is to extend those components and if you do you do not want to override existing stylings.
+> Note: Our demo examples might look simple and not very stylish. This is on purpose. They are designed to be basic so you can easily add your own styles to them to match your intended design, without dealing with styles that are already there.
 
 <p align="center">
   <a href="https://lion-web.netlify.app/guides/"><strong>Explore the Lion Guides&nbsp;&nbsp;▶</strong></a>


README.md
```diff
@@ -47,8 +47,8 @@ They provide an unopinionated, white-label layer that can be extended to your ow
 - **Modern Code:** Lion is distributes as pure es modules.
 - **Exposes functions/classes and Web Components:** Ships a functionality in it's most appropriate form.
 
-> Note: Our demos may look a little bland but that is on purpose. They only come with functional stylings.
-> This makes sense as the main use case is to extend those components and if you do you do not want to override existing stylings.
+> Note: Our demos may look a little bland, but that is on purpose. They only come with functional stylings.
+> This makes sense as the main use case is to extend those components, and if you do, you do not want to override existing stylings.
```
gerjanvangeest, line 51: My two cents, please feel free to change.

```suggestion
> The lion ui components are created to be extended, and add your own styling on top of them. But you do not want to override existing stylings, so we kept ours minimal.
```

aghArdeshir: Thanks @gerjanvangeest . I tried to keep my changes as minimal as possible. But I agree to changing it totally.

To be honest, I find your suggestion also hard to understand. I ChatGPTed a bit and here are some suggestions:

> "Our demo examples are intentionally plain and simple, allowing you to easily add your own designs. This approach helps you customize them to your liking without the hassle of changing or removing any existing styles."

> "Our demo examples are plain on purpose, making it easy for you to customize them without removing existing styles."

> "Our demos are simple so you can easily add your own style without having to change what's already there."

> "Our demo examples might look simple and not very stylish. We did this on purpose. They are designed to be basic so you can easily add your own styles to them. This way, you don't have to worry about changing styles that are already there."

Its a bit hard to decide. I'll come back to this later.

gerjanvangeest: I like the last one :)

aghArdeshir: Thanks @gerjanvangeest . I applied the last suggestion, but made some last changes to it:

> Note: Our demo examples might look simple and not very stylish. This is on purpose. They are designed to be basic so you can easily add your own styles to them to match your intended design, without dealing with styles that are already there.

Hope you like it :smiley:

gerjanvangeest (approved):

PR #2334, opened 2024-08-19
opened-changed event detail exposes opened state

rajkeshwar:
## What I Did

1. Replaced the `Event` with `CustomEvent` to include the `opened` state in the event's detail.
2. The `CustomEvent` now also references the host element in its event target.

### Note:

The `opened-changed` event previously lacked detail, and the `event.target` was `null`, making it impossible to determine the dialog's state within the event. This caused issues when attempting to control the dialog's state dynamically via the `opened` property.

A potential scenario is where a user dynamically triggers the dialog through the `opened` property, expecting to open or close the dialog accordingly. However, when the ESC key is pressed and the dialog closes, the state is not synced externally. As a result, the dialog might not reopen the next time the `opened` property is set to `true`.

By including `{ detail: { opened: true/false } }`, it becomes possible to accurately determine the dialog's state and update the `opened` property externally, ensuring it remains in sync with the actual dialog state.

### DIFF
--- a/.changeset/fuzzy-poets-brake.md
+++ b/.changeset/fuzzy-poets-brake.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+opened-changed event detail exposes opened state
--- a/docs/components/dialog/src/external-dialog.js
+++ b/docs/components/dialog/src/external-dialog.js
@@ -1,0 +1,32 @@
+import { LitElement, html } from 'lit';
+
+class DialogTriggerDemo extends LitElement {
+  static get properties() {
+    return {
+      _isOpen: { state: true },
+    };
+  }
+
+  toggleDialog(open) {
+    // eslint-disable-next-line no-return-assign
+    return () => (this._isOpen = open);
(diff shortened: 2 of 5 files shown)


gerjanvangeest:
Hi @rajkeshwar thanks for the PR 💪!

- Could you sign the CLA?
- Can you run `npm run changeset`, to add the change to our changelog?
- There is a linting issue, can you fix that?

gerjanvangeest (approved):

rajkeshwar:
> Hi @rajkeshwar thanks for the PR 💪!
> 
> * Could you sign the CLA?
> * Can you run `npm run changeset`, to add the change to our changelog?
> * There is a linting issue, can you fix that?

I have signed the CLA multiple times. The below text also says the same. 
You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/ing-bank/lion?pullRequest=2334) it.

Here is the screenshot from https://cla-assistant.io/my-cla
cla-assistant-ing-bank

PR #2335, opened 2024-08-20
feat(ajax): add interceptors for parsed response JSON objects

bashmish:
## What I did

1. added `addResponseJsonInterceptor` to be able to intercept parsed response JSON objects
2. added docs and tests for it

### DIFF
--- a/.changeset/flat-fireants-hang.md
+++ b/.changeset/flat-fireants-hang.md
@@ -1,0 +1,5 @@
+---
+'@lion/ajax': minor
+---
+
+add interceptors for parsed response JSON objects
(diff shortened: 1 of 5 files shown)


bashmish:
> LGTM. Perhaps use "forEach" instead of "for..of" when iterating through the interceptors but that's just a niggle.

as discussed in the chat, it was implemented the same way as other interceptor handlers, so should be good for code consistency :)

```js
  /**
   * @param {Response} response
   * @returns {Promise<Response>}
   */
  async __interceptResponse(response) {
    let interceptedResponse = response;
    for (const intercept of this._responseInterceptors) {
      // In this instance we actually do want to await for each sequence
      // eslint-disable-next-line no-await-in-loop
      interceptedResponse = await intercept(/** @type {CacheResponse} */ (interceptedResponse));
    }
    return interceptedResponse;
  }
```

ing-athirlwall:
Fair enough .. like the feature, let's release it.

tlouisse (approved):

tlouisse:
Approved and merged. Sorry for the delay (I thought you were able to approve this as well @ing-athirlwall)

PR #2340, opened 2024-08-21
chore: make readme logo absolute path

ayoayco:
## What I did

1. make the readme logo absolute path

👉 Small change to fix the broken link on [npmjs.org](https://www.npmjs.com/package/@lion/ui)

## Screenshot

[Screenshot 2024-08-21 at 10 34 00](https://www.npmjs.com/package/@lion/ui)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img
     width="20%"
-    src="https://raw.githubusercontent.com/ing-bank/lion/master/docs/_assets/logo.svg"
+    src="./master/docs/_assets/logo.svg"
     alt="Lion"
   />
   <h1 style="text-align: center;">Lion</h1>
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img
     width="20%"
-    src="./docs/_assets/logo.svg"
+    src="https://raw.githubusercontent.com/ing-bank/lion/master/docs/_assets/logo.svg"
     alt="Lion"
   />
   <h1 style="text-align: center;">Lion</h1>
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img
     width="20%"
-    src="./master/docs/_assets/logo.svg"
+    src="./docs/_assets/logo.svg"
     alt="Lion"
   />
   <h1 style="text-align: center;">Lion</h1>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 <p align="center">
   <img
     width="20%"
-    src="./docs/_assets/logo.svg"
+    src="https://raw.githubusercontent.com/ing-bank/lion/master/docs/_assets/logo.svg"
```
ryubro, line 4: I'm not entirely sure about this approach.

This makes the code github dependent. In other word, this code will break when we decided to move to gitlab. Ideally we don't want to create such coupling.

Ideally, we want to embed the image during the packaging stage - although it will increase the size of the package itself, it won't have an impact to the user side.

ryubro: Seems like updating this file would be a better than embedding the image: https://github.com/ing-bank/lion/blob/master/scripts/rewrite-links.js

ryubro: Or this one: https://github.com/ing-bank/lion/blob/master/packages/ui/scripts/create-docs-for-npm-publish.js

ayoayco: Hey @ryubro thanks for the review!

I checked the scripts, which is cool because it puts the relevant docs to the packages we publish. However I think it's a bit overkill to automate changing relative paths for assets especially because this is just a single instance; that is we don't have the pattern anywhere else as far as I see...

I also doubted a bit if we should link to a deployed asset in our site (like what [solidjs](https://raw.githubusercontent.com/solidjs/solid/main/README.md) does, but seeing that we are still on a Netlify I opted to link to here for now -- as I want to suggest a better more permanent domain in the near future. :)

If it helps, [I pushed one old contribution to Qwik](https://github.com/QwikDev/qwik/commit/fb22cabaadb70fa4ede9d3c615466bd824b4c492) where we also have this pattern [in their readme](https://raw.githubusercontent.com/QwikDev/qwik/main/README.md)

ayoayco:
marking draft to fix commits

README.md
ryubro: I agree that it might be an overkill, but I updated the rewrite-links just for fun and pushed to your branch.

ayoayco: Okay I don't mind, this looks useful for other purposes :) Thanks for the help!

gerjanvangeest (approved):

PR #2350, opened 2024-08-27
fix dialog aria-expand attribute

okadurin:
Add `aria-expanded` attribute to the dialog element only when it is a non-modal dialog

### DIFF
--- a/packages/ui/components/dialog/test/lion-dialog.test.js
+++ b/packages/ui/components/dialog/test/lion-dialog.test.js
@@ -1,5 +1,5 @@
 /* eslint-disable lit-a11y/no-autofocus */
-import { expect, fixture as _fixture, html, unsafeStatic } from '@open-wc/testing';
+import { expect, fixture as _fixture, html, unsafeStatic, aTimeout } from '@open-wc/testing';
 import { runOverlayMixinSuite } from '../../overlays/test-suites/OverlayMixin.suite.js';
 import '@lion/ui/define/lion-dialog.js';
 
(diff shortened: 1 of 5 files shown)


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -640,13 +640,14 @@ export class OverlayController extends EventTarget {
...
       this.__storeOriginalAttrs(this.contentNode, ['role', 'id']);
+      const isModal = this.hasBackdrop;
 
       if (this.invokerNode) {
-        this.__storeOriginalAttrs(this.invokerNode, [
-          'aria-expanded',
-          'aria-labelledby',
-          'aria-describedby',
-        ]);
+        const attributesToStore = ['aria-labelledby', 'aria-describedby'];
```
gerjanvangeest, line 646: Can we have tests inside the `OverlayController.test.js` for this?

packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -495,11 +495,11 @@ describe('<lion-input-datepicker>', () => {
       const el = await fixture(html`<lion-input-datepicker></lion-input-datepicker>`);
```
gerjanvangeest, line 495: I am surprised to see that datepicker is a modal dialog, especially when it is presented as local next to the invoker.
On small screens (when it is a bottomsheet), this should be fine, but on desktop size screens it should be a local overlay.

packages/ui/components/dialog/test/lion-dialog.test.js
```diff
@@ -166,4 +166,24 @@ describe('lion-dialog', () => {
       el.setAttribute('opened', '');
     });
   });
+
+  describe('Accessibility', () => {
+    it('adds [aria-expanded] to invoker button', async () => {
```
gerjanvangeest, line 171:
```suggestion
    it('does not add [aria-expanded] to invoker button', async () => {
```

packages/ui/components/overlays/src/OverlayController.js
okadurin: Done

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1528,6 +1528,23 @@ describe('OverlayController', () => {
       expect(ctrl.invokerNode?.getAttribute('aria-expanded')).to.equal('false');
     });
 
+    it('synchronizes [aria-expanded] on invoker when the overlay is modal', async () => {
```
gerjanvangeest, line 1531:
```suggestion
    it('does not synchronize [aria-expanded] on invoker when the overlay is modal', async () => {
```

gerjanvangeest (approved):

Issue #2351, opened 2024-08-27
Test environment (npm run test:browser) behaves differently than real browser


aghArdeshir:
## Expected behavior

I'm trying to write a test to expose the bug explained here: https://github.com/ing-bank/lion/issues/2330
The bug is clearly reproducible when I try it manually.
But when I try to expose this bad behavior as an automated test, it starts acting correctly.

## Actual Behavior

Take this reproduction link: https://studio.webcomponents.dev/edit/CdTHK8iuVNqVSuive1NH/src/index.js?p=README.md

It contains this code snippet:
```
import { LionButton } from "@lion/ui/button.js";

customElements.define("lion-button", LionButton);

["lion-button", "button"].forEach((tagName) => {
  let counter = 1;

  const theButton = document.createElement(tagName);
  theButton.textContent = `This is ${tagName}`;
  document.body.appendChild(theButton);

  theButton.addEventListener("click", () => {
    console.log(`Hello from ${tagName} - ${counter}`);
  });

  theButton.click();
  counter += 1;

  theButton.setAttribute("disabled", "true");

  theButton.click();
  counter += 1;

  theButton.removeAttribute("disabled");

  theButton.click();
  counter += 1;
});
```

After running this code snippet, the normal button logs to the console 2 times, but the lion-button logs to the console 3 times. The normal button is behaving correctly and not handling clicks when its disabled.

Now look at the new test in this PR: https://github.com/ing-bank/lion/pull/2352
- This is exactly the same code: https://github.com/ing-bank/lion/pull/2352/files#diff-7f84ff1242514e9bb95f64318f1da4b563b5af86287f2117585917210d75020a but it produces different result than in the browser
- The issue is coming from this line: https://github.com/ing-bank/lion/pull/2352/files#diff-7bfdd444e5650854e7536869ef3973956e89fbfc8562197c7c13f99cfb3047bfR38 Note that I commented it out, and now the behavior in the test and in the browser are identical for this specific test.

## Additional context

~~I guess the right approach is to remove that extra line (which imports `@webcomponents/scoped-custom-element-registry`), and instead every test that needs it, should import it independently. Importing it for all tests affects tests behavior (which is bad).~~ It gives us false confidence that our components is working fine, but in the real world, not everyone is importing `@webcomponents/scoped-custom-element-registry`)

I'm cooking an MR for this :cook:

aghArdeshir:
This PR should fix this: https://github.com/ing-bank/lion/pull/2353

PR #2353, opened 2024-08-27
Run every test with and without scoped-element-registry polyfill

aghArdeshir:
This PR fixes https://github.com/ing-bank/lion/issues/2351

- Runs every test two times: once with the scoped-element-registry polyfill and once without it.
- Uses `testGroups` to achieve this
- Why? Because that polyfill affects the behavior of components, and gives us false positives. The users of the lion library may not all have loaded that polyfill.
- Creates two suffixes for test files: `some-test.polyfill.test.js` and `some-test.no-polyfill.test.js` for tests that require to be run only with/without the polyfill. (Example is the SlotMixin tests)
- The SlotMixin test had two tests at the end of the file:
- - Moved one of them to another file and suffixed it with `.polyfill`, because that needed polyfill to pass. Also refactored it to use a simple `spy` instead of a mocked object.
- - Removed the other test because it was lying. The title of the test was `does not scope elements when polyfill not loaded` but when you actually remove the polyfill from test config, it fails! (I gave more explanation here: https://github.com/ing-bank/lion/issues/2358)
- - :question: Do I need to extract the refactor and removal of SlotMixin tests as a separate MR?
- :question:  I'd loved to be able to hint when a test fails, did it fail with or without the polyfill, but failed to do so. Any idea?

**UPDATE: Below is the original message of MR, which is not accurate anymore. This PR is re-purposed.**

~~- We should not load a polyfill globally in tests environment, because it affects how tests behave and may give us false positive and false confidence.~~
~~- Every test that needs polyfill, loads the polyfill only for itself.~~
~~- No unnecessary mocking is required. Spy would do the job in this scneario.~~
~~- Adds a util for adding a script in tests on demand.~~
~~- Removes one unnecessary ts-ignore~~

### DIFF
--- a/packages/ui/components/core/test/SlotMixin.polyfill.test.js
+++ b/packages/ui/components/core/test/SlotMixin.polyfill.test.js
@@ -1,0 +1,40 @@
+import sinon from 'sinon';
+import { defineCE, expect, fixture, unsafeStatic, html } from '@open-wc/testing';
+import { ScopedElementsMixin } from '@open-wc/scoped-elements/lit-element.js';
+import { SlotMixin } from '@lion/ui/core.js';
+import { LitElement } from 'lit';
+
+it('supports scoped elements when polyfill loaded', async () => {
+  // @ts-ignore the scoped-custom-element-registry polyfill makes sure `ShadowRoot.prototype.createElement` is defined
+  const createElementSpy = sinon.spy(ShadowRoot.prototype, 'createElement');
+
+  class ScopedEl extends LitElement {}
+
+  const tagName = defineCE(
+    class extends ScopedElementsMixin(SlotMixin(LitElement)) {
+      static get scopedElements() {
+        return {
+          // @ts-expect-error
+          ...super.scopedElements,
+          'scoped-elm': ScopedEl,
+        };
+      }
+
+      get slots() {
+        return {
+          ...super.slots,
+          template: () => html`<scoped-elm></scoped-elm>`,
+        };
+      }
+
+      render() {
+        return html`<slot name="template"></slot>`;
+      }
+    },
+  );
+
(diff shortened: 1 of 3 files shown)


okadurin:
Hi @aghArdeshir ,
Thanks a lot for the contribution! Indeed at some point we added a polyfill for all tests. The right solution would be to run all the tests with and without the polyfill to be sure the code works well in all cases. I think ideally there should be 2 configuration files like `web-test-runner-with-polyfill.config.mjs` and `web-test-runner-without-polyfill.config.mjs` which we run on every `npm run test`. To do that it will require duplicating some tests as some of them were slightly modified when the polyfill support was added. [Here](https://github.com/ing-bank/lion/commit/2fab9e5559a570b7e47f773846cb8cf45e6184f7) is one of those that was modified. 
Do you think you could propose a solution for this request?

aghArdeshir:
Intersting @okadurin . I see now what we are trying to test. Yes I may be able to look into this.

I'll let you know here if I could come up with a solution or if I eventually gave up

aghArdeshir:
@okadurin I've updated my MR and have also updated my original message (https://github.com/ing-bank/lion/pull/2353#issue-2490418176) to list what changes I made. Take another look?

tlouisse:
@aghArdeshir would you like to create a PR for this?

tlouisse:
@aghArdeshir I added a solution for this and all related tickets here: https://github.com/ing-bank/lion/pull/2407/commits
Thanks again for the help! 👍

tlouisse:
https://github.com/ing-bank/lion/pull/2407

aghArdeshir:
Hi. Saw your message late, but I see all is taken care of now :smiley: Thanks

Issue #2330, opened 2024-08-12
[LionButton] @click event listener is called when button has `disabled`  and click is trigger programatically using JS `.click()` method


jhuesos:
## Expected behavior

When a lion-button has the attribute `disabled`, the JavaScript `.click()`  method should not trigger the @click listener(s)

## Actual Behavior

Clicking with mouse does respect the disabled attribute, but when calling with the .click() it does not.

https://stackblitz.com/edit/vite-zh1vmm?file=main.js

## Additional context

Which component and version of it are you using.

aghArdeshir:
I looked into this for a few hours:
- It is hard to reproduce it in the automated tests for some reason. But it is easily reproducible in the context of a normal consumer app
- Simply putting this line in the `LionButton.js` seems to solve this issue, and I don't know why:

![image](https://github.com/user-attachments/assets/95a6129e-c74f-419a-8418-e8a1d049b8f7)

And its hard to vlaidate this fix, because it is hard to put this problem in an automated test

aghArdeshir:
I found about the reason that prevented me from writing a test for this bug: https://github.com/ing-bank/lion/issues/2351

After fixing https://github.com/ing-bank/lion/issues/2351, I should be unblocked for working on this bug with disabled lion-buttons handling clicks

aghArdeshir:
This PR: https://github.com/ing-bank/lion/pull/2354 fixes this issue

denilsonsa:
Hi! When @jhuesos showed me the bug in `lion-button`, I decided to check if the native elements would also behave the same way. So I created a small testcase showing [the behavior of native HTML elements when being clicked while disabled](https://codepen.io/denilsonsa/pen/zYVZRVX). I'm sharing the link here because I think it is relevant, and it can serve as a baseline to build a similar testcase for lion input fields.

aghArdeshir:
Thanks. I've also created this reproduction link in my other issue: https://studio.webcomponents.dev/edit/CdTHK8iuVNqVSuive1NH/src/index.js?p=README.md that shows the difference in behavior for lion button vs native button

PR #2354, opened 2024-08-27
Lion Button should not react to clicks when its disabled 

aghArdeshir:
This PR fixes: https://github.com/ing-bank/lion/issues/2330 and is blocked by: https://github.com/ing-bank/lion/pull/2353

This PR:
- exposes the wrong functionality described here: https://github.com/ing-bank/lion/issues/2330 in automated tests.
- Adds `static formAssociated = true` to the LionButton which fixes the problem at hand. I don't know what is this and why it fixes the issue, but it works in both the automated test and in my manual tests. (and does not break any other test, so I assume it should be good). I first read about it here: https://webkit.org/blog/13711/elementinternals-and-form-associated-custom-elements/, and then I found some references to it in the `@webcomponents/scoped-custom-element-registry` from `node_modules`.

image

If I found any more information about the `formAssociated`, I'll share. If anyone knows what is this and why it fixes the issue, please knowledge-share :pray:

### DIFF
--- a/packages/ui/components/button/src/LionButton.js
+++ b/packages/ui/components/button/src/LionButton.js
@@ -27,6 +27,8 @@ const isSpaceKeyboardClickEvent = (/** @type {KeyboardEvent} */ e) => e.key === 
  * - When implicit form submission should be supported on top, use LionButtonSubmit.
  */
 export class LionButton extends DisabledWithTabIndexMixin(LitElement) {
+  static formAssociated = true;
+
   static get properties() {
     return {
       active: { type: Boolean, reflect: true },
--- a/packages/ui/components/button/test-suites/LionButton.suite.js
+++ b/packages/ui/components/button/test-suites/LionButton.suite.js
@@ -171,6 +171,15 @@ export function LionButtonSuite({ klass = LionButton } = {}) {
         expect(el.getAttribute('tabindex')).to.equal('0');
       });
 
+      it('does not react to clicks when disabled (issue-2330)', async () => {
+        const clickSpy = sinon.spy();
+        const el = /** @type {LionButton} */ (
+          await fixture(html`<${tagButton} disabled @click=${clickSpy}></${tagButton}>`)
+        );
+        el.click();
+        expect(clickSpy).to.not.have.been.called;
+      });
+
       it('has a tabindex="-1" when disabled', async () => {
         const el = /** @type {LionButton} */ (
           await fixture(html`<${tagButton} disabled>foo</${tagButton}>`)
--- a/packages/ui/components/core/test/SlotMixin.test.js
+++ b/packages/ui/components/core/test/SlotMixin.test.js
@@ -8,22 +8,18 @@ import { LitElement } from 'lit';
  * @typedef {import('../types/SlotMixinTypes.js').SlotHost} SlotHost
  */
 
-// @ts-ignore
-const createElementNative = ShadowRoot.prototype.createElement;
-function mockScopedRegistry() {
-  const outputObj = { createElementCallCount: 0 };
-  // @ts-expect-error wait for browser support
-  ShadowRoot.prototype.createElement = (tagName, options) => {
-    outputObj.createElementCallCount += 1;
-    // Return an element that lit can use as render target
-    return createElementNative(tagName, options);
-  };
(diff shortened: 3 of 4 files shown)


denilsonsa:
> If I found any more information about the `formAssociated`, I'll share. If anyone knows what is this and why it fixes the issue, please knowledge-share 🙏

From [this MDN article](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/form) I got to [the official HTML spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-elementinternals-form), and after navigating for a while I got to [form-associated custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-element):

> An [autonomous custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#autonomous-custom-element) is called a form-associated custom element if the element is associated with a [custom element definition](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-definition) whose [form-associated](https://html.spec.whatwg.org/multipage/custom-elements.html#concept-custom-element-definition-form-associated) field is set to true.
>
> The [name](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-name) attribute represents the [form-associated custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-element)'s name. The [disabled](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-disabled) attribute is used to make the [form-associated custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-element) non-interactive and to prevent its [submission value](https://html.spec.whatwg.org/multipage/custom-elements.html#face-submission-value) from being submitted. The [form](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fae-form) attribute is used to explicitly associate the [form-associated custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-element) with its [form owner](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-owner).

Or, in a nutshell, from my understanding of those paragraphs: the `name`, `disabled`, `form` and `readonly` attributes only have their special semantics for custom elements that are form-associated. (And the `readonly` attribute has a few caveats listed one paragraph later.)

---

I'd suggest you to please check all lion input fields and see if they all are form-associated. For your convenience (and your sanity), there is a [list of form-associated elements](https://html.spec.whatwg.org/multipage/forms.html#form-associated-element)

aghArdeshir:
Thanks @denilsonsa really useful links :+1: . So I learned that `static formAssociated = true;` not only helps with clicks, but so many other accessibility features baked into native form elements. And we could also put it on more lion form components. :+1: 

I'm currently blocked by this MR: https://github.com/ing-bank/lion/pull/2353 because without it, I can't get these components under automated tests.

tlouisse:
Thanks for discovering the bug in LionButton and all the investigative work @aghArdeshir and @denilsonsa! 

Wrt think adding `formAssociated`as a solution to this bug: I think it unintentionally disables `.click()`:
https://github.com/webcomponents/polyfills/commit/aeb203863e92edbd48482262db8f87ba26b8f16b#diff-69baa2421e0bf5e67686e1b8184753023486289b9bc745d974495db827b31a69R151

So the best way to fix this is disabling .click in all controls. So maybe in [DisabledMixin](https://github.com/ing-bank/lion/blob/master/packages/ui/components/core/src/DisabledMixin.js) we could do:

```
click() {
  if (this.disabled) return;
  super.click();
}
```

Wrt form association in general for inputs: this, together with ElementInternals, is indeed something that we want to be incorporate in our form controls in the future, especially as they might allow us to get rid of the registration system and they will make our code base leaner/smaller in general.

tlouisse:
@aghArdeshir do you want to create a pull request for this?
- add the click function in DisabledMixin?
- rewrite your existing test and apply [DisabledMixin.test.js] (https://github.com/ing-bank/lion/blob/master/packages/ui/components/core/test/DisabledMixin.test.js)   (add expects and replace LionButton with CanBeDisabled element from test)?

aghArdeshir:
Sorry for the late reply. And thanks fr fixing this :smiley:

PR #2357, opened 2024-08-30
Proof of slot mixin test having wrong assumption about polyfill being loaded

aghArdeshir:
This is as a prof for https://github.com/ing-bank/lion/issues/2358
This PR:
- Adds `only` to only run the target test
- Removes the polyfill from web-test-runner.config
- Filters only the tests in the `core` folder so we see the proof faster

This test fails when polyfill is not loaded, but passes as soon as you add back the polyfill in the web-test-runner.config file. But the test title implies that it should run without the polyfill.


aghArdeshir:
Here is the non-test-environment of the same code with the same error. Create a `test.html` file in the root folder of the liom repo, put these in, and do `npx parcel test.html`

```
<!doctype html>
<html>
  <body>
    <!-- UNCOMMENT SCRIPT BELOW TO SEE THE MAGIC -->
    <!-- <script src="/node_modules/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js"></script> -->
    <script type="module">
      // @ts-nocheck
      /* eslint-disable */

      import { LitElement, html } from 'lit';
      import { ScopedElementsMixin } from '@open-wc/scoped-elements/lit-element.js';
      import { SlotMixin } from './packages/ui/components/core/src/SlotMixin.js';

      let defineCECounter = 0;
      function defineCE(klass) {
        const tag = `test-${defineCECounter}`;
        customElements.define(tag, klass);
        defineCECounter += 1;
        return tag;
      }

      ShadowRoot.prototype.createElement = null;
      class ScopedEl extends LitElement {}

      const tagName = defineCE(
        class extends ScopedElementsMixin(SlotMixin(LitElement)) {
          static get scopedElements() {
            return {
              // @ts-expect-error
              ...super.scopedElements,
              'scoped-el': ScopedEl,
            };
          }

          get slots() {
            return {
              ...super.slots,
              template: () => html`<scoped-el></scoped-el>`,
            };
          }

          render() {
            return html`<slot name="template"></slot>`;
          }
        },
      );

      const renderTarget = document.createElement('div');
      const el = document.createElement(tagName);

      document.body.appendChild(renderTarget);
      renderTarget.appendChild(el);

      document.body.removeChild(renderTarget);
    </script>
  </body>
</html>
```

PR #1978, opened 2023-05-10
fix(calendar): focusable disabled dates

gerjanvangeest:
## What I did

Rebased version of: https://github.com/ing-bank/lion/pull/1381

- Do not automatically force selection of a valid date
- Enables focus to disabled dates to make it more reasonable for screen readers
- Add helper functions to find next/previous/nearest enabled date

TODO:
- [x] (re)check a11y tests
- [x] Add data bounds ([issue #195](https://github.com/ing-bank/lion/issues/195))?
   - For disabled dates before minDate => aria-describe-by... element with "before first available date"
   - For disabled dates after maxDate => aria-describe-by... element with "after last available date"
- [x] check with UX integration with input-datepicker
   - What should happen on selection of disabled date?

Fix: https://github.com/ing-bank/lion/issues/194

PR #2361, opened 2024-09-06
fix the bug that disabled date becomes central date

ryubro:
## What I did

1. Add back the logic that checks the central date is available (removed from this PR: #1978 
2. Call `this.__calculateInitialCentralDate()` and `this.__createData()` in `firstUpdated` instead of `connectedCallback` because props (like `minDate`) are not assigned yet when `connectedCallback` is called.
3. Fix a few exisitng tests (check more details on the line comments)

### DIFF
--- a/packages/ui/components/calendar/test/lion-calendar.test.js
+++ b/packages/ui/components/calendar/test/lion-calendar.test.js
@@ -1278,7 +1278,9 @@ describe('<lion-calendar>', () => {
           expect(calObjWithMax.centralDayObj?.monthday).to.equal(5);
 
           const calWithDisabled = await fixture(
-            html`<lion-calendar .disableDates=${d => d.getDate() === 15}></lion-calendar>`,
+            html`<lion-calendar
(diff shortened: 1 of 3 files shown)


packages/ui/components/calendar/test/lion-calendar.test.js
```diff
@@ -465,9 +465,8 @@ describe('<lion-calendar>', () => {
         `);
 
         clock.restore();
-        expect(isSameDate(el.centralDate, new Date('2019/06/03')), 'central date').to.be.true;
-        expect(isSameDate(elSetting.centralDate, new Date('2019/07/03')), 'central date').to.be
-          .true;
+        expect(isSameDate(el.centralDate, new Date('2019/07/03'))).to.be.true;
```
ryubro, line 468: After the PR, the central date would be nearest enabled date instead of today even when the date is disabled, which is the desired behavior, I believe.

gerjanvangeest (approved):

PR #2362, opened 2024-09-06
Use model value chain instead of hard-coding formatted value and value

ryubro:
## What I did

1. Remove hard-coded formattedValue & value and make the chain-reaction of updating model value handles formatted value & value when user selects date from calendar in date picker.


gerjanvangeest:
I'm missing a changeset (`npm run changeset`), and preferably also some tests.

gerjanvangeest (approved):

PR #2369, opened 2024-09-27
fix(combobox): unexpected overlay opening

maxevilmind:
## What I did

Added condition to check if overlay should be opened, but this change breaks 17 unit tests due to differences between programmatic actions in unit tests and actual UI behavior. Verified that most of the affected scenarios function as expected in the UI.

Unit tests need further work as discussed with @okadurin. He confirmed the unexpected unit tests behavior.

Reproduction scenario of a bug I am trying to fix:
1. Override` _showOverlayCondition` with `return false` to simplify the use case when subclassers override this method and return false due to some logic.
2. Go to the multiple choice combobox and start typing
3. You will notice that the overlay opens on keydown and closes on key release. Input is cleared as a side-effect.

Please see what kind of fix could be applied to unit tests.

PR #2370, opened 2024-09-29
Proposal: Introduce Playwright for e2e testing

okadurin:
`web test runner` is not reliable when it comes to keyboard interaction and Playwright seems more advanced. It provides a high level API for filling out the text, interacts with the browser from the Node side as if a user typed the text. As a result the Keyboard events generated in the right order so we don't need guess how many `updateComplete` cycles to wait. We can consider using Playwright for testing when keyboard interaction is required.
See more details regarding the PR in the added [documentation](https://github.com/ing-bank/lion/blob/chore/playwright-test/e2e/info.md)

Here are the same tests written in `web test runner` and with Playwright for comparison:
* `doesn't select any similar options after using delete when requireOptionMatch is false`: [wtr](https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/test/lion-combobox.test.js#L665), [playwright](https://github.com/ing-bank/lion/blob/chore/playwright-test/packages/ui/components/combobox/test/lion-combobox.e2e.spec.js#L90)
* `allows new options when multi-choice when requireOptionMatch=false and autocomplete="both`: [wtr](https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/test/lion-combobox.test.js#L853), [playwright](https://github.com/ing-bank/lion/blob/chore/playwright-test/packages/ui/components/combobox/test/lion-combobox.e2e.spec.js#L138)
* `hides listbox on click/enter (when multiple-choice is false)`: [wtr](https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/test/lion-combobox.test.js#L132), [playwright](https://github.com/ing-bank/lion/blob/chore/playwright-test/packages/ui/components/combobox/test/lion-combobox.e2e.spec.js#L175)

And here is the test where it detects whether the popup window flashed during the key typing: [Combobox does not flash the menu when _showOverlayCondition returns "false"](https://github.com/ing-bank/lion/blob/chore/playwright-test/packages/ui/components/combobox/test/lion-combobox.e2e.spec.js#L11). I checked it with the [fix PR](https://github.com/ing-bank/lion/pull/2369) and the test passes.


okadurin:
[sendKeys](https://modern-web.dev/docs/test-runner/commands/#custom-commands) is a wrapper around Playwright functions. So no need to bring Playwright separately.

PR #2371, opened 2024-10-01
Fix keyboard typing

okadurin:
Replace `mimicUserTyping` in some tests with `sendKeys ` to unblock the [PR](https://github.com/ing-bank/lion/pull/2369) related to the overlay bug

### DIFF
--- a/.changeset/green-chefs-live.md
+++ b/.changeset/green-chefs-live.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': minor
+---
+
+[combobox] change mimicUserTyping test helper function async and use sendKeys() internally
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -17,6 +17,7 @@ module.exports = {
     },
     {
       files: [
+        '**/test-helpers/**/*.js',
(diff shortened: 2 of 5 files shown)


gerjanvangeest (approved):

tlouisse:
Would it be possible to solve this inside all "mimic methods" that we expose here?
https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/test-helpers/combobox-helpers.js#L40 

(so that we only change these external files and fix it inside our current abstractions)

okadurin:
> Would it be possible to solve this inside all "mimic methods" that we expose here? https://github.com/ing-bank/lion/blob/master/packages/ui/components/combobox/test-helpers/combobox-helpers.js#L40
> 
> (so that we only change these external files and fix it inside our current abstractions)

@tlouisse  Good point! I think we need to make a replacement for all keyboard interaction cases we have in our tests. At least for components that use OverlayMixin. That is because the actual source code fix is in the OverlayMixin and not in the combobox component. Otherwise we cannot be sure that the rest of the tests are not false positive.

tlouisse (approved):

PR #2372, opened 2024-10-01
Exclusive accordion

johncliffrule:
This PR adds a new property called "exclusive" , which when enabled allows only one accordion to be open at any given time.

### DIFF
--- a/.changeset/five-squids-act.md
+++ b/.changeset/five-squids-act.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+[accordion] new "exclusive" feature, allowing one opened collapsible at a time
--- a/docs/components/accordion/index.md
+++ b/docs/components/accordion/index.md
@@ -1,3 +1,3 @@
-# Accordion ||10
+# Accordion
 
 -> go to Overview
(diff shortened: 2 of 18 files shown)


gerjanvangeest (changes requested):
@johncliffrule thanks for the addition to the accordion 💪 

Can you add some tests for `exclusive`?
Can you also add a changeset (`npm run changeset`) to which helps us to add this change to the changelog.

johncliffrule:
> @johncliffrule thanks for the addition to the accordion 💪
> 
> Can you add some tests for `exclusive`? Can you also add a changeset (`npm run changeset`) to which helps us to add this change to the changelog.

Hi, you can now see tests added
And for changeset I see this message
image
 **_WARNING: Releasing a major version for @lion/ui will be its first major release.
🦋  If you are unsure if this is correct, contact the package's maintainers before committing this changeset._**

Is this fine?

johncliffrule:
> The app fails to render the dynamic examples when running via `npm run start` when navigating to `http://localhost:8001/components/accordion/use-cases/`

@tlouisse did you check this?

tlouisse:
> > The app fails to render the dynamic examples when running via `npm run start` when navigating to `http://localhost:8001/components/accordion/use-cases/`
> 
> @tlouisse did you check this?

Yes, I just pushed the fix and a changeset 👍

gerjanvangeest (approved):

PR #2380, opened 2024-10-15
chore: update/cleanup dependencies

tlouisse:
housekeeping (for better perf + security)
- updated root devDependencies
- removed unused deps

providence-analytics
- get rid of postinstall `npx patch-package` command (unneeded and potentially creates errors in on-prem contexts)
- updated to all latest deps


ryubro (approved):
👍

PR #2381, opened 2024-10-17
Merge pull request #2377 from ing-bank/changeset-release/master

tlouisse:
Version Packages

## What I did

1.

### DIFF
--- a/packages-node/providence-analytics/src/program/utils/optimised-glob.js
+++ b/packages-node/providence-analytics/src/program/utils/optimised-glob.js
@@ -109,6 +109,24 @@ export const parseGlobToRegex = memoize(
   },
 );
 
+/**
+ * @template T
+ * @param {T|T[]} value
+ * @returns {T[]}
+ */
+function ensureArray(value) {
+  return Array.isArray(value) ? value : [value];
+}
+
+/**
+ * @template T
+ * @param {T[]} arr
+ * @returns {T[]}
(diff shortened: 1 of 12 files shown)


packages-node/providence-analytics/src/program/utils/optimised-glob.js
```diff
@@ -219,7 +238,12 @@ export async function optimisedGlob(globOrGlobs, providedOptions = {}) {
     options.onlyDirectories = true;
   }
 
-  const globs = Array.isArray(globOrGlobs) ? Array.from(new Set(globOrGlobs)) : [globOrGlobs];
+  const regularGlobs = Array.isArray(globOrGlobs) ? toUniqueArray(globOrGlobs) : [globOrGlobs];
```
ryubro, line 241: Did you mean to use ensureArray here?

ryubro: Also, toUniqueArray here seems not necessary as we do it anyway on L246.

ryubro (approved):
I left some comments about possible optimization, but they are optional, of course.

packages-node/providence-analytics/src/program/utils/optimised-glob.js
tlouisse: Thanks. Fixed 👍

PR #2382, opened 2024-10-17
fix(form-core): add focusableNode to choiceGroupMixin to make _setFo…

gerjanvangeest:
…cusOnFirstErroneousFormElement function work for checkbox-group and radio-group

## What I did

1.

### DIFF
--- a/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -127,13 +127,6 @@ const ChoiceGroupMixinImplementation = superclass =>
       }
     }
 
-    /**
-     * @configure FocusMixin
-     */
-    get _focusableNode() {
-      return this.formElements[0]._focusableNode;
-    }
-
     get operationMode() {
       // @ts-ignore
       return this._repropagationRole === 'choice-group' ? 'select' : 'enter';
--- a/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -127,6 +127,13 @@ const ChoiceGroupMixinImplementation = superclass =>
       }
     }
 
+    /**
+     * @configure FocusMixin
+     */
+    get _focusableNode() {
+      return this.formElements[0]._focusableNode;
+    }
+
     get operationMode() {
       // @ts-ignore
       return this._repropagationRole === 'choice-group' ? 'select' : 'enter';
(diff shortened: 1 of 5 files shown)


packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -127,6 +127,13 @@ const ChoiceGroupMixinImplementation = superclass =>
       }
     }
 
+    /**
+     * @configure FocusMixin
+     */
```
tlouisse, line 132: The FocusMixin is not applied to the ChoiceGroupMixin. 
It's usually applied to "endpoints", that only have one focusable element. The listbox for instance, that uses active-descendant, focuses either the el with role=listbox, or the textbox in case of a combobox. 

The ChoiceGroupMixin can be applied to:
- radiogroup
- checkboxgroup
- select/listbox/combobox (only in these cases the FocusMixin is applied)

tlouisse: But maybe it's worth considering adding it indeed to the ChoiceGroupMixin 👍

gerjanvangeest: Or shall i apply these lines to the radio-group and checkbox-group?

tlouisse: Atm the FocusMixin is only designed to have one focusable element (many of these cases would break if the second elm of a choicegroup received focus):
https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/test/FocusMixin.test.js#L92

So then `_firstFocusablNode` would be a more appropriate name. 
Maybe we can try to solve it here for now? https://github.com/ing-bank/lion/blob/master/packages/ui/components/form/src/LionForm.js#L102

packages/ui/components/form/src/LionForm.js
```diff
@@ -104,10 +95,10 @@ export class LionForm extends LionFieldset {
       element.formElements.find(child => child.hasFeedbackFor.includes('error')) ||
       element.formElements[0];
 
-    if (hasFocusableChildren(firstFormElWithError)) {
-      this._setFocusOnFirstErroneousFormElement(firstFormElWithError);
-    } else {
+    if (firstFormElWithError._focusableNode) {
       firstFormElWithError._focusableNode.focus();
+    } else {
+      this._setFocusOnFirstErroneousFormElement(firstFormElWithError);
```
tlouisse, line 101: Looks like an elegant solution :)

tlouisse (approved):

PR #2396, opened 2024-10-30
localize: make sure lit-ssr can be used with LocalizeManager

tlouisse:


packages/ui/components/localize/src/LocalizeManager.js
```diff
@@ -293,18 +273,84 @@ export class LocalizeManager extends EventTarget {
...
+   */
+  _setupTranslationToolSupport() {
+    this.#localeSetByTranslationTool = documentElement.lang || null;
+  }
+
+  /**
+   * @param {string} locale
+   * @protected
+   */
+  _setHtmlLangAttribute(locale) {
```
gerjanvangeest, line 334: If we switch to protected functions being written with # why not this function too?

gerjanvangeest (approved):

packages/ui/components/localize/src/LocalizeManager.js
tlouisse: `__` -> `#` is for private functions (only meant to be used within the class. `_` (protected) may be used by Subclassers, so they cannot change.

PR #2398, opened 2024-10-31
fix(overlays): no hiding of nested overlays having `hideOnEsc` configured

tlouisse:
closes https://github.com/ing-bank/lion/pull/2198

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -1,16 +1,17 @@
 import { overlays } from './singleton.js';
 import { containFocus } from './utils/contain-focus.js';
+import { deepContains } from './utils/deep-contains.js';
 import { overlayShadowDomStyle } from './overlayShadowDomStyle.js';
 import { _adoptStyleUtils } from './utils/adopt-styles.js';
 
 /**
- * @typedef {import('@lion/ui/types/overlays.js').OverlayConfig} OverlayConfig
+ * @typedef {'setup'|'init'|'teardown'|'before-show'|'show'|'hide'|'add'|'remove'} OverlayPhase
  * @typedef {import('@lion/ui/types/overlays.js').ViewportConfig} ViewportConfig
- * @typedef {import('@popperjs/core').createPopper} Popper
+ * @typedef {import('@lion/ui/types/overlays.js').OverlayConfig} OverlayConfig
  * @typedef {import('@popperjs/core').Options} PopperOptions
  * @typedef {import('@popperjs/core').Placement} Placement
+ * @typedef {import('@popperjs/core').createPopper} Popper
  * @typedef {{ createPopper: Popper }} PopperModule
- * @typedef {'setup'|'init'|'teardown'|'before-show'|'show'|'hide'|'add'|'remove'} OverlayPhase
  */
 
 /**
@@ -102,6 +103,8 @@ async function preloadPopper() {
   // @ts-ignore [external]: import complains about untyped module, but we typecast it ourselves
   return /** @type {* & Promise<PopperModule>} */ (import('@popperjs/core/dist/esm/popper.js'));
 }
+
+const childDialogsClosedInEventLoopWeakmap = new WeakMap();
 
 /**
  * OverlayController is the fundament for every single type of overlay. With the right
@@ -498,9 +501,6 @@ export class OverlayController extends EventTarget {
         '[OverlayController] .isTooltip only takes effect when .handlesAccessibility is enabled',
       );
     }
-    // if (newConfig.popperConfig.modifiers.arrow && !newConfig.contentWrapperNode) {
-    //   throw new Error('You need to provide a .contentWrapperNode when Popper arrow is enabled');
-    // }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -4,13 +4,13 @@ import { overlayShadowDomStyle } from './overlayShadowDomStyle.js';
 import { _adoptStyleUtils } from './utils/adopt-styles.js';
 
 /**
- * @typedef {import('@lion/ui/types/overlays.js').OverlayConfig} OverlayConfig
+ * @typedef {'setup'|'init'|'teardown'|'before-show'|'show'|'hide'|'add'|'remove'} OverlayPhase
```
tlouisse, line 7: no changes here, just a "cosmetic re-ordering"

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1,24 +1,26 @@
 /* eslint-disable no-new */
+import { OverlayController, overlays, OverlayMixin } from '@lion/ui/overlays.js';
```
tlouisse, line 2: no changes here, just a "cosmetic re-ordering"

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -617,6 +619,35 @@ describe('OverlayController', () => {
         await aTimeout(0);
         expect(ctrl.isShown).to.be.true;
       });
+
+      it('does not hide when [escape] while the overlay has a child open', async () => {
```
tlouisse, line 623: this is the test that starts working after the fix

gerjanvangeest (dismissed):

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1148,7 +1145,20 @@ export class OverlayController extends EventTarget {
 
   /** @private */
   __escKeyHandler(/** @type {KeyboardEvent} */ ev) {
-    return ev.key === 'Escape' && this.hide();
+    // @ts-expect-error
+    if (ev.key !== 'Escape' || ev._hasAlreadyHandledEscape) return;
+
+    this.hide();
+
+    const shouldNotCloseParents = this.hidesOnEsc && !this.hidesOnOutsideEsc;
```
gerjanvangeest, line 1153: If the child overlay has `hidesOnOutsideEsc` set to true, does not mean the parent overlay should close.

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1148,7 +1145,20 @@ export class OverlayController extends EventTarget {
...
+
+    const shouldNotCloseParents = this.hidesOnEsc;
+    if (shouldNotCloseParents) {
+      // We could do ev.stopPropagation() here, but we don't want to hide info for
+      // the outside world about user interactions. Instead, we piggyback
+      // on the event so our parent overlay can read it.
+      // @ts-expect-error
+      // eslint-disable-next-line no-param-reassign
+      ev._hasAlreadyHandledEscape = true;
+    }
```
ryubro, line 1161: My general idea about JS event is that an event is just thrown with details no more than its type and the target then more details should be investigated from the target by the event handler.

Do you think it's better if we let this handler investigate the target to decide if it has been already handled?

tlouisse: I can make it cleaner by storing these data in a Weakmap instead of on the event. The thing here is that it passes two nested overlays and the parent one should know about the child one, so just checking the target is not a solution

tlouisse: See https://jsfiddle.net/41jq0Ltg/1/ to see what we're trying to achieve

packages/ui/components/overlays/src/OverlayController.js
tlouisse: Good catch. Fixed it (and added tests for all possible oombinations, including shadow dom)

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -1,28 +1,41 @@
...
+/**
+ * A small wrapper function that closely mimics an escape press from a user
+ * (prevents common mistakes like no bubbling or keydown)
+ * @param {Element|Document} element
+ */
+async function mimicEscapePress(element) {
+  element.dispatchEvent(
+    new KeyboardEvent('keyup', { key: 'Escape', bubbles: true, composed: true }),
+  );
+  await aTimeout(0);
```
okadurin, line 36: Can we use 
```
await sendKeys({
    up: 'Escape',
  });
```
instead?

packages/ui/components/overlays/src/OverlayController.js
ryubro: Thank you for the example! Great to see how the dialog element simplify things.

How do you think about making the OverlayManger know about parent-child relationship of overlays and handle what should happen when the ESC key was pressed?

I think nested popups often comes with different way to display which can benefit from the central manager:
- We might want to have control on the background of the popups (ex: making the half-opaque background to stack to make the background darker or just keep the background opacity)
- We might want to put the child popup so that it looks neatly stacked above the parent, especially on mobiles
- In the previous company, the designer decided to put the child popup contents in the parent popup like the iOS navigation menu (https://developer.apple.com/documentation/uikit/uinavigationcontroller)

Anyway, I'm not insisting, just a suggestion.

ryubro (approved):

packages/ui/components/overlays/src/OverlayController.js
tlouisse: I really like the scenarios you describe. It would be great to create them in a demo (as a lion example) and see what's needed to build them :) 

With all the recent platform developments (especially wrt popover) we have a great opportunity to make things more lightweight (I think we still need the manager and controller, see https://hidde.blog/popover-accessibility, https://adrianroselli.com/2023/05/brief-note-on-popovers-with-dialogs.html, etc.).
Let's refine the future of the system together and have a brainstorm session about how we can make it better with modern technology.

Currently, I have a draft doc that explains the origins (and potential future) of the overlay system. I'm also thinking about ways to make the manager opt-in. Since we're using dialog inside today we can leverage the document order (whereas before we moved contents to the body and needed the OverlayManager to keep track of their ordering). Having this dom order intact, we have a lot of opportunities css-wise and by querying the dom.

packages/ui/components/overlays/test/OverlayController.test.js
tlouisse: I tried, but didn't work for the Escape key. Maybe you can try in a separate PR if you want to?

PR #2402, opened 2024-11-06
chore: adds lion-select docs about the repeat directive

thatdudemanguy:
## What I did

1. Added documentation about updating lion-select options using the repeat directive from Lit, to fix odd user experience when using map or other types of repeating templates.


tlouisse:
Awesome, thanks 🥳

tlouisse (approved):

PR #2407, opened 2024-11-12
Chore/scoped registries improvements

tlouisse:
- [fix(core): no registration of same class twice w/o scoped-registries polyfill](https://github.com/ing-bank/lion/pull/2407/commits/bfcfd4a69fc40c5457c329e8c542228449332508)
- [chore: allow tests to run w/o polyfill](https://github.com/ing-bank/lion/pull/2407/commits/bf2a22d486332ed4c260ff799b9ce4c8c14abf6b)

- [chore: cleanup wtr config, add pkg script for running w/o polyfill](https://github.com/ing-bank/lion/pull/2407/commits/30204f5c6364adc51e807b2795b42bbc21635c8e)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "custom-elements-manifest": "npm run custom-elements-manifest --workspaces --if-present && node ./scripts/create-api-tables.mjs",
     "debug": "web-test-runner --watch --config web-test-runner-chrome.config.mjs",
     "debug:firefox": "web-test-runner --watch --config web-test-runner-firefox.config.mjs",
+    "debug:no-scoped-registries-polyfill": "npm run debug -- --no-scoped-registries-polyfill",
     "debug:webkit": "web-test-runner --watch --config web-test-runner-webkit.config.mjs",
     "format": "npm run format:eslint && npm run format:prettier",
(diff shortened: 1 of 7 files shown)


tlouisse:
closes https://github.com/ing-bank/lion/pull/2357

package.json
```diff
@@ -9,6 +9,7 @@
     "custom-elements-manifest": "npm run custom-elements-manifest --workspaces --if-present && node ./scripts/create-api-tables.mjs",
     "debug": "web-test-runner --watch --config web-test-runner-chrome.config.mjs",
     "debug:firefox": "web-test-runner --watch --config web-test-runner-firefox.config.mjs",
+    "debug:no-scoped-registries-polyfill": "npm run debug -- --no-scoped-registries-polyfill",
```
okadurin, line 12: Should we add both `with polyfill` and `without polyfill` configs to `test:browser` so that the `npm run test` cover both cases? Or is it to be done in separate PR in the future?

okadurin (approved):

package.json
tlouisse: I was thinking about that as well, but for now I kept it limited. Feel free to do a proposal in the future

PR #2412, opened 2024-11-19
feat(overlays): add contentElementToFocus config option to change the…

gerjanvangeest:
… content element that receives focus on show

## What I did

1.


docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -220,6 +220,153 @@ export const hidesOnOutsideClick = () => {
...
+    const frameId = uuid();
+    this.headerNode.setAttribute('id', `frame-header-${frameId}`);
+    this.bodyNode.setAttribute('id', `frame-body-${frameId}`);
+    this.bodyNode.setAttribute('aria-labelledby', `frame-header-${frameId}`);
+    this.bodyNode.setAttribute('aria-describedby', `frame-body-${frameId}`);
+  }
+
+  render() {
+    return html`
+      <h1 class="header">
```
tlouisse, line 275: header !== heading? :)

docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -1,7 +1,7 @@
 # Systems >> Overlays >> Configuration ||40
 
 ```js script
-import { html } from '@mdjs/mdjs-preview';
+import { html as previewHtml } from '@mdjs/mdjs-preview';
```
tlouisse, line 4: why not use one html per page?

docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -220,6 +220,153 @@ export const hidesOnOutsideClick = () => {
...
+
+class MyOverlay extends OverlayMixin(LitElement) {
+  _defineOverlayConfig() {
+    return {
+      ...withModalDialogConfig(),
+      contentElementToFocus: this._overlayContentNode?.bodyNode,
+    };
+  }
+
+  render() {
```
tlouisse, line 294: Can't we render the frame here directly, so this example is really simple? (the external component is tech debt in our extension layer and stems from a time we needed a web component to protect styling (as the contents were moved to the body and now we have the dialog element to point to the top layer)

gerjanvangeest: You mean, skip the frame component, and give the overlay 3 slots: invoker, header and content?

docs/fundamentals/systems/overlays/configuration.md
gerjanvangeest: Because then i can't load html as part of the example.

tlouisse:
Played a bit around with a pure css solution...

Would this (position sticky based) be a good one? https://jsfiddle.net/aLyz10or/ 
Maybe we can add this as a code example on how to achieve this.

Issue #2403, opened 2024-11-11
Reporting bug on "reporting bug" 📟 


takrishna:
## Expected behavior
The link in "Actual Behavior" section for "live code reproduction" is not rendering anything.

e.g., opening the below URL results in a "empty" render
https://studio.webcomponents.dev/edit/OJTxKWU06xEJGaI2AFSG

![image](https://github.com/user-attachments/assets/c7ae7a19-d7f4-4cec-b396-975d95aac801)

## Actual Behavior

Describe what happens instead. Best via a live code reproduction.
To create one click the link below and edit away.
Once the error is reproduced fork/safe and link it.

https://studio.webcomponents.dev/edit/OJTxKWU06xEJGaI2AFSG

## Additional context

Which component and version of it are you using.

rajithaeyee:
@takrishna You're absolutely right! The given sandbox is indeed missing a few important things:
1. dependencies for the lion tooltip in package.json 
2. relavant esmodule import in index.js

working example here-> https://studio.webcomponents.dev/edit/4TFVd8JS6lCLMZL3KghI/src/index.js?p=README.md

tlouisse:
Thanks both, for pointing this out 👍 
The second example still fails for me @rajithaeyee.

We can maybe move to this Stackblitz example: https://stackblitz.com/~/github.com/Knorgias/lion-demo

tlouisse:
Thanks for reporting this @takrishna and @rajithaeyee 
I updated the issue template in this pr: https://github.com/ing-bank/lion/pull/2413

PR #2413, opened 2024-11-20
fix: make sure that voiceover + safari modals are accessible

tlouisse:
## What I did

- warning when `.contentNode` is a shadowRoot host (as they break in Safari when tabindex of -1 is applied)
- extra hardening of tests by applying `isActiveElement` helper everywhere
- update the issue template with working stackblitz playground

closes https://github.com/ing-bank/lion/issues/2403

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -1121,6 +1121,15 @@ export class OverlayController extends EventTarget {
     if (this.manager) {
       this.manager.disableTrapsKeyboardFocusForAll();
     }
+
+    const isContentShadowHost = Boolean(this.contentNode.shadowRoot);
+    if (isContentShadowHost) {
+      // eslint-disable-next-line no-console
+      console.warn(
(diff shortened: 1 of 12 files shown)


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1121,6 +1121,15 @@ export class OverlayController extends EventTarget {
     if (this.manager) {
       this.manager.disableTrapsKeyboardFocusForAll();
     }
+
+    const isContentShadowHost = Boolean(this.contentNode.shadowRoot);
+    if (isContentShadowHost) {
+      // eslint-disable-next-line no-console
+      console.warn(
```
gerjanvangeest, line 1128: So the fix is just a warning?

tlouisse: Yes. Say we would be updating this in ing-web, we would be instructed to change it to a parent node in shadow dom that projects it (an error would be too breaking).

And that was exactly the solution for ing-web

PR #2414, opened 2024-11-20
fix: make web-test-runner using `document.activeElement`debuggable

tlouisse:
This PR may contain  lot of code, but it only does a few things:
- create and apply test-helper `isActiveElement` inside tests (see explanation below)
- move getDeepActiveElement to `core` folder (note the overlays entrypoint is kept backwards-compatible)
- aesthetically order imports based on line length

```js
/**
 * Readable alternative for `expect(el).to.equal(document.activeElement);`.
 * While this is readable by itself, it makes Web Test Runner hang completelt in many occasions.
 * Therefore it's better to write:
 * `expect(isActiveElement(el)).to.be.true;`
 * @param {Element} el
 * @param {{deep?: boolean}} opts
 * @returns {boolean}
 */
export function isActiveElement(el, { deep = false } = {}) {
  const activeEl = deep ? getDeepActiveElement() : document.activeElement;
  return el === activeEl;
}
```


packages/ui/components/core/test-helpers/isActiveElement.js
```diff
@@ -0,0 +1,14 @@
+import { getDeepActiveElement } from '../src/getDeepActiveElement.js';
+/**
+ * Readable alternative for `expect(el).to.equal(document.activeElement);`.
+ * While this is readable by itself, it makes Web Test Runner hang completelt in many occasions.
```
gerjanvangeest, line 4:
```suggestion
 * While this is readable by itself, it makes Web Test Runner hang completely in many occasions.
```

gerjanvangeest (approved):

PR #2421, opened 2024-11-28
do not show overlay when _showOverlayCondition is "false"

okadurin:
1. I added a test called `'does not flash the menu when _showOverlayCondition returns "false"` which fails in the master and works with the proposed PR. The test makes sure the overlay indeed does not flash when it is not expected
2. In some tests I replaced `return options.currentValue.length > 3 && super._showOverlayCondition(options);` with 
```js
  return (
    // @ts-ignore
    this.__prevCboxValueNonSelected.length > 3 && super._showOverlayCondition(options)
  );
```
because the tests like that do not check if the menu shows up after each letter typed and check it only after all text is typed. And at the very beginning `options.currentValue` is the whole option value like `Artichoke` which is a error.  `this.__prevCboxValueNonSelected` contains the actual value typed by that is a private variable which are not supposed to use. So that is a temporary fix
3. For the rest of the tests where I added `// valid` I checked the code in the browser and I can see that the overlay does not show up with on current branch and shows up on `master`.

### DIFF
--- a/.changeset/weak-hairs-jump.md
+++ b/.changeset/weak-hairs-jump.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+show overlay based on overridden "\_showOverlayCondition"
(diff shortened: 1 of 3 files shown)


gerjanvangeest (approved):
Can you add a changeset?

okadurin:
> Can you add a changeset?

Done

PR #2423, opened 2024-12-02
feat(input-stepper): add aria-valuemin, aria-valuemax and an option t…

gerjanvangeest:
…o set aria-valuetext

## What I did

1.

### DIFF
--- a/docs/components/input-stepper/use-cases.md
+++ b/docs/components/input-stepper/use-cases.md
@@ -51,10 +51,10 @@ Use `min` and `max` attribute to specify a range.
 
 ### Value text
 
-Use the `.valueText` property to override the value with a text.
+Use the `.valueTextMapping` property to override the value with a text.
 
 ```js preview-story
-export const valueText = () => {
+export const valueTextMapping = () => {
   const values = {
     1: 'first',
     2: 'second',
@@ -73,7 +73,7 @@ export const valueText = () => {
       min="1"
       max="10"
       name="value"
-      .valueText="${values}"
+      .valueTextMapping="${values}"
     ></lion-input-stepper>
   `;
 };
--- a/packages/ui/components/input-stepper/src/LionInputStepper.js
+++ b/packages/ui/components/input-stepper/src/LionInputStepper.js
@@ -36,7 +36,7 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
         type: Number,
         reflect: true,
       },
-      valueText: {
+      valueTextMapping: {
         type: Object,
         attribute: 'value-text',
       },
@@ -74,7 +74,7 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
      * The aria-valuetext attribute defines the human-readable text alternative of aria-valuenow.
      * @type {{[key: number]: string}}
      */
-    this.valueText = {};
+    this.valueTextMapping = {};
     this.step = 1;
     this.values = {
       max: this.max,
@@ -138,7 +138,7 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
       this.__toggleSpinnerButtonsState();
     }
 
-    if (changedProperties.has('valueText')) {
+    if (changedProperties.has('valueTextMapping')) {
       this._updateAriaAttributes();
     }
 
@@ -240,13 +240,14 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
     if (displayValue) {
       this.setAttribute('aria-valuenow', `${displayValue}`);
       if (
-        Object.keys(this.valueText).length !== 0 &&
-        Object.keys(this.valueText).find(key => Number(key) === this.currentValue)
+        Object.keys(this.valueTextMapping).length !== 0 &&
+        Object.keys(this.valueTextMapping).find(key => Number(key) === this.currentValue)
       ) {
-        this.setAttribute('aria-valuetext', `${this.valueText[this.currentValue]}`);
+        this.setAttribute('aria-valuetext', `${this.valueTextMapping[this.currentValue]}`);
       } else {
         // VoiceOver announces percentages once the valuemin or valuemax are used.
         // This can be fixed by setting valuetext to the same value as valuenow
+        // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow
         this.setAttribute('aria-valuetext', `${displayValue}`);
       }
     } else {
--- a/packages/ui/components/input-stepper/test/lion-input-stepper.test.js
+++ b/packages/ui/components/input-stepper/test/lion-input-stepper.test.js
@@ -292,6 +292,9 @@ describe('<lion-input-stepper>', () => {
     });
 
     it('updates aria-valuetext when stepper is changed', async () => {
+      // VoiceOver announces percentages once the valuemin or valuemax are used.
+      // This can be fixed by setting valuetext to the same value as valuenow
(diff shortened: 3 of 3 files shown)


docs/components/input-stepper/use-cases.md
```diff
@@ -49,6 +49,41 @@ Use `min` and `max` attribute to specify a range.
...
+    inputStepper.valueText = values[inputStepper.modelValue - 1];
+  }
+  const format = { locale: 'nl-NL' };
+  return html`
+    <lion-input-stepper
+      label="Quality of oranges"
+      min="1"
+      max="10"
+      name="value"
+      @model-value-changed="${ev => onModelValueChanged(ev)}"
```
tlouisse, line 81:
```suggestion
      @model-value-changed="${onModelValueChanged}"
```

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -66,6 +71,8 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
     this.formatter = formatNumber;
     this.min = Infinity;
     this.max = Infinity;
+    /** @type {string | undefined} */
```
tlouisse, line 74: Maybe some description, so that this will be picked up in our api tables?
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaValueText

docs/components/input-stepper/use-cases.md
```diff
@@ -49,6 +49,41 @@ Use `min` and `max` attribute to specify a range.
 ></lion-input-stepper>
 ```
 
+### Value text
+
+Use the `.valueText` property to override the value with a text.
+
+```js preview-story
+export const valueText = () => {
+  const values = [
```
tlouisse, line 58: Since an input step will always have a set of discrete steps and labels (value-texts) belonging to it, maybe a better api would be an index based array corresponding with the amount of steps 

Otherwise you would have to do this always (there's no use case for providing just one valueText

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -205,7 +233,19 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
     }
     decrementButton[disableDecrementor ? 'setAttribute' : 'removeAttribute']('disabled', 'true');
     incrementButton[disableIncrementor ? 'setAttribute' : 'removeAttribute']('disabled', 'true');
-    this.setAttribute('aria-valuenow', `${this.currentValue}`);
+
+    const displayValue = this._inputNode.value;
+    if (displayValue) {
+      this.setAttribute('aria-valuenow', `${displayValue}`);
+      if (!this.valueText) {
+        // VoiceOver announces percentages once the valuemin or valuemax are used.
```
tlouisse, line 241: Do we have a source for this?

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -205,7 +233,19 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
...
+        // This can be fixed by setting valuetext to the same value as valuenow
```
tlouisse, line 242: Do we have a source explaining this behaviour + workaround?

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -36,6 +36,11 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
         type: Number,
         reflect: true,
       },
+      valueText: {
+        type: Number,
```
okadurin, line 40: Since the stepper is always about numbers, do we really need to extend the API with such property? Does it make sense to just set `aria-valuenow` and `aria-valuetext` with the same value internally?

gerjanvangeest: ValueText should be a string :)

okadurin: Right :) But main question was why having this property at all? Is having number for `aria-valuetext` not enough? Like `aria-valuetext="2"`

gerjanvangeest: > The aria-valuenow property is required for the meter, scrollbar, separator (if the element is focusable), and slider roles.

> When the element's values are contained within a range but those values are not numeric (such as "small", "medium" and "large"), or they are numeric but there is value in communicating more information than just a number, aria-valuetext is used to surface the text value to assistive technologies. Only use aria-valuetext when aria-valuenow is not sufficiently meaningful for users because using aria-valuetext will prevent assistive technologies from communicating aria-valuenow.
>
> For example, for a battery indicator, it would be useful to know how many minutes are remaining, in addition to the percentage of charge remaining.
>
> ```html
> <span id="battery-label">Battery</span>
> <span role="meter" aria-labelledby="battery-label"
>   aria-valuenow="5"
>   aria-valuetext="5%, 18 minutes remaining.">
> </span>

https://www.w3.org/WAI/ARIA/apg/practices/range-related-properties/

packages/ui/components/input-stepper/src/LionInputStepper.js
gerjanvangeest: > When used with progressbar and scrollbar, assistive technologies announce the value to users as a percent. When aria-valuemin and aria-valuemax are both defined, the percent value is calculated as a position on the range. Otherwise, the actual value is announced as a percent.
>
> When the value to be announced, either the actual value or the value as a percent, may not be clear to users, the [aria-valuetext](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuetext) attribute should be used to provide a user-friendly representation of the value.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow

At least for VoiceOver the same goes up for spinbutton.

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -36,6 +36,11 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
         type: Number,
         reflect: true,
       },
+      valueText: {
+        type: Object,
+        reflect: true,
```
okadurin, line 41: Do we want to `reflect` it?

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -66,6 +70,11 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
     this.formatter = formatNumber;
     this.min = Infinity;
     this.max = Infinity;
+    /**
+     * The aria-valuetext attribute defines the human-readable text alternative of aria-valuenow.
+     * @type {[ Number: String] | []}
+     */
+    this.valueText = [];
```
okadurin, line 77:
```suggestion
    this.valueText = {};
```

okadurin (approved):

docs/components/input-stepper/use-cases.md
```diff
@@ -49,6 +49,36 @@ Use `min` and `max` attribute to specify a range.
 ></lion-input-stepper>
 ```
 
+### Value text
+
+Use the `.valueText` property to override the value with a text.
```
tlouisse, line 54: Nice, can we rename it to `.valueTextMapping` ?

packages/ui/components/input-stepper/src/LionInputStepper.js
tlouisse: Thanks. Maybe add the link in the test as well?

tlouisse (approved):

PR #2426, opened 2024-12-04
fix(input-stepper): move role=spinbutton to input node

gerjanvangeest:
## What I did

1. To make input-stepper work with NVDA we move the `role="spinbutton"` to the inputNode.
2. To make the decrease & increase buttons announce the new value, a hidden value element was created.

**Before:**

https://github.com/user-attachments/assets/051d7efd-4e19-4ba9-abee-bc5199c0dcb5

**After:**

https://github.com/user-attachments/assets/d46c6cc5-7aa9-4c8f-a7b4-9c28b75c6ccc

### DIFF
--- a/packages/ui/components/input-stepper/src/LionInputStepper.js
+++ b/packages/ui/components/input-stepper/src/LionInputStepper.js
@@ -97,8 +97,8 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
 
     this.__increment = this.__increment.bind(this);
     this.__decrement = this.__decrement.bind(this);
-    this.__boundOnEnterButton = this._onEnterButton.bind(this);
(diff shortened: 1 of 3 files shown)


packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -84,6 +97,7 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
 
     this.__increment = this.__increment.bind(this);
     this.__decrement = this.__decrement.bind(this);
+    this.__boundOnEnterButton = this._onEnterButton.bind(this);
```
tlouisse, line 100: Maybe we can bind the same name under this like we do in all components? (and align this behavior throughout the whole comp (and later write a linting rule for this :) ))?

tlouisse (approved):

PR #2430, opened 2024-12-12
Added utility method to resolve locale supplied and fallback

rajkeshwar:
## What I did

1. Implemented utility method to resolve locale files. If supplied it loads the target locale file. If it doesn't find then it strips everything from - from the filename and loads it. 

ex supplied `de-DE` loads `de-DE.js` file
supplied `de-AB` tries to load `de-AB.js` since the file doesn't exists it loads `de.js` file

### DIFF
--- a/packages/ui/components/localize/src/utils/resolveLocaleConfig.js
+++ b/packages/ui/components/localize/src/utils/resolveLocaleConfig.js
@@ -1,0 +1,20 @@
+/**
+ *
+ * @param {string} namespace Namespace for the locale
+ * @param {any} localeDir Directory of the locale
+ * @returns Promise<LocaleConfig>
+ */
+export const resolveLocaleConfig = (namespace, localeDir) => ({
+  [namespace]: (locale = 'en') => {
+    const localePath = (/** @type {string} */ loc) => [localeDir, `${loc}.js`].join('/');
+
+    const fallbackLocale = localePath(locale.replace(/-\w+/, ''));
+
+    return import(localePath(locale)).catch(reason => {
+      console.warn(`
+        Locale ${locale} not found. Reason ${reason}.
+        Loading fallback locale ${fallbackLocale}`);
+      return import(fallbackLocale);
+    });
+  },
+});
--- a/packages/ui/components/localize/test/resolveLocaleConfig.test.js
+++ b/packages/ui/components/localize/test/resolveLocaleConfig.test.js
@@ -1,0 +1,19 @@
+import { expect } from '@open-wc/testing';
+import { resolveLocaleConfig } from '../src/utils/resolveLocaleConfig.js';
+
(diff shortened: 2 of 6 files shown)


packages/ui/components/localize/src/utils/resolveLocaleConfig.js
```diff
@@ -0,0 +1,20 @@
...
+  [namespace]: (locale = 'en') => {
+    const localePath = (/** @type {string} */ loc) => [localeDir, `${loc}.js`].join('/');
+
+    const fallbackLocale = localePath(locale.replace(/-\w+/, ''));
+
+    return import(localePath(locale)).catch(reason => {
+      console.warn(`
+        Locale ${locale} not found. Reason ${reason}.
+        Loading fallback locale ${fallbackLocale}`);
+      return import(fallbackLocale);
```
gerjanvangeest, line 17: Can we add 'en' as another if the fallbackLocale does not exists at all?

packages/ui/components/localize/src/utils/resolveLocaleConfig.js
```diff
@@ -0,0 +1,20 @@
+/**
+ *
+ * @param {string} namespace Namespace for the locale
+ * @param {any} localeDir Directory of the locale
+ * @returns Promise<LocaleConfig>
+ */
+export const resolveLocaleConfig = (namespace, localeDir) => ({
```
gerjanvangeest, line 7: Can we apply the `resolveLocaleConfig()` to all lion-ui components that have translations?

packages/ui/components/localize/src/utils/resolveLocaleConfig.js
rajkeshwar: Good idea to have fallback to en. I will add that.

packages/ui/components/localize/src/utils/resolveLocaleConfig.js
tlouisse: Hi @rajkeshwar. 

Don't want to introduce a show stopper, but... keep in mind the reason for the current verbosity is to support all bundlers.

We came from a situation with oneliners like `return import('@lion/ui/combobox-translations/${myLocale}.js')`.
Rollup could handle this, Webpack couldn't.

I'm afraid that delegating the dynamic import to an external function makes it harder for all bundlers to make correct chunks.

rajkeshwar: Hi @tlouisse,

Thank you so much for the detailed explanation and for outlining the key areas to consider. To further explore this topic, I’ve started a discussion to gather input on the direction: [#2431](https://github.com/ing-bank/lion/discussions/2431).

tlouisse: cc @tmsns

PR #2432, opened 2024-12-18
remove the country telephone code from the modelValue on init

gerjanvangeest:
## The issue

The input field is prefilled with a country code. When I'm using this as an optional field in my form and on submit it's getting a validation error for the untouched input that the number is not valid.
 
Also when disabled the validations, the form is still sending the country code when I'm submitting, even when I haven't touched the field. I think it would be desirable that an optional untouched field doesn't have any value when submitting the form.

### DIFF
--- a/.changeset/plenty-poets-drop.md
+++ b/.changeset/plenty-poets-drop.md
@@ -2,4 +2,4 @@
 '@lion/ui': patch
 ---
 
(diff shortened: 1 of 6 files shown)


okadurin:
Can you add a test please that makes sure on submit the validation does not fail?

okadurin (approved):

PR #2435, opened 2025-01-03
Fix/parse number locale

vskorohod:
## What I did

1. Pass the options object to the getParseMode function.
2. Check the locale option and return 'withLocale' if its not undefined.
3. Update current unit tests, add new to check different local options.

### DIFF
--- a/packages/ui/components/localize/src/number/parseNumber.js
+++ b/packages/ui/components/localize/src/number/parseNumber.js
@@ -16,17 +16,19 @@ import { getDecimalSeparator } from './getDecimalSeparator.js';
  * parseNumber('1,234.56'); // method: heuristic => 1234.56
  *
  * @param {string} value Clean number (only [0-9 ,.]) to be parsed
- * @param {object} options
- * @param {string?} [options.mode] auto|pasted
+ * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} [options] Locale Options
  * @return {string} unparseable|withLocale|heuristic
  */
-function getParseMode(value, { mode = 'auto' } = {}) {
+function getParseMode(value, options = { mode: 'auto' }) {
   const separators = value.match(/[., ]/g);
 
   if (!separators) {
     return 'withLocale';
   }
-  if (mode === 'auto' && separators.length === 1) {
+  if (options.locale) {
+    return 'withLocale';
+  }
+  if (options.mode === 'auto' && separators.length === 1) {
     const decimalLength = value.split(`${separators}`)[1].length;
     if (decimalLength >= 3) {
       return 'withLocale';
--- a/packages/ui/components/localize/src/number/parseNumber.js
+++ b/packages/ui/components/localize/src/number/parseNumber.js
@@ -16,19 +16,19 @@ import { getDecimalSeparator } from './getDecimalSeparator.js';
  * parseNumber('1,234.56'); // method: heuristic => 1234.56
  *
  * @param {string} value Clean number (only [0-9 ,.]) to be parsed
- * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} options Locale Options
+ * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} [options] Locale Options
  * @return {string} unparseable|withLocale|heuristic
  */
-function getParseMode(value, { mode = 'auto', locale }) {
+function getParseMode(value, options = { mode: 'auto' }) {
   const separators = value.match(/[., ]/g);
 
   if (!separators) {
     return 'withLocale';
   }
-  if (locale) {
+  if (options.locale) {
     return 'withLocale';
   }
-  if (mode === 'auto' && separators.length === 1) {
+  if (options.mode === 'auto' && separators.length === 1) {
     const decimalLength = value.split(`${separators}`)[1].length;
     if (decimalLength >= 3) {
       return 'withLocale';
--- a/packages/ui/components/localize/src/number/parseNumber.js
+++ b/packages/ui/components/localize/src/number/parseNumber.js
@@ -16,19 +16,19 @@ import { getDecimalSeparator } from './getDecimalSeparator.js';
  * parseNumber('1,234.56'); // method: heuristic => 1234.56
  *
  * @param {string} value Clean number (only [0-9 ,.]) to be parsed
- * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} [options] Locale Options
+ * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} options Locale Options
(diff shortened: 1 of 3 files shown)


packages/ui/components/localize/src/number/parseNumber.js
```diff
@@ -16,17 +16,19 @@ import { getDecimalSeparator } from './getDecimalSeparator.js';
  * parseNumber('1,234.56'); // method: heuristic => 1234.56
  *
  * @param {string} value Clean number (only [0-9 ,.]) to be parsed
- * @param {object} options
- * @param {string?} [options.mode] auto|pasted
+ * @param {import('../../types/LocalizeMixinTypes.js').FormatNumberOptions} [options] Locale Options
  * @return {string} unparseable|withLocale|heuristic
  */
-function getParseMode(value, { mode = 'auto' } = {}) {
+function getParseMode(value, options = { mode: 'auto' }) {
```
ryubro, line 22: This introduces a break change in a subtle way. For whatever reason, when `{}` is passed as the options, `mode` will be `'auto'` in the previous implementation while it would be `undefined` after this change.

Maybe something like below would be safer, if we go for this route.

```suggestion
function getParseMode(value, { mode = 'auto', locale }) {
```

packages/ui/components/localize/test/number/parseNumber.test.js
```diff
@@ -94,20 +94,30 @@ describe('parseNumber()', () => {
...
-  it('uses locale to parse amount if there is only one separator e.g. 1.234 and amount of decimals is bigger then 2', () => {
+  it('uses english locale to parse amount if there is only one separator e.g. 1.234 and amount of decimals is bigger then 2', () => {
     localizeManager.locale = 'en-GB';
     expect(parseNumber('1.234')).to.equal(1.234);
     expect(parseNumber('1,234')).to.equal(1234);
     expect(parseNumber('1.2345')).to.equal(1.2345);
     expect(parseNumber('1,2345')).to.equal(12345);
+  });
 
+  it('uses dutch locale to parse amount if there is only one separator e.g. 1.234 and amount of decimals is bigger then 2', () => {
```
ryubro, line 105: I don't think the test descriptions make sense to say different x in `uses x locale to parse amount` while the if clause are exactly the same. Can we revert this separation?

packages/ui/components/localize/test/number/parseNumber.test.js
```diff
@@ -183,4 +193,73 @@ describe('parseNumber()', () => {
       }),
     ).to.equal(6000);
   });
+
+  it('with english locale and english number format', () => {
```
ryubro, line 197: These tests are to test that the `parseNumber` function uses the locale parse mode when locale is given, not to test how numbers are parsed in each locale. Probably best to put it in a single test case and putting just GB and NL should be enough as DE behaves the same to NL in these examples. Also the description should make sense with prepended `it` hence should be something like `'should use locale to parse numbers when locale was passed as the option'`.

packages/ui/components/localize/src/number/parseNumber.js
```diff
@@ -16,17 +16,19 @@ import { getDecimalSeparator } from './getDecimalSeparator.js';
...
   if (!separators) {
     return 'withLocale';
   }
-  if (mode === 'auto' && separators.length === 1) {
+  if (options.locale) {
+    return 'withLocale';
+  }
+  if (options.mode === 'auto' && separators.length === 1) {
     const decimalLength = value.split(`${separators}`)[1].length;
     if (decimalLength >= 3) {
```
ryubro, line 33: @gerjanvangeest, what was [the idea behind this logic](https://github.com/ing-bank/lion/pull/1976/files)? Just wondering, if this logic is still relevant after introducing `options.locale`.

ryubro (changes requested):
Thank you for the fix!!

I found a few minor issues. Can you have a look?

Also, please add the change log by running `npm run changeset`.

packages/ui/components/localize/src/number/parseNumber.js
vskorohod: if you check line 33, with the 55.55 value the decimal length is less than 3 and the function returns ''heuristic", not "withLocale". From my point of view, if we provide the locale, it should always return the 'withLocale' mode.

ryubro:
Also I see some pipeline is failing. Can you have a look, @vskorohod ?

vskorohod:
> Also I see some pipeline is failing. Can you have a look, @vskorohod ?

Hi @ryubro
**for lint error:** 
```
components/localize/src/number/parseNumber.js:121:48 - error TS2345: Argument of type 'FormatNumberOptions | undefined' is not assignable to parameter of type 'FormatNumberOptions'.
  Type 'undefined' is not assignable to type 'FormatNumberOptions'.

121   const parseMode = getParseMode(cleanedInput, options);
```

optional object should be optional. I can change it:
From
`function getParseMode(value, { mode = 'auto', locale }) {`
To:
`Line 22: function getParseMode(value, { mode, locale } = { mode: 'auto' }) {`

**for browser test error:**
```
 🚧 Browser logs on Firefox:
      Error: could not find controller to hide
        at hide (packages/ui/components/overlays/src/OverlaysManager.js:109:13)
        at hide (packages/ui/components/overlays/src/OverlayController.js:862:20)
        at #outsideEscKeyHandler (packages/ui/components/overlays/src/OverlayController.js:1195:12)
        at mimicEscapePress (packages/ui/components/overlays/test/OverlayController.test.js:34:11)
        at packages/ui/components/overlays/test/OverlayController.test.js:767:25
```

I have no idea. It doesn't seem to be related to my changes.

packages/ui/components/localize/src/number/parseNumber.js
gerjanvangeest: If you only look at the parser it might make sense to add this attribute. But if i look at the InputAmount, it is a good feature to be a bit more forgiven on the first input.

However when an amount has been formatted once, and then changed the parser could be a bit more strict.

I was thinking about adding another mode version: `preformatted`.
See the draft pr: https://github.com/ing-bank/lion/pull/2439

Please let me know what you think about that solution.

vskorohod: Hi @gerjanvangeest 

If user changes value from 55.555 to 55.55 with the DE locale, there is a separator.
So these lines will not cover the case, right?
```
if (!separators || mode === 'preformatted') {
    return 'withLocale';
  }
```

gerjanvangeest: Its an OR, so that should work as expected.

vskorohod: Sorry. I'm fine with this solution if it helps us to correctly format numbers with the german locale.

PR #2438, opened 2025-01-06
fix(overlays): allow Subclassers to teardown at the right moment

tlouisse:
## What I did

1.

### DIFF
--- a/packages/ui/components/overlays/src/OverlayMixin.js
+++ b/packages/ui/components/overlays/src/OverlayMixin.js
@@ -24,6 +24,8 @@ export const OverlayMixinImplementation = superclass =>
       };
     }
 
+    #hasSetup = false;
+
     constructor() {
       super();
       /**
@@ -173,22 +175,23 @@ export const OverlayMixinImplementation = superclass =>
       }
     }
 
-    /**
-     * @param {import('lit-element').PropertyValues } changedProperties
-     */
-    firstUpdated(changedProperties) {
-      super.firstUpdated(changedProperties);
+    connectedCallback() {
+      super.connectedCallback();
 
-      this._setupOverlayCtrl();
+      this.updateComplete.then(() => {
+        if (this.#hasSetup) return;
+        this._setupOverlayCtrl();
+        this.#hasSetup = true;
+      });
     }
 
     async disconnectedCallback() {
       super.disconnectedCallback();
 
-      if (!this._overlayCtrl) return;
-      if (await this.#isMovingInDom()) return;
-
-      this._teardownOverlayCtrl();
+      if (await this._isPermanentlyDisconnected()) {
+        this._teardownOverlayCtrl();
+        this.#hasSetup = false;
+      }
     }
 
     /**
@@ -238,6 +241,8 @@ export const OverlayMixinImplementation = superclass =>
 
     /** @protected */
     _setupOverlayCtrl() {
+      if (this._overlayCtrl) return;
+
       /** @type {OverlayController} */
       this._overlayCtrl = this._defineOverlay({
         contentNode: this._overlayContentNode,
@@ -253,11 +258,12 @@ export const OverlayMixinImplementation = superclass =>
 
     /** @protected */
     _teardownOverlayCtrl() {
+      if (!this._overlayCtrl) return;
+
       this._teardownOpenCloseListeners();
       this.__teardownSyncFromOverlayController();
 
-      /** @type {OverlayController} */
-      (this._overlayCtrl).teardown();
+      /** @type {OverlayController} */ (this._overlayCtrl).teardown();
     }
 
     /**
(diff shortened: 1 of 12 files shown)


tlouisse:
> Would it be possible to run the `init` callback when `connected` instead of `firstUpdated`? I find such symmetry (setting up when connected and tearing down when disconnected) would be cleaner than conditionally tearing things down.

Hey, good idea. The history of this is a bit like this:

- we used connectedCallback to setup things before. This was especially needed as "pre-dialog" we moved contents to the body

tlouisse:
> Would it be possible to run the `init` callback when `connected` instead of `firstUpdated`? 

What do you mean by the `init` callback? We could indeed reduce some overhead by doing smth like this in a base class or mixin (named the callbacks `setup`/`teardown`, since we already use these ):

Lion base class:
```js
    #hasSetup = false;

    connectedCallback() {
      super.connectedCallback();

        if (!this.#hasSetup) {
           this.setup();
        }
      });
      this.#hasSetup = true;
    }

    async disconnectedCallback() {
      super.disconnectedCallback();

      if (await this._isPermanentlyDisconnected()) {
        this.teardown();
        this.#hasSetup = false;
      }
    }
    
    /**
     * When we're moving around in dom, disconnectedCallback gets called.
     * Before we decide to teardown, let's wait to see if we were not just moving nodes around.
     * @return {Promise<boolean>}
     */
    async _isPermanentlyDisconnected() {
      await this.updateComplete;
      return !this.isConnected;
    }
    
    /**
     * Here we are sure all dom (shadow dom and light dom) has rendered.
     * We can add event listeners etc.
     * Also, we execute logic before the first lit paint (which is vital for (perceived) performance),
     * not unnecessarily delaying it
     */
    setup() {
    }
    
    /**
     * Here we are sure that a component is permanently disconnected 
     * (instead of being moved around in dom, which synchronously triggers disconnected and connectedCallback).
     * When dom nodes are moved around, we don't 
     */
    teardown() {
    }
```

In subclassers, we would just use the setup/teardown callbacks

ryubro:
@tlouisse

With `init` I meant this: https://github.com/ing-bank/lion/blob/fef94cd01607d41ddaa43ca95d2227da6e32e726/packages/ui/components/overlays/src/configurations/visibility-trigger-partials/withClickInteraction.js#L23

This callback is called in a very complex chain: `OverlayController.constructor` -> `this._init()` -> `this._handleFeatures({ phase: 'init' })` -> [`this.__visibilityTriggerHandler[phase]()`](https://github.com/ing-bank/lion/blob/fef94cd01607d41ddaa43ca95d2227da6e32e726/packages/ui/components/overlays/src/OverlayController.js#L1015) (I don't remember where `firstUpdated` came from. Maybe I meant `constructor` and my mind malfunctioned 🤪).

This new interface looks pretty straight-forward!! Do you also plan to simplify the overlay config?

ryubro:
I was wrong about another thing during the communication. The bug that the teardown callback was called while the element was still on the memory didn't happen due to tabs, but due to a routing library. I believe the router detached the element but kept the reference to the element object hence the next time it was attached to the document again, the constructor was not called. In this scenario, `!this.isConnected;` would still yield `true` and the `teardown()` will be called. `#hasSetup` would be still true (because the instance wouldn't get recreated) and will have the same bug.

I still think that it would be much simpler to make sure that what is called during `disconnected` lifecycle callback has its pair in the `connected` lifecycle callback and be called, without any added state / conditions like `hasSetup` or `isPermanatlyDisconnected`. For instance, we can detach and attach again all the event listeners during the DOM is moving around. We might have a slight overhead but wouldn't be significant enough, I believe.

tlouisse:
> I was wrong about another thing during the communication. The bug that the teardown callback was called while the element was still on the memory didn't happen due to tabs, but due to a routing library. I believe the router detached the element but kept the reference to the element object hence the next time it was attached to the document again, the constructor was not called. In this scenario, `!this.isConnected;` would still yield `true` and the `teardown()` will be called. `#hasSetup` would be still true (because the instance wouldn't get recreated) and will have the same bug.
> 
> I still think that it would be much simpler to make sure that what is called during `disconnected` lifecycle callback has its pair in the `connected` lifecycle callback and be called, without any added state / conditions like `hasSetup` or `isPermanatlyDisconnected`. For instance, we can detach and attach again all the event listeners during the DOM is moving around. We might have a slight overhead but wouldn't be significant enough, I believe.

Imho after teardown #hasSetup would be false again after teardown (see code above) and everything would work.

Would you mind to make a failing test case for this?

okadurin (approved):

packages/ui/components/overlays/src/OverlayMixin.js
```diff
@@ -173,22 +175,23 @@ export const OverlayMixinImplementation = superclass =>
...
     async disconnectedCallback() {
       super.disconnectedCallback();
 
-      if (!this._overlayCtrl) return;
-      if (await this.#isMovingInDom()) return;
-
-      this._teardownOverlayCtrl();
+      if (await this._isPermanentlyDisconnected()) {
+        this._teardownOverlayCtrl();
+        this.#hasSetup = false;
```
ryubro, line 193: I didn't notice this line: `this.#hasSetup = false`. It makes sense that the routing scenario works.

ryubro (approved):

PR #2439, opened 2025-01-07
fix(input-amount): parse amount always  on locale once the amount is …

gerjanvangeest:
…formatted

## What I did

1. Make the parser always parse based on locale, if the value is a preformatted value.

## Todo

- [x] Test edge cases
- [x] Add tests
- [x] Add changeset

### DIFF
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -45,7 +45,7 @@ jobs:
         run: npm install --ci
 
       - name: Playwright
-        run: npx playwright install --with-deps
+        run: npx playwright install --with-deps chrome firefox webkit
 
       - name: Test
         run: npm run test:browser
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -45,7 +45,7 @@ jobs:
         run: npm install --ci
 
       - name: Playwright
-        run: npx playwright install --with-deps chrome firefox webkit
+        run: npx playwright install --with-deps
 
       - name: Test
         run: npm run test:browser
--- a/packages/ui/components/form-core/src/FormatMixin.js
+++ b/packages/ui/components/form-core/src/FormatMixin.js
@@ -241,8 +241,6 @@ const FormatMixinImplementation = superclass =>
         return undefined;
       }
 
-      this.#setFormatModeForParseOrFormat('user-edit');
-
       // B) parse the view value
 
       // - if result:
@@ -251,7 +249,7 @@ const FormatMixinImplementation = superclass =>
       // Apparently, the parser was not able to produce a satisfactory output for the desired
       // modelValue type, based on the current viewValue. Unparseable allows to restore all
       // states (for instance from a lost user session), since it saves the current viewValue.
-      const result = this.parser(value, this.formatOptions);
+      const result = this.parser(value, { ...this.formatOptions, mode: this.#getFormatMode() });
       return result !== undefined ? result : new Unparseable(value);
     }
 
@@ -272,7 +270,7 @@ const FormatMixinImplementation = superclass =>
       // input into `._inputNode` with modelValue as input)
 
       if (this._isHandlingUserInput && this.hasFeedbackFor?.includes('error') && this._inputNode) {
-        return this._inputNode ? this.value : undefined;
+        return this.value;
       }
 
       if (this.modelValue instanceof Unparseable) {
@@ -282,9 +280,10 @@ const FormatMixinImplementation = superclass =>
         return this.modelValue.viewValue;
       }
 
-      this.#setFormatModeForParseOrFormat('user-edit');
-
-      return this.formatter(this.modelValue, this.formatOptions);
+      return this.formatter(this.modelValue, {
+        ...this.formatOptions,
+        mode: this.#getFormatMode(),
+      });
     }
 
     /**
@@ -540,7 +539,7 @@ const FormatMixinImplementation = superclass =>
      */
     __onPaste() {
       this._isPasting = true;
-      this.#setFormatModeForParseOrFormat('pasted', () => {
+      queueMicrotask(() => {
         this._isPasting = false;
       });
     }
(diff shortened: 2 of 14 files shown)


ryubro:
Thank you!! I think this could be definitely one of the solutions.

I have one edge cases in mind:
- User is in the locale where the decimal separator is `,` but he is familar with using `.` as the separator (because that's the default behaviour of excel and firefox), editing existing value in large (like removing all the value or leaving only one or two) would be confusing
   -  ex: user enters 99.99 and blurs the field so that  the value was formatted to 99,99, then he removes everything other than 9 and edits to 9.45. It would be 945 instead of 9.45

.github/workflows/verify-pr.yml
```diff
@@ -47,7 +47,7 @@ jobs:
       - uses: microsoft/playwright-github-action@v1
 
       - name: Playwright
-        run: npx playwright install
+        run: npx playwright install --with-deps chrome firefox webkit
```
tlouisse, line 50:
```suggestion
        run: npx playwright install --with-deps
```

packages/ui/components/input-amount/test/lion-input-amount.test.js
```diff
@@ -128,6 +128,21 @@ describe('<lion-input-amount>', () => {
...
 
+  it('parses an updated value based on locale', async () => {
+    const el = /** @type {LionInputAmount} */ (
+      await fixture(
+        html`<lion-input-amount .modelValue=${123456.78} currency="EUR"></lion-input-amount>`,
+      )
+    );
+    const { _inputNode } = getInputMembers(/** @type {* & LionInput} */ (el));
+    expect(_inputNode.value).to.equal('123,456.78');
+    _inputNode.value = '123,45';
```
tlouisse, line 139: Do we expect this when we programatically add a value? Or only when the user types? (probably the latter)

tlouisse: I probably works correctly as is atm (as the formatter triggered by `.modelValue`doesn't use the formatOptions atm), but the "pasted" approach (cleaning up the "preformatted" state) might be safer in general

tlouisse:
> Thank you!! I think this could be definitely one of the solutions.
> 
> I have one edge cases in mind:
> 
> * User is in the locale where the decimal separator is `,` but he is familar with using `.` as the separator (because that's the default behaviour of excel and firefox), editing existing value in large (like removing all the value or leaving only one or two) would be confusing
>   
>   * ex: user enters 99.99 and blurs the field so that  the value was formatted to 99,99, then he removes everything other than 9 and edits to 9.45. It would be 945 instead of 9.45

Good point :) As agreed: let's accept this improvement and list open edge cases in a new discussion

tlouisse (approved):

packages/ui/components/form-core/src/FormatMixin.js
```diff
@@ -285,6 +282,8 @@ const FormatMixinImplementation = superclass =>
         return this.modelValue.viewValue;
       }
 
+      this.#setFormatModeForParseOrFormat('user-edit');
```
ryubro, line 285: This baffles me because there is no direct link between this function name (`_callFormatter`) and 'user-edit' format mode which implies that user is editing the value. 

Will it make sense to set the format mode on more directly relevant callback (like where we focus)?

If not, how do you think about adding a comment why this (calling formatter) changes the format mode?

ryubro: Also I have a more detailed question like: won't pasting call this function?

tlouisse: > This baffles me because there is no direct link between this function name (_callFormatter) and 'user-edit' format > mode which implies that user is editing the value.

I moved the logic to check whether a user is editing an input inside. A more covering name would then be `#setFormatModeForNextParseOrFormatIfEditing`

Feel free to suggest a clearer name or way or to do a refactor

packages/ui/components/form-core/src/FormatMixin.js
```diff
@@ -269,12 +271,7 @@ const FormatMixinImplementation = superclass =>
...
       // input into `._inputNode` with modelValue as input)
 
-      if (
-        this._isHandlingUserInput &&
-        this.hasFeedbackFor?.length &&
-        this.hasFeedbackFor.includes('error') &&
-        this._inputNode
-      ) {
+      if (this._isHandlingUserInput && this.hasFeedbackFor?.includes('error') && this._inputNode) {
         return this._inputNode ? this.value : undefined;
```
ryubro, line 275: When we reach this line, `this._inputNode` is guaranteed to be `truthy`, no? If so shouldn't we do `return this.value`?

packages/ui/components/form-core/src/FormatMixin.js
ryubro: Yeah, I think it's very difficult to name the function, probably because the function does a very complex thing.

It would be great if we can do something like:
```javascript
return this.formatter(this.modelValue, { ...this.formatOptions, mode });
```

Is it only the paste that prevent us to do that?

ryubro: This line: `this.#setFormatModeForParseOrFormat('user-edit');` is confusing because we explicitly supply `user-edit` but `setFormatModeForParseOrFormat` makes its own decision.

I don't have a breaking-through idea, but maybe something like this would be easier to follow?

```javascript
_callFormatter() {
  .
  .
  this.#setFormatModeConditionally();
  .
  .
}

#setFormatModeConditionally() {
  const isUserEditingOrPasting = Boolean(
    this._isPasting || (this._isHandlingUserInput && this.__prevViewValue),
  );
  if (!isUserEditingOrPasting) return;

  if (this.formatOptions.mode === 'pasted') return;

  this.#setFormatModeTemporarily('user-edit');
}

#setFormatModeTemporarily(newMode, callbackFn) {
  this.formatOptions.mode = newMode;

  queueMicrotask(() => {
    this.formatOptions.mode = 'auto';
      callbackFn?.();
    });
  }
}
```

ryubro: After exploring the code a bit more, I was wondering if the following assumption would be correct:

- Each component supply `formatOptions`, but `mode` is set by only the `FormatMixins`.
- The only time `formatOptions` is consumed is when `this.formatter` is called.

If the above is true, can we do something like this?

```javascript
_callFormatter() {
  .
  .
  let mode = 'auto';

  if (this._isPasting) {
    mode = 'pasted';
  }

  if (this._isUserEditing) {
    mode = 'user-edit';
  }

  return this.formatter(this.modelValue, { ...this.formatOptions, mode });
}

get isUserEditing() {
  return this._isHandlingUserInput && this.__prevViewValue;
}

__onPaste() {
  this._isPasting = true;
  queueMicrotask(() => {
    this._isPasting = false;
  });
}
```

tlouisse: Makes sense imho. Will continue tomorrow

packages/ui/components/form-core/src/FormatMixin.js
tlouisse: Will also have a deeper look tomorrow

packages/ui/components/form-core/src/FormatMixin.js
tlouisse: A pull request is also welcome btw @ryubro :)

tlouisse: As I need to do a release, I added these improvements in latest commit. Great suggestions!

ryubro (approved):

packages/ui/components/form-core/src/FormatMixin.js
ryubro: Thank you!! Much easier to follow to me.

PR #2445, opened 2025-01-16
feat(dialog): add is-alert-dialog option

gerjanvangeest:
## What I did

1.

### DIFF
--- a/docs/fundamentals/systems/overlays/configuration.md
+++ b/docs/fundamentals/systems/overlays/configuration.md
@@ -67,6 +67,36 @@ export const placementGlobal = () => {
 };
 ```
 
+## isAlertDialog
+
+In some cases the dialog should act like an [alertdialog](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/), which is a combination of an alert and dialog. If that is the case, you can add `is-alert-dialog` attribute, which sets the correct role on the dialog.
+
+```js preview-story
+export const alertDialog = () => {
+  const placementModeGlobalConfig = { ...withModalDialogConfig(), isAlertDialog: true };
+  return html`
+    <demo-el-using-overlaymixin .config="${placementModeGlobalConfig}">
+      <button slot="invoker">Click me to open the alert dialog!</button>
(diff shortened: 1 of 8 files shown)


docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -67,6 +67,30 @@ export const placementGlobal = () => {
 };
 ```
 
+## isAlertDialog
+
+In some cases the dialog should act like an [alertdialog](https://www.w3.org/WAI/ARIA/apg/patterns/alertdialog/examples/alertdialog/), which is a combination of an alert and dialog. If that is the case, you can add `is-alert-dialog` attribute, which sets the correct role on the dialog.
+
+```js preview-story
+export const alertDialog = () => {
+  const placementModeGlobalConfig = { placementMode: 'global', isAlertDialog: true };
```
tlouisse, line 76: Maybe the best practice here would be to combine it with `withModalDialog()`

docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -67,6 +67,30 @@ export const placementGlobal = () => {
...
+  return html`
+    <demo-el-using-overlaymixin .config="${placementModeGlobalConfig}">
+      <button slot="invoker">Click me to open the alert dialog!</button>
+      <div slot="content" class="demo-overlay">
+        Hello! You can close this notification here:
```
tlouisse, line 81: As application of the role is dependent on content, maybe the content should reflect this a bit more.
```suggestion
        Are you sure you want to perform this action?
```

docs/fundamentals/systems/overlays/configuration.md
```diff
@@ -67,6 +67,30 @@ export const placementGlobal = () => {
...
+        <button
+          class="close-button"
+          @click="${e => e.target.dispatchEvent(new Event('close-overlay', { bubbles: true }))}"
+        >
+          ⨯
+        </button>
+      </div>
```
tlouisse, line 88: Maybe add a  "No" and "Yes" button here?

PR #2457, opened 2025-01-23
fix(calendar): remove duplicate information for screen-reader user

gerjanvangeest:
## What I did

1.  Remove `aria-labelledby="month year"` from table. Each day already announces the month and year in the dayTemplate. So it got announced double.
2. Remove `weekdayName` from dayTemplate - The weekday is also the table header, so if you move from one text element to another while using the screen-reader, you already hear the weekday announced. 
3. Added a "." to dayTemplate message, to make the message more clear.

### DIFF
--- a/packages/ui/components/calendar/src/utils/dataTemplate.js
+++ b/packages/ui/components/calendar/src/utils/dataTemplate.js
@@ -15,10 +15,15 @@ export function dataTemplate(
         month => html`
           <table role="grid" data-wrap-cols aria-readonly="true" class="calendar__grid">
             <thead>
-              <tr>
+              <tr role="row">
                 ${weekdaysShort.map(
                   (header, i) => html`
-                    <th class="calendar__weekday-header" scope="col" aria-label="${weekdays[i]}">
+                    <th
+                      role="columnheader"
+                      class="calendar__weekday-header"
+                      scope="col"
+                      aria-label="${weekdays[i]}"
+                    >
                       ${header}
                     </th>
                   `,
@@ -28,7 +33,7 @@ export function dataTemplate(
             <tbody>
               ${month.weeks.map(
                 week => html`
-                  <tr>
+                  <tr role="row">
                     ${week.days.map(day =>
                       dayTemplate(day, { weekdaysShort, weekdays, monthsLabels }),
                     )}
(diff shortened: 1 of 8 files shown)


packages/ui/components/calendar/src/utils/dayTemplate.js
```diff
@@ -11,10 +11,10 @@ const lastDaysOfYear = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
  * @param {{ weekdays: string[], monthsLabels?: string[] }} opts
  */
 export function dayTemplate(day, { weekdays, monthsLabels = defaultMonthLabels }) {
-  const { dayNumber, monthName, year, weekdayName } = getDayMonthYear(day, weekdays, monthsLabels);
+  const { dayNumber, monthName, year } = getDayMonthYear(day, weekdays, monthsLabels);
 
   function __getFullDate() {
```
okadurin, line 16: Optional
```suggestion
  function __getMonthAndYearDate() {
```

okadurin (approved):

packages/ui/components/calendar/src/utils/dataTemplate.js
```diff
@@ -13,23 +13,12 @@ export function dataTemplate(
...
-          <table
-            role="grid"
-            data-wrap-cols
-            aria-readonly="true"
-            class="calendar__grid"
-            aria-labelledby="month year"
-          >
+          <table role="grid" data-wrap-cols aria-readonly="true" class="calendar__grid">
             <thead>
-              <tr role="row">
```
tlouisse, line 24: Note we should keep the roles. I think this discussion pops up every now and then again :)

The reason is: lion is made witg extending it in mind. That means overriding css. 
If you change the display prop of tables, it loses its semantics

tlouisse (changes requested):
Please keep the roles.
Maybe we should add some comments to prevent this change from being requested in te future again

packages/ui/components/calendar/src/utils/dataTemplate.js
gerjanvangeest: I didn't know about this one, oke will rollback that part and add a note somewhere

tlouisse: Do we also have a rationale about other changes? Like removing aria-labelledby and removing weekday names?

tlouisse: If its duplicate? How exactly? And for which screen readers?

gerjanvangeest: Let me list them in the description of the PR.

PR #2458, opened 2025-01-24
Fix indeterminate checkbox edge cases

ryubro:
## What I did

There were a few problems in the LionCheckboxIndeterminate:
1. When the element was connected, the last child was ignored because the `_setOwnCheckedState()` was called by `form-element-register` event **before** the same event was handled by the checkbox group which adds the element emitting the event to the `formElements`
2. When it had disabled child and all the other children were checked, clicking the indeterminate checkbox didn't respond, because the checked state was `false` although, in practice, it was fully checked
3. When the custom element's `indeterminate` property was `true` and the same property of the input node was also `true`, and clicking it should keep it indeterminated (like when one of the children is disabled and the others are mixed with checked and unchecked), the `indeterminate` property of the custom element doesn't change but the same property of the input node changes to `false` (seems like browser's default behavior) because no syncing could happen without the custome element's `indeteminate` property changing.

This PR fixes those issues.

### DIFF
--- a/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -89,9 +89,11 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
         break;
       default: {
         this.indeterminate = true;
-        const disabledElements = subCheckboxes.filter(checkbox => checkbox.disabled);
+        const disabledUncheckedElements = subCheckboxes.filter(
+          checkbox => checkbox.disabled && checkbox.checked === false,
+        );
         this.checked =
-          subCheckboxes.length - checkedElements.length - disabledElements.length === 0;
+          subCheckboxes.length - checkedElements.length - disabledUncheckedElements.length === 0;
       }
     }
     this.updateComplete.then(() => {
--- a/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
+++ b/packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
@@ -30,9 +30,6 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
         type: Boolean,
         reflect: true,
       },
-      forceSyncIndeterminate: {
-        type: Number,
-      },
       mixedState: {
         type: Boolean,
         reflect: true,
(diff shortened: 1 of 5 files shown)


packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -30,6 +30,9 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
         type: Boolean,
         reflect: true,
       },
+      forceSyncIndeterminate: {
```
okadurin, line 33: Could you update tests and docs for this public API?

gerjanvangeest: Should it be public?

packages/ui/components/checkbox-group/test-suites/CheckboxIndeterminate.suite.js
```diff
@@ -384,6 +442,65 @@ export function runCheckboxIndeterminateSuite(customConfig) {
...
+      `)
+      );
+      const elIndeterminate = /**  @type {LionCheckboxIndeterminate} */ (
+        el.querySelector(`${cfg.tagString}`)
+      );
+      const { _subCheckboxes, _inputNode } = getCheckboxIndeterminateMembers(elIndeterminate);
+
+      // Act
+      _inputNode.click();
+      await elIndeterminate.updateComplete;
```
okadurin, line 495: It might be minor or unrelated to this PR, but when clicking the parent checkbox, it gets `checked` for a short time (right before `await elIndeterminate.updateComplete;`). That is because `packages/ui/components/form-core/src/choice-group/ChoiceInputMixin.js` is listening to the `click` event and set it to `checked`. We might want to add a click listener in this file and run `preventDefault()` when the state stay indeterminate

packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -259,7 +287,7 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
   /** @param {import('lit-element').PropertyValues } changedProperties */
   updated(changedProperties) {
     super.updated(changedProperties);
-    if (changedProperties.has('indeterminate')) {
+    if (changedProperties.has('indeterminate') || changedProperties.has('forceSyncIndeterminate')) {
```
okadurin, line 290: We also could get rid of `forceSyncIndeterminate` completly in the code to improve readability
```suggestion
    if (changedProperties.has('indeterminate') ||  || changedProperties.has('checked')) {
```

gerjanvangeest:
fixes: https://github.com/ing-bank/lion/issues/1967

packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
ryubro: Thank you, updated as you suggested with some comments above (as, checking `checked` to sync indeterminate might not seem obvious).

okadurin:
Could you add a test with a use case that covers `__removeFromSubCheckboxes`?

okadurin:
When there is one of children checkboxes disabled checked and the other child checkbox is unchecked, the parent should be unchecked. Now it is checked

packages/ui/components/checkbox-group/src/LionCheckboxIndeterminate.js
```diff
@@ -97,9 +87,12 @@ export class LionCheckboxIndeterminate extends LionCheckbox {
         this.indeterminate = false;
         this.checked = false;
         break;
-      default:
+      default: {
         this.indeterminate = true;
-        this.checked = false;
+        const disabledElements = subCheckboxes.filter(checkbox => checkbox.disabled);
+        this.checked =
+          subCheckboxes.length - checkedElements.length - disabledElements.length === 0;
```
ryubro, line 94: I was assuming that disabledElements are always unchecked here. Thank you for pointing this out, @okadurin!!

okadurin (approved):

PR #2461, opened 2025-01-29
fix(select-rich): allow arrowLeft and arrowRight to change the value …

gerjanvangeest:
…when navigateWithinInvoker is true and dropdown is closed

## What I did

1.

### DIFF
--- a/packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
+++ b/packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
@@ -188,7 +188,7 @@ describe('lion-select-rich interactions', () => {
   });
 
   describe('Invoker Keyboard navigation Windows/Linux', () => {
-    it('navigates through list with [ArrowDown] [ArrowUp] keys checks the option while listbox unopened', async () => {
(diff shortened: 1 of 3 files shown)


okadurin (approved):

packages/ui/components/select-rich/test/lion-select-rich-interaction.test.js
```diff
@@ -86,26 +104,91 @@ describe('lion-select-rich interactions', () => {
...
+      mimicKeyPress(el, 't', 'KeyT');
+      expect(el.checkedIndex).to.equal(1);
+
+      mimicKeyPress(el, 'u', 'KeyU');
+      expect(el.checkedIndex).to.equal(2);
+    });
+  });
 
+  describe('Invoker Keyboard navigation Windows/Linux', () => {
+    it('navigates through list with [ArrowDown] [ArrowUp] keys checks the option while listbox unopened', async () => {
```
tlouisse, line 191:
```suggestion
    it('navigation through list with [ArrowDown] [ArrowUp] keys checks the option while listbox unopened', async () => {
```

tlouisse (approved):

PR #2472, opened 2025-02-06
feat(providence): allow to run with swc; make all analyzers compatible with swc

tlouisse:
- Allow to provide `parser` to providence config (choose between oxc and swc)
- make all analyzers compatible with both oxc and swc
- make oxcTraverse fully compatible with both oxc and swc


gerjanvangeest (approved):

tlouisse:
closing, already done (and ocx in meantime)

PR #2474, opened 2025-02-10
fix(select-rich): block arrow key interaction when singleOption is set

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/select-rich/test/lion-select-rich.test.js
+++ b/packages/ui/components/select-rich/test/lion-select-rich.test.js
@@ -342,11 +342,17 @@ describe('lion-select-rich', () => {
       expect(elSingleoption.opened).to.be.false;
     });
 
-    it('stays closed with [ArrowUp] or [ArrowDown] key in readonly mode', async () => {
+    it('stays closed with [ArrowUp] or [ArrowDown] key in readonly mode or has a single option', async () => {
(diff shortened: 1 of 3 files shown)


packages/ui/components/select-rich/test/lion-select-rich.test.js
```diff
@@ -342,19 +342,31 @@ describe('lion-select-rich', () => {
...
+          <lion-option .choiceValue=${10}>Item 1</lion-option>
+        </lion-select-rich>
+      `);
+
       elReadOnly.dispatchEvent(new KeyboardEvent('keyup', { key: 'ArrowDown' }));
       expect(elReadOnly.opened).to.be.false;
 
       elReadOnly.dispatchEvent(new KeyboardEvent('keyup', { key: 'ArrowUp' }));
       expect(elReadOnly.opened).to.be.false;
+
```
tlouisse, line 364: Should these events be fired from focusable element (invokerNode)?

And should they bubble? (Or wrap them in a mimic method that uses sendKeys?)

gerjanvangeest: The function is on the `lion-select-rich` directly. It already has an todo next to the function to move it to the invoker.
It wouldn't hurt to wrap them in a mimic method. But let's do that in a future task.

PR #2481, opened 2025-02-24
fix(combobox): do not submit form if overlay of combobox is opened an…

gerjanvangeest:
…d ENTER is hit

## What I did

1.


packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1134,7 +1134,7 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
         }
         break;
       case 'Enter':
-        if (this.multipleChoice && this.opened) {
+        if (this.opened) {
```
okadurin, line 1137: Could you add tests?

okadurin (approved):

PR #2485, opened 2025-02-28
feat: migrated the navigation metadata from inline MD title decorations to frontmatter

paulcodiny:
Currently the order and parenting in the navigation menu is controlled by a special syntax, added to a title of the document:
```
# Button >> Overview ||10
```

Internally this format is converted to frontmatter, which is understood by Eleventy:
```
---
parts: [ 'Overview', 'Button' ],
title: 'Button: Overview',
eleventyNavigation: {
    key: 'Button: Overview,
    title: 'Overview',
    order: 10,
    parent: 'Button'
}
---
```

The benefit of this approach is that is removes a boilerplate text. Note how many times  words"Button" and "Overview" are duplicated above.

The downside is that it happens every time which slightly slows down the building time. Apart from it it decreases compatibility of the Lion MD files with other systems because they need a to support extended syntax of the markdown.


tlouisse (approved):

tlouisse:
Thanks @paulcodiny!
Great step to make it easier to move to Astro, while re-using the content 💪

Issue #2483, opened 2025-02-27
LionInputDate: 3-digit year is accepted as valid


radioflyer042104:
## Expected Behavior

Dates with 3-digit years would be invalidated

Date values like "27/12/202" will be invalidated.  This happens in the default parser (@lion/ui/components/localize/src/date/parseDate.js).

parseDate.js line 72:  
`const parsedDate = new Date(new Date(correctedYear, month - 1, day));`

Passing a dateString, "27/12/202", to the parser will return a new Date (27-12-0202).

This might also be a bug because, by looking at the code of the default parser, it seems that the intent was to only accept dates from 1950 onwards.

## Actual Behavior

Date values like "27/12/202" are accepted and a new Date (27-12-0202) is returned by the default parser.

## Additional context

Using @lion/ui: 0.7.8

PR #2486, opened 2025-03-03
fix(parseDate): Prevents 3-digit year from being accepted as valid (#2483)

imShubh03:
## What I did

1. Fixed an issue where 3-digit years (e.g., `27/12/202`) were incorrectly parsed as valid dates.
2. Ensured that years below `1000` are treated as invalid.
3. Added a check to return `undefined` for such cases.

@tlouisse Please review and verify.


gerjanvangeest:
This PR aimed to fix the same as: https://github.com/ing-bank/lion/pull/2489
That one is already merged, so I will close this PR.

PR #2489, opened 2025-03-09
fix: bug that accepted 3 digit year

SebayK:
### 1. Functionality Added: Year correction logic

Details: The function now includes logic to correct years less than 50 by adding 2000 to the year value.
Example:
Input: parseDate('1/1/49')
Output: new Date('2049/01/01')

### 2. Validation Added: Year format validation
Details: The function now checks if the year format is correct before proceeding with date parsing.
Example:
Input: parseDate('12-12-0000')
Output: undefined

###  3.Test Cases Added: Comprehensive tests for the new logic
Details: Added tests to ensure the year correction and validation logic works as expected.
Test: expect(equalsDate(parseDate('1/1/49'), new Date('2049/01/01'))).to.equal(true);

### DIFF
--- a/packages/ui/components/localize/src/date/parseDate.js
+++ b/packages/ui/components/localize/src/date/parseDate.js
@@ -1,6 +1,7 @@
 import { getDateFormatBasedOnLocale } from './getDateFormatBasedOnLocale.js';
 import { getLocalizeManager } from '../getLocalizeManager.js';
 import { addLeadingZero } from './utils/addLeadingZero.js';
+import { isCorrectYearFormat } from './utils/checkYearFormat.js';
 
 /**
  * @typedef {import('../../types/LocalizeMixinTypes.js').FormatDateOptions} FormatDateOptions
@@ -63,12 +64,16 @@ export function parseDate(dateString, options) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
+  const parsedYear = parsedString.split('/')[0];
+  if (!isCorrectYearFormat(String(parsedYear))) {
+    return undefined;
+  }
   let correctedYear = year;
   if (year < 50) {
     correctedYear = 2000 + year;
   }
 
-  const parsedDate = new Date(new Date(correctedYear, month - 1, day));
+  const parsedDate = new Date(correctedYear, month - 1, day);
   // Check if parsedDate is not `Invalid Date` or that the date has changed (e.g. the not existing 31.02.2020)
   if (
     year > 0 &&
--- a/packages/ui/components/localize/src/date/parseDate.js
+++ b/packages/ui/components/localize/src/date/parseDate.js
@@ -64,8 +64,8 @@ export function parseDate(dateString, options) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
-  const parsedYear = parsedString.split('/')[0];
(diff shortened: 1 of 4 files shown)


packages/ui/components/localize/test/date/parseDate.test.js
```diff
@@ -91,4 +91,17 @@ describe('parseDate()', () => {
     expect(parseDate('02.31.2020')).to.equal(undefined); // non existing date
     expect(parseDate('31.12.2020')).to.equal(undefined); // day & month switched places
   });
+
+  it('returns undefined for incorrect year format', () => {
+    expect(parseDate('12-12-0000')).to.equal(undefined);
+    expect(parseDate('12-12-999')).to.equal(undefined);
+  });
+
+  it('corrects years less than 50 to 2000+year', () => {
```
gerjanvangeest, line 100: How are these tests different from: https://github.com/ing-bank/lion/pull/2489/files#diff-f200d619b2da83b90a6dcd06f60f7e523d437ab862c56f11b7df89361171f2c3R26 ?

packages/ui/components/localize/src/date/parseDate.js
```diff
@@ -63,12 +64,16 @@ export function parseDate(dateString, options) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
+  const parsedYear = parsedString.split('/')[0];
```
gerjanvangeest, line 67: Maybe keep the same way of writing as the line above.

```suggestion
  const [yearString] = parsedString.split('/');
```

packages/ui/components/localize/test/date/parseDate.test.js
SebayK: That's true they are the same, removed

packages/ui/components/localize/src/date/parseDate.js
```diff
@@ -63,12 +64,16 @@ export function parseDate(dateString, options) {
   }
 
   const [year, month, day] = parsedString.split('/').map(Number);
+  const [yearString] = parsedString.split('/');
+  if (!isCorrectYearFormat(yearString)) {
+    return undefined;
+  }
   let correctedYear = year;
   if (year < 50) {
```
tlouisse, line 72: It seems we explicitly allowed this use case here? Will it still be able to reach this code or has it determined incorrect by now?

tlouisse (approved):

packages/ui/components/localize/test/date/parseDate.test.js
```diff
@@ -91,4 +91,9 @@ describe('parseDate()', () => {
     expect(parseDate('02.31.2020')).to.equal(undefined); // non existing date
     expect(parseDate('31.12.2020')).to.equal(undefined); // day & month switched places
   });
+
+  it('returns undefined for incorrect year format', () => {
+    expect(parseDate('12-12-0000')).to.equal(undefined);
```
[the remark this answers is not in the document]
tlouisse: I know this is not related to your PR, but just existent behavior.. So maybe: 
```suggestion
    // TODO: we need a > 0 check for day and month, but maybe consider >= 0 for year?
    expect(parseDate('12-12-0000')).to.equal(undefined);
```

tlouisse:
closes https://github.com/ing-bank/lion/issues/2483

PR #2499, opened 2025-03-24
Revert "feat: migrated the navigation metadata from inline MD title d…

ryubro:
…ecorations to frontmatter "

This reverts commit 11bbc5fff5595da897d5f7f79456cf033c5e9fb2.

## What I did

We reuse the document of Lion in another product, and this change breaks that part.

I can imagine a few approaches:
- preprocess the lion document before reusing it in the other product
- programatically process the rocket-compatible documents into Astro compatible document before feeding it to the Astro engine

None of above is done yet, we should revert the change at the moment.


okadurin (approved):

tlouisse:
See https://github.com/ing-bank/lion/pull/2499 for a fix

PR #2503, opened 2025-03-30
Feature/mdjs sandbox demos

rajkeshwar:
## What I did

1. Developed sandbox component to show full example
2. Configured snippets to dynamically generate Codepen example
3. Configured snippets to dynamically generate Stackblitz example
4. Created separate demos.md file to show the full lit examples

sandbox-example


takrishna:
I generally like the idea.. Providing an embedded solution within the Lion package would better support enterprise use cases, where security and compliance constraints often prevent internal libraries from being exposed to external, hosted editors.

PR #2507, opened 2025-04-02
fix: preselection in combobox

vbabu35:
## What I did

1. fix preselection based on change in modelValue
2. _syncToTextboxCondition in combobox to return true if modelValue differs from oldModelValue

### DIFF
--- a/packages/ui/components/combobox/src/LionCombobox.js
+++ b/packages/ui/components/combobox/src/LionCombobox.js
@@ -1177,7 +1177,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
   // eslint-disable-next-line no-unused-vars
   _syncToTextboxCondition(modelValue, oldModelValue, { phase } = {}) {
     return (
-      this.autocomplete === 'inline' || this.autocomplete === 'both' || phase === 'overlay-close'
+      modelValue !== oldModelValue ||
+      this.autocomplete === 'inline' ||
+      this.autocomplete === 'both' ||
+      phase === 'overlay-close'
     );
   }
 


packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -2440,14 +2440,14 @@ describe('lion-combobox', () => {
...
+      el.autocomplete = 'list';
       el.setCheckedIndex(0);
       expect(_inputNode.value).to.equal('');
 
       el.setCheckedIndex(-1);
-      el.autocomplete = 'list';
+      el.autocomplete = 'none';
       el.setCheckedIndex(0);
-      expect(_inputNode.value).to.equal('');
+      expect(_inputNode.value).to.equal('Artichoke');
```
tlouisse, line 2450: This seems a bit contradictory tbh.

See:
- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-none/
- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

```
The nature of possible values presented by a popup and the way they are presented is called the autocomplete behavior. Comboboxes can have one of four forms of autocomplete:
```

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1175,7 +1175,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
   // eslint-disable-next-line no-unused-vars
   _syncToTextboxCondition(modelValue, oldModelValue, { phase } = {}) {
```
tlouisse, line 1182: I think we mainly need to make sure that, when text is manually entered, the original autocomplete behaviour is preserved.

See:
- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-none/
- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

> The nature of possible values presented by a popup and the way they are presented is called the autocomplete behavior. Comboboxes can have one of four forms of autocomplete:
> 
> **1. No autocomplete**: The combobox is editable, and when the popup is triggered, the suggested values it contains are the same regardless of the characters typed in the combobox. For example, the popup suggests a set of recently entered values, and the suggestions do not change as the user types.

tlouisse: It seems that we only want the syncing to be done when interaction has finished (phase === 'overlay-close') or, when a value is set programatically. 
The latter case seems to be not covered. 
Can you maybe look into that?

I think in the tests, we should keep the current ones the same, but emphasize they focus on manual entry.
So instead of calling `el.setCheckedIndex`, they should use `mimicUserTyping`.

Syncing only needs to happen on:
- `autocomplete=inline|both` or when the user leaves the field (phase === 'overlay-close')
- so explicitly not on `autocomplete=none|list`

And then we need separate tests for programmatic interaction.
- regardless of user interaction, the textbox should be synced with `setCheckedIndex` (or preferably, as this is how a dev would usually interact, via `el.modelValue`)

tlouisse (changes requested):

packages/ui/components/combobox/src/LionCombobox.js
vbabu35: I have made changes accordingly. Please check

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -2294,28 +2313,19 @@ describe('lion-combobox', () => {
 
     describe('Subclassers', () => {
       it('allows to configure autoselect', async () => {
```
tlouisse: this test should not be changed

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -2748,12 +2759,6 @@ describe('lion-combobox', () => {
           async function performChecks(autocompleteMode) {
             await el.updateComplete;
 
-            el.formElements[0].click();
```
tlouisse, line 2757: can this be enabled again?

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -2718,7 +2729,7 @@ describe('lion-combobox', () => {
              */
             // eslint-disable-next-line class-methods-use-this
             _getTextboxValueFromOption(option) {
-              return option.label;
+              return option?.label || this.modelValue?.viewValue;
```
tlouisse: Was this change made on purpose?

packages/ui/components/combobox/test/lion-combobox.test.js
vbabu35: enabled

packages/ui/components/combobox/test/lion-combobox.test.js
vbabu35: Reverted most but autocomplete list cannot be used here since with the fix the and overriding autoSelect, by setting checkedIndex the input value is updated to the modelValue and for list since there is no autocomplete in mimicUserTypingAdvanced, the hasSelection will be false for list thereby making the input value to Viictoria Plum while simulating keypress

PR #2510, opened 2025-04-25
chore: wireit actions v2

tlouisse:
## What I did

1.


ryubro (approved):
Updating the version of wireit was also required. I pushed the updated dependency.

PR #2511, opened 2025-04-25
feat(input-stepper): add alignToStep property to align value with ste…

movrack:
Adding property alingToStep on input-stepper.
When property enabled, clicking on + or - button will verify the next step to reach and align on step multiple if user did enter custom value in between.

**example**
min 0
max 100
step 10
current value 55

click +  ==> 60 instead of 65
click - ==> 50 instead of 55

if min is not 0:

min 3
max 100
step 10
current value 55

click +  ==> 63 instead of 65
click - ==> 53 instead of 55

## What I did

1. Implement the property
2. Add test
3. Write documentation

I made multiple try with the changeset but I don't find how to do it correctly.

### DIFF
--- a/packages/ui/components/input-stepper/src/LionInputStepper.js
+++ b/packages/ui/components/input-stepper/src/LionInputStepper.js
@@ -265,12 +265,17 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
    * @private
    */
   __increment() {
-    const { step, min, max } = this.values;
+    const { step, max } = this.values;
+    let { min } = this.values;
+    if (min === Infinity) {
+      min = 0;
+    }
+
     let newValue = this.currentValue + step;
 
     if ((this.currentValue + min) % step !== 0) {
       // If the value is not aligned to step, align it to the nearest step
-      newValue = Math.floor(this.currentValue / step) * step + step + min;
+      newValue = Math.floor(this.currentValue / step) * step + step + (min % step);
     }
 
     if (newValue <= max || max === Infinity) {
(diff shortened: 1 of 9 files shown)


packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -266,7 +277,13 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
    */
   __increment() {
     const { step, min, max } = this.values;
-    const newValue = this.currentValue + step;
+    let newValue = this.currentValue + step;
+
+    if (this.alignToStep && (this.currentValue + min) % step !== 0) {
```
gerjanvangeest, line 282: This is native behaviour of the `<input type="number">`, so maybe we should not provide an option and just make this default. @ryubro @okadurin ?

ryubro: Yes, it makes better sense to me but we're introducing a break change. Maybe we should offer an option that is opposite to `alignToStep`, for those possible consumers who are using the current behavior as a feature?

gerjanvangeest: @movrack, we discussed it internally and we do not think this that big of a breaking change, we can defend this as a bug fix.
So can you remove the `alignToStep` property and make this the default behavoir?

gerjanvangeest:
@movrack One more request, can you add a changeset `npm run changeset`, so the change ends up in our changelog?

gerjanvangeest (approved):

movrack:
Looks like tests are not working but not related to my changes.
Any idea what it could be ?

movrack:
Fixed and adding tests to try others limits/ edge cases with min below 0 for examples.

packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -265,8 +265,19 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
    * @private
    */
   __increment() {
-    const { step, min, max } = this.values;
-    const newValue = this.currentValue + step;
+    const { step, max } = this.values;
+    let { min } = this.values;
+    if (min === Infinity) {
```
gerjanvangeest, line 270: Can we align both functions?

I prefer the `const stepMin = min !== Infinity ? min : 0;` approach, thats more readable.

PR #2512, opened 2025-04-25
Update the wireit chaching version

ryubro:
From `google/wireit@setup-github-actions-caching/v1` to `google/wireit@setup-github-actions-caching/v2`.


ryubro:
Duplicate of #2510

PR #2520, opened 2025-05-12
chore: add @html-eslint

yeonjuan:
## What I did

Hi! 👋
I'm a maintainer of the [html-eslint](https://github.com/yeonjuan/html-eslint) plugin, which brings HTML linting support to ESLint.

While exploring the ing-bank/lion project, I tried applying html-eslint and found that it works well here. This PR adds HTML linting and includes a few fixes based on the results — such as adding missing type="button" to buttons (to prevent unintended form submissions) and removing unnecessary whitespace.

If this seems useful, I’d be happy to discuss further or help tune the lint rules for your setup. Would you be interested? 😊


packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -1069,7 +1069,7 @@ describe('lion-combobox', () => {
       const tag = unsafeStatic(tagName);
 
       const el = /** @type {LionCombobox} */ (
-        await fixture(html` 
+        await fixture(html`
```
yeonjuan, line 1072: 스크린샷 2025-05-12 오후 11 35 06

PR #2521, opened 2025-05-13
Fix: Allow editing of negative numbers in LionInputAmount

gitraiwaqas:
### What was the problem?

After entering a negative number into the `LionInputAmount` component, the dash character was converted into a Unicode minus sign (`\u2212`) during formatting. This caused the `unmatchedInput` regex in `parsers.js` to fail, making it impossible to edit the number later without triggering an error.

### What did I change?

- Replaced the Unicode minus sign (`\u2212`) with a standard dash (`-`) before parsing input, using a new variable to avoid mutating function parameters (ESLint rule).
- Updated `unmatchedInput` logic in `parsers.js` to handle sanitized input.
- Minor adjustment in `formatters.js` to prevent reintroducing Unicode minus during formatting.

### Testing

- ✅ Manually tested with positive and negative numbers.
- ✅ Switching focus works without triggering errors.
- ✅ Editing negative values now behaves as expected.
- ✅ Ran `npm run lint` and `npm run test` — all passed.

---

Let me know if any changes or improvements are needed. Thanks for reviewing!

### DIFF
--- a/packages-node/providence-analytics/test-node/program/analyzers/e2e/find-classes_-_importing-target-project_0.0.2-target-mock__1828311700.json
+++ b/packages-node/providence-analytics/test-node/program/analyzers/e2e/find-classes_-_importing-target-project_0.0.2-target-mock__1828311700.json
@@ -1,0 +1,220 @@
+{
+  "meta": {
+    "searchType": "ast-analyzer",
+    "analyzerMeta": {
+      "name": "find-classes",
+      "requiredAst": "oxc",
+      "identifier": "importing-target-project_0.0.2-target-mock__1828311700",
+      "targetProject": {
+        "mainEntry": "./target-src/match-imports/root-level-imports.js",
+        "name": "importing-target-project",
+        "version": "0.0.2-target-mock",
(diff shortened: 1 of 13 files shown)


gerjanvangeest:
Hi @gitraiwaqas thank you for the contribution 💪!

Can you add some tests, to make sure it won't break in the future?
And can you also add a changeset `npm run changeset`?

gitraiwaqas:
Thanks for the feedback! 🙌

I've now added:

- ✅ Unit tests for negative number parsing with both Unicode minus and standard dash
- ✅ A changeset with a patch bump for `@lion/ui`

Let me know if you need anything else!

packages-node/providence-analytics/test-node/program/analyzers/e2e/find-classes_-_importing-target-project_0.0.2-target-mock__1828311700.json
```diff
@@ -0,0 +1,220 @@
+{
```
gerjanvangeest, line 1: Can you remove these files from the PR, they are not part of this fix.

PR #2522, opened 2025-05-14
chore(providence): ignore auto generated e2e test files

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages-node/providence-analytics/.gitignore
+++ b/packages-node/providence-analytics/.gitignore
@@ -1,3 +1,4 @@
 /providence-output
 providence-input-data
 /.nyc_output
+/test-node/program/analyzers/e2e


packages-node/providence-analytics/.gitignore
```diff
@@ -1,3 +1,4 @@
 /providence-output
 providence-input-data
 /.nyc_output
+/test-node/program/analyzers/e2e
```
tlouisse, line 4: These actually serve to validate the test and need to be committed from time to time

tlouisse (commented):
Rejecting, as we need to be able to commit them when analyzer output changes

tlouisse (commented):
Rejecting, as we need to be able to commit them when analyzer output changes

PR #2525, opened 2025-05-16
Fix: Replace Unicode minus sign with dash in LionInputAmount

gitraiwaqas:
## What was the problem?

When entering a negative number into the LionInputAmount component, the dash character (`-`) was transformed into a Unicode minus sign (`\u2212`) during formatting. This caused parsing errors in `parsers.js`, making it impossible to edit the number later without triggering an exception.

---

## What did I change?

✅ Replaced the Unicode minus sign with a standard dash before parsing input to ensure correct behavior.  
✅ Avoided mutating function parameters directly (ESLint rule).  
✅ Updated the regex logic in `parsers.js` to match sanitized input.  
✅ Prevented reintroducing the Unicode minus during formatting.

---

## ✅ Tests Added

- Added test cases to cover:
  - Formatting of negative values.
  - Parsing behavior after formatting.
  - Editing behavior of negative inputs.

---

## 📦 Changeset Created

- `@lion/ui` was patch bumped with a summary explaining the fix.

---

## 🧹 Clean PR

- Verified no unrelated files are included.
- This is a **cleaned-up version** of my previous PR (which was closed due to fork cleanup).

Thank you for your time and review 🙌

### DIFF
--- a/packages/ui/components/input-amount/src/parsers.js
+++ b/packages/ui/components/input-amount/src/parsers.js
@@ -31,14 +31,17 @@ function round(value, decimals) {
  * @param {FormatNumberOptions} [givenOptions] Locale Options
  */
 export function parseAmount(value, givenOptions) {
-  const unmatchedInput = value.match(/[^0-9,.\- ]/g);
+  // Replace Unicode minus with normal minus before parsing
+  const sanitizedValue = value.replace(/\u2212/g, '-');
+
+  const unmatchedInput = sanitizedValue.match(/[^0-9,.\- ]/g);
   // for the full paste behavior documentation:
   // ./docs/components/input-amount/use-cases.md#paste-behavior
   if (unmatchedInput && givenOptions?.mode !== 'pasted') {
     return undefined;
   }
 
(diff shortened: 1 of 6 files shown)


packages/ui/components/input-amount/src/parsers.js
```diff
@@ -31,14 +31,17 @@ function round(value, decimals) {
  * @param {FormatNumberOptions} [givenOptions] Locale Options
  */
 export function parseAmount(value, givenOptions) {
-  const unmatchedInput = value.match(/[^0-9,.\- ]/g);
+  // Replace Unicode minus with normal minus before parsing
+  const sanitizedValue = value.replace(/\u2212/g, '-');
+
+  const unmatchedInput = sanitizedValue.match(/[^0-9,.\- ]/g);
```
gerjanvangeest, line 37: Inside the `parseNumber()` function we process the minus signs and set the Unicode value. So all we have to do is whitelist it here and it works. Also no need to change the formatter.

```suggestion
  const unmatchedInput = sanitizedValue.match(/[^0-9,.\-\u2212 ]/g);
```

PR #2526, opened 2025-05-20
fix(input-tel): use full locale to translate country names in dropdown

Hzunax:
## What I did

1. Pass full locale to `new Intl.DisplayNames` constructor.

## Issue

The constructor accepts either a language (e.g. `en`, `nl`, ...) or the full locale (e.g. `en-GB`, `nl-BE`, ...).
Instead, we're passing only the region (e.g. `GB`, `BE`, ...), which is not a valid param for the constructor.
Sometimes that works because language and region are the same (e.g. `fr-FR`, `de-DE`), but that is not always the case.

- [Intl locales argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locales_argument)
- [Intl.Locale](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale)

If the specified locale is not recognized, the translations default to English.

## Reproduce

1. Go to the demo of [input-tel-dropdown extension](https://lion.js.org/components/input-tel-dropdown/extensions/)
2. Open the dropdown and check the translations.
3. Open the browser dev tools to change the `lang` of the html tag to `nl-BE`.
4. Open the dropdown again.
5. The countries are shown in English.
6. Change the `lang` attribute to `fr-FR`
7. Open the dropdown again
8. The countries are shown in French

image

## Fix

By using the full locale in the constructor for `Intl.DisplayNames`, the countries are translated correctly to the browser language:

image

### DIFF
--- a/packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown.suite.js
+++ b/packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown.suite.js
@@ -253,7 +253,7 @@ export function runInputTelDropdownSuite({ klass } = { klass: LionInputTelDropdo
         // @ts-ignore
         const originalLocale = el._localizeManager.locale;
         // @ts-ignore
-        el._localizeManager.locale = 'nl-NL';
+        el._localizeManager.locale = 'fr-BE';
         await el.updateComplete;
         const dropdownNode = el.refs.dropdown.value;
         const templateDataForDropdown = /** @type {TemplateDataForDropdownInputTel} */ (
@@ -267,7 +267,7 @@ export function runInputTelDropdownSuite({ klass } = { klass: LionInputTelDropdo
                 {
                   countryCode: 31,
                   flagSymbol: '🇳🇱',
-                  nameForLocale: 'Nederland',
+                  nameForLocale: 'Pays-Bas',
                   nameForRegion: 'Nederland',
                   regionCode: 'NL',
                 },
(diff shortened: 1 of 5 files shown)


Hzunax:
I have now also added a fix for the `nameForRegion`, which was displayed incorrectly.
I created a `regionCodeToLocale` function, which estimates the full locale based on a region code, in order to pass it to the `new Intl.DisplayNames` constructor.

packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown.suite.js
```diff
@@ -239,6 +239,74 @@ export function runInputTelDropdownSuite({ klass } = { klass: LionInputTelDropdo
...
+          .preferredRegions="${['PH']}"
+          ></${tag}> `);
+        const spy = sinon.spy(
+          /** @type {typeof LionInputTelDropdown} */ (el.constructor).templates,
+          'dropdown',
+        );
+        // @ts-ignore
+        const originalLocale = el._localizeManager.locale;
+        // @ts-ignore
+        el._localizeManager.locale = 'nl-NL';
```
ryubro: Can we test with `fr-BE`?

Hzunax: Updated test case to set locale to `fr-BE`.

PR #2530, opened 2025-06-13
fix(overlays): don't call _setupOverlayCtrl() in OverlayMixin if the component is not connected

denilsonsa:
In our application, there is a very complicated logic that creates a bunch of form fields based on some data. It's very complicated, we hate that logic; but it exists, it works, and it is being used in production.

Somewhere in that logic, under a very specific case, a datepicker form field is created, connected, and then immediately disconnected. Don't ask me why, I said it is complicated.

This commit fixes a bug in OverlayMixin. The `connectedCallback` calls `_setupOverlayCtrl()` asynchronously, after a `Promise.then()`. Thus, by the time this setup function is called, the component has already been disconnected (and `disconnectedCallback()` has already been called).

The fix is simple: don't try to setup the overlay if the component is no longer connected.

### DIFF
--- a/packages/ui/components/form-core/src/FocusMixin.js
+++ b/packages/ui/components/form-core/src/FocusMixin.js
@@ -198,6 +198,7 @@ const FocusMixinImplementation = superclass =>
      * @private
      */
     __teardownEventsForFocusMixin() {
+      if (!this._focusableNode) return;
       this._focusableNode?.removeEventListener(
         'focus',
         /** @type {EventListenerOrEventListenerObject} */ (this.__redispatchFocus),
--- a/packages/ui/components/overlays/src/OverlayMixin.js
+++ b/packages/ui/components/overlays/src/OverlayMixin.js
@@ -179,6 +179,7 @@ export const OverlayMixinImplementation = superclass =>
       super.connectedCallback();
 
       this.updateComplete.then(() => {
+        if (!this.isConnected) return;
         if (this.#hasSetup) return;
         this._setupOverlayCtrl();
         this.#hasSetup = true;


packages/ui/components/form-core/src/FocusMixin.js
```diff
@@ -198,6 +198,7 @@ const FocusMixinImplementation = superclass =>
      * @private
      */
     __teardownEventsForFocusMixin() {
+      if (!this._focusableNode) return;
```
ryubro, line 201: Trying to understand it better. What does this code change? What happens without this line?

packages/ui/components/overlays/src/OverlayMixin.js
```diff
@@ -179,6 +179,7 @@ export const OverlayMixinImplementation = superclass =>
       super.connectedCallback();
 
       this.updateComplete.then(() => {
+        if (!this.isConnected) return;
```
ryubro, line 182: By reading the code, I see that `_teardownOverlayCtrl()` will be called after `_setupOverlayCtrl()` is called even when the component is disconnected immediately, which sounds logical to me.

Can you describe the issue of calling `_setupOverlayCtrl()` when it's already disconnected in more details?

denilsonsa: Happy flow:

```mermaid
sequenceDiagram
    connectedCallback->>+func: super.connectedCallback()
    func->>-connectedCallback: super.connectedCallback()
    connectedCallback-->>+func: this.updateComplete.then()
    func-->>-connectedCallback: this.updateComplete.then()
    connectedCallback->>+func: this._setupOverlayCtrl()
    func->>-connectedCallback: this._setupOverlayCtrl()

    disconnectedCallback->>+func: super.disconnectedCallback()
    func->>-disconnectedCallback: super.disconnectedCallback()
    disconnectedCallback-->>+func: await this._isPermanentlyDisconnected()
    func-->>-disconnectedCallback: await this._isPermanentlyDisconnected()
    disconnectedCallback->>+func: this._teardownOverlayCtrl()
    func->>-disconnectedCallback: this._teardownOverlayCtrl()
```

Unhappy flow:

```mermaid
sequenceDiagram
    connectedCallback->>+func: super.connectedCallback()
    func->>-connectedCallback: super.connectedCallback()
    connectedCallback-->>+func: this.updateComplete.then()

    disconnectedCallback->>+func: super.disconnectedCallback()
    func->>-disconnectedCallback: super.disconnectedCallback()
    disconnectedCallback-->>+func: await this._isPermanentlyDisconnected()
    func-->>-disconnectedCallback: await this._isPermanentlyDisconnected()
    disconnectedCallback->>+func: this._teardownOverlayCtrl()
    func->>-disconnectedCallback: this._teardownOverlayCtrl()

    func-->>-connectedCallback: this.updateComplete.then()
    connectedCallback->>+func: this._setupOverlayCtrl()
    func->>-connectedCallback: this._setupOverlayCtrl() crashes
```

That's because `this.updateComplete.then()` is a Promise receiving a callback (an arrow function). The `connectedCallback()` function my return and finish before that arrow function is called. Thus, by the time that callback arrow function is called, the component isn't in the `connectedCallback()` anymore, and things may have changed. In special, it may happen that the component isn't even connected anymore.

(Note: these sequence diagrams are not strictly correct, specially because the `.then()` function will return immediately. But these diagrams should still be good enough to understand the control flow.)

How to reproduce this bug? I wish I knew. I took me hours of debugging, in the middle of our complicated application code with many thousands of lines. I recognize that providing a minimal testcase is the best practice when reporting a bug, but I'm sorry. You're welcome to contact me privately and I can try reproducing this bug again on our large corporate application, and even do a video call with screen sharing.

packages/ui/components/form-core/src/FocusMixin.js
denilsonsa: As described in [the second commit of this PR](https://github.com/ing-bank/lion/pull/2530/commits/80f6b1c9b3c6cbfbc544677e5bb98021794eaf41), this line solves the following error:

> Uncaught TypeError: this._focusableNode is null

There is a file input class that overloads the `get _focusableNode()` property.

There is a very specific case in the complicated application where this file input component gets created, and immediately destroyed.

This means it gets created, it gets connected (`connectedCallback` was certainly called), and then it gets disconnected (`disconnectedCallback` is being called), and it happens before the `firstUpdated` ever happens. In this case, we are removing event listeners from a descendant element (i.e. child element or grand-child element) that hasn't been fully constructed. Thus, `this._focusableNode` returns null (as expected), but the teardown routine tries to remove event listeners from it (and it breaks).

ryubro: Yes, I know, but all the following code has optional chaining (like this `this._focusableNode?.removeEventListener` instead of this `this._focusableNode.removeEventListener`), so the `_focusableNode` being `null` wouldn't yield any errors, no?

packages/ui/components/overlays/src/OverlayMixin.js
ryubro: Wow, thanks for the diagram. Very clear and making the communication way easier. 

I noticed that I was mistaken during thinking. I somehow thought that in the unhappy flow executing `this.#hasSetup = false;` in the disconnected callback is to guard `this._setupOverlayCtrl()` is called. After reading it more thoroughly, now I see that it's to guard `this._setupOverlayCtrl()` is called twice.

Your code looks good. Can I ask you to add a comment of the intention of the code? Something like:
```js
// This prevents the code crash in the rare case that the component is discunnected after connected was called and before this.updateComplete is resolved
```

Also, have you tried to write a test to reproduce your case?

packages/ui/components/form-core/src/FocusMixin.js
denilsonsa: Well, indeed. I'm not sure anymore why I added that line. I've added it at 18:30, so I was already tired of deep debugging for hours. And it was one month ago, so I already forgot.

I may have got confused because I have both `node_modules/@lion/form-core/src/FocusMixin.js` (which is old) and `node_modules/@lion/ui/components/form-core/src/FocusMixin.js` (which is new and fixed). I'll try to reproduce and debug it. Maybe I can't because this one is already fixed.

packages/ui/components/overlays/src/OverlayMixin.js
ryubro: Or a simple isolated reproduction would be very nice.

tlouisse: Thanks for the spot-on analysis @denilsonsa!

This failing test proves the issue: https://github.com/ing-bank/lion/pull/2546

tlouisse (approved):

Issue #2529, opened 2025-06-12
[LionSelectRich] disabled options should not get focus


gerjanvangeest:
## Expected behavior

I would expect a disabled option to be skipped when navigating with the keyboard, as happens in a native select.

## Actual Behavior

[When an option is disabled](https://lion.js.org/components/select-rich/use-cases/#disabled-select) it can still receive keyboard focus, and can be clicked upon via [Enter].

gerjanvangeest:
The error occurs if a single option is disabled, not the whole select.

<img width="324" alt="Image" src="https://github.com/user-attachments/assets/d6f91c65-a686-4c13-803c-a1cb72a8d5f4" />

shiralizadeh:
@Dozom Nice!

Dozom:
@shiralizadeh 
Hey, I noticed that some tests started failing after the pull request:

 ❌ <lion-select-rich> integration with ListboxMixin > ListboxMixin > Interactions > Disabled Options > does not skip disabled options but prevents checking them
 ❌ <lion-select-rich> integration with ListboxMixin > ListboxMixin > Interactions > Disabled Options > does not check disabled options when selection-follow-focus is enabled

To be honest, I'm not sure how to solve this yet -  it's only my second contribution to open source. I did see that these tests are coming from: lion-select-listbox-suite-integration.test.js, so hopelly that helps.

Open to any suggestions — thanks!

Dozom:
@shiralizadeh Thanks a lot, I really appreciate it. I’ll be waiting.

PR #2531, opened 2025-06-22
fix: ignore disabled form elements when navigating options of Select Rich

Dozom:
…Select Rich (fix issue #2529: [LionSelectRich] disabled options should not get focus bug)

## What I did

1. I added !this.formElements[i].disabled to skip disabled elements when moving with arrows or cursor in the function __getEnabledOption of ListboxMixin inherited by LionSelectRich.
2. I tested locally editing the overview.md (not included in this PR) in select-rich inside docs > components > select-rich.

### DIFF
--- a/packages/ui/components/listbox/src/ListboxMixin.js
+++ b/packages/ui/components/listbox/src/ListboxMixin.js
@@ -869,7 +869,11 @@ const ListboxMixinImplementation = superclass =>
 
       // Try to find the next / previous option
       for (let i = currentIndex + offset; until(i); i += offset) {
-        if (this.formElements[i] && !this.formElements[i].hasAttribute('aria-hidden')) {
+        if (
+          this.formElements[i] &&
+          !this.formElements[i].hasAttribute('aria-hidden') &&
+          !this.formElements[i].disabled
+        ) {
           return i;
         }
       }
@@ -879,7 +883,11 @@ const ListboxMixinImplementation = superclass =>
       if (this.rotateKeyboardNavigation) {
         const startIndex = offset === -1 ? this.formElements.length - 1 : 0;
         for (let i = startIndex; until(i); i += offset) {
-          if (this.formElements[i] && !this.formElements[i].hasAttribute('aria-hidden')) {
+          if (
+            this.formElements[i] &&
+            !this.formElements[i].hasAttribute('aria-hidden') &&
+            !this.formElements[i].disabled
+          ) {
             return i;
           }
         }


packages/ui/components/select-rich/src/LionSelectRich.js
```diff
@@ -50,6 +50,66 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
...
+
+  /**
+   * @param {number} currentIndex
+   * @param {number} offset
+   * @returns {number}
+   */
+  _getPreviousEnabledOption(currentIndex, offset = -1) {
+    return this.__getEnabledOption(currentIndex, offset);
+  }
+
```
ryubro: These look the same with the methods in its parent class. Why did you add these?

packages/ui/components/select-rich/src/LionSelectRich.js
```diff
@@ -50,6 +50,66 @@ export class LionSelectRich extends SlotMixin(ScopedElementsMixin(OverlayMixin(L
...
+  /**
+   * @param {number} currentIndex
+   * @param {number} offset
+   * @returns {number}
+   */
+  __getEnabledOption(currentIndex, offset) {
```
ryubro: I guess this is the same with the method in the parent, other than checking the disabled state. Did you consider updating the method in the parent, instead of overriding it?

packages/ui/components/select-rich/src/LionSelectRich.js
danielnagercarpio: You're right — these methods are the same as in the parent class, there's no point to maintain duplicate code. That was my mistake; I'll remove them.

packages/ui/components/select-rich/src/LionSelectRich.js
danielnagercarpio: @ryubro 
I considered it, but to be honest, I limited the solution to just the fix so it wouldn't affect other parts of the code. Would you prefer that I make this change in the ListboxMixin instead of LionSelectRich?

I saw LionSelectRich (the main objective), LionCombobox, and MdCombobox could be affected, given the mixin chain.

danielnagercarpio:
@ryubro I modified a test that was written before the recent behavior change.
Could you or @tlouisse (the author of the original test) review it?
I updated it because disabled options should now be omitted from the result when using the arrow up or down keys.

packages/ui/components/select-rich/src/LionSelectRich.js
ryubro: I see that you updated the listbox mixin. Did you check if the new behavior makes sense in those other affected components?

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -1305,7 +1305,7 @@ export function runListboxMixinSuite(customConfig = {}) {
       });
 
       describe('Disabled Options', () => {
-        it('does not skip disabled options but prevents checking them', async () => {
```
ryubro, line 1308: Having this test, it seems the original code intended not to skip the disabled options. Probably we want to know why, before updatinging it.

@gerjanvangeest , @tlouisse do you remember anything about it? Maybe so that the screen reader still reads the disabled option?

ryubro: I had a brief check.

We actually want this to be focusable so that the SR can read the disabled options as well, just like the others can see the disabled options on screen. At the moment we just say "dimmed" when the option is disabled. We can improve the text better here, but excluding them from focusilbe element seems not a correct approach.

We really appreciate your contribution, but I think we should close this PR.

If you have a suggestion how we can improve the SR user's experience, please feel free to share.

Screenshot 2025-07-15 at 16 32 41

danielnagercarpio: Thank you @ryubro!

No problem! I’ll give it some thought, and if I come up with something, I’ll share it. I wasn’t aware of the screen reader aspect—thanks for pointing it out. I’m going to keep contributing to this project, as it’s helping me learn a lot, and I truly appreciate you taking the time to review my code—it’s making me a better professional.

danielnagercarpio: @ryubro 
First of all, I have no problem to close the PR.

Here we go 😄

I'm not sure why this behavior was implemented. From my point of view, allowing selection of a disabled option doesn’t seem very practical. Maybe it's used in production, but I’m not aware of any specific cases.

Looking at W3Schools examples for SR usage, the default HTML behavior omits the value of disabled options. So I don’t see a strong reason to keep it.

I don’t mind closing the PR if necessary — just wanted to share my thoughts. If this was done intentionally, maybe it was for a special case.

My suggestion is to update the test and remove the disabled option (which is the main commit here). In my opinion, having a disabled item doesn’t help the user navigate "quickly" through the interface. As your example shows, if there’s only one valid option and four dimmed ones — and the valid one is last — the user might still have to scroll for a long time using the keyboard.

That's all, have a nice day!

Regrets,

Daniel.

ryubro: Thank you for sharing your thought.

I think both approaches make sense. On one hand, it sounds unnecessary to make an unselectible option focusable, on the other hands knowing that some options are available but not valid in a certain situation can be useful info, even for the SR users.

At the moment, we leaned toward the later approach. We always welcome and appreciate our consumers' feedback and contribution.

I'll close the PR for now.

PR #2534, opened 2025-07-03
feat(input-tel): use first preferred region to set a default region

Hzunax:
## What I did

1. Move the `preferredRegions` property from LionInputTelDropdown to LionInputTel
2. In the `__calculateActiveRegion`, check if `preferredRegions` exist and if so, take the first value in that list and set it as the active region. It comes in 3rd place in order of priority, after user input and before Locale.

## Issue

There is no way of passing a default region to the input-tel.

## Reproduce

1. Go to the demo of [input-tel-dropdown preferred regions](https://lion.js.org/components/input-tel-dropdown/use-cases/#preferred-regions)
2. Clear the input
3. The active region is now based on the locale (for me that was en-GB), while the preferred regions are NL and DE

image
image

## Fix

By using the preferred regions, a default is set which will be more in line with expected user input.

image
image


tlouisse (approved):

tlouisse:
Thanks for the PR @Hzunax!

Issue #2492, opened 2025-03-19
[lion-input-amount] Editing negative numbers is not possible


vorosgy:
After entering a negative number into a LionInputAmount, the dash character turns into a minus sign, but the LionInputAmount is unable to parse that later, so if the user tries to edit the number, it results in an error.

## Expected behavior

Editing a negative number in a LionInputAmount should be possible.

## Actual Behavior

https://lion.js.org/components/input-amount/overview/
Enter a negative number, switch focus, then edit it. An error message appears.

## Additional context

I think the problem is in packages/ui/components/input-amount/src/parsers.js, namely in the unmatchedInput regexp which does not match the minus sign.

gerjanvangeest:
Great find @vorosgy , are you willing to do a PR to fix this?

vorosgy:
Right now I can't even run the tests for some reason, so I have to say no. But if that helps, adding the minus sign to unmatchedInput seems to solve the problem.

gitraiwaqas:
Hi, I'd like to work on this issue as my first contribution. I've already reproduced the bug and tested a local fix involving the `unmatchedInput` regex update.

May I proceed with creating a PR?

Thanks!

gerjanvangeest:
Yes, @gitraiwaqas please do so 💪!

gerjanvangeest:
Fixed in https://github.com/ing-bank/lion/pull/2538

PR #2538, opened 2025-07-15
fix: editing negative numbers

bilalozlu:
## What I did

1. fixed the 'editing negative number' problem by adding (\u2212) to unmatchedInput regex inside parseAmount function. 
2. added test for the negative numbers both with \u2212 and '-'.

issue: #2492


gerjanvangeest:
Hi @bilalozlu can you add a changeset `npm run changeset` so the change ends up in our changelog?

gerjanvangeest (approved):

PR #2539, opened 2025-07-17
fix(combobox): add support for disabled and readonly states with corresponding tests

sainzrow:
## What I did

1. Switched to toggleAttribute function for disabled and readOnly attributes on combobox and input field.
2. Added tests and updated documentation for this options


denilsonsa:
This looks like a duplicate of #2536, which I opened a few days ago. But your PR looks way more complete than mine.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1213,8 +1217,12 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
...
   __setComboboxDisabledAndReadOnly() {
     if (this._comboboxNode) {
-      this._comboboxNode.setAttribute('disabled', `${this.disabled}`);
-      this._comboboxNode.setAttribute('readonly', `${this.readOnly}`);
+      this._comboboxNode.toggleAttribute('disabled', this.disabled);
+      this._comboboxNode.toggleAttribute('readOnly', this.readOnly);
+    }
+    if (this._inputNode) {
+      this._inputNode.toggleAttribute('disabled', this.disabled);
+      this._inputNode.toggleAttribute('readOnly', this.readOnly);
```
ryubro: Before this, we updated only `_comboboxNode` (although was not implemented correctly). Can you explain why this was added?  What was the issue before toggling attribute in the input node?

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -502,6 +502,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
     if (this._selectionDisplayNode) {
       this._selectionDisplayNode.comboboxElement = this;
     }
+
+    if (this.disabled || this.readOnly) {
+      this.__setComboboxDisabledAndReadOnly();
```
ryubro, line 507: What's the idea of adding this in constructor? What issue have you experienced before adding it here?

sainzrow: This was necessary because when first rendered, the input fields nested in the combobox structure didn't have the disabled nor readonly attributes and was incorrectly rendering a enabled field in case for example of a readonly combobox configuration .

packages/ui/components/combobox/src/LionCombobox.js
sainzrow: I've added those in order to make the input fields self consistent with the disabled and the readonly states.
Before this, those attributes were not handled at input level and this could have caused issues when the parent attributes were not aligned with the input field

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -3431,4 +3437,70 @@ describe('lion-combobox', () => {
...
+
+      // Try to open overlay by clicking input
+      _inputNode.dispatchEvent(new Event('click', { bubbles: true, composed: true }));
+      await el.updateComplete;
+      expect(el.opened).to.be.false;
+    });
+  });
+
+  describe('Disabled state removed', () => {
+    it('does not open overlay or allow input when disabled', async () => {
```
ryubro: This description doesn't reflect the content of the test, right?

packages/ui/components/combobox/src/LionCombobox.js
ryubro: Thanks

packages/ui/components/combobox/src/LionCombobox.js
ryubro: Thanks

packages/ui/components/combobox/src/LionCombobox.js
ryubro: > those attributes were not handled at input level and this could have caused issues when the parent attributes were not aligned with the input field

One more question, not aligning means, when parent attribute is disabled, input is not, right? Did you just find it annoying or had some practical problems as well?

sainzrow: There was some strange behaviour on ios webkit where the input was considered as contenteditable when it wasn't.

packages/ui/components/combobox/test/lion-combobox.test.js
sainzrow: Yep, sorry.
Fixing it rn.

packages/ui/components/combobox/src/LionCombobox.js
ryubro: Thank you.

Now we have `disabled` and `readonly` for both input and its wrapper.

I think it would be better if we remove `disabled` from its wrapper or replace it with `aria-disabled="true"`.

How do you think?

packages/ui/components/combobox/test/lion-combobox.test.js
```diff
@@ -3431,4 +3437,70 @@ describe('lion-combobox', () => {
       });
     });
   });
+
+  describe('Disabled state', () => {
```
ryubro, line 3441: Can you add a11y tests like here: https://github.com/ing-bank/lion/blob/226f0cf003588b7062c590da551a3070e7940f82/packages/ui/components/dialog/test/lion-dialog.test.js#L191

packages/ui/components/combobox/src/LionCombobox.js
sainzrow: How can we remove it from the wrapper if, in order to apply the disabled state, we need to pass the attribute:

`<lion-combobox ?disabled=${disabled}>...</lion-combobox>`

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1213,8 +1217,12 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
    */
   __setComboboxDisabledAndReadOnly() {
     if (this._comboboxNode) {
-      this._comboboxNode.setAttribute('disabled', `${this.disabled}`);
-      this._comboboxNode.setAttribute('readonly', `${this.readOnly}`);
+      this._comboboxNode.toggleAttribute('disabled', this.disabled);
+      this._comboboxNode.toggleAttribute('readOnly', this.readOnly);
```
tlouisse:
```suggestion
      this._comboboxNode.toggleAttribute('readonly', this.readOnly);
      this._comboboxNode.setAttribute('aria-readonly', `${this.readOnly}`);      
```

packages/ui/components/combobox/src/LionCombobox.js
tlouisse, line 1225:
```suggestion
      this._inputNode.toggleAttribute('readOnly', this.readonly);
```

tlouisse: Quite complex with all those different elements that can get disabled :)

To be sure we're on the same page:

- there's the combobox host (lion-combobox component). This is not focusable and does not communicate anything to the screen reader. It has a disabled attr/prop for styling and configuration (end user configures it and rest happens under the hood) purposes
- there's the `._comboboxNode` [role=combobox] (due to browser/SR difference ,this can be the wrapper div[role=combox] around the textbox, or the textbox itself). 
  - In case [role=combobox] is the wrapping div, we need aria-disabled="true" (semantics) on the div, and disabled on `_inputNode` (for operability)
  - in case [role=combobox] is textbox (`._comboboxNode === .inputNode`) we only need `disabled` there (as it works for semantics + operability)
  

So I think this approach would work:

- applying both disabled + aria-disabled on comboboxNode
- applying disabled on _inputNode (it might be already applied if ._comboboxNode === ._inputNode)


Same applies for readOnly.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1213,8 +1217,12 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
    */
   __setComboboxDisabledAndReadOnly() {
     if (this._comboboxNode) {
-      this._comboboxNode.setAttribute('disabled', `${this.disabled}`);
-      this._comboboxNode.setAttribute('readonly', `${this.readOnly}`);
+      this._comboboxNode.toggleAttribute('disabled', this.disabled);
```
tlouisse, line 1224:
```suggestion
      // Since `._comboboxNode` can either be `<div "role=combobox">` or `<input>`,
      // we need to cater for both scenarios (aria for semantics, "native attr" for operability)
      this._comboboxNode.toggleAttribute('disabled', this.disabled);
      this._comboboxNode.setAttribute('aria-disabled', `${this.disabled}`);
```

packages/ui/components/combobox/src/LionCombobox.js
tlouisse, line 1234:
```suggestion
      // N.B. in case ._inputNode === ._comboboxNode (we have <input role="combobox">)
      // this value has already been set above. This is fine, as a toggle with boolean flag is idempotent.
      this._inputNode.toggleAttribute('disabled', this.disabled);
      this._inputNode.toggleAttribute('readOnly', this.readOnly);
```

packages/ui/components/combobox/test/lion-combobox.test.js
sainzrow: Added also accessibility tests for disabled and readonly states

tlouisse (approved):

PR #2561, opened 2025-08-18
amount-dropdown small fixes

thatdudemanguy:
## What I did

1. exported missing types
2. removed an unnecessary opinionated styling which annoyed subclassers

### DIFF
--- a/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
+++ b/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
@@ -101,9 +101,6 @@ export class LionInputAmountDropdown extends LionInputAmount {
     const refs = {
       dropdown: {
         ref: this.refs.dropdown,
-        props: {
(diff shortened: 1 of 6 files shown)


packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
```diff
@@ -102,7 +102,7 @@ export class LionInputAmountDropdown extends LionInputAmount {
       dropdown: {
         ref: this.refs.dropdown,
         props: {
```
gerjanvangeest: Maybe just remove the whole section?

thatdudemanguy: Yeah was doubting here, fine by me. I'll update the type as well.

thatdudemanguy: done.

PR #2570, opened 2025-09-12
Feat/lazy loaded singleton

ryubro:
## What I did

1. Add a function that defers initiating and registering manager until a getter or setter is called
2. Made icon manager and overlay manager singleton

After the PR, the types of singleton managers is not `any` anymore but the type of the relevant manager. This might generate type errors as the type became stricter.

### DIFF
--- a/.changeset/poor-penguins-wait.md
+++ b/.changeset/poor-penguins-wait.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': minor
+---
+
+Make the manager singletons get instantiated and registered lazily.
--- a/packages/ui/components/icon/src/icons.js
+++ b/packages/ui/components/icon/src/icons.js
@@ -1,4 +1,16 @@
-import { singletonManager } from 'singleton-manager';
+import { singletonManager, lazifyInstantiation } from 'singleton-manager';
 import { IconManager } from './IconManager.js';
 
-export const icons = singletonManager.get('@lion/ui::icons::0.x') || new IconManager();
+/**
+ * @returns {IconManager}
+ */
+function getIconManager() {
+  if (!singletonManager.has('@lion/ui::icons::0.x')) {
+    const iconManager = new IconManager();
+    singletonManager.set('@lion/ui::icons::0.x', iconManager);
+  }
+
+  return singletonManager.get('@lion/ui::icons::0.x');
+}
+
+export const icons = lazifyInstantiation(getIconManager);
--- a/packages/ui/components/icon/test/hammer.svg.js
+++ b/packages/ui/components/icon/test/hammer.svg.js
@@ -1,2 +1,5 @@
-export default /** @param {(strings: TemplateStringsArray, ... expr: string[]) => string} tag */ tag =>
-  tag`<svg focusable="false" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" data-test-id="svg-hammer"><path d="M.476 12.915L6.7 6 9 8.556l-6.223 6.915a1.542 1.542 0 0 1-1.15.529 1.54 1.54 0 0 1-1.15-.53c-.636-.706-.636-1.85 0-2.555zm12.638-9.031L16 6.863 12.866 10 4 .919 9.35 0l1.912 1.972.251-.251c.52-.52 2.4 1.363 1.88 1.882l-.279.28z"></path></svg>`;
+// @ts-nocheck
+export default /** @type {import("lit").TemplateResult} */ (
+  tag =>
+    tag`<svg focusable="false" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" data-test-id="svg-hammer"><path d="M.476 12.915L6.7 6 9 8.556l-6.223 6.915a1.542 1.542 0 0 1-1.15.529 1.54 1.54 0 0 1-1.15-.53c-.636-.706-.636-1.85 0-2.555zm12.638-9.031L16 6.863 12.866 10 4 .919 9.35 0l1.912 1.972.251-.251c.52-.52 2.4 1.363 1.88 1.882l-.279.28z"></path></svg>`
+);
--- a/packages/ui/components/icon/test/lion-icon.test.js
+++ b/packages/ui/components/icon/test/lion-icon.test.js
@@ -183,13 +183,19 @@ describe('lion-icon', () => {
     try {
       icons.addIconResolver(
         'foo',
-        // eslint-disable-next-line no-promise-executor-return
-        () => new Promise(resolve => setTimeout(() => resolve(heartSvg), 10)),
+        () =>
+          /** @type {Promise<TemplateResult>} */ (
+            // eslint-disable-next-line no-promise-executor-return
+            new Promise(resolve => setTimeout(() => resolve(heartSvg)))
+          ),
       );
       icons.addIconResolver(
         'bar',
-        // eslint-disable-next-line no-promise-executor-return
-        () => new Promise(resolve => setTimeout(() => resolve(hammerSvg), 4)),
+        () =>
+          /** @type {Promise<TemplateResult>} */ (
+            // eslint-disable-next-line no-promise-executor-return
+            new Promise(resolve => setTimeout(() => resolve(hammerSvg)))
+          ),
       );
       const el = await fixture(html`<lion-icon icon-id="foo:lorem:ipsum"></lion-icon>`);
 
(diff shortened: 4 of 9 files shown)


packages/ui/components/helpers/src/lazyfyManager.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * @type {<T>(func: () => T) => T}
+ */
+const lazyfyManager = registerSingleton => {
```
tlouisse:
```suggestion
const lazifyManager = registerSingleton => {
```

packages/ui/components/helpers/src/lazyfyManager.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * @type {<T>(func: () => T) => T}
```
tlouisse:
```suggestion
 * Lazifies the instantiation of singletons. It does so by wrapping the singleton in a Proxy, which delays
 * instantiation until one of its members gets called/set for the first time.
 * This is important in cases where a singleton needs to be overridden on an app level:
 * let's say there are two versions of @lion/ui in an app that need to synchronize their locale.
 * We want to make sure that the latest LocalizeManager instance (localize) of the latest lion version (that has the latest features and is backwards compatible with earlier versions) is registered before others. 
 * We see often that this complicates setup/index files (the need of extra files to make sure logic gets executed before side-effectful imports. Also, build processes like Rollup tend to merge files and therefore hoist imports and change execution order.
 This method aims to solve above problems.
 * @type {<T>(func: () => T) => T}
```

tlouisse (changes requested):
Looks good, few remarks

packages/ui/components/icon/test/lion-icon.test.js
```diff
@@ -183,13 +183,19 @@ describe('lion-icon', () => {
     try {
       icons.addIconResolver(
         'foo',
-        // eslint-disable-next-line no-promise-executor-return
-        () => new Promise(resolve => setTimeout(() => resolve(heartSvg), 10)),
+        () =>
+          /** @type {Promise<TemplateResult>} */ (
+            // eslint-disable-next-line no-promise-executor-return
+            new Promise(resolve => setTimeout(() => resolve(heartSvg), 10))
```
tlouisse: Can you explain why these timeouts are needed?

packages/ui/components/helpers/src/lazyfyManager.js
tlouisse: Maybe we can also give it a more precise name, like `lazifySingletonInstantiation`

tlouisse: Also, we should consider providing this functionality in the singleton-manager package (where we could call it `lazifyInstantiation`).
As the helpers entrypoint only exposes dev and demo helpers (and sibling exports of this entrypoint would end up in micro frontend builds in our extension layer atm).

.changeset/poor-penguins-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+Make the localize manager get instantiated and registered lazily
```
tlouisse:
```suggestion
Make the manager singletons get instantiated and registered lazily
```

.changeset/poor-penguins-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
```
tlouisse:
```suggestion
'@lion/ui': minor
```

packages/ui/components/icon/src/icons.js
```diff
@@ -1,4 +1,17 @@
...
 import { IconManager } from './IconManager.js';
 
-export const icons = singletonManager.get('@lion/ui::icons::0.x') || new IconManager();
+/**
+ * @returns {IconManager}
+ */
+function getIconManager() {
+  if (!singletonManager.has('@lion/ui::icons::0.x')) {
+    const iconManager = new IconManager();
+    singletonManager.set('@lion/ui::icons::0.x', iconManager);
```
tlouisse, line 10: This (registration by default) deviates a bit from previous behaviour. Not necessarily bad though, as we now reached parity with how we did `getLocalizeManager` (and we will release this as a minor anyway)

packages/ui/components/icon/test/hammer.svg.js
```diff
@@ -1,2 +1,5 @@
-export default /** @param {(strings: TemplateStringsArray, ... expr: string[]) => string} tag */ tag =>
-  tag`<svg focusable="false" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" data-test-id="svg-hammer"><path d="M.476 12.915L6.7 6 9 8.556l-6.223 6.915a1.542 1.542 0 0 1-1.15.529 1.54 1.54 0 0 1-1.15-.53c-.636-.706-.636-1.85 0-2.555zm12.638-9.031L16 6.863 12.866 10 4 .919 9.35 0l1.912 1.972.251-.251c.52-.52 2.4 1.363 1.88 1.882l-.279.28z"></path></svg>`;
+// @ts-nocheck
```
tlouisse, line 1: this was needed bc TS complained, I assume?

packages/ui/components/localize/src/getLocalizeManager.js
```diff
@@ -59,16 +60,20 @@ import { LocalizeManager } from './LocalizeManager.js';
  *
  * @returns {LocalizeManager}
  */
-export function getLocalizeManager() {
-  if (singletonManager.has('@lion/ui::localize::0.x')) {
-    return singletonManager.get('@lion/ui::localize::0.x');
+function getLocalizeManager() {
```
tlouisse, line 64: As the `localize` export in https://github.com/ing-bank/lion/blob/master/packages/ui/components/localize/src/singleton.js has the exact same behaviour, maybe we can deprecate `getLocalizeManager` and adjust the singleton.js file, so that it will be exactly the same as with overlays and icons.
(this means this file can be removed, as we only need backw. compat on the entrypoints).

In a separate PR we can then get rid of the internal use of `getLocalizeManager`

ryubro: My first attempt was to do something like, to keep it consistent with others:
```js
const  localize = lazyfyManager(getLocalizeManager);
const getLocalizeManagerLazily = () => localize;

export { localize, getLocalizeMangerLazily as getLocalizeManager }
```

Then this test started to fail: https://github.com/ing-bank/lion/blob/02d0106a069fcf9e420d5875be1dac302a3223b3/packages/ui/components/localize/test/getLocalizeManager.test.js#L25-L30
because once any of the assessors or setters was triggered, `localize` points to the instance that was created even after `singletonManager._map.clear();` is called in the `beforeEach` func.

Maybe I could add a method to reset such thing inside `lazyfyManager`, but I thought maybe we can leave it out of the scope of this PR keeping the interface exactly the same as before.

packages/ui/components/icon/test/lion-icon.test.js
ryubro: This is not added by me (I just updated the type), but reading the comments, it wanted to test a race conditions with two icon resolver with different time to resolve.

packages/ui/components/helpers/src/lazyfyManager.js
ryubro: Oh, I didn't know about singleton-manager package. Is it only consumed inside @lion/ui?

packages/ui/components/localize/src/getLocalizeManager.js
tlouisse: For clarity: deprecate the export (we only use it internally as a way to configure lazifyManager)

packages/ui/components/icon/test/hammer.svg.js
ryubro: Yes. It worked much better when the type was defined as TemplateResult in the tests, but this svg.js file was complaining that this is not a TemplateResult. I could dig deeper but as this is like a resource for the test, I didn't put more effort than putting the ts comment.

ryubro: This happens because now the singleton variables are typed with the corresponding managers. Before they were `any`.

packages/ui/components/helpers/src/lazyfyManager.js
tlouisse: Also in our extension layer (and by some of our consumers)

okadurin (approved):

PR #2574, opened 2025-09-18
fix(margin); attempt to fix issue 2572

ruud:
## What I did
Set negative margin on contentWrapperNode to compensate for the extra margin on body so that the overlay content does not shift when a scrollbar appears/disappears. This is an alternative to setting margin-right style on document.body.

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -847,7 +847,6 @@ export class OverlayController extends EventTarget {
       case 'before-show':
         this.__bodyClientWidth = document.body.clientWidth;
         this.__bodyClientHeight = document.body.clientHeight;
-        this.__bodyMarginRightInline = document.body.style.marginRight;
         this.__bodyMarginBottomInline = document.body.style.marginBottom;
         break;
       case 'show': {
@@ -863,22 +862,20 @@ export class OverlayController extends EventTarget {
           document.body.clientWidth - /** @type {number} */ (this.__bodyClientWidth);
         const scrollbarHeight =
           document.body.clientHeight - /** @type {number} */ (this.__bodyClientHeight);
-        const newMarginRight = this.__bodyMarginRight + scrollbarWidth;
         const newMarginBottom = this.__bodyMarginBottom + scrollbarHeight;
         // @ts-expect-error [external]: CSS not yet typed
         if (window.CSS?.number && document.body.attributeStyleMap?.set) {
           // @ts-expect-error [external]: types attributeStyleMap + CSS.px not available yet
-          document.body.attributeStyleMap.set('margin-right', CSS.px(newMarginRight));
-          // @ts-expect-error [external]: types attributeStyleMap + CSS.px not available yet
           document.body.attributeStyleMap.set('margin-bottom', CSS.px(newMarginBottom));
         } else {
-          document.body.style.marginRight = `${newMarginRight}px`;
           document.body.style.marginBottom = `${newMarginBottom}px`;
         }
+        // set negative margin on contentWrapperNode to compensate for the extra margin on body
+        // so that the overlay content does not shift when a scrollbar appears/disappears
+        this.contentWrapperNode.style.marginLeft = `-${scrollbarWidth}px`;
         break;
       }
       case 'hide':
-        document.body.style.marginRight = this.__bodyMarginRightInline || '';
         document.body.style.marginBottom = this.__bodyMarginBottomInline || '';
         break;
       /* no default */


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -863,22 +862,20 @@ export class OverlayController extends EventTarget {
...
-          document.body.attributeStyleMap.set('margin-right', CSS.px(newMarginRight));
           // @ts-expect-error [external]: types attributeStyleMap + CSS.px not available yet
           document.body.attributeStyleMap.set('margin-bottom', CSS.px(newMarginBottom));
         } else {
-          document.body.style.marginRight = `${newMarginRight}px`;
           document.body.style.marginBottom = `${newMarginBottom}px`;
         }
+        // set negative margin on contentWrapperNode to compensate for the extra margin on body
+        // so that the overlay content does not shift when a scrollbar appears/disappears
+        this.contentWrapperNode.style.marginLeft = `-${scrollbarWidth}px`;
```
tlouisse: Looks good!
We should do the same as well for marginTop:

```js
this.contentWrapperNode.style.marginTOp = `-${scrollbarHeight}px`;
```

For bonus points: 
- we can remove the `if (window.CSS?.number && document.body.attributeStyleMap?.set)` checks and just get for the "lowest common denominator solution" (that works well and is most readable).
-`window.getComputedStyle` shouldn't be guarded in an if... (it's widely supported)

tlouisse: For more bonus points: this is one of the few areas in our overlay code that is not tested. Ideally we have a test where we test the difference between overlay position from window:

```js
const viewportOffset = el.getBoundingClientRect();
const top = viewportOffset.top;
cosnt left = viewportOffset.left;

```


We should then force scrollbars for all browsers by injecting some css:

```js

function ensureScrollbar() {
  const style = document.createElement("style");
  // N.B. needs to be done for non webkit browser as well still...
  style.textContent = `::-webkit-scrollbar { -webkit-appearance: none; width: 25px; background-color: grey; }; ::-webkit-scrollbar-thumb { background-color: rgba(0,0,0,.5); } body { height: 10000px; width: 10000px; }`;
  document.head.appendChild(style);
  
  const cleanup = () => {
     document.head.removeChild(style);
  }
  
  return cleanup;
}

// ===========

const cleanupEnsureScrollbar = ensureScrollbar();

// test

cleanupEnsureScrollbar();

```
- 

```js
function ensureScrollbar() {
  const style = document.createElement("style");
  style.textContent = `::-webkit-scrollbar { -webkit-appearance: none; width: 25px; background-color: grey; }; ::-webkit-scrollbar-thumb { background-color: rgba(0,0,0,.5); } body { height: 10000px; }`;
  document.head.appendChild(style);
}

ensureScrollbar();
```

ruud: Yes, good suggestion. I can add the same for the vertical part.

Not sure though about injecting CSS if not required.

tlouisse: > Not sure though about injecting CSS if not required.

Ah sorry if I didn't make it clear. That part is purely for the test (and should be removed again at the end of the test)

tlouisse: @ruud 
Hmm, I was too quick with my previous assumption (that we need to prevent a "jump" of the dialog itself)

As the overlay was already placed inside a container with `position: fixed`, the body marginRight and -bottom did not affect the overlay positioning => so what we try to prevent is a jump of the body when the scrollbar on the page disappears. The only place to do that, is on the body itself.

So: we need to keep the current behavior, as it prevents a jump of the body page when a dialog opens.

On a page that "paints" the full width of the page we might see a white "gap" (that fills the "void" the scrollbar left behind).

However, most websites have a little space (look at this github page for instance) so, this goes unnoticed.

TLDR; it's not really solvable in Lion. It should be solved in your app.

In case of the gap on the scrollbar leaves behind is too visible, it's best to mitigate the contrast of the gap in your app. For instance:
- make sure the body color matches that of the page 
- make the backdrop less opaque

ruud: Why does the scrollbar disappear?
So, the negative margin (right & bottom) changes are not solving this too?

tlouisse: > Why does the scrollbar disappear?

Bc we want to prevent scroll of the page. This is desired behavior

> So, the negative margin (right & bottom) changes are not solving this too?

No, since they have a parent with `position: fixed`

tlouisse: As you rightfully pointed out @ruud, we can also try padding instead of margin.
This would mean the background color "blends in" by default

ruud: I only changed the margin-right into padding-right, because margin-bottom is working as it should

tlouisse: Ok, could you do the same for paddingBottom? (to keep things consistent and generic)
As the same problem can occur for other layouts with a horizontal scrollbar.

And then the unit test should be updated. Can you then also add a changeset (`npm run changeset`)?

ruud: Ok, changed the marginBottom to paddingBottom too, like was done for marginRight

PR #2575, opened 2025-09-24
More tests to make sure "overlays-scroll-lock" CSS class is removed on dialog teardown

okadurin:


packages/ui/components/dialog/test/lion-dialog--enable-scroll-on-dialog-removal.test.js
```diff
@@ -0,0 +1,141 @@
...
+import { sendKeys, setViewport } from '@web/test-runner-commands';
+import '@lion/ui/define/lion-dialog.js';
+
+/**
+ * @typedef {import('../src/LionDialog.js').LionDialog} LionDialog
+ * @typedef {import('lit').TemplateResult} TemplateResult
+ */
+const fixture = /** @type {(arg: TemplateResult) => Promise<LionDialog>} */ (_fixture);
+
+describe.skip('lion-dialog', () => {
```
tlouisse, line 12: Is this skipped because the test is failing and needs a fix to be added later?

okadurin: Yes, at the moment the skipped test is failing because of the bug we have in the code. And the idea to fix it later. Removing `skip` will be one of the proofs that the fix works

gerjanvangeest: Maybe add a TODO here, so we won't forget

PR #2578, opened 2025-10-01
Astro integration

paulcodiny:
## What I did

Together with the rest of the gang we migrated the portal to Astro.

1. Added Astro side by side with the current Rocket setup
2. Preserved backwards compatibility with the documents from the current portal
3.


tlouisse:
Thanks for all the great work @paulcodiny 🙌

PR #2579, opened 2025-10-02
Fix/listbox mixin

okadurin:
ListboxMixin should not move around the drafts of slottable nodes

### DIFF
--- a/packages/ui/components/combobox/test/lion-combobox.test.js
+++ b/packages/ui/components/combobox/test/lion-combobox.test.js
@@ -414,7 +414,7 @@ describe('lion-combobox', () => {
       expect(el.validationStates.error).to.have.property('MatchesOption');
     });
 
-    it('kepes slottable provided in `slots` getter as direct host child', async () => {
+    it('keeps slottable provided in `slots` getter as direct host child', async () => {
       class MyEl extends SlotMixin(LionCombobox) {
         // @ts-ignore
         get slots() {
--- a/packages/ui/components/overlays/test/OverlayController.test.js
+++ b/packages/ui/components/overlays/test/OverlayController.test.js
@@ -799,6 +799,9 @@ describe('OverlayController', () => {
           const { parentOverlay, childOverlay } = await createNestedEscControllers(parentContent);
           await mimicEscapePress(childOverlay.contentNode);
 
+          // without this line, the test is unstable on FF sometimes
+          await aTimeout(0);
+
           expect(parentOverlay.isShown).to.be.false;
           expect(childOverlay.isShown).to.be.true;
 


packages/ui/components/listbox/src/ListboxMixin.js
```diff
@@ -42,12 +42,101 @@ import { LionOptions } from './LionOptions.js';
...
+    }
+    if (isInsideSlotComment) {
+      if (c.textContent.includes('_end_slot_')) {
+        isInsideSlotComment = false;
+      }
+      return;
+    }
     if (!isNamedSlottable) {
       target.appendChild(c);
     }
```
tlouisse: Nice solution 💪
As this is tightly coupled to a private implementation of SlotMixin, can we put `_moveDefaultSlottablesToTarget` inside SlotMixin and call it from ListboxMixin?
It will be easier to maintain like this.

Maybe smth like this would work best:



```suggestion
function moveDefaultSlottablesToTarget(source, target) {
  for (const childNode of source.childNodes) {
    const isNamedSlottable = childNode.hasAttribute?.('slot');
    if (isNamedSlottable) continue;
    const isNamedSlotMixinComment = childNode.nodeType === Node.COMMENT_NODE && childNode.textContent.matches(/^_(start|end)_slot_/) && ! childNode.textContent.matches(/^_(start|end)_slot__$/); // make sure it's not the default slot
    if (isNamedSlotMixinComment) continue;
   
    target.appendChild(childNode);
```

tlouisse: I think we also need to be sure that we do not accidentally skip the default slot: https://github.com/ing-bank/lion/blob/master/packages/ui/components/core/src/SlotMixin.js#L144 (in this case the name is '').
Will update example above.

packages/ui/components/core/test/SlotMixin.test.js
```diff
@@ -134,6 +135,140 @@ describe('SlotMixin', () => {
...
+      get slots() {
+        return {
+          ...super.slots,
+          _optionalLabel: () => ({
+            template: this.renderOptionalLabel
+              ? html`<span id="labelOptionalId">(Optional)</span>`
+              : html`${nothing}`,
+            renderAsDirectHostChild: true,
+          }),
+          '': () => html`<span id="defaultSlotLabelId">default</span>`,
```
tlouisse: Default slots should be moved as well indeed. They will with this test implementation (as they are not surrounded by comments). 

However, they won't when configured as a [SlotRerenderObject](https://lion.js.org/fundamentals/systems/core/slotmixin/#slotrerenderobject). The test should explicitly check this, as I expect this to fail with the current implementation of `moveDefaultSlottablesToTarget`

packages/ui/components/core/test/SlotMixin.test.js
```diff
@@ -134,6 +135,140 @@ describe('SlotMixin', () => {
...
+  // The reson the test is skipped now is because of the bug
+  it.skip('should not move nodes provided in `slots` and should move projection like lion-option', async () => {
+    const nodeLabel = document.createElement('span');
+    nodeLabel.id = 'nodeLabel';
+    class MyEl extends SlotMixin(LionCombobox) {
+      // @ts-ignore
+      get slots() {
+        return {
+          ...super.slots,
+          _optionalLabel: () => ({
```
tlouisse: N.B. the problem only occurs with lazy rendered slots from SlotMixin that are being moved. Can we call this `_lazyRenderedSlot` for clarity?

packages/ui/exports/core.js
```diff
@@ -1,7 +1,7 @@
 export { DisabledMixin } from '../components/core/src/DisabledMixin.js';
 export { DisabledWithTabIndexMixin } from '../components/core/src/DisabledWithTabIndexMixin.js';
 export { ScopedStylesController } from '../components/core/src/ScopedStylesController.js';
-export { SlotMixin } from '../components/core/src/SlotMixin.js';
+export { SlotMixin, moveDefaultSlottablesToTarget } from '../components/core/src/SlotMixin.js';
```
tlouisse: This suggests it is a universal utility that can be used in any context for all Lion consumers. However, it only works with SlotMixin. 
Exposing it means increasing the maintenance/support surface, where we would rather keep its exposure limited (also since the need for it will disappear once cross root aria is available: https://lion-web.netlify.app/_merged_assets/_static/crossRootAriaLion.pdf)

packages/ui/components/core/test/SlotMixin.test.js
```diff
@@ -1,7 +1,8 @@
 import { defineCE, expect, fixture, fixtureSync, unsafeStatic, html } from '@open-wc/testing';
 import { SlotMixin } from '@lion/ui/core.js';
-import { LitElement } from 'lit';
+import { LitElement, nothing } from 'lit';
 import sinon from 'sinon';
+import { LionCombobox } from '@lion/ui/combobox.js';
```
tlouisse: It's a good habit to avoid circular deps between "entrypoints". Also, it makes the SlotMixin test more reliable as we know the tested behavior is not affected by any of the internals of LionCombobox

packages/ui/components/core/test/SlotMixin.test.js
okadurin: > Default slots should be moved as well indeed

Did you mean they should *not* be moved? Like everything provided via `slots` should *not* be moved

> as they are not surrounded by comments

They actually are surrounded by comment and with the suggested in the PR implementation, everything thing between `start` and `end` is ignored not not moved which is good. So even though the generated `<span id="defaultSlotLabelId">default</span>` does not have any named slottable that is not a problem since it is surrounded by comments. 
The only slottable that is not surrounded by comments is `[nodeLabel](https://github.com/ing-bank/lion/blob/60e24a238d1ac14423f1735d2e70a08409ecf4af/packages/ui/components/core/test/SlotMixin.test.js#L159)` for some reason. But that is also fine because the function ignores the elements that have slot attribute.

tlouisse: > Did you mean they should not be moved? Like everything provided via slots should not be moved

The name`moveDefaultSlottablesToTarget` suggests they _should_ be moved.
Howver, saying slots created by SlotMixin should not be touched does make sense indeed.
(as the use case is to bring default slottables created by user to the right target).

Maybe the actual name should be `_moveUserProvidedDefaultSlottablesToTarget`


Indeed, a [regular template](https://github.com/ing-bank/lion/blob/master/packages/ui/components/core/src/SlotMixin.js#L245) also creates [slot comment nodes](https://github.com/ing-bank/lion/blob/master/packages/ui/components/core/src/SlotMixin.js#L144), so the test was correct on this point (it's been a while since I wrote this, so I didn't know from the top of my mind :))

tlouisse (approved):

PR #2597, opened 2025-10-27
Fix/nested overlays [WIP]

okadurin:
Changes in the PR
---
1. There is a race condition between `Escape key handler` [in OverlayController](https://github.com/ing-bank/lion/blob/b74d312bc06e2b9c7131331709982a572e3d23b8/packages/ui/components/overlays/src/OverlayController.js#L1213C3-L1213C18) set for every overlay `on show` and in subclassers of OverlayMixin (such as tooltip, select-rich, combobox). The main idea in the PR is to stop setting `this.opened = false` on Escape press in every OverlayMixing subclasser. Instead we let OverlayController hide the overlays. Then we synchronize OverlayController with OverlayMixin by setting this.opened = false in [here](https://github.com/ing-bank/lion/pull/2597/files#diff-35ee755d91d1a8583176185e503850766bd3c0132b743b7ac0345b0b93049b53R191).
2. Note, I updated the test in [here](https://github.com/ing-bank/lion/pull/2597/files#diff-676e0eb2ae5ed6a625a70c5c7d2ae2b6cc993e137989ee6f004957651e613b1aR1271) because the test was false positive. The code in non-test environment does not clear the value as the test suggested

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -1215,7 +1215,6 @@ export class OverlayController extends EventTarget {
 
     const hasPressedInside =
       event.composedPath().includes(this.contentNode) ||
-      event.composedPath().includes(this.referenceNode) ||
       (this.invokerNode && event.composedPath().includes(this.invokerNode)) ||
       deepContains(this.contentNode, /** @type {HTMLElement|ShadowRoot} */ (event.target));
 
--- a/packages/ui/components/overlays/test/OverlayController.test.js
+++ b/packages/ui/components/overlays/test/OverlayController.test.js
@@ -745,6 +745,35 @@ describe('OverlayController', () => {
         expect(parentOverlay.isShown).to.be.true;
         expect(childOverlay.isShown).to.be.false;
       });
+
+      describe('When checking whether Escape is pressed from inside of the overlay', () => {
+        it('should hide when Escape is pressed from inside `contentNode`', async () => {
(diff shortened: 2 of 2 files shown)


packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -1282,10 +1281,12 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
    */
   _onKeyUp(ev) {
     const lastKey = ev && ev.key;
-    this.opened = this._showOverlayCondition({
-      lastKey,
-      currentValue: this._inputNode.value,
-    });
+    if (lastKey !== 'Escape') {
+      this.opened = this._showOverlayCondition({
```
ryubro, line 1285: It seems `this._showOverlayCondition` checks the `lastKey` already. Would it be better if we handle the logic in `_showOverlayCondition`?

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1215,6 +1215,7 @@ export class OverlayController extends EventTarget {
 
     const hasPressedInside =
       event.composedPath().includes(this.contentNode) ||
+      (this.invokerNode && event.composedPath().includes(this.invokerNode)) ||
```
ryubro, line 1218: It seems the `__escKeyHandler()` was designed to close only the most nested overlay already. Just wondering, without this added condition, what was the problem?

PR #2606, opened 2025-11-06
chore: deploy to gh pages

ayo-run:
## What I did

1. add [`gh-pages`](https://www.npmjs.com/package/gh-pages) as a dev dependency, to facilitate easy deployment
2. add instructions in our readme on how collaborators can deploy

After a project member sets the [Pages repository setting](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site), we should have http://ing-bank.github.io/lion up (see [deployment on my fork](https://ayoayco.github.io/lion)). The `/next` route is messed up, but will be fixed once I point the domain to the gh-pages deployment. Will be done via a PR to [js-org](https://github.com/js-org/js.org) as my next step, but they require that the static site is up first.

**Update**: [GH Pages deployment](http://ing-bank.github.io/lion) now up, you can check on the PR to point the domain to it here: https://github.com/js-org/js.org/pull/10304

Thanks! 🙏🏼

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "license": "MIT",
   "scripts": {
     "build": "npm run build:astro && npm run build:rocket && npm run build:astro:update-links && mv dist _site/next",
-    "build:astro": "ASTRO_BASE_URL=/next npx astro build --force",
+    "build:astro": "astro telemetry disable && ASTRO_BASE_URL=/next astro build --force",
     "build:astro:update-links": "node ./src/utils/rewrite-links.mjs",
     "build:clean": "find src/pages -path \"*/_demos/*.js\" -type f -delete && rm -rf node_modules/_lion_docs && npm run build",
     "build:rocket": "rocket build",
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+# temp files (vim/nvim)
+*~
+*swo
+*swp
+
 # build output
 dist/
 # generated types


package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "scripts": {
     "build": "npm run build:astro && npm run build:rocket && npm run build:astro:update-links && mv dist _site/next",
-    "build:astro": "ASTRO_BASE_URL=/next npx astro build --force",
+    "build:astro": "astro telemetry disable && ASTRO_BASE_URL=/next astro build --force",
```
tlouisse, line 7: nice one

package.json
```diff
@@ -15,6 +15,8 @@
     "debug:firefox": "web-test-runner --watch --config web-test-runner-firefox.config.mjs",
     "debug:no-scoped-registries-polyfill": "npm run debug -- --no-scoped-registries-polyfill",
     "debug:webkit": "web-test-runner --watch --config web-test-runner-webkit.config.mjs",
+    "predeploy": "npm run build",
```
tlouisse, line 18: can you explain the idea here?

ayo-run: the `deploy` script calls this `predeploy` thingy :) -- so we can have other stuff here like copy assets and such

we can think of it as the "build stage" in netlify before the actual deployment stage

PR #2610, opened 2025-11-07
fix(ValidateMixin): no dispatch validators on readOnly form items

geniuspacs:
## What I did

1. I prevent validation on readOnly fields which uses ValidateMixin


geniuspacs:
Hi @gerjanvangeest I signed the CLA but for any reason appears like I didn't 😑

PR #2619, opened 2025-11-14
chore: update urls to use new one

ayo-run:
## What I did

1. update all occurrences of old domain to use the new site URL; except the ones in `ui/_legacy-changelogs`

### DIFF
--- a/packages-node/providence-analytics/CHANGELOG.md
+++ b/packages-node/providence-analytics/CHANGELOG.md
@@ -219,7 +219,7 @@
 
   **We also unfixed lion's dependencies (now using caret ^) on its own packages**, because it caused a lot of problems with duplicate installations for end users as well as subclassers and its end users. Both of these changes may affect subclassers in a breaking manner, hence the minor bump.
 
-  Be sure to [read our Rationale on this change](https://lion.js.org/docs/rationales/versioning/) and what this means for you as a user.
+  Be sure to [read our Rationale on this change](https://lion.js.org/fundamentals/rationales/versioning/) and what this means for you as a user.
 
 ## 0.12.4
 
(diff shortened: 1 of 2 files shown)


packages/ajax/CHANGELOG.md
```diff
@@ -192,7 +192,7 @@
 
   **We also unfixed lion's dependencies (now using caret ^) on its own packages**, because it caused a lot of problems with duplicate installations for end users as well as subclassers and its end users. Both of these changes may affect subclassers in a breaking manner, hence the minor bump.
 
-  Be sure to [read our Rationale on this change](https://lion-web.netlify.app/docs/rationales/versioning/) and what this means for you as a user.
+  Be sure to [read our Rationale on this change](https://lion.js.org/docs/rationales/versioning/) and what this means for you as a user.
```
ayo-run: This link is broken, we need to update or remove.

packages-node/providence-analytics/CHANGELOG.md
```diff
@@ -219,7 +219,7 @@
 
   **We also unfixed lion's dependencies (now using caret ^) on its own packages**, because it caused a lot of problems with duplicate installations for end users as well as subclassers and its end users. Both of these changes may affect subclassers in a breaking manner, hence the minor bump.
 
-  Be sure to [read our Rationale on this change](https://lion-web.netlify.app/docs/rationales/versioning/) and what this means for you as a user.
+  Be sure to [read our Rationale on this change](https://lion.js.org/docs/rationales/versioning/) and what this means for you as a user.
```
ayo-run: This link is broken, we need to update or remove.

packages/ajax/CHANGELOG.md
ryubro: If the url is invalid, should we leave it as it was?

As this is a release note, probably removing wouldn't be a good idea. And keeping the old url might signal the readers that this used to be valid. Changing URL to the one that doesn't exist looks a bit weird to me.

tlouisse: it should be https://lion.js.org/fundamentals/rationales/versioning/

ayo-run: Thanks @tlouisse I will use this one.

ryubro (approved):

PR #2621, opened 2025-11-15
fix(LionInputStepper): improve handling of decimal step values and al…

### LINKED ISSUE #2615: input-stepper stops after 3 counts (up or down)
## Expected behavior

Input stepper to work with decimal numbers.

## Actual behavior

With the code below, the stepper reaches  (from 0) only up to 0.3 with repeated clicks on `+`. I cannot make it go above 0.3.

```html
<lion-input-stepper
          name="foo"
          label="Bar"
          min="0"
          max="9"
          step="0.1"
          value="0"
        ></lion-input-stepper>
```

vikri-odoo:
## What I did

1. Reproduced the issue with step value 0.1
2. Wrote the test cases for increment and decrement with value 5.5 to confirm that it's working for the edge cases
3. Provided fix using epsilon as JavaScript gives 0.3 % 0.1 = 0.09999999999999998

## This closes #2615

### DIFF
--- a/.changeset/four-rabbits-cover.md
+++ b/.changeset/four-rabbits-cover.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(LionInputStepper): improve handling of decimal step values and alignment closes #2615


MiB-1:
@gerjanvangeest  this fixes #2615

gerjanvangeest:
@MiB-1 Looks good :)
Can you add a changeset (`npm run changeset`) so the change ends up in our changelog?

PR #2624, opened 2025-11-18
fix(pagination): remove unnecessary ellipsis when count equals visibl…

### LINKED ISSUE #2589: [LionPagination] Ellipsis shown when visible pages is 5 and page count is 6
## Expected Behavior

No ellipsis when all pages fit:
- `count = 6`: `[1, 2, 3, 4, 5, 6]`

## Actual Behavior

Ellipsis shown unnecessarily:
- `[1, 2, 3, 4, 5, '...', 6]`

Image

### Possible cause

- The original condition `finish <= visiblePages + 1` was correct.
- The issue came from the fallback logic:
  - `finish - current <= 3` was too strict and triggered ellipsis even when all pages could fit.
  - It didn’t account for the case where `count = visiblePages + 1`.

TudorGR:
fix(pagination): remove unnecessary ellipsis when count equals visiblePages + 1

Fixes #2589

- Changed condition from `count > visiblePages` to `count > visiblePages + 1`
- This allows one extra page to be displayed without ellipsis (e.g., [1,2,3,4,5,6] instead of [1,2,3,4,5,'...',6])
- Added test case to verify ellipsis is not shown when count=6 and visiblePages=5

## What I did

1. **Identified the issue**: The `__calculateNavList()` method in `LionPagination.js` was using `count > visiblePages` which caused it to show ellipsis even when there was only one extra page beyond the visible pages setting.

2. **Applied minimal fix**: Updated line 199 in `packages/ui/components/pagination/src/LionPagination.js` from `if (this.count > this.__visiblePages)` to `if (this.count > this.__visiblePages + 1)` with an explanatory comment.

3. **Added test coverage**: Created a new test case in `packages/ui/components/pagination/test/lion-pagination.test.js` to verify that no ellipsis is shown when count=6 and visiblePages=5.

4. **Created changeset**: Added a changeset file following the project's contribution guidelines for proper changelog generation.

5. **Verified the fix**: 
   - All existing tests pass (3310 tests on Chromium)
   - New test passes
   - Linting passes with no errors
   - Manual verification confirms the fix works as expected

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/ui/components/pagination/src/LionPagination.js
+++ b/packages/ui/components/pagination/src/LionPagination.js
@@ -196,7 +196,7 @@ export class LionPagination extends LocalizeMixin(LitElement) {
     const finish = this.count;
     // If there are more pages then we want to display we have to redo the list each time
     // Else we can just return the same list every time.
-    // Allow one extra page without ellipsis (e.g., count=6 when visiblePages=5)
+    // Allow up to 2 extra pages without ellipsis (e.g., count=6 or count=7 when visiblePages=5)
     if (this.count > this.__visiblePages + 2) {
       // Calculate left side of current page and right side
       const pos3 = this.current - 1;
--- a/packages/ui/components/pagination/test/lion-pagination.test.js
+++ b/packages/ui/components/pagination/test/lion-pagination.test.js
@@ -178,5 +178,31 @@ describe('Pagination', () => {
       // There should be 8 nav items: previous button + 6 page buttons + next button
       expect(navItems.length).to.equal(8);
     });
+
+    it('should not show ellipsis when count is visiblePages + 2 (count=7, visiblePages=5)', async () => {
+      const el = await fixture(html` <lion-pagination count="7" current="1"></lion-pagination> `);
+      // Check that no ellipsis is rendered (no <span> elements with '...')
+      const spans = Array.from(
+        /** @type {ShadowRoot} */ (el.shadowRoot).querySelectorAll('li span'),
+      );
+      expect(spans.length).to.equal(0);
+
+      // There should be 9 nav items: previous button + 7 page buttons + next button
+      const navItems = Array.from(/** @type {ShadowRoot} */ (el.shadowRoot).querySelectorAll('li'));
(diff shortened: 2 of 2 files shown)


MiB-1:
@gerjanvangeest I think we should do `count > visiblePages + 2` because it's giving the same problem when we use count as 7.
Screenshot 2025-11-20 at 9 48 14 AM

packages/ui/components/pagination/src/LionPagination.js
```diff
@@ -196,7 +196,8 @@ export class LionPagination extends LocalizeMixin(LitElement) {
     const finish = this.count;
     // If there are more pages then we want to display we have to redo the list each time
     // Else we can just return the same list every time.
-    if (this.count > this.__visiblePages) {
+    // Allow one extra page without ellipsis (e.g., count=6 when visiblePages=5)
+    if (this.count > this.__visiblePages + 1) {
```
gerjanvangeest:
```suggestion
    if (this.count > this.__visiblePages + 2) {
```

gerjanvangeest (changes requested):

packages/ui/components/pagination/test/lion-pagination.test.js
```diff
@@ -164,4 +164,19 @@ describe('Pagination', () => {
...
+      // Check that no ellipsis is rendered (no <span> elements with '...')
+      const spans = Array.from(
+        /** @type {ShadowRoot} */ (el.shadowRoot).querySelectorAll('li span'),
+      );
+      expect(spans.length).to.equal(0);
+
+      // There should be 8 nav items: previous button + 6 page buttons + next button
+      expect(navItems.length).to.equal(8);
+    });
+  });
```
gerjanvangeest, line 207: Can we also have a test for 
- `should not show ellipsis when count is visiblePages + 2 (count=7, visiblePages=5)`
- `should show ellipsis when count is visiblePages + 3 (count=8, visiblePages=5)`?

PR #2629, opened 2025-11-19
feat(LionInputStepper): implement self-destructing output content for…

### LINKED ISSUE #2602: [LionInputStepper] update sr-only text from `input-stepper__value` to output element
## Expected behavior

Currently we have a `div` which contains the `valueText`, which is used to have the decrement and increment buttons announce the updated value.

A [w3 example for spinbutton](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/quantity-spinbutton/) makes use of an [output element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/output). This comes with the benefit that works the same as an `aria-live` attribute, so we can remove some logic.

So change:

```html
<div class="input-stepper__value">${this.__valueText}</div>
```

to:

```html
<output
  class="input-stepper__value"
  for=".."
>${this.__valueText}</output>
```

And remove the `_onEnterButton()` and `_onLeaveButton()` logic.

Also add a `self-destruct` mechanism (also shown in the same example), so the text-content doesn't pop-up in the a11y dom tree, when navigating through all the text-nodes.

vikri-odoo:
## What I did

1. from
`<div class="input-stepper__value">${this.__valueText}</div>` 
to
`<output class="input-stepper__value"  for="..">${this.__valueText}</output>`
2. remove the `_onEnterButton()` and `_onLeaveButton()` logic.

closes #2602

### DIFF
--- a/.changeset/giant-ties-attack.md
+++ b/.changeset/giant-ties-attack.md
@@ -1,0 +1,8 @@
+---
+'@lion/ui': patch
+---
+
+feat(LionInputStepper): implement self-destructing output content for value display
+
+1. from <div class="input-stepper__value">${this.__valueText}</div> to <output class="input-stepper__value"  for="..">${this.\_\_valueText}</output>
+2. remove the \_onEnterButton() and \_onLeaveButton() logic.


packages/ui/components/input-stepper/src/LionInputStepper.js
```diff
@@ -427,7 +441,9 @@ export class LionInputStepper extends LocalizeMixin(LionInput) {
   /** @protected */
   _inputGroupTemplate() {
     return html`
-      <div class="input-stepper__value">${this.__valueText}</div>
+      <output for="${this.fieldName}" data-self-destruct="2000" class="input-stepper__value"
```
gerjanvangeest: The `for` should point to the input field.

```suggestion
      <output for="${this._inputId}" data-self-destruct="2000" class="input-stepper__value"
```

PR #2636, opened 2025-12-02
chore: Modified test for failing

SebayK:
In `OverlaysManager.test.js`, replaced calls to `hide()` with `teardown()` for dialog elements to simulate element removal from the DOM, ensuring scroll lock classes are properly cleared.

Updated assertions to check that both `overlays-scroll-lock` and `overlays-scroll-lock-ios-fix` classes are removed from `document.body` after teardown.

### DIFF
--- a/.changeset/nervous-carrots-lay.md
+++ b/.changeset/nervous-carrots-lay.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+remove overlay-scroll-lock properly after overlay teardown


packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -1443,6 +1446,10 @@ export class OverlayController extends EventTarget {
     if (this.#isRegisteredOnManager()) {
       this.manager.remove(this);
     }
+
```
SebayK, line 1449: For me it should be done in next iteration. We should clean up the DOM after teardown

Issue #2608, opened 2025-11-06
[OverlayController] _handleZIndex never executes - 'setup' phase is unused


hyakt:
## Expected behavior

The `_handleZIndex` method should set z-index to '1' for local overlays during initialization when the content node doesn't have an explicit z-index. 
This ensures local overlays appear on top of elements with no z-index that appear later in the DOM.

## Actual Behavior

The z-index logic never executes because `_handleZIndex` checks for phase === 'setup', but this phase is never used anywhere in the codebase.

Code reference: https://github.com/ing-bank/lion/blob/eb8b7bd73fda44f2b07f9fead15d803b029d7935/packages/ui/components/overlays/src/OverlayController.js#L680-L691

Suggested fix:  
From `if (phase === 'setup')` to `if (phase === 'init')`

## Additional context

This affects local overlays like tooltips when they don't have explicit z-index values set.

SebayK:
i can take it

PR #2640, opened 2025-12-09
fix(zIndex-execution-during-local-init): update zIndex handling durin…

SebayK:
fix(zIndex-execution-during-local-init): update zIndex handling during initialization and extend tests

## What I did

1. update zIndex handling during initialization
2. extend zIndex tests in OverlayController.test.js

### DIFF
--- a/packages/ui/components/overlays/src/OverlayController.js
+++ b/packages/ui/components/overlays/src/OverlayController.js
@@ -8,7 +8,7 @@ import { _adoptStyleUtils } from './utils/adopt-styles.js';
  * @typedef {'init'|'teardown'|'before-show'|'show'|'hide'} OverlayPhase
  * @typedef {import('@lion/ui/types/overlays.js').ViewportConfig} ViewportConfig
  * @typedef {import('@lion/ui/types/overlays.js').OverlayConfig} OverlayConfig
- * @typedef {import('@lion/overlays').OverlaysManager} OverlayManager
+ * @typedef {import('@lion/overlays').OverlaysManager} OverlaysManager
  * @typedef {import('@popperjs/core').Options} PopperOptions
  * @typedef {import('@popperjs/core').Placement} Placement
  * @typedef {import('@popperjs/core').createPopper} Popper


SebayK:
Resolve #2608

packages/ui/components/overlays/src/OverlayController.js
```diff
@@ -5,9 +5,10 @@ import { overlayShadowDomStyle } from './overlayShadowDomStyle.js';
 import { _adoptStyleUtils } from './utils/adopt-styles.js';
 
 /**
- * @typedef {'setup'|'init'|'teardown'|'before-show'|'show'|'hide'|'add'|'remove'} OverlayPhase
+ * @typedef {'init'|'teardown'|'before-show'|'show'|'hide'} OverlayPhase
  * @typedef {import('@lion/ui/types/overlays.js').ViewportConfig} ViewportConfig
  * @typedef {import('@lion/ui/types/overlays.js').OverlayConfig} OverlayConfig
+ * @typedef {import('@lion/overlays').OverlaysManager} OverlayManager
```
tlouisse:
```suggestion
 * @typedef {import('@lion/overlays').OverlaysManager} OverlaysManager
```

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -229,47 +229,112 @@ describe('OverlayController', () => {
         await ctrl.show();
         // @ts-expect-error find out why config would/could be undfined
         expect(ctrl.content.style.zIndex).to.equal(`${ctrl.config.zIndex + 1}`);
+        expect(ctrl.manager.shownList).to.have.lengthOf(1);
```
tlouisse: PR looks good to me. 
Any reason we should check for the shownList length here in these tests, instead of in the manager? https://github.com/ing-bank/lion/blob/master/packages/ui/components/overlays/test/OverlaysManager.test.js#L83

SebayK: f I understand correctly, if there is more siblings, it always should have higher zIndex than the previous sibling by one. It's probably a little more than that. Let me know

tlouisse:
@SebayK we discussed this in person, but I leave a comment here for transparency...

This one needs a bit more investigation. I see we applied zIndex on 3 nodes (should have been one I think).

Also: now that we moved to dialog, (n case it's configured as modal) it paints to the top layer. It will win from every other layer regardless of zIndex (that's what I expect). I think this should somehow be reflected in the tests. (this will play an even bigger role once we move to popovers for (local) overlays)

PR #2641, opened 2025-12-09
fix(ChoiceGroupMixin): initialModelValue is not managed properly

geniuspacs:
## What I did

1. I fix a bug which causes _initialModelValue is returned undefined on ChoiceGroupMixin

### DIFF
--- a/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
+++ b/packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
@@ -261,6 +261,7 @@ const ChoiceGroupMixinImplementation = superclass =>
         return this[property];
       }
 
+      // TODO: consider a shared approach for all value getters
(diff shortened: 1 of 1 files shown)


packages/ui/components/form-core/src/choice-group/ChoiceGroupMixin.js
```diff
@@ -260,7 +260,16 @@ const ChoiceGroupMixinImplementation = superclass =>
       ) {
         return this[property];
       }
-      return this.formElements.filter(el => _filterFn(el, property)).map(el => el.property);
+
+      const filteredFnElements = this.formElements.filter(el => _filterFn(el, property));
```
tlouisse, line 265:
```suggestion
      // TODO: consider a shared approach for all value getters
      const filteredFnElements = this.formElements.filter(el => _filterFn(el, property));
```

gerjanvangeest (approved):

PR #2645, opened 2025-12-19
fix: allow pasting value with currency for input-amount

okadurin:
Before the fix when there is a cell in Excel which is formatted as Accounting, Currency: Zl, Decimal 2 and the whole cell is copied/pasted, the input-amount shows `100`. After the fix the displayed value is `1` as expected

### DIFF
--- a/packages/ui/components/input-amount/src/parsers.js
+++ b/packages/ui/components/input-amount/src/parsers.js
@@ -31,14 +31,15 @@ function round(value, decimals) {
  * @param {FormatNumberOptions} [givenOptions] Locale Options
  */
 export function parseAmount(value, givenOptions) {
-  const unmatchedInput = value.match(/[^0-9,.\-\u2212 ]/g);
+  const trimmedValue = value.trim();
+  const unmatchedInput = trimmedValue.match(/[^0-9,.\-\u2212 ]/g);
   // for the full paste behavior documentation:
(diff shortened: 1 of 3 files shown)


packages/ui/components/input-amount/src/parsers.js
```diff
@@ -31,14 +31,15 @@ function round(value, decimals) {
  * @param {FormatNumberOptions} [givenOptions] Locale Options
  */
 export function parseAmount(value, givenOptions) {
-  const unmatchedInput = value.match(/[^0-9,.\-\u2212 ]/g);
+  const trimmedValue = value.trim();
+  const unmatchedInput = trimmedValue.match(/[^0-9,.\-\u2212 ]/g);
```
ryubro, line 35: So before the fix, this value would have been `null`, right? How that can make `' 1,00 zł '`  to 100?

okadurin: unmatchedInput was an array of `z` and `l`

tlouisse: Before and after indeed.
The fix is in the fact that providing `'1,00 zł'` to parseNumber instead of `' 1,00 zł '` returns `1` instead of `100` (the preceding space was a problem)

ryubro: Ah, OK Thanks!

PR #2656, opened 2026-01-08
changed currency of Bulgaria to Euro and proper changesets

VivekSingla20:
## What I did

This pull request makes a major update to the currency mapping for Bulgaria in the `@lion/ui` package. Bulgaria's currency is now set to Euro (EUR) instead of Bulgarian Lev (BGN).

Currency mapping update:

* Changed the currency code for Bulgaria (`BG`) from `BGN` to `EUR` in `currencyUtil.js`.
* Documented the major change in `.changeset/nine-beers-pump.md`.

### DIFF
--- a/.changeset/nine-beers-pump.md
+++ b/.changeset/nine-beers-pump.md
@@ -1,5 +1,5 @@
 ---
-'@lion/ui': major
+'@lion/ui': patch
 ---
 
 The currency for Bulgaria (BG) is now set to Euro(EUR) instead of Bulgarian Lev(BGN)


.changeset/nine-beers-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': major
```
gerjanvangeest: This is not a major change, a fix would be sufficient i would say.

```suggestion
'@lion/ui': patch
```

gerjanvangeest (approved):

PR #2657, opened 2026-01-08
fix(currency): fixed Bulgria currency to eur in drop down

### LINKED ISSUE #2654: Update currencies input-amount-dropdown since Bulgaria is now using the Euro
Bulgaria stopped using the Lev and adopted the Euro as its currency since 1st of January 2026

Update the java script from: https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-amount-dropdown/src/currencyUtil.js

To get the latest `countryToCurrencyList` values.
Make sure BG is updated to Eur.

SahilJat:
## What I did
closes #2654 
1. Updated the currency in Input-amount-dropdown for Bulgaria from Lev to EUR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42298,7 +42298,7 @@
     },
     "packages/ui": {
       "name": "@lion/ui",
-      "version": "0.15.2",
+      "version": "0.15.3",
       "license": "MIT",
       "dependencies": {
         "@bundled-es-modules/message-format": "^6.2.4",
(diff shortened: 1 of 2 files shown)


gerjanvangeest:
Hi @SahilJat thank you for your effort, however [someone else created a PR ](https://github.com/ing-bank/lion/pull/2656)already before you, with the same change.
So i'll close your PR again.

PR #2658, opened 2026-01-08
Fix that ListboxMixin doesn't work very well with ScopedStylesMixin when the host component is cached

ryubro:
## What I did

When a component uses `ListboxMixin` and its options are using `ScopedStylesMixin`, unnecessary `_onListboxContentChanged()` is triggered when the component is disconnected then reconnected, due to the mutation observer that `ListboxMixin` uses. This cause a subtle mal-behavior.

I updated so that the mutation observer ignores adding and removing of style tags.

### DIFF
--- a/.changeset/clever-emus-add.md
+++ b/.changeset/clever-emus-add.md
@@ -2,4 +2,4 @@
 '@lion/ui': patch
 ---
 
-Make ListboxMixin works well with ScopedStylesContoller when cache is used
+Make ListboxMixin works well with ScopedStylesController when cache is used


packages/ui/components/core/test-helpers/getCachedFixture.js
```diff
@@ -0,0 +1,44 @@
+import { defineCE, fixture, html, unsafeStatic } from '@open-wc/testing';
+import { LitElement, nothing } from 'lit';
+import { cache } from 'lit/directives/cache.js';
+import { createRef } from 'lit/directives/ref.js';
+
+export async function getCachedFixture(render) {
```
ryubro, line 6: Maybe we can apply this function to other places where we test the cached behavior, but that would make this PR too big. I think it deserves a separate PR.

packages/ui/components/combobox/src/LionCombobox.js
```diff
@@ -965,7 +965,10 @@ export class LionCombobox extends LocalizeMixin(OverlayMixin(CustomChoiceGroupMi
           if (isInlineAutoFillCandidate) {
             const textboxValue = this._getTextboxValueFromOption(option);
             const stringValues =
-              textboxValue && typeof textboxValue === 'string' && typeof curValue === 'string';
+              typeof textboxValue === 'string' &&
+              textboxValue !== '' &&
+              typeof curValue === 'string' &&
+              curValue !== '';
             const beginsWith =
               stringValues && textboxValue.toLowerCase().indexOf(curValue.toLowerCase()) === 0;
```
ryubro, line 973: A side fix. Before, we haven't checked if curValue is empty. `<any_string>.indexOf('')` is always `0`, so when the `curValue` is empty and somehow the runtime reaches this part of the code, `beginsWith` becomes always `true` which is not our intention, I believe.

.changeset/clever-emus-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+Make ListboxMixin works well with ScopedStylesContoller when cache is used
```
okadurin:
```suggestion
Make ListboxMixin works well with ScopedStylesController when cache is used
```

ryubro: Thank you, done!

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -1735,6 +1741,17 @@ export function runListboxMixinSuite(customConfig = {}) {
         await el.updateComplete;
         expect(spy).to.have.been.calledTwice;
       });
+
+      it('does not calls "_onListboxContentChanged" during connect and disconnected', async () => {
+        const renderEl = refObj => html`<${wrappingTag} ${ref(refObj)}></${wrappingTag}>`;
```
okadurin, line 1746: Could the API be improved so that it looks as follows:
```js
const renderEl = html`<${wrappingTag}></${wrappingTag}>`;
const { el, show, hide } = await getCachedFixture(renderEl);

``` 
?

packages/ui/components/combobox/src/LionCombobox.js
okadurin: Could we have a separate PR for this with the tests that cover the use cases please?

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
tlouisse: I agree. The api (especially the need for ref) is a bit hard.
If we avoid the ref (by adding a parent element we get the child from here: https://github.com/ing-bank/lion/pull/2658/changes#diff-d8b52d9e6096f14efb4891a2c493273c45907c276475b24d4fe87a52d2cdbd7cR29) we can get this done

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
tlouisse, line 1541: N.B. Not related to the PR, but this template seems to have options provided?

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -1547,8 +1549,12 @@ export function runListboxMixinSuite(customConfig = {}) {
         expect(el.querySelector('[role=listbox]')).to.equal(_listboxNode);
 
         expect(el.formElements.length).to.equal(2);
-        expect(_listboxNode.children.length).to.equal(2);
-        expect(_listboxNode.children[0].tagName).to.equal(cfg.optionTagString.toUpperCase());
+        // When we use ScopedStylesController, lisboxNode can have style tag along with options
+        const childrenOtherThanStyle = Array.from(_listboxNode.children).filter(
+          child => child.nodeName !== 'STYLE',
+        );
+        expect(childrenOtherThanStyle.length).to.equal(2);
```
tlouisse, line 1556: We could also consider to opt-in to `instanceof LionOption || potentialOption.getAttribute('role') === 'option'`

tlouisse: In that case, we should provide a filter fn to `moveUserProvidedDefaultSlottablesToTarget`
See https://github.com/ing-bank/lion/blob/master/packages/ui/components/listbox/src/ListboxMixin.js#L871C11-L871C52

packages/ui/components/core/test-helpers/getCachedFixture.js
tlouisse: This is a great idea indeed, as we have more places where 

Then we could also tackle the comment from Oleksii: https://github.com/ing-bank/lion/pull/2658/changes#r2682461536

Even elaborating further on the right abstraction: are the problems related to the cache directly or the fact that we run (dis)connectedCallback?
See https://github.com/webcomponents/gold-standard/wiki, especially https://github.com/webcomponents/gold-standard/wiki/Reattachment


That also reminds us we need to move on with our own [Golden component](https://github.com/orgs/ing-bank/projects/5) :)

tlouisse: Made a follow-up ticket for the cache/reattachment test-helper: https://github.com/ing-bank/lion/issues/2669

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
tlouisse: (a.k.a. don't see the relation between the description and conent of the test)

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
tlouisse: That would make the solution more generic. For the most common case, the change in ScopedStylesController will do.

packages/ui/components/combobox/src/LionCombobox.js
ryubro: To reach this part of the code, `_handleAutocomplete` should be called and `userIntendsInlineAutoFill` should be `true`. When things go well, we shouldn't reach this part of the code when `curValue` is an empty string, I believe (and probably that's why we didn't check if `curValue` is empty). We were reaching this part because of the bug that this PR is fixing and once this PR is merged, I don't think we can easily emulate a test case.

I'm fixing it here because I found this logic gap during the inspection.

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
ryubro: Do you mean using `querySelector` or some kind of DOM traversal (like node.children[0])?

Maybe a bit broader topic, but do we see using `ref` a bad practice? The API might look a bit complicated than just passing the template, but it makes sense to use `ref` for this use case, no? I am a bit hesitant to add an unecessary parent so that we can traverse DOM while we have ref directives just for such use case. It also gives more flexibility to target other than the top element of the template, and ref keeps getting the DOM reference more declarative.

In general, I think we can benefit from embracing `ref` a bit more, but especially considering that this is a test util (hence will be consumed only by contributors), wouldn't it be a good case to try ref?

tlouisse: Let's take this discussion to the [helper](https://github.com/ing-bank/lion/issues/2669). My personal take:
- most important thing is the api for the consumer. 
  - the more aligned we make it with the existing fixture, the better.
  - if we can avoid importing and passing the ref in all tests for which we want to test consistent behavior during re-attachment (maybe all components), we should make setup as simple as possible
- the less deps/imports, the better (from test-helper perspective. Although for non-prod deps like test-helpers, it's more of a nice to have. Again, from a security perspective (recent npm supply chain attacks), it's a good general habit to go full https://e18e.dev. 
- `ref` can have a readability advantage when we don't have a web component (before it existed, we used [renderLitAsNode](https://github.com/search?q=repo%3Aing-bank%2Flion%20renderLitAsNode&type=code) (we should update those demos))
- querySelector would be fine, but its' an implementation detail. The most important thing is that it's stable and abstracted away by our helper

packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
ryubro: It seems updating `moveUserProvidedDefaultSlottablesToTarget` wouldn't work. After doing this, the test (https://github.com/ing-bank/lion/blob/c5e2eb37699108d8c7e360a908dc264fc5cbfe1e/packages/ui/components/form-integrations/test/model-value-consistency.test.js#L561) fails, because we wrap the options with div (otherwise only the first one is appended to the element) and div is not either LionOption nor having role of option.

Maybe we can change the test a bit to make it work, but now I'm thinking maybe filtering the slottables by the instanceof and role might be too fragile. I guess we should keep this part dumb.

tlouisse: Agree, the whitelist might not really work as this is considered correct:

```html
<lion-listbox role="listbox">
  <div>
    <lion-option role="option"></lion-option>
    <lion-option role="option"></lion-option>
  </div>  
</lion-listbox>
```

We could consider filtering out disallowed tags, but better just wait till we can use cross root aria: https://github.com/web-platform-tests/interop/issues/1011 (then we don't need to move around dom nodes)

PR #2660, opened 2026-01-08
Collapsible - add invoker-position attribute

michal-perlakowski:
This adds an invoker-position attribute. When invoker-position is set to "bottom", it renders the invoker below the content, instead of above it.

### DIFF
--- a/packages/ui/components/collapsible/test/lion-collapsible.test.js
+++ b/packages/ui/components/collapsible/test/lion-collapsible.test.js
@@ -145,32 +145,24 @@ describe('<lion-collapsible>', () => {
         expect(content).to.have.attribute('aria-labelledby', invoker?.id);
       });
     });
+  });
 
-    describe('[invoker-position]', () => {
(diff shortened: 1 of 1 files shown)


packages/ui/components/collapsible/src/LionCollapsible.js
```diff
@@ -30,10 +30,21 @@ export class LionCollapsible extends LitElement {
...
+      invokerPosition: {
+        type: String,
+        reflect: true,
+        attribute: 'invoker-position',
+      },
     };
   }
 
   render() {
+    if (this.invokerPosition === 'bottom') {
```
gerjanvangeest, line 42: @michal-perlakowski can you add a test for this?

michal-perlakowski: Done

michal-perlakowski: @gerjanvangeest I updated the tests. Can you rerun the workflow?

PR #2661, opened 2026-01-12
fix(listbox): remove resetInteractionStates from clear function

gerjanvangeest:
## What I did

1.


packages/ui/components/listbox/test-suites/ListboxMixin.suite.js
```diff
@@ -589,6 +589,53 @@ export function runListboxMixinSuite(customConfig = {}) {
...
+
+        el.setCheckedIndex(0);
+        await el.updateComplete;
+        expect(el.modelValue).to.deep.equal('10');
+
+        el.clear();
+        await el.updateComplete;
+        expect(el.modelValue).to.equal('');
+
+        const el2 = await fixture(html`
```
okadurin: Could you move it to the separate `it(`?

PR #2673, opened 2026-01-15
fix: update pug-code-gen

okadurin:
This PR might fix https://github.com/ing-bank/lion/pull/2671 . 
Note I checked `npm run start` and Rocket works fine

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12687,29 +12687,6 @@
         "node": ">=0.4.0"
       }
     },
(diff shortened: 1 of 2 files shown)


tlouisse:
> This PR might fix #2671 . Note I checked `npm run start` and Rocket works fine

Nice. I think this allows us to remove https://github.com/ing-bank/lion/pull/2671

PR #2682, opened 2026-01-20
feat(input-file): add status messages for screenreader users

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/input-file/src/LionInputFile.js
+++ b/packages/ui/components/input-file/src/LionInputFile.js
@@ -1,6 +1,6 @@
 import { LionField } from '@lion/ui/form-core.js';
 import { LocalizeMixin } from '@lion/ui/localize.js';
-import { css, html } from 'lit';
+import { css, html, nothing } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { ScopedElementsMixin } from '../../core/src/ScopedElementsMixin.js';
 import { FileHandle, MAX_FILE_SIZE } from './FileHandle.js';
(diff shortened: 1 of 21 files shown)


packages/ui/components/input-file/src/LionInputFile.js
```diff
@@ -872,6 +880,23 @@ export class LionInputFile extends ScopedElementsMixin(LocalizeMixin(LionField))
...
+    const outputElement = /** @type {HTMLElement} */ (this.shadowRoot?.querySelector('output'));
+
+    const timeoutValue = outputElement?.dataset?.selfDestruct
+      ? Number(outputElement.dataset.selfDestruct)
+      : 5000;
+    clearTimeout(this.timer);
+    if (outputElement) {
+      this.timer = setTimeout(() => {
+        if (outputElement.parentNode) {
+          this._statusMessage = nothing;
```
tlouisse, line 893: I know this component doesn't really comply with our form-controls, but did we explore the possibility of connecting it with the feedback mechanism we have in the ValidateMixin?

gerjanvangeest: The InputFile has validation messaging in place. This serves only of a different goal.
These are messages for screen readers, to inform them about the status. Therefore we use the output element (which works like an automatic aria-live region), place the messages and remove them again afterwards.
The same thing we do inside the input-stepper.

tlouisse: The ValidateMixin handles all types of vaidations/feedback (error,warning,success,info etc.). 
- from a consumer perspective, it allows control of the UX (supply visibility conditions)
- instead of having aria-live functionality, it's a bit more nuanced: https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/test/FormControlMixin.test.js#L310
- in order for aria-live functionality to work, we need to create an idref in the same dom (the input is in light dom now, the output is in shadow dom). Current code will not work

packages/ui/components/input-file/src/LionSelectedFileList.js
```diff
@@ -210,7 +212,7 @@ export class LionSelectedFileList extends LocalizeMixin(ScopedElementsMixin(LitE
           padding-inline-start: 0;
         }
 
-        .sr-only {
+        .u-sr-only {
```
tlouisse: Think we can stick to our current convention (LionInputRange etc.) of not prefixing with "u-". This comes from [ITCSS](https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/#utility-namespaces-u-), but we don't apply it elsewhere in the codebase, so maybe it's best to go with the less opinionated "sr-only" (and apply it in LionCalendar as well)

gerjanvangeest: Yes, i saw we use both atm, I thought we wanted to go with the `u-`, but i can easily revert this

PR #2702, opened 2026-02-13
fix(LionInputStepper): prevent click on increase/decrease buttons whe…

geniuspacs:
## What I did

1. When LionInputStepper has readonly attribute is still possible to click on increase/decrease buttons at prefix/suffix slots, so you can change the input stepper value even if is a readonly field. This commit prevents click on this scenario.

### DIFF
--- a/.changeset/yellow-rocks-drum.md
+++ b/.changeset/yellow-rocks-drum.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(LionInputStepper): prevent click on increase/decrease buttons when is readonly
--- a/docs/components/input-stepper/use-cases.md
+++ b/docs/components/input-stepper/use-cases.md
@@ -127,3 +127,29 @@ The value is always aligned with the defined step size when using the increase o
   value="55"
 ></lion-input-stepper>
 ```
+
+### Disabled
+
+`disabled` attribute will be delegated to the native `<input>` to make it disabled.
+
+This field **will not be included** in the parent fieldset or form's `serializedValue`.
+
+```html preview-story
+<lion-input-stepper name="year" disabled value="10">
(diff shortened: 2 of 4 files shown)


okadurin (approved):

gerjanvangeest:
Any clue why `?disabled=${this.disabled || this.readOnly}` in the html template doesn't work?
https://github.com/ing-bank/lion/blob/9b3e42bceb8a94e37416be50abbb3882636dbd9f/packages/ui/components/input-stepper/src/LionInputStepper.js#L413

Can you remove it?

geniuspacs:
> Any clue why `?disabled=${this.disabled || this.readOnly}` in the html template doesn't work?
> 
> https://github.com/ing-bank/lion/blob/9b3e42bceb8a94e37416be50abbb3882636dbd9f/packages/ui/components/input-stepper/src/LionInputStepper.js#L413
> 
> 
> 
> Can you remove it?

It seems like standard defined that readOnly not apply for buttons or submit. It is only valid for some inputs. Even if you set a button as readOnly, this will not prevent the click event.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly

gerjanvangeest (approved):

geniuspacs:
> Any clue why `?disabled=${this.disabled || this.readOnly}` in the html template doesn't work?
> 
> https://github.com/ing-bank/lion/blob/9b3e42bceb8a94e37416be50abbb3882636dbd9f/packages/ui/components/input-stepper/src/LionInputStepper.js#L413
> 
> 
> 
> Can you remove it?

It seems like the way we are rendering buttons inside _getIncrementButtonNode for example is rendering node out of reactive render lit lifecycle and is causing that is rendering the DOM element but is not sync with the component

Issue #2558, opened 2025-08-11
[OverlayManager][iOS] Scrolling back to wrong position when closing overlay


igomezal:
## Expected behavior

When closing an overlay in iOS, I expect to be at the same position of the page as I was when opening the overlay.

## Actual Behavior

When opening an overlay in iOS, the page scroll to the top as it is applying a position: fixed to avoid issues with the inputs as explained at https://github.com/ing-bank/lion/issues/1905 and when it is closed the page scrolls back until the invokerNode is visible but not back to the same position the user expected to be.

So it is strange for the user to the content to be in the middle of the screen when opening the overlay and once he close it, the content he was on is now at the bottom as it is there when the invoker starts to be visible.

## Additional context

@lion/overlays v0.27.5

SebayK:
I can take it

PR #2712, opened 2026-02-26
Fix/2558/scrolling back to wrong position when closing overlay

SebayK:
#2558

### DIFF
--- a/.changeset/dark-cars-sin.md
+++ b/.changeset/dark-cars-sin.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+restoring scroll position after closing dialog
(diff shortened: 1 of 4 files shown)


SebayK:
https://github.com/user-attachments/assets/8919c1e8-8be1-41d5-9fd3-d1afd8bbda6a

SebayK:
@tlouisse could you check ?

tlouisse:
> @tlouisse could you check ?

Hi @SebayK, had a quick check yesterday.
In the movie above I would expect to see the arrows and invoker again after closing the dialog?

Also mind that open source webkit might not be exactly the same as macos, let alone ios (see https://github.com/ing-bank/lion/issues/2558).

In the code, we explicitly check for iOS, so without either mocking it or running it on actual iOS, the fix code will not be executed in ordinary webkit (neither in macos or or other browsers).

The proof should be:
- run the test in iOS simulator (or device on your local network). You can do that by copying the url
- the test should fail without the code change
- the test should pass with the code change

PR #2737, opened 2026-03-26
fix: 🐛 Dynamic arialabel for input date picker button invoker

movrack:
For A11Y reason, the arialabel must be different if the user use multiple datepicker in the same page.

✅ Closes: #2735

## What I did

1. Add fieldName as parameter of the translation key
2. Update all translations with the fieldName parameter
3. Add test with 2 datepicker and check the aria-label are different
4. Update documentation


docs/components/input-datepicker/use-cases.md
```diff
@@ -96,7 +96,46 @@ Faulty prefilled input will be cleared
...
 export const faultyPrefilled = () => html`
-  <lion-input-datepicker label="Faulty prefiiled" .modelValue="${new Date('30/01/2022')}">
+  <lion-input-datepicker label="Faulty prefilled" .modelValue="${new Date('30/01/2022')}">
   </lion-input-datepicker>
 `;
 ```
+
+## Accessibility
+
+For improved accessibility, it is highly advisable to provide a `name`, `label`, or `fieldName` attribute to your input so the aria-label is set correctly.
```
gerjanvangeest: It is mandatory for an input field to have a label, so this will need a rewrite. And the first 2 examples can be removed.
The third example is the default behaviour, so its also not adding anything.

packages/ui/components/input-datepicker/translations/de.js
```diff
@@ -1,3 +1,3 @@
 export default {
-  openDatepickerLabel: 'Datumswähler öffnen',
+  openDatepickerLabel: 'Öffnen Sie den "{fieldName}"-Wähler',
```
gerjanvangeest: Keep it short 

```suggestion
  openDatepickerLabel: '"{fieldName}"-Wähler öffnen',
```

docs/components/input-datepicker/use-cases.md
movrack: See in [9ebdb40](https://github.com/ing-bank/lion/pull/2737/commits/9ebdb40a5e570f612fd549a6fbd6e68107c023bd)

movrack:
Hello,

Any idea when it can be integrated ?

gerjanvangeest (approved):

PR #2739, opened 2026-03-31
fix(input-amount-dropdown): update _isEmpty method to only check amount property for emptiness

adascaliteiradu:


packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
```diff
@@ -398,14 +398,14 @@ export class LionInputAmountDropdown extends LionInputAmount {
...
-   * @param {string} modelValue
+   * We only consider the amount part for emptiness checks.
+   * Missing, null, empty-string or NaN amount is considered empty.
    * @return {boolean}
    * @protected
    */
   _isEmpty(modelValue = this.modelValue) {
-    return super._isEmpty(modelValue) || this.currency === this.__initializedCurrencyCode;
+    const amount = modelValue?.amount;
+    return amount == null || amount === '' || Number.isNaN(amount);
```
gerjanvangeest: Can we maybe reuse the code from the [original isEmpty function](https://github.com/ing-bank/lion/blob/9b3e42bceb8a94e37416be50abbb3882636dbd9f/packages/ui/components/form-core/src/FormControlMixin.js#L629)

```suggestion
       const isNumberValue = typeof value === 'number' && (value === 0 || Number.isNaN(value));
      return !value && !isNumberValue;
```

packages/ui/components/input-amount-dropdown/test/LionInputAmountDropdown.test.js
gerjanvangeest, line 187: Can we add a test that shows the feedback for?

adascaliteiradu: Done.

packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
adascaliteiradu: Updated.

gerjanvangeest:
@adascaliteiradu can you sign the CLA?

gerjanvangeest (approved):

PR #2756, opened 2026-05-12
feat: Add custom min/max labels to input range

zacharyrad:
## What I did

1. Added option for input-range to have custom min-max labels. 

image
image
image

### DIFF
--- a/packages/ui/components/input-range/src/LionInputRange.js
+++ b/packages/ui/components/input-range/src/LionInputRange.js
@@ -38,6 +38,14 @@ export class LionInputRange extends LocalizeMixin(LionInput) {
         type: Boolean,
         attribute: 'no-min-max-labels',
       },
+      minLabel: {
+        type: String,
+        attribute: 'min-label',
+      },
+      maxLabel: {
+        type: String,
+        attribute: 'max-label',
+      },
     };
   }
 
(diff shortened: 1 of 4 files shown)


gerjanvangeest:
Hi @zacharyrad thank you for the contribution!
Could you sign the CLA?
And could you add a changes (`npm run changeset`), so the change ends up in our changeling?

zacharyrad:
> Hi @zacharyrad thank you for the contribution! Could you sign the CLA? And could you add a changes (`npm run changeset`), so the change ends up in our changeling?

Hi @gerjanvangeest, thanks for checking my PR! I've signed the CLA and added a changeset

gerjanvangeest:
Hi @zacharyrad, sorry it took some time to look at this PR again.
I've added some code to notify screen-reader users also of the min/maxLabel.
Could you add some unit tests for this logic, if not please also let us know than I'll look into that at some point myself?

zacharyrad:
Hi @gerjanvangeest, apologies for also taking a while to look back at this. I've added the unit tests for the recent changes you made. 😄

gerjanvangeest (approved):

packages/ui/components/input-range/src/LionInputRange.js
```diff
@@ -115,6 +144,16 @@ export class LionInputRange extends LocalizeMixin(LionInput) {
     if (changedProperties.has('step')) {
       this._inputNode.step = `${this.step}`;
     }
+
+    if (changedProperties.has('modelValue')) {
```
okadurin, line 148: Could you explain this use case? Why do we show `aria-valuetext` only for min/max cases? And only when custom labels are on. Should we instead show the current value in all cases?

gerjanvangeest: By default the value is read out, only in these case of min/max-label we the a text alternative.

packages/ui/components/input-range/src/LionInputRange.js
```diff
@@ -157,12 +196,16 @@ export class LionInputRange extends LocalizeMixin(LionInput) {
           ? html`
               <div class="input-range__limits">
                 <div>
-                  <span class="sr-only">${this.msgLit('lion-input-range:minimum')} </span
-                  >${formatNumber(this.min)}
+                  <span class="sr-only">${this.msgLit('lion-input-range:minimum')} </span>${this
+                    .minLabel
```
okadurin, line 200: Should we say that `this.minLabel` is by default `formatNumber(this.min)` and it could be overridden by a developer by providing custom attribute/property for it? Then this piece of code would be 
```
<div><span class="sr-only">${this.msgLit('lion-input-range:minimum')} </span>${this.minLabel}</div>
```

zacharyrad: No, I don't think it should be the default since its an optional value.

okadurin (approved):

PR #2758, opened 2026-05-21
chore: more button for menu

okadurin:


gerjanvangeest (approved):

packages/ui/components/menu/src/MoreButtonMenuMixin.js
```diff
@@ -0,0 +1,295 @@
...
+        this.hideMoreButton();
+
+        if (this.doItemsFit()) {
+          return;
+        }
+
+        this.displayMoreButton();
+        const hiddenItemsCount = this.hideItemsOneByOneInMainMenuUntilTheyFit();
+        this.moveHiddenItemsFromMainMenuToMoreButtonMenu(hiddenItemsCount);
+      }, 50);
```
tlouisse, line 293: Any reason for this number? :)

Did we also consider aligning with lit's internal reactive cycle?

packages/ui/components/menu/src/MoreButtonMenuMixin.js
```diff
@@ -0,0 +1,295 @@
...
+            ?.focus();
+        }
+      });
+    }
+
+    getMoreButtonSlotProjection() {
+      return this.querySelector('[slot="more-button"]');
+    }
+
+    getMoreButtonMenu() {
```
tlouisse, line 165: Convention is:
```suggestion
    get _moreButtonMenuNode() {
```

tlouisse:
Looks great @okadurin 👍  Will merge it to our prerelease branch and might do a few tweaks (make methods private instead of public etc.)

packages/ui/components/menu/src/MoreButtonMenuMixin.js
okadurin: That is a random number for debouncing the `resize` event

tlouisse: Ah indeed, maybe we can make a small debounce util

tlouisse (approved):

PR #2769, opened 2026-06-01
fix: make sure more menu items can be tabbed

tlouisse:
## What I did

1.

### DIFF
--- a/packages/ui/components/menu/src/MoreButtonMenuMixin.js
+++ b/packages/ui/components/menu/src/MoreButtonMenuMixin.js
@@ -78,7 +78,7 @@ export const MoreButtonMenuMixin = superclass =>
       this._listNode.appendChild(moreButtonWrapper);
 
       const isElementDirectFocusableItemUnderMoreButtonMenu = (/** @type {Element} */ element) =>
-        moreButtonMenu.contains(element);
+        this.listItems.includes(element) && moreButtonMenu.contains(element);
 
       moreButtonMenu.addEventListener(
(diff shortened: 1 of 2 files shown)


packages/ui/components/menu/src/MoreButtonMenuMixin.js
```diff
@@ -107,46 +75,46 @@ export const MoreButtonMenuMixin = superclass =>
...
 
-      const isElementDirectFocusableItemUnderMoreButtonMenu = element => {
-        const targetTagName = element?.tagName;
-        return (
-          element?.parentElement?.parentElement === moreButtonMenu &&
-          (targetTagName === 'BUTTON' || targetTagName === 'A')
-        );
-      };
+      const isElementDirectFocusableItemUnderMoreButtonMenu = (/** @type {Element} */ element) =>
+        moreButtonMenu.contains(element);
```
okadurin: `.contains` includes also `L2` items. But the variable is for direct L1 children. Also is it safe not to check if the element is `button` or a `tag`? Could it be that a div or something is a target for bubbled `focusin` event

tlouisse: Yes, makes sense. Thought we already filtered on this.listItems, but we didn't. Will put it back

tlouisse: Pushed a fix

okadurin (approved):

PR #2770, opened 2026-06-01
fix(menu): add checked attribute to LionItem

gerjanvangeest:
## What I did

1.

### DIFF
--- a/packages/ui/components/menu/src/InteractiveListMixin.js
+++ b/packages/ui/components/menu/src/InteractiveListMixin.js
@@ -22,7 +22,7 @@ import {
 
 // TODO: consider renaming to FocusGroupMixin
 
-// TODO: make all available in controller/directve (same logic with an elegent prop-to-host-mapping)
+// TODO: make all available in controller/directive (same logic with an elegant prop-to-host-mapping)
 
 /**
  * @param {Element} potentialFocusable
@@ -55,12 +55,24 @@ function isFocusableElement(potentialFocusable) {
 export class LionItem extends LitElement {
   /** @type {any} */
   static get properties() {
-    return { type: String };
+    return {
+      type: { type: String },
+    };
   }
 
   get _invokerNode() {
-    return Array.from(this.children).find(child => child.slot === 'invoker');
+    return /** @type {HTMLElement[]} */ (Array.from(this.children)).find(
+      child => child.slot === 'invoker',
+    );
   }
+
+  static styles = [
+    css`
+      :host {
+        display: block;
+      }
+    `,
+  ];
 
   constructor() {
     super();
@@ -80,14 +92,6 @@ export class LionItem extends LitElement {
       this.setAttribute('role', this._computedType);
     }
   }
-
-  static styles = [
-    css`
-      :host {
-        display: block;
-      }
-    `,
-  ];
 
   render() {
     return html`
@@ -332,11 +336,10 @@ const InteractiveListMixinImplementation = superclass =>
     setCheckedIndex(index) {
       const item = this.listItems[index];
       if (!item) return;
-
       if (!this.multipleChoice) {
         // Uncheck all
         this.listItems.forEach(listItem => {
-          setChecked(listItem, false);
+          setChecked(listItem, true);
         });
         setChecked(this.listItems[index]);
       } else {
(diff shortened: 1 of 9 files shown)


packages/ui/components/menu/src/utils/listItemInteractions.js
```diff
@@ -34,29 +38,48 @@ export function toggleDisabled(item) {
...
+  if (role?.endsWith('checkbox') || role?.endsWith('radio')) {
+    ariaAttr = 'checked';
+  } else if (role === 'option' || role === 'treeitem') {
+    ariaAttr = 'selected';
+  }
+  let text;
+  if (item instanceof HTMLAnchorElement) {
+    text = 'page';
+  }
+  if (/** @type {LionCheckboxIndeterminate} */ (item).mixedState) {
```
tlouisse, line 62: Here we should also support native checkboxes and "synthetic" lightweight ones. I will look into it later

packages/ui/components/menu/src/utils/listItemInteractions.js
```diff
@@ -34,29 +38,48 @@ export function toggleDisabled(item) {
...
+    // }
   }
 }
 
 /**
  * @param {HTMLElement} item
  */
 export function toggleChecked(item) {
-  setChecked(item, !isChecked(item));
+  setChecked(item, isChecked(item));
```
tlouisse: 🤔

This seems not correct? We need a negation here?

packages/ui/components/menu/src/InteractiveListMixin.js
```diff
@@ -332,11 +336,10 @@ const InteractiveListMixinImplementation = superclass =>
     setCheckedIndex(index) {
       const item = this.listItems[index];
       if (!item) return;
-
       if (!this.multipleChoice) {
         // Uncheck all
         this.listItems.forEach(listItem => {
-          setChecked(listItem, false);
+          setChecked(listItem, true);
```
tlouisse, line 342: did we replace all occurrences of setChecked? (N.B. time for unit tests)

gerjanvangeest: found one more

tlouisse (approved):

packages/ui/components/menu/src/LionMenu.js
```diff
@@ -209,6 +213,18 @@ export class LionMenu extends MultiLevelListMixin(LitElement) {
     this._activateOnTypedChars = true;
     /** @configure DisclosureMixin */
     this.invokerInteraction = 'click';
+    /** @protected */
+    this._onPopState = this._onPopState.bind(this);
```
tlouisse: Now that we have class fields, we can also define the fn like this (it will be bound to this automatically)

```
_onPopState = () => {}
```

packages/ui/components/menu/src/LionMenu.js
```diff
@@ -247,4 +263,31 @@ export class LionMenu extends MultiLevelListMixin(LitElement) {
       }
     }
   }
+
+  /** @protected */
+  _onPopState() {
+    this._setLocationAsCheckedIndex(document.location);
```
tlouisse: This should be guarded for types that support aria-current-page only?

packages/ui/components/menu/src/LionMenu.js
```diff
@@ -247,4 +263,31 @@ export class LionMenu extends MultiLevelListMixin(LitElement) {
...
+    this._setLocationAsCheckedIndex(window.location);
+  }
+
+  /**
+   * @param {Location} location
+   */
+  _setLocationAsCheckedIndex(location) {
+    this.listItems.forEach(item => {
+      if (location.href.includes(item.href)) {
+        setChecked(item);
```
tlouisse: Nice. Maybe we can call it `__syncCurrentPageWithLocationHref` ? (and do not call it for listitems that do not support aria-current-page).

Also, we should make sure to uncheck the prev current-page

packages/ui/components/menu/src/LionMenu.js
```diff
@@ -209,6 +213,18 @@ export class LionMenu extends MultiLevelListMixin(LitElement) {
     this._activateOnTypedChars = true;
     /** @configure DisclosureMixin */
     this.invokerInteraction = 'click';
+    /** @protected */
+    this._onPopState = this._onPopState.bind(this);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    window.addEventListener('popstate', this._onPopState);
```
tlouisse: Think this can be applied on InteractiveListMixin level?

PR #2784, opened 2026-06-16
chore(menu): add more test suites

gerjanvangeest:
## What I did

1.


packages/ui/components/menu/test-suites/InteractiveListMixin.suite.js
```diff
@@ -255,7 +256,7 @@ export function runInteractiveListMixinSuite(customConfig) {
 
       it('can set checked state to "mixed" for role="menuitemcheckbox"', async () => {
         const el = await fixture(html`
-          <${tag}>
+          <${tag} multiple-choice>
```
tlouisse, line 259: is this needed for the test to pass?

tlouisse: makes sense for checkbox indeed

tlouisse (approved):

PR #2787, opened 2026-06-17
Add readonly to switch 

ing-athirlwall:
## What I did

1. Added readonly to LionSwitchButton

### DIFF
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -35,10 +35,10 @@ jobs:
       - uses: google/wireit@setup-github-actions-caching/v2
       - uses: actions/checkout@v4
 
-      - name: Setup Node 22.x
+      - name: Setup Node 24.x
         uses: actions/setup-node@v4
         with:
-          node-version: 22.x
+          node-version: 24.x
           cache: npm
 
       - name: Install Dependencies
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -35,10 +35,10 @@ jobs:
       - uses: google/wireit@setup-github-actions-caching/v2
       - uses: actions/checkout@v4
 
-      - name: Setup Node 24.x
+      - name: Setup Node 22.x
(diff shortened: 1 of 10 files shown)


.github/workflows/verify-pr.yml
```diff
@@ -35,10 +35,10 @@ jobs:
       - uses: google/wireit@setup-github-actions-caching/v2
       - uses: actions/checkout@v4
 
-      - name: Setup Node 24.x
```
tlouisse, line 38: Now that we updated Playwright, is this still needed to let ci pass?

tlouisse (approved):

packages/ui/components/overlays/test/OverlayController.test.js
```diff
@@ -875,6 +878,9 @@ describe('OverlayController', () => {
           const { parentOverlay, childOverlay } = await createNestedEscControllers(parentContent);
           await mimicEscapePress(childOverlay.contentNode);
           await waitUntil(() => !parentOverlay.isShown);
+
+          // TODO: This is the failing line ("sometimes").
+          // The child overlay is not staying shown after the parent overlay is hidden.
           await waitUntil(() => childOverlay.isShown);
```
tlouisse, line 884: Did we try:

```
if (!childOverlay.isShown) {
  await waitUntil(() => childOverlay.isShown);
}
```

tlouisse (approved):

packages/ui/components/overlays/test/OverlayController.test.js
ing-athirlwall: The problem is that when the test fails the child overlay "isShown" never returns true - it's own dialog container must have  display 'none' . This can happen when updateConfig on the overlay is called.

.github/workflows/verify-pr.yml
ing-athirlwall: Is this why it was failing? outdated Playwright?
I'll try taking this out ..

tlouisse (approved):

PR #2797, opened 2026-06-24
chore: add more test-suites

gerjanvangeest:
## What I did

1.


packages/ui/components/menu/src/LionTree.js
```diff
@@ -22,7 +22,7 @@ export class LionTree extends MultiLevelListMixin(LitElement) {
     super.updated(changedProperties);
 
     if (changedProperties.has('level')) {
-      if (this.level > 0) {
```
okadurin, line 25: Why do we change this?

okadurin (approved):

PR #2803, opened 2026-07-03
fix(FormControlMixin): make possible invert group two order

geniuspacs:
## What I did

1. Adding a property to make possible set the feedback template above input template on second group

### DIFF
--- a/.changeset/silent-garlics-rescue.md
+++ b/.changeset/silent-garlics-rescue.md
@@ -1,0 +1,5 @@
+---
+'@lion/ui': patch
+---
+
+fix(FormControlMixin): make possible invert group two order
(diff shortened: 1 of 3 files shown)


tlouisse:
@geniuspacs this is something we usually expect to be handled in the extension layer. 
Lion focuses on the generic, functional part of the component (including functional styling), but this is related to a specific styling decision from a design system.

Would that work for you (doing it in your extension layer)?

geniuspacs:
> @geniuspacs this is something we usually expect to be handled in the extension layer. 
> 
> Lion focuses on the generic, functional part of the component (including functional styling), but this is related to a specific styling decision from a design system.
> 
> 
> 
> Would that work for you (doing it in your extension layer)?

That totally makes sense. Thanks for feedback!

PR #2807, opened 2026-07-08
Feat/tooltip longpress on mobile

Knorgias:
## What I did

1. Suppressed `focusin` triggered by a tap:`lastPointerWasTouch` flag used instead of `:focus-visible` alone, which some browsers apply on tap too.
2. Suppressed the `click` event that fires after a completed longpress, so an invoker that also opens an overlay on tap does not trigger both interactions at once.
3. Added unit tests for both fixes and a matching use-case demo in the docs.

https://github.com/user-attachments/assets/d00cd68c-406a-40d7-a6a8-a12dfa9e181c

### DIFF
--- a/docs/components/tooltip/use-cases.md
+++ b/docs/components/tooltip/use-cases.md
@@ -169,6 +169,39 @@ export const longpress = () => html`
 `;
 ```
 
+## Mobile: tooltip on an invoker that also opens an overlay
+
+When an invoker opens an overlay on tap, the two interactions do not collide:
+
+- **Tap**: dialog opens, tooltip does not show.
+- **Long-press 500ms**: tooltip shows, click is suppressed so the dialog does not open.
+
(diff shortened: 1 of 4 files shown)


docs/components/tooltip/use-cases.md
```diff
@@ -169,6 +169,39 @@ export const longpress = () => html`
 `;
 ```
 
+## Mobile: tooltip on an invoker that also opens an overlay
```
gerjanvangeest: I would expect this to be an h3 ( `### Mobile: tooltip on an invoker that also opens an overlay` ) as its part of the previous heading.

docs/components/tooltip/use-cases.md
okadurin: > pick a phone preset,

Could this statement be removed? There are different reasons:
* `phone preset` could be read as a phone component in ing-web
* it's really an optional part, isn't is? If a user doesn't pick a phone preset, the default option works just fine

okadurin (approved):

PR #2809, opened 2026-07-09
chore: lint phone-number lib version so that we are up to date with changes across all countries

tlouisse:
## What I did

prevent we unrightfully say a phonenumber is invalid (like we did before after the recent addition of "494" in Australia)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "lint": "run-p lint:*",
     "lint:eslint": "eslint --ext .js,.html .",
     "lint:markdownlint": "git ls-files '*.md' | xargs markdownlint --ignore '{.github/**/*.md,.changeset/*.md,**/CHANGELOG.md,packages/ui/_legacy-changelogs/*.md}'",
+    "lint:phone-number-up-to-date": "node ./scripts/lint-phone-number-lib.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "lint": "run-p lint:*",
     "lint:eslint": "eslint --ext .js,.html .",
     "lint:markdownlint": "git ls-files '*.md' | xargs markdownlint --ignore '{.github/**/*.md,.changeset/*.md,**/CHANGELOG.md,packages/ui/_legacy-changelogs/*.md}'",
+    "lint:phone-number-up-to-date": "node ./scripts/lint-phone-number-lib.js",
```
gerjanvangeest, line 28: If we automate the phone number check, maybe we can do something similar for the currency list?
https://github.com/ing-bank/lion/blob/master/packages/ui/components/input-amount-dropdown/src/currencyUtil.js

gerjanvangeest (approved):

package.json
tlouisse: Good idea :) Will first merge this one

tlouisse: It seems to be based on a Java lib that we cannot easily automate in our repo, so needs some investigation.

PR #2812, opened 2026-07-10
Emit `user-selected-date-changed` on `keyup` instead of `keydown`

okadurin:


ryubro (approved):
After a lengthy discussion, I think it would be easier and simpler to use keyup in datepicker popup level.

PR #2838, opened 2026-07-30
fix(input-datepicker): set value even if modelValue creates a validat…

gerjanvangeest:
…ion error

## What I did

1.

### DIFF
--- a/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
@@ -371,14 +371,6 @@ export class LionInputDatepicker extends ScopedElementsMixin(
         this.modelValue = undefined;
       } else {
         this.modelValue = selectedDate;
-        /**
-         * In case a custom validator is used, the dates inside the calendar will not be disabled.
-         * The calendar selection will update the modelValue, but not the view value, so we need
-         * to set it manually
-         */
-        // if (this.hasFeedbackFor?.includes('error') && this.serializedValue) {
-        //   this.value = this.formatter(this.deserializer(this.serializedValue));
-        // }
       }
 
       this._isHandlingUserInput = false;
--- a/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
@@ -130,6 +130,10 @@ export class LionInputDatepicker extends ScopedElementsMixin(
     );
   }
 
+  get _isEditingTextField() {
+    return this.focused && this._isHandlingUserInput;
+  }
+
   constructor() {
     super();
     /** @private */
@@ -372,9 +376,9 @@ export class LionInputDatepicker extends ScopedElementsMixin(
          * The calendar selection will update the modelValue, but not the view value, so we need
          * to set it manually
          */
-        if (this.value === '' && this.hasFeedbackFor?.includes('error') && this.serializedValue) {
-          this.value = this.serializedValue;
-        }
+        // if (this.hasFeedbackFor?.includes('error') && this.serializedValue) {
+        //   this.value = this.formatter(this.deserializer(this.serializedValue));
+        // }
       }
 
       this._isHandlingUserInput = false;
--- a/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
+++ b/packages/ui/components/input-datepicker/src/LionInputDatepicker.js
@@ -367,6 +367,14 @@ export class LionInputDatepicker extends ScopedElementsMixin(
         this.modelValue = undefined;
       } else {
         this.modelValue = selectedDate;
+        /**
+         * In case a custom validator is used, the dates inside the calendar will not be disabled.
+         * The calendar selection will update the modelValue, but not the view value, so we need
+         * to set it manually
+         */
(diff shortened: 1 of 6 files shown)


packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -455,6 +456,35 @@ describe('<lion-input-datepicker>', () => {
         expect(el.validationStates.error).to.have.property('IsDate');
       });
 
+      it('shows a date even if the validator throws an error', async () => {
```
tlouisse:
```suggestion
      it('syncs view value from calendar if custom validator is in error state', async () => {
```

packages/ui/components/input-datepicker/src/LionInputDatepicker.js
```diff
@@ -367,6 +367,14 @@ export class LionInputDatepicker extends ScopedElementsMixin(
         this.modelValue = undefined;
       } else {
         this.modelValue = selectedDate;
+        /**
+         * In case a custom validator is used, the dates inside the calendar will not be disabled.
+         * The calendar selection will update the modelValue, but not the view value, so we need
+         * to set it manually
+         */
+        if (this.value === '' && this.hasFeedbackFor?.includes('error') && this.serializedValue) {
```
tlouisse: Should we only sync if view value is empty?

tlouisse: Also, the serializedValue is always ISO format (we want locale dependent formattedValue)

packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -455,6 +456,35 @@ describe('<lion-input-datepicker>', () => {
...
+        await elObj.selectMonthDay(12);
+        await el.updateComplete; // safari take a little longer
+
+        expect(el.hasFeedbackFor).to.include('error');
+        expect(el.validationStates).to.have.property('error');
+        expect(el.validationStates.error).to.have.property('AlwaysInvalid');
+
+        expect(isSameDate(/** @type {Date} */ (new Date(el.value)), elObj.calendarEl.selectedDate))
+          .to.be.true;
+        expect(
```
tlouisse, line 504: We should check if formattedValue is set here (compatible with locale).

packages/ui/components/input-datepicker/test/lion-input-datepicker.test.js
```diff
@@ -455,6 +456,35 @@ describe('<lion-input-datepicker>', () => {
...
+          isSameDate(
+            // @ts-ignore [allow-protected] in test
+            /** @type {Date} */ (new Date(el._inputNode.value)),
+            elObj.calendarEl.selectedDate,
+          ),
+        ).to.be.true;
+      });
```
tlouisse, line 511: There should also be another check to see if the value updates when view value is not empty

packages/ui/components/input-datepicker/src/LionInputDatepicker.js
tlouisse: It seems this can better be solved in FormatMixin, as the problem is that we do not want to format when the user interacts with the textbox ("rewarding" a to be corrected input is confusing UX (formatting is usually perceived as confirmation of correct input)).

The current check is: `if (this._isHandlingUserInput && this.hasFeedbackFor?.includes('error') && this._inputNode)` 
https://github.com/ing-bank/lion/blob/master/packages/ui/components/form-core/src/FormatMixin.js#L294

`_isHandlingUserInput` does not work as a check for the calendar though, as the user is not currently interacting with the input.


What we can do in FormatMixin:

```js
// ...

get _isEditingTextField() {
  // for backwd. compat
  return this._isHandlingUserInput;
}

// ...


if (this._isEditingTextField && this.hasFeedbackFor?.includes('error') && this._inputNode)
```

In LionInputDatepicker


```js
get _isEditingTextField() {
  return this._inputNode.matches(':focus');
}
```

gerjanvangeest: > It seems this can better be solved in FormatMixin, as the problem is that we do not want to format when the user interacts with the textbox ("rewarding" a to be corrected input is confusing UX (formatting is usually perceived as confirmation of correct input)).

I disagree with this, If you select a date from a picker it would be strange if it shows it in a wrong format.

Changing the check inside the `formatMixin` makes other generic tests fail, as they will need to have 
`el.focussed = true` as well when they test the `LionInputDatepicker` format.

tlouisse: > If you select a date from a picker it would be strange if it shows it in a wrong format.

I would agree with that statement, but this is the case with the latest changes, right? (the right format is the formattedValue, and we show it since user does not interact with text field. Basically same behavior as prefilling with erroneous parseable value).

tlouisse: > Changing the check inside the formatMixin makes other generic tests fail, as they will need to have
el.focussed = true as well when they test the LionInputDatepicker format.

That's because the mimicUserInput helper function in the tests does not focus the input. Fixed it in my latest commit

tlouisse (approved):

PR #2846, opened 2026-08-13
Sync value upwards

TyC-vB:
## What I did

1. Override set modelValue and get modelValue
2. _initModelValueBasedOnDropdown checks currency and amount properties instead of modelValue itself
2. [amount-dropdown-integrations.test.js] Add parser and getExpectedModelValueForEmptyInputNode
3. [FormatMixin.suite.js] Add parser and getExpectedModelValueForEmptyInputNode to interface
4. Update tests to be more generic

### DIFF
--- a/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
+++ b/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
@@ -97,7 +97,7 @@ export class LionInputAmountDropdown extends LionInputAmount {
    * For a FormControl to function properly, its modelValue, serializedValue and  viewValue/formattedValue
    * should always translate in two directions. This is expected by `_callParser` method. It guards for empty
    * viewValues. As the viewValue (`.value`) only concerns the amount part and not the currency part in
-   * InputAmountDropdown, this contract is broken. 
+   * InputAmountDropdown, this contract is broken.
    * Below, we compensate for this.
    */
   set modelValue(value) {
--- a/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
+++ b/packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
@@ -100,6 +100,7 @@ export class LionInputAmountDropdown extends LionInputAmount {
    * InputAmountDropdown, this contract is broken.
    * Below, we compensate for this.
    */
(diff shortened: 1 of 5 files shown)


packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
```diff
@@ -73,12 +73,41 @@ export class LionInputAmountDropdown extends LionInputAmount {
...
+    };
+  }
+
+  /** @type {import('../types/index.js').AmountDropdownModelValue} */
+  _modelValue;
+
+  /**
+   * The FormatMixin _syncValueUpwards method creates a hard link between this.value and this.modelValue.
+   * The amount dropdown class value and modelValue are different. Overwriting set modelValue makes sure
+   * that the modelValue remains the correct type with associated properties.
```
tlouisse:
```suggestion
   * For a FormControl to function properly, its modelValue, serializedValue and  viewValue/formattedValue
   * should always translate in two directions. This is expected by `_callParser` method. It guards for empty
   * viewValues. As the viewValue (`.value`) only concerns the amount part and not the currency part in
   * InputAmountDropdown, this contract is broken. 
   * Below, we compensate for this.
```

packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
```diff
@@ -73,12 +73,41 @@ export class LionInputAmountDropdown extends LionInputAmount {
...
+        hasChanged: hasChangedAmountDropdownModelValue,
+      },
+      preferredCurrencies: { type: Array },
+      allowedCurrencies: { type: Array },
+      __dropdownSlot: { type: String },
+    };
+  }
+
+  /** @type {import('../types/index.js').AmountDropdownModelValue} */
+  _modelValue;
```
tlouisse: we should make it private (`__modelValue`), as we don't want Subclassers to override it

Also add `@private` jsdoc tag.

tlouisse (changes requested):

packages/ui/components/input-amount-dropdown/src/LionInputAmountDropdown.js
TyC-vB: Done

tlouisse (approved):

PR #2850, opened 2026-09-03
Fix/tel dropdown

okadurin:
Fix syncing between the tel dropdown value and its input. Fix allows recognize the country code for cases when the number is too short

### DIFF
--- a/packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown--parentheses.suite.js
+++ b/packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown--parentheses.suite.js
@@ -36,11 +36,11 @@ const telDropdownConfig = {
 /**
  * @param {{ klass:LionInputTelDropdown, config: {} }} config
  */
-export function runInputTelDropdownParenthesesSuite({
+
+export function runInputTelDropdownParenthesesSuite(
   // @ts-ignore
-  klass = LionInputTelDropdown,
-  config = telDropdownConfig,
-}) {
+  { klass, config } = { klass: LionInputTelDropdown, config: telDropdownConfig },
+) {
   // @ts-ignore
   const tagName = defineCE(/** @type {* & HTMLElement} */ (class extends klass {}));
   const tag = unsafeStatic(tagName);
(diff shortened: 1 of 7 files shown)


packages/ui/components/input-tel-dropdown/test-suites/LionInputTelDropdown--parentheses.suite.js
```diff
@@ -0,0 +1,103 @@
...
+      it('updates dropdown value', async () => {
+        const dropdown = config.getTelDropdownInvokerEl(el);
+        // @ts-ignore
+        await waitUntil(() => dropdown?.checkVisibility());
+        const input = el.querySelector('input');
+        expect(input).to.exist;
+        // @ts-ignore
+        input.value = '';
+        // @ts-ignore
+        input.value = '(+91) 99451 46400';
```
gerjanvangeest: Why putting first a full phone number and then remove it again?
Doesn't it also works if you add less characters?

packages/ui/components/input-tel/src/LionInputTel.js
```diff
@@ -325,9 +324,7 @@ export class LionInputTel extends LocalizeMixin(LionInput) {
 
     // 2. Try to derive action region from user value
     const regex = /[+0-9]+/gi;
-    const value = !(this.modelValue instanceof Unparseable)
```
gerjanvangeest, line 328: Isn't the removal of this check breaking any code?

gerjanvangeest: Maybe add a test with Unparseable value?

gerjanvangeest (approved):

packages/ui/components/input-tel/src/LionInputTel.js
okadurin: Doing a test would be tricky. So what happens is that the previous version of `awesome-phonenumber` parsed `(+91) 9` slightly different than the current version. Previously that particular input was set to modelValue as Unparseable and not it's not. The fix makes it more robust by parsing the value anyways.

PR #2857, opened 2026-09-15
Guard `showsFeedbackFor` custom converter

matijs:
Guard the custom `showsFeedbackFor` converters instead of assuming there's always a value.

In a Storybook environment where `@wc-toolkit/storybook-helpers` is used we were getting errors because either Storybook or the helper seemed to remove the `shows-feedback-for` attribute resulting in `null` ending up in the converter.

This fixes this issue by guarding `value`.


tlouisse:
🙏 Thanks

tlouisse (approved):

